Issue #270, opened 2015-07-06
Map issues with Firefox


archesprojectbot:
Originally reported by: **Bojan Kastelic (Bitbucket: [bojankd64](http://bitbucket.org/bojankd64), GitHub: [bojankd64](http://github.com/bojankd64))**

----------------------------------------

With Mozilla Firefox, v. 39 (Windows and OS X, haven't tried on Linux yet), all markers have some (about 100 pixels) vertical offset. All clicks reponds on markers about 100 pixels below. If there is no other marker below clicked area, click on marker has no effect.

This issue is present only in Firefox 39 (38 works fine), on other browsers (Safari, Chrome) this is no issue.

----------------------------------------
- Bitbucket: https://bitbucket.org/arches/arches3/issue/283

jmunowitch:
Upgrading Openlayers from 3.1.1 to 3.2.0 fixed this issue.
#427 should fix this issue, along with #426.

PR #288, opened 2016-01-12
change to firefox compatible event properties

mradamcox:
modify two functions so that firefox can interpret the pixel position correctly.

### DIFF
--- a/arches/app/media/js/views/map.js
+++ b/arches/app/media/js/views/map.js
@@ -118,9 +118,9 @@ define([
                 self.trigger('viewChanged', view.getZoom(), extent);
             });
 
-
             this.map.on('click', function(e) {
-                var clickFeature = self.map.forEachFeatureAtPixel(e.pixel, function (feature, layer) {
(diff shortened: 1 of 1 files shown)


mradamcox:
also, this should close [issue 270](https://github.com/archesproject/arches/issues/270), but I'm not sure if there's an automated way to do that.

mradamcox:
So I'm not sure why the travis tests failed, I've only just learned about what Travis CI is.  I only changed a few, non-build-related lines, so does that mean these tests will fail on every pull request?

apeters:
Hi Adam,
We have a couple of tests that are failing at the moment that aren't
related (as far as I know) to what you pushed.   Any time you push it
triggers a build.  Until we fix those 2 tests it will always fail.
So, I wouldn't worry.
Cheers,
Alexei


Director of Web Development - Farallon Geographics, Inc. - 971.227.3173

mradamcox:
Ok, that's what I was thinking, just wanted to double-check.

PR #348, opened 2016-03-18
Model Upgrade

jmunowitch:
These are the latest changes to the Data Model as discussed between @apeters and @adamlodge on or around March 11th

### DIFF
--- a/arches/app/models/models.py
+++ b/arches/app/models/models.py
@@ -50,12 +50,12 @@ class BranchMetadata(models.Model):
 
 class Card(models.Model):
     cardid = models.UUIDField(primary_key=True, default=uuid.uuid1)  # This field type is a guess.
-    name = models.TextField()
-    title = models.TextField()
+    name = models.TextField(blank=True, null=True)
+    title = models.TextField(blank=True, null=True)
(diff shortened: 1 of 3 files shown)


arches/app/models/models.py
```diff
@@ -172,19 +175,21 @@ class Meta:
...
         managed = True
         db_table = 'forms_x_card'
-        unique_together = (('form', 'parentcard'),)
+        unique_together = (('form', 'card'),)
 
 
 class NodeGroup(models.Model):
     nodegroupid = models.UUIDField(primary_key=True, default=uuid.uuid1)  # This field type is a guess.
-    cardinality = models.TextField()
+    cardinality = models.TextField(blank=True, null=True)
```
apeters: I think every nodegroup needs to specify a cardinality, maybe we can specify a default value as well

adamlodge: You’re right.  I’ll make sure the change gets implemented.  

Adam  

--  
Adam Lodge
Geospatial Systems Consultant
Farallon Geographics
415.317.6625

jmunowitch:
@apeters: The update has been made.

PR #431, opened 2016-05-04
421 graph metadata form

robgaston:
some small fixes for graph metadata work


robgaston:
coverage decreased 😢 
This is because I deleted one previously unused, but covered line.  I didn't add any untested python lines, thus coverage went down.  Accordingly, I am merging this in despite this.

Issue #447, opened 2016-05-23
Graph Designer Landing Page: Manage Button UI behavior


dwuthrich:
Graph/Resource and its icon should remain highlighted even when user's cursor moves over the Manage button

dwuthrich:
meets requirement

dwuthrich:
meets requirement

PR #468, opened 2016-05-24
Graph manager fixes

robgaston:
fixes issue #447


robgaston:
fixes issue #465 - removing the js file referenced therein, thus nothing to document

PR #491, opened 2016-06-01
Ko components

robgaston:


robgaston:
This is a pretty simple modification of our existing cards/forms/widgets feature that simply reworks widgets (which were previously just django templates (ie markup w/o js) as ko components (ie js and markup, bound with a viewmodel).  I believe this will be useful for developing more complex widgets and training outside developers on widget development.  Here are some notes (pro, con & other):

## pros:
- can use js to manipulate both viewmodel and element (for a map, for instance) (http://knockoutjs.com/documentation/component-registration.html#a-createviewmodel-factory-function)
- can attach functions to viewmodel for use in widget template
- can internally require other resources (such as custom bindings) so form page doesn’t need to keep track of those
- can easily train other devs on building ko components using ko docs
- ko has already thought this problem through, and it seems reasonable to use their solution rather than build our own, which will inevitably be less flexible
- makes widget registration and development easier

## cons:
- slightly more verbose for simple widgets (needs js file)

## other:
- templates are raw text, so cannot use django template tags; this still seems nice just to keep coupled templates w/ js and keep them simple.  If we wanted to, we could load these templates via django’s template engine.
- localization cannot be done in templates, but shouldn’t be a problem as the entire widget record is passed to the component, so we can simply attach any localized strings to the widget object provided by the server

I would like to merge this work in, but some review would be appreciated.

PR #740, opened 2016-08-22
Validation data model changes

robgaston:


arches/app/models/migrations/0001_initial.py
```diff
@@ -263,6 +263,8 @@ class Migration(migrations.Migration):
             fields=[
                 ('datatype', models.TextField(primary_key=True, serialize=False)),
                 ('iconclass', models.TextField()),
+                ('config', django.contrib.postgres.fields.jsonb.JSONField(blank=True, db_column='config', null=True)),
```
robgaston: @ryan86 I think we should rename this column from `config` to `defaultconfig`

arches/app/models/models.py
```diff
@@ -87,7 +87,9 @@ class DDataType(models.Model):
     datatype = models.TextField(primary_key=True)
     iconclass = models.TextField()
     defaultwidget = models.ForeignKey(db_column='defaultwidget', to='models.Widget')
-
+    config = JSONField(blank=True, null=True, db_column='config')
```
robgaston: See above note re: renaming `d_data_types.config` to `d_data_types.defaultconfig`

robgaston:
@ryan86 we should also add the linking records to `validations_x_datatypes` for the 'required' validation in `db_data.sql` (this validation should be related to all `d_data_types` records except 'semantic'.

robgaston:
Looks good to me!

Issue #528, opened 2016-06-09
Graph Manager: Move Node Form alerts to global alert panel


dwuthrich:
## Expected Behavior
User alert messages should be displayed in the global alert panel

## Current Behavior
Node form is used to display alerts in the Graph Manager

robgaston:
fixed in #531

dwuthrich:
Alert displays as expected.  However, canceling from the alert form dismisses the Node CRUD form.  Instead, form should stay open but edits should be discarded

robgaston:
I think what we want to happen is the following:

- if the user clicks 'cancel' (meaning, cancel the change in selection) or the 'x' button the form should remain open with the original node selected and the edits remain in place
- if the user clicks okay, then the edits are discarded (as explained in the alert itself) and the users action that triggered the alert proceeds
- if the user triggered the alert by selecting another node (either through the graph or the node list), then the newly specified node should be selected and loaded into the node form

dwuthrich:
meets requirements

PR #757, opened 2016-08-31
fixing graph manager alert behavior

robgaston:
re: #528


apeters:
I restarted the build, as it looks like IE didn't respond in time for the test to pass

Issue #749, opened 2016-08-29
Graph relationship management UI unclear


azaroth42:
In the Graph Manager, when creating a relationship between two Graphs, the label for the dropdown of possible relationships to select is "Relation to parent". This is confusing, as the relationship is actually _from_ the parent to the child that was just added, rather than from the child to the parent.

As a data modeler, I create two graphs, an E7 Activity and an E40 Legal Body that can carry out that Activity and want to link the two together with the P14 carried out by relationship.  In order to add this relationship, I have select the Legal Body node, and say that the relationship from that node to the parent is carried out by, which is not correct.

Requested Change:  Modify the UI template to say "Relationship from Parent"

Nice to Have:  Dynamically generate the UI to say "Relationship from [name of parent node]"

Nice to Have: Below the dropdown, dynamically generate text:  "[name of parent node] [name of relationship] [name of currently selected node]", e.g. `Activity p14_carried_out_by Legal Body`

jmunowitch:
Thanks @azaroth42. We are taking a look at this and should be adding it to our current Sprint.

dwuthrich:
meets requirements

Issue #746, opened 2016-08-25
Force selection of a root class in graph settings form


robgaston:
When creating a graph, if the user selects an ontology,  but not a root class the graph manager page results in an error.

We should force selection of a root class on the settings form when an ontology is selected to prevent this state from occuring.

dwuthrich:
Implementation defaults to first entry in the root node list.  Better would be to not set a default value but instead confirm that root node is not null on save

Issue #737, opened 2016-08-22
Filter list of validations in node and widget forms based on the node’s data type


robgaston:
In Graph Manager > node form and in Card Designer > widget form

Using the foreign key added to the `validations` table as part of #735

robgaston:
Perhaps we can support generalized validations with a null value in the foreign key to `d_data_types`.  These validations should show up for all nodes.

apeters:
Functionally works, but for datatypes that don't define any validations, should we even show the dropdown?  Also, when there's an empty list, the styling of the dropdown appears like it's trying to show something but that it's chopped off (see screen shot).  

![screen shot 2016-08-31 at 9 49 18 pm](https://cloud.githubusercontent.com/assets/823773/18155685/d59215d4-6fc4-11e6-96f7-323b4a547232.png)

Will it be clear to the user that there is nothing to select and that typing something in is not an option?

PR #762, opened 2016-09-01
Card graph bugs

robgaston:
re: #749, #746, #737


arches/app/templates/views/graph/card-configuration/component-forms/widget-configuration.htm
```diff
@@ -76,6 +76,7 @@ <h4 class="title-block-title" style=""> {% trans "Widget Manager" %} </h4>
 
     <div id="widget-crud-validation" class="tab-pane fade">
 
+        <!-- ko if: $parent.validations().length > 0 -->
```
apeters: should we remove this, or rather just disable?

PR #772, opened 2016-09-08
Fixes #771

### LINKED ISSUE #771: Validation of graph fails with non-semantic root
We are enforcing the rule that the root node of a graph must be non-semantic.  This is only true when the graph is a "Resource", but the validation currently enforces this rule regardless of whether the graph is a "resource" or not.

ryan86:
Fixes graph validation bug that wouldn't allow user to save graph if the top node was not semantic. The top node only needs to be semantic if you are creating a resource.

### Description of Change
Bug fix

### Issues Solved
#771 

### Types of changes

### Checklist


ryan86:
Fix bug in top node validation.

PR #924, opened 2016-10-12
Bower component install

robgaston:


robgaston:
@apeters what do you think of this?

robgaston:
@apeters I'd tried to work it that way initially, but it wasn't working so well w/ our CI integrations.  I think I've got it all sorted, but for now, developers will just have to run `bower install` manually as a setup step.  We can try to integrate it into our python setup scripts, but I don't see this as absolutely necessary.  The recipe for setting it up is in [`arches/install/install_arches.sh`](https://github.com/archesproject/arches/blob/bower_component_install/arches/install/install_arches.sh) and [`.travis.yml`](https://github.com/archesproject/arches/blob/bower_component_install/.travis.yml).  If you've already got node & bower installed on your system, all you should need to do is run `bower install`.

Issue #1237, opened 2016-11-18
Card cardinality not persisted when importing a branch/resource model 


apeters:
#### Steps to Reproduce (for bugs)

1. Create a simple branch with a single string node
2. In the Card Manager, set the card cardinality to "n" and save the card
3. Export the branch
4. View the branch in a text editor and confirm that the nodegroups property has a single nodegroup and that the cardinality is set to "n"
5. Delete the branch from Arches
6. Import the branch from the .json file
7. Open the card manager and observe that the branch cardinality is set to "1"

apeters:
fixed in the latest

PR #1279, opened 2016-11-30
fixes for bug where cardinality isn't persisted on import of a json f…

apeters:
re #1237


apeters:
this was fixed by #1294

Issue #1256, opened 2016-11-23
Export to arches.json doesn't properly include forms_x_cards data


apeters:
export a graph with more then one form and inspect the json, 
notice there is only 1 form id being referenced in the forms_x_cards array

dwuthrich:
meets requirements

PR #1281, opened 2016-11-30
properly export forms_x_cards, re #1256

apeters:
re #1256


apeters:
this was fixed by #1294

PR #1457, opened 2017-01-09
Unify top navbars with help: restructure Resource Editor base template

mradamcox:
### Description of Change
Small template restructuring to improve consistency for help panel where it appears on different pages. Specifically, the ep-toolbar is moved from the header block in templates/views/resource/resource-base.htm to the content block, allowing the help panel to vertically fill the whole view. This is to match the existing template structure that is found in templates/views/graph/graph-base.htm.

### Types of changes

- [ x] New feature (non-breaking change which adds functionality)

### Checklist

### Further comments
I've recommended to @dwuthrich that this same ep-toolbar template code be used in all base pages, e.g. in the Arches Designer landing page, and the jury is currently out.


mradamcox:
There is a different nav bar on some of the other pages that does use that block, I believe. I'm going to make a new branch to see about refactoring more navbars, and maybe then the header block can be removed everywhere...

Issue #1178, opened 2016-11-09
Add 'Edit' shortcut for authenticated users viewing a resource report


mradamcox:
### User Story 
As a user, I'd like to a quick way to edit a resource if I see something amiss while viewing a resource report, like the EDIT button in v3. Currently, I don't see how to edit an existing resource.

dwuthrich:
closing ticket as v4 appears to implement this functionality

PR #1509, opened 2017-01-25
Further top navbar unification

mradamcox:
### Description of Change
This is a fairly substantial reworking of the top navbar that exists in all pages that inherit from base-manager. The contents of the bar is now defined dynamically by altering context variables at the view class level.

### Issues Solved
No specific issue existed for this overall update, but #1178 (adding an edit shortcut to the resource report) is taken care of.

### Types of changes

### Checklist

### Further comments
This creates a single unified top navbar for all of the base-manager pages. It remains to be seen how this will be incorporated into the rest of the site, i.e. those pages that a not-logged in user will be viewing. Mostly because these pages aren't made yet. It's likely that the same bar could be used.


mradamcox:
Thanks Alexei, made all of these fixes. Also talked with @chiatt about the search-base-manager template that was added after I initially created this branch. I'm removing the knockout-driven navbar he had put in that template in favor of the one we now have in all pages that inherit from base-manager.

Issue #1520, opened 2017-01-27
Empty extensions list results in loading null ontology


azaroth42:
If I take out all of the extensions, the current internal serialization of the list is deserialized to [''], which then calls add_ontology() with an empty data_source.

Either the serialization should result in a real empty list, or run_loader() should check if the extension is true before calling add_ontology()

azaroth42:
Closed by #1521 ?

apeters:
fixed by #1538

PR #1521, opened 2017-01-27
Ontology load to config ; closes #1519

### LINKED ISSUE #1519: Move initial ontologies to config?
In trying to manage the set of terms available within the system, I replaced the CIDOC ontology with a new version that includes the 6.2.2 Monetary_Amount and related terms and removed the unused CRM extensions.  Assuming that we can't dynamically reload ontologies (?) I recreated the installation from scratch.  During packages setup, the initial migration errors as it cannot find the extensions here:

https://github.com/archesproject/arches/blob/master/arches/app/models/migrations/0001_initial.py#L25

Could these file paths be exported to settings.py, overridable in settings_local.py?
Or if there's required to be a master ontology with extensions, pick the master and then assume all other .xml files in the directory are extensions?

If there's a better way to manage this, please let me know :)

azaroth42:
### Description of Change

Move initial load of ontology to settings to enable override at install time.

### Issues Solved

#1519 
#1520 

### Types of changes

### Checklist

### DIFF
--- a/arches/app/models/migrations/0001_initial.py
+++ b/arches/app/models/migrations/0001_initial.py
@@ -26,16 +26,10 @@ def forwards_func(apps, schema_editor):
     # We get the model from the versioned app registry;
     # if we directly import it, it'll be the wrong version
 
-    path_to_ontologies = os.path.join(settings.ROOT_DIR, 'db', 'ontologies', 'cidoc_crm')
-    extensions = [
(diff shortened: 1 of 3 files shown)


apeters:
Thanks for the PR @azaroth42, I ended up merging your code (with a few minor changes) in with this PR #1538

PR #1538, opened 2017-02-01
Azaroth42 master

apeters:
re #1519


azaroth42:
:+1: Many thanks @apeters!

PR #1576, opened 2017-02-03
set iconclass from graph

mradamcox:


dwuthrich:
If the resource model does have an icon class, then nav bar renders as expected.  If the resource model doesn't have an icon class then no icon is rendered at all and the title is mis-aligned.

Desired behavior: if no icon is defined in icon class, default to "chevron-circle-right".  

Note also that the RDM nav bar does not use the icon for the RDM

mradamcox:
New commits should address all of these issues. As new pages are created (import/export), remember that the icon class should be set in the `context['nav']['icon']` property of the view class for each page, nowhere else.

dwuthrich:
Appears that if no icon is defined in the resource model, nav bar mis aligns text and does not display an icon

mradamcox:
Hm, I'm not able to reproduce that. This is what happens locally for me, are we looking at the same part of the app? Which raises another question: I'm assuming chevron-arrow-right should be used to replace the question mark as shown in the first part of this gif?
unspecificied icon

dwuthrich:
That gif is helpful.  I'm looking at Resource Editor nav bar (e.g.: editing an instance of the resource model)

mradamcox:
K, I have the same thing there though.
unspecificied icon-editor
(I chopped that one up to speed up the page load)

dwuthrich:
Interesting.  I'm not seeing the same behavior.  But go ahead and commit to master.  I'll issue another ticket if the issue still persists.

PR #1584, opened 2017-02-06
1580 delete resource index

chiatt:
Added deletion of a resource instances index, related resources and related resources indexes to the delete method of the resource view.


chiatt:
@apeters I was thinking that this functionality should be moved into the Resource model, or maybe we should override the delete method on the ResourceInstance model.

Issue #1702, opened 2017-02-27
Update virtualenv to 15.1.0


jmunowitch:
### Bug Report

Installing Arches on Windows with virtualenv 13.1.2 requires an upgrade to setuptools and pip within the activated virtual environment.  Installing Arches on windows with virtualenv 15 does not cause these issues.

Our documentation will be updated in https://github.com/archesproject/arches/issues/1696. 

arches/setup.py needs to be updated to remove deprecated syntax `--no-use-wheel`
virtualenv/virtualenv.py needs to be updated to the latest version of `virtualenv` (15.1.0)

Issue #1696, opened 2017-02-27
Update Install Documentation for Arches


jmunowitch:
As we are going through the install process for the Liverpool workshop on Windows 10 machines, we are running into a few hiccups that should be documented:

- [x] Considering upgrading the virtualenv version used to 15.*
- [x] Pip upgrade needs to happen within the virtual environment: `install --upgrade pip` (Dependant on virtualenv update)
- [x] Setuptools upgrade needs to happen, also within the virtual environment: `install --upgrade setuptools` (Dependant on virtualenv update)
- [x] On Windows, Shapely does not install through`python setup.py install`.  It needs to be installed via the Wheel that is on this site: http://www.lfd.uci.edu/~gohlke/pythonlibs/#shapely
- [x] Update Prereq for JDK.  With Windows, the JAVA_HOME needs to be added to PATH
- [x] Update documentation on the new version of ElasticSearch (See link [here](https://www.elastic.co/guide/en/elasticsearch/reference/5.2/windows.html))
- [x] Consider linking to the MacOS install gist I created [here](https://gist.github.com/jmunowitch/35eb1d20a0b9cbf86f4b0eb4b167d43a)

jmunowitch:
In our V[3 documentation](http://arches3.readthedocs.io/en/latest/getting-started/#installing-arches) and [V4 documentation](https://github.com/archesproject/arches/wiki/Dev-Installation), it instructs the user to pip install virtualenv. In the V4 documentation, the version of `virtualenv` is pinned to 13.1.2 which includes the following python libraries by default:
```
Package    Version
---------- -------
pip        7.1.2
setuptools 18.2
```

With `virtualenv`version 15.1.0, the following python libraries are included:
```
Package    Version
---------- -------
pip        9.0.1  
setuptools 34.3.0 
```

Since we seem to have to update pip and setuptools in the Windows install by hand, would it be worth using `virtualenv` version 15.1.0?

Also of note, there is a separate process that could install virtualenv alltogether in the [setup.py](https://github.com/archesproject/arches/blob/master/arches/setup.py#L82) that calls the [virtualenv.py](https://github.com/archesproject/arches/blob/master/virtualenv/virtualenv.py) file to install a virtual environment. this method is is currently pinned to version 1.11.2, but the method here is not really documented anywhere.

What should the suggested install for `virtualenv` be?

jmunowitch:
[...]

PR #1748, opened 2017-03-02
1650 docker support

veuncent:
Initial files needed to run Arches in Docker

### Description of Change

Docker is the world’s leading software containerization platform: https://www.docker.com/  

Some advantages of Docker:  

    - Eliminates the installation process from the end-user side: one can simply download the Arches Docker image from hub.docker.com and run the container.  
    - Keeps your system clean and provides a stable environment that is the same wherever you run it.  
    - Enhances security: the container your app is running in is isolated from the host machine.  

### Types of changes

### Checklist
- [x] I have added necessary documentation (if appropriate)
--> See [docker/Readme.md](docker/Readme.md)

### To-do:
- [x] Fix unit tests: Tests are not detected by `python manage.py test tests --pattern="*.py" --settings="tests.test_settings"`. Needs fixing.
- [x] Check missing time_wheel_config: Getting a 404 on: `/resource/search/time_wheel_config HTTP/1.1" 404`
- [x] Develop Readme and workflow that distinguishes between Docker for Arches core and Docker for Arches end users (custom projects).
- [x] Add remote debugging to the Image. *Note: added all that is necessary to enable remote debugging for Visual Studio Code. Added directions for this in docker/Readme.md, as well as for debugging with Pycharm.


apeters:
I notice you're using postgres 9.5 and mapnik 3.0 and I we're using 9.6 and 2.0 respectively

jmunowitch:
It looks like some versions are a bit off. This is totally our fault because we haven't updated the appropriate install scripts/documentation for this. 

Here are some updated library versions:
- Virtualenv should now be 15.1.0 (See: https://github.com/archesproject/arches/issues/1696 and https://github.com/archesproject/arches/issues/1702)
- Elasticsearch is now pinned to 5.2, it should be set to the appropriate version in the [pip install](https://github.com/archesproject/arches/pull/1748/files#diff-3254677a7917c6c01f55212f86c57fbfR96) (See: https://github.com/archesproject/arches/commit/8081c17bf401a1700efcc831a252ddef6cd21ad4)

apeters:
Additionally you have ip addresses hard-coded in settings_local.py.  Those should probably be removed.

veuncent:
Hard-coded ip addresses: these are actually services provided by AWS to check an EC2 instance's private ip (here used to check if server is running in AWS at all).

After discussion Mapnik binaries will be removed altogether. 

Other versions will be upgraded in following commit.

veuncent:
I pushed a number of new commits to my docker fork. Most notable changes are 
- Changing versions according to discussion
- Remove Mapnik binaries
- Removed unnecessary `pip install` commands
- Added convenient cleanup script
- Moved `setup.py develop` to entrypoint.sh, so it is only run when container starts, not during image build
- Moved init script into entrypoint.sh --> `manage.py packages -o setup` is run if psql cannot find existing db (can be forced with a flag). Please let me know what you think of this. It looks a bit scary, but it makes the whole ordeal a lot simpler. The psql command looks safe enough to me.

veuncent:
- Updated Readme to reflect new Organization name `archesproject`
- Updated docker-compose.yml to work as production environment.
- Updated Readme to reflect this and how to use it for dev

veuncent:
Sure thing

arches/app/utils/data_management/sparql_providers/aat_provider.py
```diff
@@ -26,7 +26,7 @@
 class AAT_Provider(Abstract_Provider):
 
     def __init__(self, **kwargs):
-        super(AAT_Provider, self).__init__('http://vocab.getty.edu/sparql.json', **kwargs)
```
apeters: This needs to be changed back to use the whole fully qualified url (eg: http://vocab.getty.edu/sparql.json)
This will fail without the protocol specified.

PR #1762, opened 2017-03-06
1548 help updates

mradamcox:
###Description of Change

This is a first but complete draft of all in-app help related to the Arches Designer tools. The one exception is the Card Designer help, wherein the section describing card/card container settings can not be completed until a few outstanding tickets have been addressed.

###Issues Solved

New help content in the following pages:

Arches Designer landing page
Card Manager
Card Designer
Menu Manager
Menu Designer
Report Manager
Report Designer
Function Manager
(i.e. all of the separate templates that compose the Arches Designer. Settings and Graph Manager are already in master)

### Types of changes

### Checklist
- [x] I have added necessary documentation (if appropriate)

### Further comments

This is new pull request to take place of the original, in order to better manage some merge conflicts.


mradamcox:
from @adamlodge 

Below are a bunch of comments that can serve as fodder for a discussion between us Adams.

Arches Designer landing page
About Arches Designer:
Suggested reword:
…the structure of your database, the details of your data entry forms, and how data are presented in reports. To begin…
Terminology and Components
—> Resource Models
…a data entry user must decide to which resource model a given resource belongs. …
—> Branches
- n/a
—> Graph
- I would remove the third and fourth sentences. They are more about resource models than graphs.
- (Truly a question, not a statement): What do you think about replacing “logical framework” with “logical data model”? Also, what do you think about ordering? Should Graph come before Branch and Resource Model?
—> Cards
- We should talk about this. What you wrote is inaccurate in the strictest sense, but this may be a good candidate for sacrificing precision in favor of clarity. The dubious statement is: “One branch = one card”.
—> Menus
- Rewrite: Menus are groupings of Cards associated to a given resource model. They allow for an organized, thematic approach to data entry.
—> Reports
- Rewrite: Reports allow you to control how the attributes within a resource model will be presented to the public.
—> Functions
- Functions are behind-the-scenes operations that are run on some user-initiated activity. Examples might be: calculate a value administrative area based on resource geometry location, or dynamically concatenate name and name type into a single string when viewing a report.

Putting It All Together
Very well written. However, I am concerned about the assertion that one branch = one card for two reasons: 1> at the db level, a card record is created for every nodegroup within a branch. So, for instance, "Place 2" branch actually has 6 cards in it, 2> we envision a completely seperate set of cards being created for every node group for use in the mobile app.

Card Manager

Card Designer
This section acknowledges that there are many cards in a given branch. Concerned about consistency.

Menu Manager
Since there was a demo in the Card Manager, I wonder if a demo is also appropriate... for consistency if nothing else.

Menu Designer
Looks good.

Report Manager

Partial Rewrite:
To begin, choose a template from the left panel. The differing templates respectively prioritize display of tabular attributes, a map showing location of the resource, or an image associated with the resource.

You will notice that by default, a new Report is set as Inactive. This allows the Report to remain hidden from users until you have finished configuring it. To configure a Report, click on it and you'll be brought to the Report Designer.

Report Designer
Feels kinda thin, but I have zero experience actually making reports, so I dont have any useful input. Probably we should get the HIP reports built in order to learn some lessons to share here.

Function Manager

It might be worth saying in the document that functions are envisioned as the hook through which developers can easily customize Arches capabilities. As such, a given Arches implementation may have custom built and registered to support a cusomtization. However, these three functions are delivered with Arches by defualt to support common needs.

I think it might be misleading to say that functions only fire on save operations. In your name/name type example, it is technically true that the function fires on save, but the result of the code is only saved in the index, and not in the DB.

mradamcox:
From @veuncent Some ideas regarding Arches documentation / Help section

General

Consider adding a 'view full size' link to the images (click to enlarge).
You could highlight key words in a paragraph to navigate the eye through it
Perhaps even add links/anchors to the description of the mentioned keywords so you can switch back and forth between definitions.
To go even further, it would be awesome if on hover over a key word or paragraph header an arrow would show what section on the arctual Arches page you are talking about (where applicable, e.g. a Card).
Consider adding a breadcrum at the top of the Help pages, allowing the user to navigate to a parent Help section (preferably without leaving the actual Arches page). Handy e.g. in cases where the user is looking at a specific Resource Model in the Designer: The user could be made aware that there is a more general help page about the Resource Designer on the /graph page.
/graph/

Terminology and Components

Paragraphs are nice and short.
Expanding and collapsing the headers cleans the screen up nicely.
Under "Graph": To me the words 'logical framework' don't mean much. Is there a synonym? Or a more 'everyday' term?
Under "Graph": The term 'nodes' is not included in this glossary. Consider adding a paragraph or explaining it in-line.
Under "Graph": This sentence is a little confusing:
"Graphs comprise nodegroups--collections of one or more nodes that either represent data to be collected, or serve as connections to other nodes."
--> Maybe something like: "Graphs consist of nodegroups. Nodegroups are collections of one or more nodes that either represent data to be collected, or serve as connections to other nodes."
When you speak of a "Branch's Graph", could you instead also just say "Branch"? Makes it simpler.
Under "Cards": The sentence
"it is where you define how information will be collected for each node"
could be made simpler by speaking of data entry instead of information collection, something like:
"it is where you define how information will be entered by your data entry users. It is a part of a data entry form."
Menus, Reports and Functions are clear and to the point.
Consider including images of a graph and branch. Also the process of copying a branch to a graph.
Visualisation always helps a bunch when explaining stuff.
Alternatively, consider refering to the 'Putting it all Together' paragraph below, so users are aware of those images.
Putting it All Together

This sentence could possibly be improved:
"Now consider what happens when a Branch is added to a Resource Model. Its component parts are copied into the Resource Model, and in the Resource Model you will add the Cards to Menus to organize the data entry process."
--> perhaps: "Now consider what happens when a Branch is added to a Resource Model. Both the Branch's Graph and its Cards are copied into that Resource Model. Next, in order to define what a data entry user will see in the web form, you will add the Branch's Cards to the Resource Model's Menus."
/graph/.../settings

Just to be sure you know: the link in "To learn more about what this means, read Ontologies in Arches." still points to a draft in Google Docs.
Piece on Ontology (in gdocs) explains things well. Good to have a simple definition of ontologies, as it is quite an abstract term for the layman.
Icons: is it possible for the user to add custom icons?

Issue #1346, opened 2016-12-07
Map Widget: scroll bar persists after overlay menu is removed.


mradamcox:
Found this while using the widget in the Data Widgets and Cards context, but also happens in the Resource Editor.

![map widget scroll bar](https://cloud.githubusercontent.com/assets/10427268/20983309/93f29c7a-bc81-11e6-86bc-149a4b39a852.gif)

Windows 10 / Chrome

robgaston:
@mradamcox looks like a platform specific thing; can you post details on OS & browser, please?

mradamcox:
Sure, I'm just using Chrome on the dev.archesproject machine.

robgaston:
@mradamcox sorry, I meant your local OS, looks like Windows, but not sure which version...

mradamcox:
on windows 10.

mradamcox:
Perhaps related: just opened the dev site on firefox for the first time, and found more scroll bar issues. For example:

![image](https://cloud.githubusercontent.com/assets/10427268/20983606/bc0a87e4-bc82-11e6-82a0-98c769625fc2.png)

A) scroll bar along tools menu
B) this one shows up for me in Chrome also but really isn't an issue overall.
C) main scroll bar for form is not functional at all.

I'm going to turn this into a separate issue actually...

mradamcox:
Please see related issue #1348 .

chiatt:
@mradamcox This should be cleared up now. Let me know if you see any unwanted scroll bars on the map or elsewhere.

mradamcox:
all good!

PR #1899, opened 2017-03-23
1346 scroll bars

chiatt:
Set vertical scrolling to `auto` in elements where the scroll bar was showing up unnecessarily. re #1346


chiatt:
@mradamcox   I didn't see the persistence of the scroll bars on my Windows vm, but I went ahead and made some css changes so that vertical scroll bars don't show up on Windows browsers (Chrome and FF) when there is no overflow.  

I'm on Windows 10, Chrome Version 56.0.2924.87 (64-bit), FF 52.0.1 (32-bit)

mradamcox:
all good!

Issue #1740, opened 2017-03-01
Create IIIF viewer and annotation widget


apeters:
Use Case:
1. A user is provided a list of available IIIF(http://iiif.io/) services to select from.
2. A user selects a an IIIF service and then the widget shows the image in the IIIF viewer widget.
3. A user can then add annotations.  A single annotation can consist of many different things but most likely would include a spatial object that references the image selected in step 2, as well as any number of additional metadata (text, rich-text, dates, etc...)

Mockup and more details to follow.

robgaston:
This will require implementing a custom data type as well, most likely.

dwuthrich:
meets requirements

PR #1940, opened 2017-04-07
z-index fix for iiif widget

robgaston:
re: #1740


chizzel36:
Coverage remained the same at 38.105% when pulling d23f791 on 1740_iiif_widget into b89560a on master.

Issue #1499, opened 2017-01-20
Business Data Import Validation


ryan86:
Create method for validating business data on import. This should output some contextual information about the errors that occurred and where to find them.

It is unclear as of now whether this will occur as arches saves tiles or will be performed on the source file format (csv, shapefile, arches.json, etc).

chiatt:
@apeters There are more places where import validation could be implemented, but I think in the coming weeks we can add those in as errors arise (particularly while testing v3 to v4 data migration).

apeters:
There are a couple of issues:
1. the error message specifies an invalid path to the .log file

> ***** Errors occured during import. For more information, check resource import error log: arches/arches/logs/resource_import.log

2. Bounding box validation doesn't cause an error to be reported (we're missing the `errors.append` statement)
3. For geometry validation I'd put this line within the try/except to catch error in serializing and converting to GEOSGeometry:
> geom = GEOSGeometry(JSONSerializer().serialize(feature['geometry']))
4. There is no validate method defined for Concept List, Boolean, Date, File, and File List datatypes

apeters:
Export for a point produces this:
GEOMETRYCOLLECTION (POINT (-101.4553730175889257 39.8188997091764634))
but import throws an error when trying to import it.

Also, when importing the date "345" which is clearly invalid, the system says that there was an error importing it, but the import still seems to occur (as '0345-01-01').

Here's the graph and test data that I've been using:
[Validation_Graph.zip](https://github.com/archesproject/arches/files/1020425/Validation_Graph.zip)

apeters:
meets requirements

PR #2072, opened 2017-05-15
1499 business data import validation

ryan86:
Added import statement for distutils.
re #1499

### Types of changes

### Checklist


apeters:
Export for a point produces this:
GEOMETRYCOLLECTION (POINT (-101.4553730175889257 39.8188997091764634))
but import throws an error when trying to import it.

Also, when importing the date "345" which is clearly invalid, the system says that there was an error importing it, but the import still seems to occur (as '0345-01-01').  

Here's the graph and test data that I've been using:
[Validation_Graph.zip](https://github.com/archesproject/arches/files/1020422/Validation_Graph.zip)

PR #2144, opened 2017-06-14
Fix 'Discard Edits' bug that creates an empty branch when no branches are present

azerbini:
originalItem was undefined until addItem method triggered. Now
originalItem is set to null from initialize stage


azerbini:
Sorry, tried to push to wrong branch, a.

PR #2145, opened 2017-06-14
Several bug fixes/improvements to v.3

azerbini:
This PR contains several v.3 improvements and bug fixes:

**Bug fixes**

- Resolve problem with 'Discard edits' creating an empty branch when the method addItem has not yet been triggered
- Fix mouse position bug with Firefox v. > 40 by porting app to OpenLayers 3.20.0
- Fix duplication of child_entities which have no sub child_entities and value =''. This occurred with highly nested resource graphs.

 **Improvements**

- Improved JSON exporter
- Introduce full-screen controls on maps
- Introduce new clustering method on maps to improve performance at scale

### DIFF
--- a/arches/app/media/css/package.css
+++ b/arches/app/media/css/package.css
@@ -1,1 +1,20 @@
 /*put package specific styles here...*/
+
+/*Map to full screen
+------------------------------------*/
+.map:-moz-full-screen {
+height: 100%;
+}
+.map:-webkit-full-screen {
+height: 100%;
+}
+.map:-ms-fullscreen {
+height: 100%;
+}
+.map:fullscreen {
+height: 100%;
+}
+
+.ol-full-screen {
+z-index:900;   
+}
--- a/arches/app/media/js/map.js
+++ b/arches/app/media/js/map.js
@@ -23,7 +23,8 @@ require([
             'click .visibility-toggle': 'visibilityToggle',
             'click .on-map-toggle': 'onMapToggle',
             'click .layer-zoom': 'layerZoom',
-            'click .cluster-item-link': 'clusterItemClick'
+            'click .cluster-item-link': 'clusterItemClick',
+            'change:resolution': 'ZoomChange'
         },
         initialize: function(options) {
             var self = this;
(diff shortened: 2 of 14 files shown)


mradamcox:
I was just checking out the fullscreen as implemented in the EAMENA installation, and I noticed that the popup shows up underneath the fullscreen map. I recently encountered this issue on a different project, and found out that the webkit fullscreen forces itself to the max z-index (even if you try to "important" over it)
Screenshot below is from dev tools viewing EAMENA map in fullscreen.
image
The [fix I used](https://github.com/mradamcox/hkstairs/commit/d5c23df4dcc9ad4bbfa8b5646e6e264779c4b52b) was just to set the thing you want on top of it also to the max z-index. Then it gets rendered on top.

Anyway, I can't test all these fixes now, but thought I would mention that as an aside.

azerbini:
Thanks for this @mradamcox, I had completely forgotten the resource popups, partly because I have not yet done much on full screen layouts, and I plan to do a lot more (I am planning a fullscreen version of the Resource Manager to make the most of our big Mac screens). Your solution is good, but I have noticed that it doesn't work on Firefox and this (https://stackoverflow.com/questions/18578244/displaying-elements-other-than-fullscreen-element-html5-fullscreen-api) suggests that similar problems may be present in IE.

mradamcox:
Good call, I see that comment now. My own project wasn't tested in IE nor Firefox, but good to know. Unfortunately,  as someone suggests, changing the container element may actually be pretty tricky, given the complexity of these html templates...

mradamcox:
Hey @azerbini looking at the test that is failing in the travis build, it seems like it's something in one of the authority files? It may be that the test needs to be updated to pass correctly.

azerbini:
Something to do with the stable/3.x branch authority files, @mradamcox, nothing that comes from my end. I will keep pushing stuff onto this big PR, and once I am done committing everything I have to commit to this branch, I will see if I can figure out where the problem lies.

jmunowitch:
Hi @azerbini: It looks like we are waiting for you on this PR.

Is there anything left on your end?

azerbini:
@jmunowitch Hi, sorry for the late reply. I do have quite a bit more that I could add now, including Drag&Drop KML to map filter, v3 advanced search, etc. Do you want me to keep using the same PR? It will be some time until I am able to push new commits though, since I need to strip the project-specific code first.

mradamcox:
@jmunowitch @azerbini @apeters I'd recommend closing this PR for now, and re-introducing it later, potentially in smaller pieces. It's the first PR on the stable/3.x branch in a long time, so it may be nice to kick off any PRs into that branch with a very small commit, so we can figure out all of these failing tests.

mradamcox:
Ok, closing now.

PR #2195, opened 2017-06-28
2069 geocoder results

chiatt:
### Description of Change
Replaced geocoder proxy classes with knockout components.
Added a django model for geocoders and made it available in the admin interface.
Switched the default geocoder to mapbox.


arches/app/media/js/views/components/geocoders/mapzen.js
```diff
@@ -0,0 +1,57 @@
+define(['knockout', 'mapbox-gl', 'arches', 'views/components/geocoders/base-geocoder', 'geocoder-templates'],
+function (ko, mapboxgl, arches) {
+    return ko.components.register('views/components/geocoders/mapzen', {
+        viewModel: function(params) {
+            BaseGeocoderViewModel.apply(this, [params]);
+            var self = this;
+            this.placeholder = params.placeholder || ko.observable('Locate a Place or Address');
+            this.anchorLayerId = params.anchorLayerId;
+            this.apiKey = params.api_key || arches.mapboxApiKey
```
robgaston: @chiatt Shouldn't this reference `arches.mapzenApiKey`?

arches/app/templates/views/components/geocoders/mapbox.htm
```diff
@@ -0,0 +1,49 @@
+{% load i18n %}
```
robgaston: I think that these two templates (`mapbox.htm` and `mapzen.htm`) are identical, no? I think we could just create a single `geocoder.htm` template that both of these components could use.

arches/app/media/js/views/components/geocoders/mapzen.js
chiatt: Ah - good catch

arches/app/templates/views/components/geocoders/mapbox.htm
chiatt: Yeah - I think they are. I'll go ahead and combine them.

chiatt:
@robgaston Ok - I've consolidated the geocoder templates into `geocoder.htm`.

Issue #2282, opened 2017-07-24
Not including an image for a saved search hangs the search page


apeters:

chiatt:
meets requirements

Issue #1636, opened 2017-02-15
Implement node group level permissions in forms


apeters:

dwuthrich:
@chiatt @apeters User account belonging to guest group (all nodes read only by default), but with account privileges set to read, edit/create, delete for the HER Place nodes (in the Assets model in the disco_data test dataset) cannot access data entry forms; instead is re-directed to log-in page.  See: http://recordit.co/6onKENkpeu

dwuthrich:
@chiatt guest user can now access Arches designer UI

apeters:
@dwuthrich to achieve this currently a user has to be part of the "Resource Editor" group.  Once part of that group, you can set the permissions as you outlined above and everything should work as expected.

dwuthrich:
@apeters if this is true then its not clear why we have a "Crowdsource Editor" group.

chiatt:
meets requirements

PR #2297, opened 2017-07-25
instead of using the "Resource Editor" group to determine eligibility…

apeters:
…ch page, re #2282, re #1636

### DIFF
--- a/arches/app/media/js/views/base-manager.js
+++ b/arches/app/media/js/views/base-manager.js
@@ -42,6 +42,7 @@ define([
                     return graph.isresource;
                 });
             });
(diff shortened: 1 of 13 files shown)


apeters:
To confirm that this works, you should be able to do the following:
1. A user with write privileges on ANY nodegroup should see have access to the "Manage Data" buttons on the splash screen
2. Only the resources associated with those nodegroups above should be visible in /resource and in the left hand nav bar
3. A user with delete only privilege on a nodegroup would be able to view a form but not create a new resource (and obviously only be able to delete the associated nodegroup data)

PR #2041, opened 2017-05-08
#2037 Some .htm templates call http links

veuncent:
### Description of Change

Removed "http:" part of "http://" links.

### Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Checklist

- [ ] Unit tests pass locally with my changes 
- [ ] I have added tests that prove my fix is effective or that my feature works --> **n/a**
- [ ] I have added necessary documentation (if appropriate) --> **n/a**

Issue #2037, opened 2017-05-05
Some .htm templates call http links


veuncent:
### Bug Report
On some pages css and/or js files are loaded using http URLs, e.g.: 
`<link rel="stylesheet" href="http://cdnjs.cloudflare.com/ajax/libs/select2/3.5.0/select2-bootstrap.min.css" type="text/css">`
(in arches/app/templates/rmd.htm)

This causes trouble when running Arches under a strict https protocol. Most modern browsers refuse to load any http content when the main site is running under https.

#### Expected Behavior
I would expect front end media to be loaded using `//` instead of `http://` to avoid the described problem. 

#### Current Behavior
Affected files:
`arches/app/templates/rmd.htm`
`tests/fixtures/viewer.html` (less problematic)

#### Your Arches Information

* Version used: master branch, SHA d99722e3b917f3fbbfdcd03d64201d9be97c4a7c 
* Browser Name and version: Chrome Version 58.0.3029.96 (64-bit)

#### Context
I would like to run Arches under strict https in production.

veuncent:
P.s. I will create a pull request for this.

chiatt:
@veuncent I searched for other files that were also using http. I think they are all covered, but let me know if you see any others.

PR #2358, opened 2017-08-15
#2037 Some .htm templates call http links

veuncent:
This is a replacement for PR #2041 

### Description of Change

Removed "http:" part of "http://" links.

### Types of changes

### Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works --> **n/a**
- [ ] I have added necessary documentation (if appropriate) --> **n/a**

### DIFF
--- a/arches/app/utils/data_management/sparql_providers/aat_provider.py
+++ b/arches/app/utils/data_management/sparql_providers/aat_provider.py
@@ -26,7 +26,7 @@ from abstract_provider import Abstract_Provider
 class AAT_Provider(Abstract_Provider):
 
     def __init__(self, **kwargs):
-        super(AAT_Provider, self).__init__('http://vocab.getty.edu/sparql.json', **kwargs)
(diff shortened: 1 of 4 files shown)


arches/app/utils/data_management/sparql_providers/aat_provider.py
```diff
@@ -26,7 +26,7 @@
 class AAT_Provider(Abstract_Provider):
 
     def __init__(self, **kwargs):
-        super(AAT_Provider, self).__init__('http://vocab.getty.edu/sparql.json', **kwargs)
+        super(AAT_Provider, self).__init__('//vocab.getty.edu/sparql.json', **kwargs)
```
robgaston: @veuncent I think that in this case we should retain the protocol (`http:`) as this particular request is sent from the server, rather than the client

veuncent: Got it, done!

robgaston: Thanks!

Issue #2363, opened 2017-08-17
Enable 'export' property in mapping file


ryan86:
### Bug Report
The 'export' property in the csv mapping file currently does not determine which nodes can and can/cannot be exported.

#### Expected Behavior
The 'export' property should determine whether or not business data from a given node can be exported.

#### Steps to Reproduce (for bugs)
1. Create a mapping file, add a file_field_name value, but leave 'export' false
2. Export business data for that node and observe that business data still exports

dwuthrich:
meets requirements

PR #2364, opened 2017-08-18
Enable 'export' property in export mapping json.

ryan86:
Enable property in export mapping json. Now export respects the 'export' property in the mapping json. If export is set to false for a given node that business data will not be exported and vice versa.

### Issues Solved
#2363 

### Types of changes

### Checklist


jmunowitch:
The code from this PR is already in master.

PR #2400, opened 2017-09-11
add user confirmation input to command, add -y/--yes to force

mradamcox:
### Description of Change
Added a user input prompt to confirm the remove_resource procedure. Added a `-y` `--yes` flag to allow command to run without user input.

### DIFF
--- a/arches/management/commands/packages.py
+++ b/arches/management/commands/packages.py
@@ -373,13 +373,10 @@ class Command(BaseCommand):
         """
         # resource_remover.delete_resources(load_id)
         if not force:
-            opt = raw_input("all resources will be removed from your database. continue? y/n ")
-        else:
-            opt = "y"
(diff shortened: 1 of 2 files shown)


mradamcox:
reason for using -y/--yes is that -f is already used for format. I would personally much prefer -f/--force, but it isn't available at this time. refactoring to a function that can be used in an if statement at the beginning of each function would be great and not too hard.

The argument is added to each function that is called. At that level, changing the default is as easy as `operation(force=True)` or `operation(force=settings.MODE=='DEV').

mradamcox:
of course, --force could be used by itself, but I thought it would be better to keep to the short-long (??) argument pair convention

mradamcox:
Personally I think it is more responsible to have the default always be `False`, as it's really not that hard to add `-y` to the end of any command.

PR #2544, opened 2017-10-11
Fixed automatic loading of system settings local

chiatt:
### Description of Change
Added logic at the end of setup_db to overwrite system settings if a user has a system settings local file.


mradamcox:
Thanks @chiatt!

PR #2560, opened 2017-10-13
Allow package load from directory

mradamcox:
### Description of Change
previously a package could only be loaded from a zip file, or a url to a github zip archive. This is not ideal during development, so I changed the function to allow loading from a directory.

### Types of changes

### Checklist


mradamcox:
looks like we have an old merge commit that was authored on a server, so no CLA. closing this for now.

PR #2635, opened 2017-10-28
add new management command that finds a uuid in your db

mradamcox:
UPDATE: please check comments below.

This management command will take a given uuid and return information about what that uuid refers to in your database. At present, all it does is return the name of the table that the uuid can be found in, and the contents of the corresponding row. It would be great to instantiate an actual object and return its properties as well, but that would be a lot more work, as I didn't see any standard way of linking a django model with an arches model.

This command could become the basis of a new view/url in the future as well, which would return a json response and/or help create a top-level utility which would take in any uuid and return the corresponding arches object.

Thanks @adamlodge for the kernal of sql to get this going.

##### USAGE

input:
`python manage.py whatisthis b7112550-43d3-4f44-bdf4-2f50f2f81433`

output:

    {
      "createdtime": "2017-10-26 15:55:26.894000-05:00",
      "graphid": "c67216bf-8cc2-11e7-883c-06ed184dc22c",
      "resourceinstanceid": "b7112550-43d3-4f44-bdf4-2f50f2f81433",
      "legacyid": "b7112550-43d3-4f44-bdf4-2f50f2f81433",
      "table name": "resource_instances"
    }

input:
`python manage.py whatisthis ad342550-43d3-4f44-bdf4-2f50f2f81433`

output:

    -- this uuid could not be found in your database

input:
`python manage.py whatisthis b7112550-43`

output:

    -- invalid uuid

### DIFF
--- a/arches/management/commands/whatisthis.py
+++ b/arches/management/commands/whatisthis.py
@@ -1,22 +1,17 @@
 from django.core import management
 from django.core.management.base import BaseCommand, CommandError
-# 
 import django.apps
-# import psycopg2
-import arches
-import json
 import uuid
-import time
 
 class Command(BaseCommand):
 
-    help = 'finds a uuid throughout your database and prints info from postgres'
+    help = 'finds any arches objects whose primary key is the input uuid and'\
+        'returns these objects in the form of a list.'
 
     def add_arguments(self, parser):
         parser.add_argument("uuid",help='input the uuid string to find')
 
     def handle(self, *args, **options):
-
         self.find_uuid(options['uuid'])
 
     def find_uuid(self,in_uuid):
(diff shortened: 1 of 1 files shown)


mradamcox:
@adamlodge pointed out that the SQL we were using in the initial commit for this required that the pg user have high-level permissions to access the information_schema. These permissions are not something that is guaranteed in Arches, and shouldn't be just for this operation.

So the new commit is a different approach that references all Django models and returns the one that matches. This is faster, and ultimately more useful. However, I don't know if it's the _fastest_ way to handle this so it may not be a good enough function yet to work into general arches code.

Usage is basically the same as described above, but now the function returns an object made from the given uuid. So, from elsewhere in the app, you should be able to use (though not tested yet):

    from django.core.management import call_command
    ob = call_command('whatisthis', 'b7112550-43d3-4f44-bdf4-2f50f2f81433')
        # do something with new ob

adamlodge:
@apeters... let Adam C or I know if you have questions.

mradamcox:
One unresolved issue with this is that UUIDs can actually be reused throughout the Arches db (I didn't know about this at the time I made the command). At present, the command only returns the first in a list of matches. Given that two or more objects in the db may use the same UUID for their pk, the command would best be altered to return a list.

apeters:
I could be mistaken, but I don't think the same UUID is used to refer to more than 1 item.  If that's not the case we should make every effort to fix that.

mradamcox:
I certainly agree that UUIDs _shouldn't_ be reused as primary keys, so I'm really not sure how to resolve the fact that @ryan86 said that they **might** be reused in some cases.

apeters:
@mradamcox I think there is one instance where we use the same UUID and it's that the root node in a nodegroup will share the same ID with that nodegroup.

EG: Root Node UUID == NodeGroup UUID

I think that if you handle for that case then this would be a nice utility

mradamcox:
Ok, I can now confirm that yes that is the case: a Node Group's pk is also used for the associated node. In that circumstance, the output of this command is now
```
================================================================================
This UUID is the primary key for 2 objects:
--------------------------------------------------------------------------------
NodeGroup object
_state
  <django.db.models.base.ModelState object at 0x0000000006AAC518>
cardinality
  1
legacygroupid
  None
nodegroupid
  0e8fdef0-4148-11e7-8330-c4b301baab9f
parentnodegroup_id
  0e8ff31c-4148-11e7-998a-c4b301baab9f
--------------------------------------------------------------------------------
Node object
_state
  <django.db.models.base.ModelState object at 0x0000000006AAC898>
config
  None
datatype
  semantic
description
  Represents a single node in a graph
graph_id
  ff623370-fa12-11e6-b98b-6c4008b05c4c
isrequired
  False
issearchable
  True
istopnode
  False
name
  Project Extent
nodegroup_id
  0e8fdef0-4148-11e7-8330-c4b301baab9f
nodeid
  0e8fdef0-4148-11e7-8330-c4b301baab9f
ontologyclass
  None
================================================================================
```
Again, this is primarily just for help during development and debugging. I haven't tried to optimize it for general use within the code-base.

Issue #2646, opened 2017-11-01
Docker: always run setup_db from arches root folder


veuncent:
### Bug Report

Currently the workdir is not set in `entrypoint.sh` (Docker) before running setup_db.
When the workdir is changed, e.g. by a custom arches project, then `setup_db` is run from the wrong folder, leading to unpredictable behavior (e.g. django migration order gets messed up).

Additionally, Django migrations are currently run on every startup (when `MODE=DEV`). Change this to run only once by incorporating the `run_migrations` command in the `setup_arches` command.

PR #2649, opened 2017-11-01
Set workdir for `setup_db`, change when run_migrations is called

veuncent:
Set workdir to arches root before running `setup_db` AND `set_dev_mode`. 
Move run_migrations to `setup_arches` command, re #2646

### Issues Solved

Workdir was not set in entrypoint.sh (Docker) before running setup_db.
When the workdir is changed, e.g. by a custom arches project, then setup_db is run from the wrong folder, leading to unpredictable behavior (e.g. django migration order gets messed up).

Additionally, Django migrations were being run on every startup (when MODE=DEV). Changed this to run only once by incorporating the `run_migrations` command in the `setup_arches` command.

### Types of changes

### Checklist

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -64,12 +64,13 @@ init_arches() {
 		setup_arches
 	fi
 
-	init_arches_projects
+	init_arches_project
 }
 
 
+# Setup Postgresql and Elasticsearch
 setup_arches() {
-	# Setup Postgresql and Elasticsearch (this deletes your existing database)
+	cd_arches_root
 
 	echo "" && echo ""
 	echo "*** Initializing database ***"
@@ -113,6 +114,7 @@ setup_arches() {
 		fi
 	fi
 
+	run_migrations
 }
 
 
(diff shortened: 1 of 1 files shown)


docker/entrypoint.sh
```diff
@@ -64,12 +64,13 @@ init_arches() {
...
 
-	init_arches_projects
+	init_arches_project
 }
 
 
+# Setup Postgresql and Elasticsearch
 setup_arches() {
-	# Setup Postgresql and Elasticsearch (this deletes your existing database)
+	cd_arches_root
```
jmunowitch: Have you tested this with Projects?

It seems like this should be APP_FOLDER as opposed to ARCHES ROOT.

veuncent:
Yes this is tested with a Project and it works fine.

It doesn't seem to be a requirement to run `setup_db` from a Project if it
exists. Rather, getting the base Arches installed before moving on to
setting up the Project removed an issue with unexpected order of Django
migrations.

After `setup_db` the script continues to run `run_migrations` from the
Project folder.

Issue #2732, opened 2017-11-10
Required node validation fails when validating required data in a child tile


ryan86:
Business data import breaks when trying to validate required node data in a child tile/child node.

Add tests for required node validation

mradamcox:
Works great, and good explanatory log message if an error is encountered.

PR #2733, opened 2017-11-10
add resource model filter to acquisition of required nodes

mradamcox:
### Description of Change
adding a resource model id filter to the code that gets a list of required nodes to check. Previously, if any required nodes existed in any branches or resource models (even if the nodes were not actually associated with the resource model into which data was being imported).

related to (but not a full fix for) #2732, because error still occur if required nodes _are_ attached to that resource model.

Also, this demonstrates the use of the `mapping` object, which could probably utilized more thoroughly throughout `CSVReader.import_business_data()`.


mradamcox:
the little change that this PR included has been added from a different branch in the meantime, so I'm going to close this.

Issue #2532, opened 2017-10-10
Boolean datatype displays "true" or "false" instead of label values in reports


apeters:
**In the form it displays like this:** 
<img width="308" alt="screen shot 2017-10-10 at 11 24 25 am" src="https://user-images.githubusercontent.com/823773/31403333-ac4b5316-adad-11e7-96c0-640788467c0b.png">

**But in the report it displays like this:**
<img width="398" alt="screen shot 2017-10-10 at 11 24 45 am" src="https://user-images.githubusercontent.com/823773/31403334-ac6971d4-adad-11e7-8ab4-04058d31232a.png">

ryan86:
@apeters do we also want to use the true and false labels in advanced search or should those stay values stay 'True' and 'False'?

apeters:
@ryan86 I think we should use the labels, 
as a user I could set up the labels to be "Black/White", or "Strong/Weak", so I would expect those to show up in adv. search.

PR #2895, opened 2018-01-08
2532 boolean display value

ryan86:
Use boolean true and false labels instead of 'true' and 'false' in widgets, search, and reports. Move trueLabel and falseLabel config to node config form and node config object in db.

### Issues Solved
#2532 

### Types of changes

### Checklist


arches/app/media/js/views/components/widgets/radio-boolean.js
```diff
@@ -31,6 +31,13 @@ define(['knockout', 'underscore', 'viewmodels/widget'], function (ko, _, WidgetV
                 }
             }
 
+            if (this.value() === true) {
```
robgaston: This should probably use a computed instead, something like this:
```
this.displayValue = ko.computed(function () {
    if (this.value() === true) {
        return this.node.config.trueLabel;
    } else if (this.value() === false) {
        return this.node.config.falseLabel;
    }
}, this);
```

robgaston: This entire block, that is, (ln 34-40)

arches/app/models/migrations/2532_boolean_datatype_display_value.py
```diff
@@ -0,0 +1,59 @@
...
+from django.core import management
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('models', '2402_node_to_node'),
+    ]
+
+    operations = [
+            migrations.RunSQL("""
```
robgaston: We should also add to this migration some SQL to remove these configuration values from `widgets.defaultconfig` (and add them back in reverse)

robgaston: we probably also want to use something other than an empty string for the default labels; it appears that previously we used "Yes" and "No"

Issue #2908, opened 2018-01-11
Bump pyshp from 1.2.3 to 1.2.12 in /arches/install 


jmunowitch:
Re PR: https://github.com/archesproject/arches/pull/2684

Also: Is pyshp needed if we can use built in GeoDjango: https://github.com/archesproject/arches/issues/2689

Bump pyshp from 1.2.3 to 1.2.12.

Issue #2689, opened 2017-11-08
Investigate if pyshp is needed


jmunowitch:
@mradamcox suggested that pyshp functionality could be replaced with the builtin GeoDjango functionality.  In an effort to reduce dependancies, we should look into it.

mradamcox:
I don't know what it is used for, but perhaps if geodjango can replace pyshp, it could also replace shapely.

mradamcox:
Here's how easy it is to use geodjango utils. This code is probably 95% of what is needed to eliminate `pyshp` and `Shapely` from the Arches codebase.

```
from django.contrib.gis.gdal import DataSource
shp = r"C:\arches\AllPublicLandsFL.shp"
ds = DataSource(shp)
layer = ds[0]
crs = layer.srs
for feat in layer:
    name = feat.get("NAME")
    geom = feat.geom.wkt
```

https://docs.djangoproject.com/en/1.11/ref/contrib/gis/layermapping/, http://www.paolocorti.net/2009/04/01/a-day-with-geodjango/, etc.

PR #2922, opened 2018-01-11
#2689 replace all pyshp and shapely operations with django.contrib.gi…

mradamcox:
…s.gdal

These two libraries were only used in one file, `importer.py`, though shapely was imported in `datatypes.py` but not utilized there.

### Issues Solved
#2689
#2912 
#2908
#2923 

### Types of changes

### Checklist

### Further comments
To test this you will need to pip uninstall pyshp and shapely and then import some shapefiles. I tested with 3 shapefiles which were loaded into a project with this package installed https://github.com/archesproject/arches-sample-data. I used point, line, and polygon (which included a donut polygon for good measure) shapefiles.

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -19,7 +19,6 @@ from django.contrib.gis.geos import fromstr
 from django.contrib.gis.geos import Polygon
 from django.core.exceptions import ValidationError
 from django.db import connection, transaction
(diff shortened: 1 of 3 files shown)


mradamcox:
Because pyshp loaded shapefiles differently than the geodjango utils do, I decided to fix #2923 in this same branch and PR. It was an error with certain zip files not being working with the previous unzipping method.

Issue #258, opened 2015-05-28
RDM: Alt/pref label


archesprojectbot:
Originally reported by: **Phil Carlisle (Bitbucket: [Philbox](http://bitbucket.org/Philbox), GitHub: [Philbox](http://github.com/Philbox))**

---

The alternative and preferred labels are not displaying correctly.

Changed the preference for meters/metres to display en-GB as pref en-US as alternative.

meters still appearing in hierarchy

---
- Bitbucket: https://bitbucket.org/arches/arches3/issue/271

chiatt:
Each concept can have one preflabel per language. If the preflabel's language disagrees with the language defined in the application settings (LANGUAGE_CODE = ''), the RDM will default to an altlabel in the language defined in the settings. In other words, if the application language is `en-US`, then a user will see the `en-US` label, even if it is an alt label (trumping a pref label in a different language). This ensures that the language that people read in the RDM is consistent with what is in the settings.  

To display labels in another language in the RDM, the LANGUAGE_CODE in settings.py would need to be changed to the desired code.

PR #2965, opened 2018-01-22
258 v4.1.x inaccurate pref labels

chiatt:
### Description of Change
Ensures that if there is only one preflabel, regardless if that label is in the default language or not, that preflabel is displayed in the rdm.  re #258


chiatt:
In discussion with @apeters  it was decided not to implement these changes.

PR #3040, opened 2018-02-13
Settings system

afeliu:
If the value of system setting in database is null, keep the value of the setting in settings file


afeliu:
reopen pull request

PR #3276, opened 2018-04-30
Southlake enhancements

robgaston:
enhancements to Arches that resulted from April 2018 Southlake retreat


robgaston:
@apeters thanks! I've reverted that commit

Issue #3353, opened 2018-05-08
Search result count for the anonymous user is lower than expected


chiatt:
### User Story
The search result count for the anonymous user, despite having read access to every node group, is lower than it is for the admin user.  If the anonymous user has read permissions on every nodegroup, the search result count should be the same as it is for the admin user.

chiatt:
To QA:

Verify that the 'anonymous' user sees fewer search results than 'admin'. With the lincoln data loaded and logged in as 'admin', the following search should return about 29 resource instances. Logged in as 'anonymous', it should return 0.

Search for Lincoln Heritage Asset - Artefacts in 1801

Using Lincoln Heritage Asset - Artefact descriptors function, reindex the instances.

Re-load the above search for the 'anonymous' user. The count should be the same as it is for the 'admin' user

dwuthrich:
meets requirements

PR #3355, opened 2018-05-09
Fixes search result count for users without reviewer status

chiatt:
Adds provisional flag for date ranges indexed from concept dates. re #3353

### DIFF
--- a/arches/app/datatypes/concept_types.py
+++ b/arches/app/datatypes/concept_types.py
@@ -55,7 +55,7 @@ class BaseConceptDataType(BaseDataType):
                 min_date = ExtendedDateFormat(date_range['min_year']).lower
                 max_date = ExtendedDateFormat(date_range['max_year']).upper
                 if {'gte': min_date, 'lte': max_date} not in document['date_ranges']:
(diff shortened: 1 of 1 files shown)


chiatt:
To QA:

1. Verify that the 'anonymous' user sees fewer search results than 'admin'. With the lincoln data loaded and logged in as 'admin', the following search should return about 29 resource instances. Logged in as 'anonymous', it should return 0.

    [Search for Lincoln Heritage Asset - Artefacts in 1801](http://localhost:8080/search?temporalFilter=%7B%22fromDate%22%3A%221801-01-01%22%2C%22toDate%22%3A%221801-01-01%22%2C%22dateNodeId%22%3A%22%22%2C%22inverted%22%3Afalse%7D&typeFilter=%5B%7B%22graphid%22%3A%22343cc20c-2c5a-11e8-90fa-0242ac120005%22%2C%22name%22%3A%22Lincoln%20Heritage%20Asset%20-%20Artefact%22%2C%22inverted%22%3Afalse%7D%5D&no_filters=false&page=1)

2. Using Lincoln Heritage Asset - Artefact descriptors function, reindex the instances.

3. Re-load the above search for the 'anonymous' user. The count should be the same as it is for the 'admin' user

Issue #3526, opened 2018-06-18
add delete_mapbox_layer command


mradamcox:
**Is your feature request related to a problem? Please describe.**
Adding map layers must be done through the command line, and tileserver layers can be removed from the command line, but mapbox layers must be removed through the admin UI. This is awkward as you need to have the dev server up and running to remove basemaps, and it makes a programmatic workflow more cumbersome.

**Describe the solution you'd like**
Add a `delete_mapbox_layer` command option that performs just like `delete_tileserver_layer`.

mradamcox:
added, works as needed.

PR #3530, opened 2018-06-18
add operation for deleting mapbox layer #3526

mradamcox:
Please see https://github.com/archesproject/arches/pull/3529

This is a cherry-pick of the same commit into a branch that was made from master.

### DIFF
--- a/arches/management/commands/packages.py
+++ b/arches/management/commands/packages.py
@@ -22,6 +22,7 @@ import urllib, uuid, glob
 from datetime import datetime
 from django.core import management
 from django.core.management.base import BaseCommand, CommandError
+from django.core.exceptions import ObjectDoesNotExist
 from django.utils.module_loading import import_string
 from django.db import transaction, connection
 from django.db.utils import IntegrityError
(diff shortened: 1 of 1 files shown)


arches/management/commands/packages.py
```diff
@@ -1088,7 +1091,15 @@ def delete_tileserver_layer(self, layer_name=False):
                 tileserver_layer.map_layer.delete()
                 tileserver_layer.map_source.delete()
                 tileserver_layer.delete()
-
+    
+    def delete_mapbox_layer(self, layer_name=False):
+        if layer_name != False:
+            with transaction.atomic():
+                mapbox_layer = models.MapLayer.objects.get(name=layer_name)
+                for layerdef in mapbox_layer.layerdefinitions:
```
robgaston: @mradamcox since a single source may be used by many layers, it seems like you may find exceptions as a result of trying to delete sources that you have already deleted if you simply iterate over the layer definitions as you are doing here. 
I think we should instead create a set of unique source names, and then iterate over those and call delete on the models.  something like this might be a starting point (note: untested): 
```py
source_names = set([layerdef['source'] for layerdef in mapbox_layer.layerdefinitions])
for source_name in source_names:
    mapbox_layer_source = models.MapSource.objects.get(name=source_name)
    mapbox_layer_source.delete()
```

mradamcox: Great catch on this, thanks. I'm finding that some layer defs don't actually have a `source` key, only `[u'paint', u'minzoom', u'type', u'id', u'layout']`. Thoughts on that? I assume those can just be skipped because they don't reference anything in the database.

mradamcox: Also, I see that the sources are suffixed with the name of the layer, `[u'contours-Topo', u'hillshading-Topo', u'openmaptiles-Topo']` for example, where Topo is the name of the mapbox layer. This suggests that two different mapbox layers will never share the same source. Is that correct?

mradamcox:
@robgaston added a commit to handle this better, also see my comment above

Issue #2636, opened 2017-10-28
Define and implement improved db user privileges architecture for Arches


adamlodge:
### User Story
In prod settings, it is good practice to create an alternative db user (besides "postgres") through which Arches accesses the database.  However, there is currently no guidance (documentation or a template script) for creating a user with the privileges required to make Arches work.  

On the implementation side, we should think through what exactly these privileges are, define a role that wraps up these privileges, and implement unit tests that ensure that the privileges  assigned to the role keep up with system requirements.

#### Acceptance Criteria
- [ ] Identify required db privileges
- [ ] Create an Arches role and implement required privileges to it
- [ ] Create and (probably by default) use an alternative db user that is associated with the Arches role 
- [ ] Unit Tests passed
- [ ] Documentation (point to postgres documentation on how to associate a user to a role)

#### Definition of Done
An implementor can create a custom db user account, and associate that user with a pre-defined db role that has all appropriate access to the db necessary to execute Arches functionalities.

chiatt:
Hey @benosteen, does your work address any of the criteria that @adamlodge has mentioned in the description above?

benosteen:
The exception handling I've added to the CreateExtension method helps
"paint over" the problem with the `python manage.py migrate` step. It means
that when the first migration operation fails (due to not having the
superuser postgres permission), it logs the error rather than crashing out
of the whole process. It doesn't fix it! It just means that the original
functionality carries on unaffected, but in the case where you are
deploying it without superuser rights, it doesn't foul up. You just have to
make sure the extensions is available. The PR I put in is just step one of
this.

chiatt:
Following #3719 this ticket will be completed by archesproject/arches-docs#115

PR #3682, opened 2018-07-13
Update Packages.py to respect set postgres database name

jmunowitch:
### Types of changes

### Description of Change

Currently, setup_db hard-codes that the postgres database is used to connect. This fix changes that.

### Checklist

### DIFF
--- a/arches/management/commands/packages.py
+++ b/arches/management/commands/packages.py
@@ -762,7 +762,7 @@ class Command(BaseCommand):
 
         truncate_db.create_sqlfile(db_settings, truncate_path)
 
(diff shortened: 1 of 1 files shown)


mradamcox:
Perhaps this is related to https://github.com/archesproject/arches/issues/2636. @jmunowitch, I had kind of thought that the `postgres` user was needed in this case because of some of the truncate commands. Do you know what specific privileges the user _does_ need to run these commands?

jmunowitch:
This needs a bit more vetting. The idea behind this PR was to allow a postgreSQL instance (Specifically a cloud based instance) that doesn't have the postgres user to be able to run setup_db.

PR #4110, opened 2018-09-13
Eamena 27 bulk upload view - improve validation strategy

mradamcox:
This is a big overhaul of the Excelreader command, which splits all of the validation functions into standalone operations. This is paired with a new strategy for chaining ajax calls to each of these validation functions, which simulates an asynchronous user experience, with feedback coming to the UI each time a specific function completes.

Refactoring the Excelreader command allowed for some improvements in efficiency. For example, validating the concepts in a 200 resource input file used to take 6 minutes, and now takes 2 seconds.

This PR does not yet address the load wait a user would experience during the actual import of the data, only during the validation operations.


mradamcox:
whoops, wrong fork, sorry..... :(

Issue #4173, opened 2018-10-09
[Docker] File compression feature should not rearchive when already exists


veuncent:
**Is your feature request related to a problem? Please describe.**
Currently the new static file compression feature in our docker entrypoint does not care whether the archive already exists, thus compressing all static files every time the container starts.

**Describe the solution you'd like**
Archiving should only be done once for each file.

PR #4142, opened 2018-09-26
4141 compress static files in Docker entrypoint

veuncent:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
Adds a step to the Docker entrypoint script in which static files are compressed.

### Issues Solved
Makes requests faster at runtime. 

### Checklist

-   [x] Unit tests pass locally with my changes
~-   [ ] I have added tests that prove my fix is effective or that my feature works~
-   [x] I have added necessary documentation (if appropriate)

### Further comments

This PR is for #4141.

PR #4174, opened 2018-10-09
Revert 4142

veuncent:
REVERT https://github.com/archesproject/arches/pull/4142

For #4173:
Compress files only if archive does not exist.
Run shell command in a more efficient loop.
Add .html to file types to compress.

### Types of changes

### Issues Solved
Prevents entrypoint from compressing all files every time it starts.
#


veuncent:
@jmunowitch  Did I miss anything?

Issue #4186, opened 2018-10-12
Update the JSON-LD export


benosteen:
The JSON-LD export should be semantically valid outside of an Arches context and be modelled in terms of RDF and how RDF handles types of data such as numbers, booleans, resource URIs, multiples and so on. The current export puts all of the 'leaf' node information (Arches Datatype information, serialised in plain JSON) as a literal value for a given node. It also reifies all the basic datatypes (number, boolean, etc) instead of exporting a single triple as expected.

chiatt:
Hey @benosteen, I see this ticket is in progress, but unassigned. Would you happen to be working on it?

benosteen:
Actually, I think that this specific ticket is wrapped up now with the PR
that pushed in branch 4186_somethingsomething. Should be closed now I think.

chiatt:
Thanks @benosteen

PR #4187, opened 2018-10-12
4186 jsonld export

benosteen:
### Types of changes

### Description of Change
Adds a datatype sensitive JSON-LD export, such that each datatype is exported in a way consistent with RDF/Linked Data expectations. 

### Related Issue

Issue #4186 

### Checklist

### Further comments

Issue #4186 highlights the broad issue with the current JSON-LD style export. Fundamentally, the JSON-LD was a second JSON DB serialisation alongside the existing Arches-specific JSON export. What is required is a JSON-LD export (and import) that will interoperate with non-Arches systems, and will expect the data within the datatypes to be represented in a conventional manner.


azaroth42:
Before in merging, should refactor to use app.datatypes.base and .datatypes as regular functions rather than a hash based on the name in the database.

benosteen:
Ah sure. The particular bit of functionality I ported from the original. As
you can see, it should raise an Exception if it is run, but so far we've
not found what configuration of a arches resource/model/etc we need to
export to trigger this code. None of the fixture code triggers it!

Ben

Issue #4202, opened 2018-10-17
Create plugin model and migration


robgaston:
we need a model to store plugins added to Arches
should be similar to widgets table; will additionally need an icon for navigation menu

apeters:
@robgaston to test this it would be nice to have some documentation on how to register a plugin and the form of the json file that needs to be uploaded

apeters:
@robgaston found in another ticket the way to register a new plugin via sql.  It would still be nice for this to be documented.

Meets requirements

PR #4216, opened 2018-10-22
adds plugins to admin, adds plugin commands

robgaston:
re: #4202


arches/management/commands/plugin.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Commands for managing Arches plugins
+
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument('operation', nargs='?')
+
+        parser.add_argument('-s', '--source', action='store', dest='source', default='',
+                            help='Widget json file to be loaded')
```
apeters: This still references the word 'Widget' in the help section, the other argument has this issue as well

PR #4419, opened 2019-01-11
4186 traceback

azaroth42:
* Add traceback module to api to handle printing exceptions.
* remove models. from Graph in post, as imported at top level
* remove the embedding of p2_has_type / classified_as in jsonld writer

### DIFF
--- a/arches/app/datatypes/concept_types.py
+++ b/arches/app/datatypes/concept_types.py
@@ -285,10 +285,11 @@ class ConceptListDataType(BaseConceptDataType):
     def to_rdf(self, edge_info, edge):
         g = Graph()
         c = ConceptDataType()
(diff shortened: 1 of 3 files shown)


robgaston:
@azaroth42 I made a made changes here:
- I moved from sending the traceback in the response body, to putting it in the application log, and put a generic message in the response (so as not to inadvertently expose anything to the client)
- I modified the way we are handling the incoming graph id (or slug) for better model performance; the model that you fixed the code to use is a proxy model that does more work than we need here, so I used the default model (`models.GraphModel`) instead.
- I did a little work to shave off some unnecessary querying from my code.

I'm going to merge this now into `benosteen-4186_jsonld_export` - thanks!

PR #4471, opened 2019-01-25
support for v3 - v4 data migration workflow

mradamcox:
### Types of changes

### Description of Change
This PR contains all of the code to support the v3-v4 migration workflow that is currently documented here during development:  https://arches-dev.readthedocs.io/en/document_v3-v4_workflow/v3-to-v4-migration.html. The intention is for it to be included in the upcoming 4.4 release. As such, those docs should soon be published to https://arches.readthedocs.io/en/latest/v3-to-v4-migration.html as well.

Only three new files are added to core Arches; no old code has been modified. Specifically, two new files in arches/app/utils/ (`v3utils.py` and `v3migration.py`), and a single management command file (`v3.py`) in arches/management/commands.

I have also added a set of tests specifically geared toward testing the v3 management command workflow (as is a specific series of commands that the user must run in sequence, with some manual file editing in between). Basic testing of the actual data import is also included.

### Issues Solved

#

### Checklist

-   [x] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: GCI and Legion GIS
*   Found by: @ 
*   Tested by: @mradamcox 
*   Designed by: @mradamcox


mradamcox:
One note on the tests I added: There is a new setting in test_settings.py called `LOAD_V3_DATA_DURING_TESTS` which should remain set to False. In a developer's `settings_local.py` file, alongside `test_settings.py`, they can set this variable to `True` to enable a more in depth set of tests which will actually load some sample v3 data (not just convert it). This loading adds about a minute to the tests, so I figured it would be best not to run it in Travis CI all the time.

mradamcox:
Merging now, as per discussion in Farallon standup meeting, Feb. 19th.

Issue #4456, opened 2019-01-18
Default Search button location is unintuitive


adamlodge:
Rich made the point that after making edits to his resource models, he wanted to go back to main map screen (the search page) to do more dispaly and editing.  It was not at all obvious to him that the magifying glass on the upper left was the button to push.  Instead, he was looking in the panel on the left where all of the other tools, plugins, etc... are located.  

I would propose adding a way to access the search button on the left panel.

azaroth42:
I have frequently wondered how to get to search before remembering that it's the magnifying glass in the top right. So :+1: to the issue.

adamlodge:
meets reqt

PR #4496, opened 2019-01-31
4456 moving search button

whatisgalen:
### Description of Change
Adds second search icon to left nav bar re #4456

### DIFF
--- a/arches/app/templates/base-manager.htm
+++ b/arches/app/templates/base-manager.htm
@@ -245,26 +245,15 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
                                         </li>
                                         {% endif %}
 
-                                        
-                                        <li>
-                                            <a href="{% url 'search_home' %}" class="ep-tools ep-tools-right" data-bind="click:function () { navigate('{% url 'search_home' %}') }">
-                                                <div class="" data-placement="bottom" data-toggle="" data-original-title='{% trans "Search" %}'>
-                                                    <i class="ion-search"></i>
-                                                    <span class="menu-title">
-                                                            <strong>{% trans "Search" %}</strong>
-                                                        </span>
-                                                </div>
+                                        <li{% if "views/search" in main_script %} class="active-sub"{% endif %}>
+                                            <a href="{% url 'search_home' %}" data-bind="click:function () { navigate('{% url 'search_home' %}') }">
+                                                <i class="fa fa-search"></i>
+                                                <span class="menu-title">
+                                                    <strong>{% trans "Search" %}</strong>
+                                                </span>
                                             </a>
-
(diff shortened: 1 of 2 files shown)


arches/app/templates/base-manager.htm
```diff
@@ -124,6 +124,7 @@ <h5 class="hidden-xs">{% trans "Welcome" %}, {{ user.first_name|default:user.use
                         <a href="{% url 'search_home' %}" class="ep-tools ep-tools-right" data-bind="click:function () { navigate('{% url 'search_home' %}') }">
                             <div class="" data-placement="bottom" data-toggle="tooltip" data-original-title='{% trans "Search" %}'>
                               <i class="ion-search"></i>
+
```
robgaston: was there a reason to add this line?

arches/app/templates/base-manager.htm
```diff
@@ -244,6 +245,25 @@ <h5>{% trans "for more documentation, visit" %} <a href="https://arches.readthed
                                         </li>
                                         {% endif %}
 
+
```
robgaston: I'd say let's reduce this to a single linefeed between each `<li>`

arches/app/templates/base-manager.htm
```diff
@@ -244,6 +245,25 @@ <h5>{% trans "for more documentation, visit" %} <a href="https://arches.readthed
                                         </li>
                                         {% endif %}
 
+                                        
+                                        <li>
```
robgaston: we need to use the `active-sub` class here when the user is on the search page, by checking the `main-script` like so:
```
{% if "views/search" in main_script %} class="active-sub"{% endif %}
```

arches/app/templates/base-manager.htm
```diff
@@ -244,6 +245,25 @@ <h5>{% trans "for more documentation, visit" %} <a href="https://arches.readthed
...
+                                            <a href="{% url 'search_home' %}" class="ep-tools ep-tools-right" data-bind="click:function () { navigate('{% url 'search_home' %}') }">
```
robgaston: the classes added here (`class="ep-tools ep-tools-right"`) are causing some rendering issues - I think you should just remove them

arches/app/templates/base-manager.htm
```diff
@@ -244,6 +245,25 @@ <h5>{% trans "for more documentation, visit" %} <a href="https://arches.readthed
...
+                                                <div class="" data-placement="bottom" data-toggle="" data-original-title='{% trans "Search" %}'>
+                                                    <i class="ion-search"></i>
```
robgaston: instead of using the `ion-search` icon that we use in the header, lets use `fa fa-search` to achieve consistency amongst the glyphs in the left nav bar.

PR #4509, opened 2019-02-04
sets download limit for each resource, fixes bug in MSM

### LINKED ISSUE #4448: Mobile Survey Manager download limit should apply to each model
Currently, the MSM allows users to define the number of instances to download as part of the survey.  However, when multiple models are configured as part of the survey, its quite possible for just instances of a single model to reach the download limit.

To insure that instances of all models in the survey are downloadable, the limit should be applied to all models in the survey not just the first arbitrarily chosen model

whatisgalen:
### Description of Change
Fixes #4448 by iterating through resource types to download up to the limit for each type.

### DIFF
--- a/arches/app/models/mobile_survey.py
+++ b/arches/app/models/mobile_survey.py
@@ -343,9 +343,12 @@ class MobileSurvey(models.MobileSurveyModel):
         resource instances relevant to a mobile survey. Takes a user object which
         is required for search.
         """
-        query = self.datadownloadconfig['custom']
+        query = self.datadownloadconfig['custom'] 
         resource_types = self.datadownloadconfig['resources']
-        instances = {}
+        from pprint import pprint as pp
+        
+        all_instances = {}
+
         if query in ('', None) and len(resource_types) == 0:
             print "No resources or data query defined"
         else:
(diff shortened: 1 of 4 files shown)


robgaston:
@chiatt could you please take an initial look at this PR when you get a chance?  thanks!

arches/app/models/mobile_survey.py
```diff
@@ -369,19 +372,29 @@ def collect_resource_instances_for_couch(self):
             search_res_json = search.search_results(request)
             search_res = JSONDeserializer().deserialize(search_res_json.content)
             try:
-                instances = {hit['_source']['resourceinstanceid']: hit['_source'] for hit in search_res['results']['hits']['hits']}
-                # if we didn't get our limit of resource instances using a spatial filter
-                # let's try to get resource instances that don't have spatial data
-                if len(instances.keys()) < int(self.datadownloadconfig['count']):
-                    request.GET['mapFilter'] = '{}'
-                    request.GET['resourcecount'] = int(self.datadownloadconfig['count']) - len(instances.keys())
+                for res_type in resource_types:
```
chiatt: It looks like iteration is taking place only for non-spatial searches. The iteration block also needs to include the spatial search as well.

PR #4527, opened 2019-02-07
ensures that spatial instances are prioritized, DL limit applied to each resource model

whatisgalen:
### Description of Change
Re: #4448 fixes download limit to apply to each resource model included in survey; incorporates requested changes to correctly prioritize spatially filtered resource instances before non-spatial.


chiatt:
@whatisgalen I made a couple of minor adjustments, but this is great. Thanks!

Issue #4535, opened 2019-02-08
Enhancement: Allow user with read-only privs on higher cards to create/edit tiles deeper in graph


adamlodge:
I have a use case where user1 will create tiles at a higher level in the graph and user2 will create tiles that are deeper in the graph.  User2 has read access to the higher level tiles created by user1.  At present, the resource editor tree disallows user2 from accessing the (previously-generated) tiles and cards necessary to create the deeper tiles for which he has read-write access. 

We should relax the resource editor tree to allow user2 to "read" the tiles created by user1 so that he (user2) can access the cards necessary to add/edit the tiles he is responsible for.

Below is a screenshot example of permissions for a problematic graph:
![Screen Shot 2019-02-07 at 5.53.33 PM.png](https://images.zenhubusercontent.com/56f150f76caf56c9d38e16f8/4a35f8fc-05bf-435a-b103-9537a33e85ae)

adamlodge:
meets requirement

PR #4543, opened 2019-02-12
4535 galen make child cards editable below ro parent

whatisgalen:
### Description of Change

re: #4535 
++ enables users who lack Write permission for a given (parent-) card to still access/edit child-cards for which they *do* have Write permission
++ adds a conditional behavior to the UI so that buttons in edit/add resource interface are disabled if a user lacks Write permission for any particular card at any level, also adds a banner at the top of the UI to inform user they lack that card-specific permission


whatisgalen:
@robgaston Looks like I can't assign it from here :( might be related to my permission-level

robgaston:
@whatisgalen this looks good to me...
@chiatt can you please take a look at this and provide your input, esp. re: provisional tiles/edits?

chiatt:
@whatisgalen Could you disable widgets that a user does not have permissions to edit?

Issue #4350, opened 2018-12-12
Enhancement: Add ability for end-users to roll up/make invisible downstream portions of a given report


adamlodge:
This will allow for creation of reports that exclude data that is irrelevant to the business problem that an end-user is working on.  It would also be cool for the roll-up details to be encoded in the url so it is easy to revisit the report without having to re-define the roll up sections.

adamlodge:
Could you enhance this to also support rolling up individual tiles?... Particularly in cases where there is a cardinality of n tiles.

adamlodge:
meets reqt

PR #4581, opened 2019-02-16
4350 galen report ui border

whatisgalen:
### Description of Change
re: #4350 (update of #4514 , without the other branch that got accidentally mixed in with these changes)
-- adds border and light background to tiles in the report view when expanded iff the tile has no children

soliciting @adamlodge @robgaston for input on further changes


arches/app/templates/views/components/cards/default.htm
```diff
@@ -393,7 +393,7 @@ <h3 class="panel-title help-panel-title" style="">
             <i class="fa report-expander print-hide" data-bind="css: {'fa-angle-down': reportExpanded(), 'fa-angle-right': !reportExpanded()}, click: function () { reportExpanded(!reportExpanded()) }"></i>
             <!-- ko if: reportExpanded() -->
             <!-- ko foreach: { data: self.preview ? [card.newTile] : card.tiles, as: 'tile' } -->
-                <div class="rp-card-section" data-bind="css: {'provisional': tile.provisionaledits() !== null && tile.userisreviewer === false, 'fullyprovisional': tile.isfullyprovisional()}">
```
robgaston: @whatisgalen yeah, this seems a little heavy handed to me... maybe we should remove the borders & background color and just use an `<hr>` between every tile in the list... add something like this here...
```html
<!-- ko if: $index() !== 0 --><hr><!-- /ko -->
```
(this would also mean we could remove the other additions from this PR, but we might think about styling the `hr` so that it looks nice & is obvious, but not heavy-handed)

whatisgalen: so I adjusted the border to just be bottom; you'll see a thick left-border commented out in my css. I think it's worth uncommenting just to see what it looks like. I erred on 'less is more', but interested in other opinions. For some reason I couldn't get an <hr> to show (used your snippet as a guide), so I fell back on just using border-bottom.
@robgaston @adamlodge

robgaston: @whatisgalen i think that `childless` is not really how we want to distinguish data here... that is, we want to create some distinction between tiles of the same card. we want to be able to create this distinction regardless of whether they have child cards or not. this is why i think that checking the tile index is a more scalable solution. I would like to see the styling that you have proposed applied in a way that doesn't respect the length of `cards`, but rather the number of tiles in the report (ie `$index` in my example or similar)

whatisgalen: Sounds good, I'll go ahead and change so that it's looking at index

arches/app/media/css/arches.css
```diff
@@ -7703,6 +7703,21 @@ a.mega-dropdown-toggle.disabled {
     padding-left: 15px;
 }
 
+/* .childlessborder {
```
robgaston: @whatisgalen can you please remove this commented out code, as it seems to be cruft

arches/app/media/js/viewmodels/card.js
```diff
@@ -332,6 +332,9 @@ define([
             }
         });
 
+        // this.childless = ko.pureComputed(function() {
```
robgaston: @whatisgalen can you please remove this commented out code, as it seems to be cruft

arches/app/templates/views/components/cards/default.htm
```diff
@@ -428,13 +430,15 @@ <h3 class="panel-title help-panel-title" style="">
                                     params: {
                                         state: 'report',
                                         preview: self.preview,
-                                        card: card
+                                        card: card,
+                                        showHeaderLine: $index !== 0
```
robgaston: @whatisgalen I think that the condition here should look something like this (as per your previous commit) so that there is also an hr between the tile data itself and the first card (if the tile itself has data):
```js
self.card.model.get("widgets")().length > 0 || $index() !== 0 
```

Issue #4707, opened 2019-03-21
`python manage.py migrate` requires postgres superuser account


benosteen:
**Describe the bug**

(Working towards a solution to #2636.)

From the Postgres documentation:

> Loading an extension requires the same privileges that would be required to create its component objects. For most extensions this means superuser or database owner privileges are needed. The user who runs CREATE EXTENSION becomes the owner of the extension for purposes of later privilege checks, as well as the owner of any objects created by the extension's script

When deploying Arches, especially with shared database provisioning (eg Amazon RDS), it is poor practice to run an application with superuser-level permissions. 

For deployment, it is better for the insufficient permissions exception to be caught and logged, with the deployment process handling anything that requires elevated permissions in a more controllable, inspectable method (eg via a .sql file, rather than relying on running a command with the Arches codebase).

**Desired Behavior**

Catch any `django.db.Error` Exceptions thrown by the CreateExtension migration operation. This will allow `python manage.py migrate` to complete successfully with a normal user with logged warnings, and assume that the deployment process correctly adds in the required extensions to the DB prior to running the application. (For example, the extension could be added to the `public` schema prior to spinning up the Arches instance.)

When run with a db superuser account, the migrate functionality should remain as it is.

mradamcox:
@benosteen What do you think is the best way forward for this ticket? With your fix, the migration command will not error out if run by a non-superuser so as you say the issue is kind of solved.

If we want to go further, I was thinking we could more clearly document the postgis template creation, and advise that users also install the `uuid-ossp` extension into it. That would mean that the Arches database (when made from the postgis template which is how the new `setup_db` command works) will have both of those extensions and no CreateExtensions would be needed, so they could be fully removed from the migrations. Going that route, it may also be good to check for these extensions prior to migrations being run, and cancel the operation if they are not present...

benosteen:
I'm firmly in the camp that any extraneous one-off db set up (eg steps
beyond creating tables and functions an app needs) should not be done by a
code regularly run within the app's environment. Stand-alone script for
setup, divorced from the app!

I would greatly prefer the migrate to check to see if the template is
correct and present, or hard fail otherwise. To put it into
container/deployment context, the Arches container is at best 1.5GB in
size. If all you want as a sys/devops, is to get some DB clusters set up
and ready to deploy Arches instances to when you need (asynchronously),
having to pull a large container and run a fully configured multi-layered
environment (cluster env -> container env -> python env -> arches
[...]

PR #4708, opened 2019-03-21
Catching db errors in the CreateExtension class re #4707

benosteen:
### Types of changes

### Description of Change

Catching `django.db.Error`s thrown when the CreateExtension operation is run with a non-superuser postgres user. A warning is logged, but it is intended that the extension will be made available to the non-superuser postgres user through another means (eg template inclusion, or via the `public` schema.)

### Issues Solved

#4707

### Checklist


robgaston:
Looks good!

Issue #4596, opened 2019-02-20
Indicate in the graph designer and resource editor which cards participate in the descriptor function


chiatt:
It would be great if there were some indication as to which cards participate in one or more descriptor functions as these are often key tiles that a user wants to be sure to create.

chiatt:
This would be useful in the Collector manager to help users select the descriptor cards for a mobile project. If we update the resource instance widget to allow the creation of new instances, it would also help a user in the widget configuration to identify those cards that need to be completed in order to recognize a newly created instance.

dwuthrich:
meets requirements

PR #4788, opened 2019-04-19
re #4596 gets the function data filtered by graph via the view, compu…

whatisgalen:
…tes whether a given node or card is associated with the applied function in viewmodels for the various templates in graph designer and resource editor; adds a css class.

### Description of Change

re #4596 checks nodes and cards (in graph designer) and cards (resource editor) against whether the arches "out of the box" descriptor function has been applied for that resource model, identifies via inline h4 tags and tree node styling which cards/nodes are associated with the descriptor function.

### DIFF
--- a/arches/app/media/js/viewmodels/card.js
+++ b/arches/app/media/js/viewmodels/card.js
@@ -332,15 +332,21 @@ define([
                 });
             },
             isFuncNode: function() {
-                var appFuncDesc, appFuncName, nodegroupId;
+                var appFuncDesc = false, appFuncName = false, nodegroupId = null;
                 if(params.appliedFunctions && params.card) {
-                    appFuncDesc = self.appliedFunctions[0]['config']['description']['nodegroup_id'];
-                    appFuncName = self.appliedFunctions[0]['config']['name']['nodegroup_id'];
-                    nodegroupId = params.card.nodegroup_id;
-                    if(nodegroupId == appFuncDesc) {
-                        return "* This card data will show as the resource description.";
-                    } else if(nodegroupId == appFuncName) {
-                        return "* This card data will show as the resource name.";
+                    for(var i=0; i < self.appliedFunctions.length; i++) {
+                        if(self.appliedFunctions[i]['config']['description']['nodegroup_id']) {
+                            appFuncDesc = self.appliedFunctions[i]['config']['description']['nodegroup_id'];
+                        }
+                        if(self.appliedFunctions[i]['config']['name']['nodegroup_id']) {
+                            appFuncName = self.appliedFunctions[i]['config']['name']['nodegroup_id'];
+                        }
+                        nodegroupId = params.card.nodegroup_id;
+                        if(nodegroupId === appFuncDesc) {
+                            return "* This card data will show as the resource description.";
+                        } else if(nodegroupId === appFuncName) {
+                            return "* This card data will show as the resource name.";
+                        }
                     }
                 }
                 return false;
--- a/arches/app/media/js/viewmodels/card.js
+++ b/arches/app/media/js/viewmodels/card.js
@@ -71,6 +71,7 @@ define([
         var multiselect = params.multiselect || false;
         var isWritable = params.card.is_writable || false;
         var selection;
+        var appliedFunctions = params.appliedFunctions;
         if (params.multiselect) {
             selection = params.selection || ko.observableArray([]);
         } else {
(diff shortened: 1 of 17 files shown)


arches/app/media/js/viewmodels/card.js
```diff
@@ -328,6 +330,20 @@ define([
                     loading: loading,
                     cardwidgets: params.cardwidgets,
                 });
+            },
+            isFuncNode: function() {
+                var appFuncDesc, appFuncName, nodegroupId;
+                if(params.appliedFunctions && params.card) {
+                    appFuncDesc = self.appliedFunctions[0]['config']['description']['nodegroup_id'];
```
robgaston: @whatisgalen this will result in errors if any of the properties referenced (`self.appliedFunctions[0]['config']['description']['nodegroup_id']`) are undefined. as such, we should probably first check for their presence in the conditional at line 336. Also, this always just checks the first function in the array.  We should probably iterate over `self.appliedFunctions` instead.

arches/app/media/js/views/graph/graph-designer/card-tree.js
```diff
@@ -370,7 +372,21 @@ define([
...
             },
             selection: selection,
-            filter: filter
+            filter: filter,
+            isFuncNode: function() {
+                var appFuncDesc, appFuncName, nodegroupId;
+                if(params.card) {
+                    nodegroupId = params.card.nodegroup_id;
+                    if(this.appliedFunctions().length > 0) {
+                        appFuncDesc = this.appliedFunctions()[0]['config']['description']['nodegroup_id'];
```
robgaston: this section has the same problems as described above re: `isFuncNode` in card.js,  ie keying on properties that may not be present and assuming the function of interest is at index 0 of the array.

arches/app/media/js/views/graph/graph-designer/graph-tree.js
```diff
@@ -85,6 +86,30 @@ define([
...
         },
 
+        /**
+         * Returns a boolean to indicate whether this node participates in descriptor function
+         * @param {object} node - a node in the tree 
+         */
+        isFuncNode: function(node) {
+            var appFuncDesc, appFuncName;
+            if(this.appliedFunctions().length > 0) {
+                appFuncDesc = this.appliedFunctions()[0]['config']['description']['nodegroup_id'];
```
robgaston: this section has the same problems as described above re: isFuncNode in card.js, ie keying on properties that may not be present and assuming the function of interest is at index 0 of the array.

arches/app/media/js/views/graph/graph-designer/node-form.js
```diff
@@ -33,6 +33,29 @@ define([
                 return self.graphModel.get('isresource') && node && node.istopnode;
             });
             this.restrictedNodegroups = options.restrictedNodegroups;
+            this.appliedFunctions = options.appliedFunctions;
+
+            this.isFuncNode = function() {
+                var node = self.node();
+                var appFuncDesc = this.appliedFunctions()[0]['config']['description']['nodegroup_id'];
```
robgaston: this section has the same problems as described above re: isFuncNode in card.js, ie keying on properties that may not be present and assuming the function of interest is at index 0 of the array.

arches/app/media/js/views/resource/new-editor.js
```diff
@@ -145,6 +147,27 @@ define([
         reviewer: data.userisreviewer,
         graphiconclass: data.graphiconclass,
         relationship_types: data.relationship_types,
+        appliedFunctions: appliedFunctions(),
+        isFuncNode: function() {
+            var node = false;
+            var appFuncDesc, appFuncName;
+            if(this.appliedFunctions.length > 0) {
+                appFuncDesc = this.appliedFunctions[0]['config']['description']['nodegroup_id'];
```
robgaston: this section has the same problems as described above re: isFuncNode in card.js, ie keying on properties that may not be present and assuming the function of interest is at index 0 of the array.

robgaston: Does this code really need to be duplicated in so many places? can't we just always use `isFuncNode` on the card viewmodel?

whatisgalen: it looks like isFuncNode in new-editor.js wasn't actually being used, so I nixed that one. Regarding just having one isFuncNode method in the card viewmodel and calling to that from the other viewmodels, I tried to assign something like a "this.isFuncNode: CardViewModel.isFuncNode" (with and w/o parentheses) and get "not a function" errors. The other thing I wasn't sure about was the fact that I'm relying on different inputs like `params`,`self`, and `node` and it wasn't clear how to reconcile that if I was just going to export that one function to the other viewmodels.

Issue #4800, opened 2019-04-24
Docker Hub docker-compose.test.yml error


benosteen:
**Describe the bug**
The build host that Docker Hub uses to test docker images runs with a setting that newer versions of Elasticsearch fail over when they run their bootstrap checks.

**To Reproduce**
Trigger an automated build on Docker Hub. The docker-compose-test.yml stack will attempt to run the test suite but will fail as the Elasticsearch container will exit on boot and become unavailable.

**Expected behavior**
The test results from a build on Docker Hub should be the same as if the test results are run locally.

Issue #2746, opened 2017-11-15
ubuntu xenial not properly prepared for elasticsearch with install script


mradamcox:
### User Story
currently, neither https://github.com/archesproject/arches/blob/master/arches/install/ubuntu_xenial_setup.sh nor the `python manage.py packages -o es install` modify the `vm.max_map_count` setting. This used to happen when the installation included the rpm and debian packages (up to this commit https://github.com/archesproject/arches/commit/14c9233b35d50bb140a551c41b48fb791bd00112#diff-585893c6dce669430322947316a9f89bL43).

Because this setting modification is OS-specific, it should be taken care of in the ubuntu install script, no in the manage.py command.

I think, though don't have the time to try right now, simply adding this would do it, but I don't know if it's the best approach:

`sudo echo vm.max_map_count=262144 > /etc/sysctl.d/99-elasticsearch.conf`

chiatt:
It might be good to take care of this issue when we update the install script with the latest version of ES. re #2691

mradamcox:
This seems to be taken care of now with ES7 on master branch. Closing.

PR #4801, opened 2019-04-24
Adding env var to ES to stop bootstrap checks

benosteen:
### Types of changes

### Description of Change
It causes Elasticsearch to run in a developmental mode, causing some preflight checks to become logged warnings, rather than critical errors. This is to avoid the Elasticsearch service from closing pre-emptively on Docker Hub, as the test host DH uses may have the vm.max_map_count set too low, or a related soft error for a test instance.

### Issues Solved

#4800 

### Checklist

### DIFF
--- a/docker-compose-test.yml
+++ b/docker-compose-test.yml
@@ -51,6 +51,7 @@ services:
         - "9200:9200"
         - "9300:9300"
       environment:
+        - discovery.type=single-node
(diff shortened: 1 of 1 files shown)


mradamcox:
Seems that this is related to https://github.com/archesproject/arches/issues/2746

PR #4830, opened 2019-05-10
Passes url params from the plugin view to the plugin component. 

chiatt:
re #4821


chiatt:
Decided on front-end solution after discussion with @robgaston

Issue #4803, opened 2019-04-26
Remove psql from Arches code


mradamcox:
**Is your feature request related to a problem? Please describe.**
The problem is that there is no need to have `psql` installed as django can handle all of the postgres connections that the app needs. Requiring `psql` as Arches currently does is an unnecessary hindrance to deployment. For example, if you want to deploy Arches and use a postgres installation on a separate server, you would still need to have psql (which is usually installed with postgres) installed on the Arches server anyway.

**Describe the solution you'd like**
Anywhere that psql is explicitly called, Arches should use built-in Django utilities instead, or psycopg2.

**Describe alternatives you've considered**
The alternative is to leave psql in the code and continue to require its installation.

**Additional context**
Typically, this is not an issue because psql is installed along with postgres. However, coding psql into Arches therefore assumes that postgres has been installed on the server as Arches is running on, which is a very naive deployment architecture.

mradamcox:
This is now done.

Issue #3719, opened 2018-07-19
Running setup_db expects PostgreSQL superuser


jmunowitch:
**Describe the bug**

When running setup_db on a non superuser, Arches fails to clean the DB before continuing:

https://github.com/archesproject/arches/blob/master/arches/db/install/truncate_db.py#L21
Using `DROP` and `CREATE` for a database requires a superuser.

On a multi-tenancy environment (multiple instances of arches for one PostgreSQL instance), having a superuser's credentials on each arches instance is not best practice.

Changing these commands to ones that do not require a superuser would fix this. Dropping every table in the DB may work.

**Expected behavior**
The ability to run setup_db without having superuser PostgreSQL credentials

**Your Arches Information**

*   Version used: 4.3.x,

**Ticket Background**
*   Found by: @jmunowitch

jmunowitch:
Related to https://github.com/archesproject/arches/issues/2636

mradamcox:
That command actually already does use the Arches-defined username, however, it tries to connect to the `postgres` database, so that may actually be the problem. Not to say that superuser status isn't required by Arches (I believe it is, and probably shouldn't be) but another issue with this command is that it's mixing a username and a database name.

mradamcox:
This is now fixed, in `master` and in `stable/4.4.x`.

PR #4833, opened 2019-05-14
3719 setup db run without superuser

mradamcox:
### Types of changes

### Description of Change
This PR creates a new management command

    python manage.py setup_db

which closely mimics the existing

    python manage.py packages -o setup_db

in behavior, but functions differently underneath. In this new command, a first attempt connection is made to postgres using username and password credentials stored in new settings `PG_SUPERUSER` and `PG_SUPERUSER_PW`. If these settings don't exist or are blank (which is their new default state) then the DATABASES['default'] username and password are used as before.

Next a check is made to test whether the connected user is a PG superuser or not. If the user is a superuser, drop and create database actions are taken and then migrations run just like before. However, if the user is not a superuser, another suite of commands are run which will fully reset the database but do not require superuser status. Specifically, `flush` (to remove the admin and other non-Arches objects from the db), unapply all Arches migrations, and finally drop all Arches models tables.

The result is that the Arches database user need not be a postgres superuser which was previously a requirement. (As mentioned in various tickets, this previous requirement was at best non-standard and at worst a security risk.)

Of course, the very first time the database is created it must be done by a superuser, but after that a normal postgres user should be sufficient for not only normal use but also for development purposes.

### Issues Solved
solves #3719

good progress toward #2636 and #4803

and builds on #4707 

### Checklist

#### Ticket Background
*   Sponsored by: GCI
*   Found by: @mradamcox @benosteen 
*   Tested by: @mradamcox 
*   Designed by: @mradamcox @benosteen 

### Further comments

As for documentation, this has led me to believe that we need a separate "Postgres prep" section which should have some snippets that users can copy/paste to prepare their Postgis template better. Previously that template was not used in the setup_db CREATE DATABASE command, but it is now. This means that if it had the `uuid-ossp` extension in it (in addition to the postgis extension of course) then the CreateExtension operations in the migrations (the subject of #4707) would be unnecessary altogether.


mradamcox:
I should clearly note here that after this PR, the original `python manage.py packages -o setup_db` command will display a deprecation warning in the console which directs users to use the new command and then exits. I have also set up the travis deployment scripts so that all passes, but have not touched any docker stuff.

mradamcox:
This is the output right now if someone uses the old setup_db command:

```
(ENV) C:\arches\my_project>python manage.py packages -o setup_db
operation: setup_db

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Warning: This command is now deprecated. From now on please use

    python manage.py setup_db [--force]

The --force argument will suppress the interactive confirmation prompt.
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

(ENV) C:\arches\my_project>
```

Are you saying that you would prefer one of the following: either this old command should be removed altogether and cause an error (it would be something like "invalid choice: setup_db"), or this deprecation message should include a date of removal and then continue to work?

I think I personally like the latter, and we could say the command will be removed in 4.5.x... That way I think these commits could safely be added to 4.4.x and released with 4.4.2.

Once the old command is fully removed, I think it will be the right time to remove a few other obsolete things from the codebase like the whole truncate_db.py situation, and a few psql-related items.

mradamcox:
Alright, I've made that correction and then did a little more work on other print statements. And fixed an issue I found with how I'd written the pg_terminate_backend command. I also removed all of the code that is now unnecessary, which ultimately led to removing the check for `psql` because that client is technically not necessary anymore.

I think it would be wise to build in a more explicit check for postgres when later work is done on this or related setup commands, but ultimately the check should be whether Arches has the ability to connect to postgres, not whether it is installed locally.

Issue #4664, opened 2019-03-06
Update resource instance widget to support resource instance creation


chiatt:
Currently the resource instance widget only allows a user to select instances that have already been saved in arches prior to visiting the page with the resource instance widget. The widget should instead allow user to create a new instance and then relate that instance to the respective node.

apeters:
The resource instance widget should have a way to "add a new instance" inline (as an entry in the dropdown, or something similar).  The ui would then present a key card (defined by a config in the widget) for the resource that can be used to add an entry in the resource instance dropdown.

chiatt:
While working on this widget, keep an eye out for this issue: globaldigitalheritage/arches-3d#173

robgaston:
@ryan86 I made some review comments on #4844 which need follow up.  thanks!

dwuthrich:
meets requirements

PR #4844, opened 2019-05-22
Add ability to create related resource in related resource card, re #…

ryan86:
…4664

### Types of changes

### Description of Change
Adds ability to create a new resource from the related resource dropdown in the related resource instance widget.

### Issues Solved

#4664 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/viewmodels/resource-instance-select.js
+++ b/arches/app/media/js/viewmodels/resource-instance-select.js
@@ -11,7 +11,7 @@ define([
         params.configKeys = ['placeholder'];
         this.multiple = params.multiple || false;
         this.value = params.value || undefined;
-        this.disable = params.disable || function(){return false};
+        this.disable = params.disable || function(){return false;};
         this.disableMessage = params.disableMessage || '';
         this.showCard = ko.observable(false);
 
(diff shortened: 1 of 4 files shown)


arches/app/templates/views/components/widgets/resource-instance-select.htm
```diff
@@ -20,6 +20,14 @@
         </div>
     </div>
 </div>
+<!-- ko if: newTileStep -->
+<div style="position: absolute; background: white;" data-bind="with: newTileStep, visible: showCard">
```
robgaston: @ryan86 can you please use classes rather than inline styles?

Also, why use `showCard` here? couldn't you instead just use `newTileStep` and set it to null when the user is done adding the new resource?

arches/app/media/js/viewmodels/resource-instance-select.js
```diff
@@ -5,19 +5,22 @@ define([
     'arches',
 ], function(ko, $, WidgetViewModel, arches) {
     var nameLookup = {};
+    require(['views/components/workflows/new-tile-step']);
```
robgaston: @ryan86 why are you requiring this module here?  require is asynchronous, so this seems dangerous.  I would just add this dependency to the list of dependencies at the top of this script.

arches/app/templates/views/components/widgets/resource-instance-select.htm
ryan86: Inline styles and references to showCard have been removed.

Issue #4879, opened 2019-06-04
package.json changes breaks Docker container build


benosteen:
**Describe the bug**
Moving package.json and removing other related files has caused the Docker build to break.

**To Reproduce**
1. Get Arches master branch
2. `docker image build .`
3. First of likely a number of related errors: `COPY failed: stat /var/lib/docker/tmp/docker-builder824149968/package.json: no such file or directory`

**Expected behavior**
`docker build .` should complete, and for the container to perform as it did before the change.

chiatt:
You resolved this in the latest, right @benosteen?

PR #4882, opened 2019-06-04
Dockerfile and docker-compose config updated due to package.json location changes

benosteen:
### Types of changes

### Description of Change
Updated the Dockerfile and entrypoint.sh script to reference and use the new location for the `.yarnrc` and `package.json` files for `yarn install` and related calls. Also, updates the `docker-compose.yml` file to include the postgresql setup script required to set up the database template and extensions for Arches projects, as the new setup script no longer attempts to perform this super user action

### Issues Solved

#4879 

### Checklist


Dockerfile
```diff
@@ -130,9 +130,12 @@ RUN . ../ENV/bin/activate \
     && pip install -e . --no-binary :all:
 
 # Install Yarn components
-COPY ./package.json ${ARCHES_ROOT}/package.json
```
veuncent: This copy command was initially used to run yarn install before copying the complete code base into the container (COPY . ${ARCHES_ROOT}). The advantage of that is you don't have to run yarn install every time you make little changes to the code base (unless package.json changes).

docker-compose.yml
```diff
@@ -68,6 +68,7 @@ services:
       volumes:
           - postgres-data:/var/lib/postgresql/data
           - postgres-log:/var/log/postgresql
+          - ./arches/install/init.sql:/docker-entrypoint-initdb.d/init.sql # to set up the DB template
```
veuncent: Not sure if this is the right solution, but I may be missing something (it's late here). What if you have a database running in aws? Would you need to manually run this script on that db before you can let the Arches container run setup_db? Could this script be incorporated in setup_db instead?

arches/install/init.sql
```diff
@@ -0,0 +1,11 @@
+CREATE DATABASE template_postgis_20 WITH ENCODING 'UTF8' LC_COLLATE='en_US.utf8' LC_CTYPE='en_US.utf8';
```
mradamcox: When I tried this on windows, it failed saying that the locale name was invalid. Turns out Windows encoding is completely different `English_United States.1252`. So I recommend leaving this file as is, but renaming it something like `init-unix.sql`.

docker-compose.yml
benosteen: This is letting the postgres container set up the template that Arches projects use to base themselves on. If the script was in setup_db, then the Arches container would have to have superuser postgresql credentials to be able to add the two extensions, which is something that should only happen once. For context, by IT policy, no app will ever get that sort of level of access to the provisioned DB. Having the required template be installed as part of deployment/provisioning, rather than when an app is being set up is not a big change, but is pretty crucial in a shared environment.

A fallback could be the entrypoint.sh looking for superuser creds in the environment variables, and running the sql script if they are present?

benosteen: This addition to docker-compose is to allow for local deployment with pg in a 'nearby' container under the control of docker-compose.

Dockerfile
benosteen: There has been some changes to where and how this is done in master. https://github.com/archesproject/arches/pull/4878 which broke the automated setup in docker and the entrypoint. The line is not necessary, as package.json is now in the arches/install directory and copied over in a previous command, along with the yarnrc, etc all within the install dir again.

docker-compose.yml
veuncent: Got it. Not forcing the use of superuser credentials seems a good move to me.

I'm still not entirely clear on how the deployment process will go on a non-local environment. As far as I can remember, currently when running Arches in a clean environment with an empty, newly provisioned Postgres db, all the user needs to do is run the Arches container and the db will be setup in a completely automated way. Would we lose this convenience with this new Postgis extensions script? 

I do think your idea of looking for superuser credentials and running the script sounds good, so that at least we have a simpler (yet less secure) way to get people started. Could that be incorporated into this PR?

By the way, I do not see any directions regarding this provisioning script in the regular installation manual:
https://arches.readthedocs.io/en/stable/installation/
Should this be mentioned there?

Dockerfile
veuncent: What I meant to say was that in the current order of commands it seems that every time a developer makes a change to any file in the `${ARCHES_ROOT}` folder, the Docker build will have to run the `yarn install` (and also `pip install`) command. This should not be necessary of you run those before doing `COPY . ${ARCHES_ROOT}`. Those steps will then be cached and reused the next time you build the image. 

(And for that to work you need to copy `package.json` separately from `COPY . ${ARCHES_ROOT}` ).

benosteen: So add a copy of package.json, yarn.lock, and .yarnrc to the /arches/install dir and run `yarn install` before the `COPY . ${ARCHES_ROOT}` step?

docker-compose.yml
mradamcox: Hi @veuncent , yes, it should definitely be mentioned there. I actually have a whole lot of install-related docs tickets grouped together in a project because I think a big overhaul of that process is warranted at this point. https://github.com/archesproject/arches-docs/projects/1

However, this change is not deployed yet in any published release, so if anything it should just be in the `latest` docs, not the `master` ones.

Dockerfile
veuncent: Exactly!

benosteen: I've just pushed the change to move the yarn install higher, to help it be cached as discussed.

benosteen:
@veuncent do you have any issue with this PR? It might be sensible to get the docker run and install steps working again as it is currently broken, and then addressing any further issue with a new issue/PR?

chiatt:
I'm going to merge this PR and as @benosteen mentioned, we can write a new issue for any additional changes.

veuncent:
Sorry, I was on holiday. I will look at the changes asap, @benosteen

PR #4905, opened 2019-06-13
Galen consultations prj 27 add ref number wkflw step

whatisgalen:
### Types of changes

-   [x ] New feature (non-breaking change which adds functionality)

### Description of Change
Adds a new step type for workflows allowing multiple tiles to be saved during that workflow step

#### Ticket Background
See PR in consultations-prj: https://github.com/archesproject/consultations-prj/pull/54
Original ticket: https://github.com/archesproject/consultations-prj/issues/27

### Further comments

While the scope of the ticket did not explicitly state the need for a generic, multi-tile workflow step, using the mockups it was clear that this is what was required given that other steps would likely need to make use of handling multiple tiles.


whatisgalen:
this will probably throw a require.js error if it's merged without the companion branch: https://github.com/archesproject/arches/pull/4905

robgaston:
@whatisgalen can you please remove the many _.DS_Store files added here? we don't want them in the repo and they make it more difficult to review the actual functional changes in this PR.  Thanks!

arches/app/media/js/viewmodels/card.js
```diff
@@ -374,6 +374,10 @@ define([
         this.isChildSelected = ko.computed(function() {
             return isChildSelected(this);
         }, this);
+        // this.isChildSelected = ko.computed({
```
robgaston: @whatisgalen let's remove this commented block that

arches/app/media/js/views/components/workflows/new-multi-tile-step.js
```diff
@@ -0,0 +1,73 @@
+define([
```
robgaston: @whatisgalen looks like there are a number of unused imports here which should be removed.  It looks like only `'knockout'`, `'views/components/workflows/new-tile-step'`, and `'viewmodels/alert'` are actually being used in this module.

arches/app/templates/views/components/cards/default.htm
```diff
@@ -33,7 +33,7 @@
         }">
             <li role="treeitem" class="jstree-node" data-bind="css: {'jstree-open': (cards.length > 0 && expanded), 'jstree-closed' : (cards.length > 0 && !expanded()), 'jstree-leaf': cards.length === 0}, event: {'dragstart': function () { console.log('dragging...') }}">
                 <i class="jstree-icon jstree-ocl" role="presentation" data-bind="click: function(){expanded(!expanded())}"></i>
-                <a class="jstree-anchor" href="#" tabindex="-1" data-bind="click: function () { self.form.selection($data)}, css:{'jstree-clicked': selected, 'child-selected': isChildSelected(), 'filtered-leaf': card.highlight()}">
+                <a class="jstree-anchor" href="#" tabindex="-1" data-bind="click: function () { self.form.selection($data); console.log(self); console.log($data);}, css:{'jstree-clicked': selected, 'child-selected': isChildSelected(), 'filtered-leaf': card.highlight()}">
```
robgaston: @whatisgalen let's remove the console log statements here and below at ln 363

arches/app/media/js/views/components/workflows/new-multi-tile-step.js
```diff
@@ -0,0 +1,73 @@
...
+    /** 
+     * A generic viewmodel for workflow steps that can add multiple tiles
+     * @name NewMultiTileStepViewModel
+     **/
+
+    function NewMultiTileStepViewModel(params) {
+        NewTileStepViewModel.apply(this, [params]);
+        var self = this;
+
+        this.remove = function(tile) {
```
robgaston: @whatisgalen your `remove` method here removes the tile from the local array, but it still remains on the server; I think what we actually want to do here is delete the tile on the server then remove it from the array.  There is a method on the `tile` viewmodel called `deleteTile` which you can use for this as we do here: https://github.com/archesproject/arches/blob/master/arches/app/media/js/views/resource/new-editor.js#L234-L236

PR #4934, opened 2019-06-20
rm select-workflow template and viewmodels, obsolete plugin (was inte…

whatisgalen:
…nded to stub out init-workflow, now separate)

### Description of Change
Removes workflow-select template and viewmodels from core arches. No dependencies.


whatisgalen:
I think a migration will be needed to remove this as a plugin (from the database)

Issue #5410, opened 2019-10-18
Resource import fails with "Object of type UUID is not JSON serializable"


apeters:

chiatt:
Works in the latest

PR #5413, opened 2019-10-18
fix for resource impport failing, re #5410

apeters:
re #5410


apeters:
fix import of business data failing

Issue #5474, opened 2019-10-28
Implement celery to manage tasks


chiatt:

apeters:
this works as expected

apeters:
fixed in the latest

PR #5517, opened 2019-11-05
Implements celery in Arches

chiatt:
If celery is available, sync will run in a celery task, re #5474


arches/install/arches-templates/project_name/settings.py-tpl
```diff
@@ -147,12 +148,17 @@ APP_TITLE = 'Arches | Heritage Data Management'
 COPYRIGHT_TEXT = 'All Rights Reserved.'
 COPYRIGHT_YEAR = '2019'
 
+CELERY_BROKER_URL = 'amqp://guest:guest@localhost'
+CELERY_ACCEPT_CONTENT = ['json']
+CELERY_RESULT_BACKEND = 'django-db' # Use 'django-cache' if you want to use your cache as your backend
```
beatboxchad: It looks like you can use CouchDB as the result backend without issue in Celery 4.x -- I dunno how Couch fits into Arches in general, but if you can count on it being available, it'd probably be a good default. Anything like that is tons more performant than the Django DB for production, _especially_ when it comes to tasks like import/export of business data. You end up with double the I/O as intermediate task data is rapidly read/written to the very same database the business data is stored in, possibly at the same time as it is being imported or exported. It might not OOM the server, but it would constitute a major bottleneck.

beatboxchad: Looks like you can also use ElasticSearch. Neat!

chiatt: Thanks @beatboxchad. Unfortunately we can't count on CouchDB being available. However, that's good to know. The default backend is the ORM just for simplicity sake. Admins will be able to adopt something more performant based on their needs and resources.

beatboxchad: Awesome,

Looks like I need to better understand the Arches ecosystem. Another important point -- I think I'm conflating the result backend with the task broker.

Thanks for taking the time!

PR #5675, opened 2019-12-11
fix datatype_to_rdf tests; deal with missing class in data; #5579

### LINKED ISSUE #5579: Allow preference order for external concept URIs
**Is your feature request related to a problem? Please describe.**

The export functionality for concepts can only have a single preferred concept scheme (settings.PREFERRED_CONCEPT_SCHEME, used [here](https://github.com/archesproject/arches/blob/master/arches/app/datatypes/concept_types.py#L162)).

If there are multiple vocabularies in use that have been imported into Arches, then the algorithm will look for that one ... and then otherwise use the internal UUID.

**Describe the solution you'd like**

Allow the preferred scheme to instead be an ordered list of preferences, with the internal UUID being the fallback option at the end if none match.

**Other Context**

Discussed in ARM WG meeting, 2019-11-19. Useful for Historic England, Linked Art.

azaroth42:
This PR does the following:

* changes settings.PREFERRED_CONCEPT_SCHEME to be an ordered list of preferred schemes to test for.  This closes #5579, and is needed for referring to both AAT and CRM properties as concepts.
* Fixes the datatype_to_rdf tests to reflect the new RdfWriter's add_edge_to_graph function. The change is to use get_rdf_uri() on the datatype to calculate its own URI, such that resource-instances can have concepts and vice versa without having to duplicate the code artificialy
* Deals with incoming data for import when a node does not have a defined class, but there is only one possible property/class combination that it could match.
* Takes out some unnecessary debugging time()s
* Replaces hardcoded localhost with correct settings.ARCHES_NAMESPACE_FOR_EXPORT

### Types of changes

### Issues Solved

#5579

### Checklist

#### Ticket Background
*   Found by: @apeters somewhat by brute force of pushing to master ;)

### DIFF
--- a/arches/app/datatypes/concept_types.py
+++ b/arches/app/datatypes/concept_types.py
@@ -156,14 +156,16 @@ class ConceptDataType(BaseConceptDataType):
         ext_ids = [
             ident.value for ident in models.Value.objects.all().filter(concept_id__exact=c.conceptid, valuetype__category="identifiers")
         ]
-        for id_uri in ext_ids:
-            if str(id_uri).startswith(settings.PREFERRED_CONCEPT_SCHEME):
-                return URIRef(id_uri)
+        for p in settings.PREFERRED_CONCEPT_SCHEMES:
+            for id_uri in ext_ids:
+                if str(id_uri).startswith(p):
+                    return URIRef(id_uri)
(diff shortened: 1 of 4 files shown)


apeters:
Changes look good, but I merged into https://github.com/archesproject/arches/tree/5671_fix_broken_build instead

Issue #5384, opened 2019-10-15
When reindexing the database, there appears to be a mismatch with the number of concepts indexed


apeters:
When running: `python manage.py es index_database`
This status message appears (when using the consultations package)
> Indexing concepts
Status: Failed, In Database: 16946, Indexed: 16776, Took: 7 seconds

whatisgalen:
Still seeing this error on latest master

apeters:
this is best tested with the consultations package because it has orphaned concepts

chiatt:
Works as expected

PR #5747, opened 2019-12-20
fix reindex of concepts when there are orphaned concepts in the db, re #5384

apeters:
re #5384


apeters:
this is best tested with the consultations package because it has orphaned concepts

Issue #5876, opened 2020-01-30
Missing table error when loading a package to a project without a database


chiatt:
If a user loads a package to a project without a database, django will fire a ProgrammingError exception.

apeters:
@chiatt how would table NOT be present?  I don't understand when this could happen.

apeters:
@chiatt is there a way to test this?

apeters:
this is fixed in the latest

PR #5877, opened 2020-01-30
Handles for missing table error, re #5876

chiatt:
Handles for missing table error that occurs when loading a package to a project before the project's database is available, re #5876


apeters:
@chiatt I don't understand when this could happen.  When would the database table NOT be available?

Issue #6054, opened 2020-04-05
Add tests for datatype validation


chiatt:
- [x] string
- [x] boolean
- [ ] number
- [ ] date
- [ ] concept
- [ ] concept list
- [ ] domain
- [ ] domain list
- [ ] node to resource 
- [ ] file list
- [x] geojson-feature-collection
- [ ] annotation
- [x] URL
- [x] reference

chiatt:
Closing because this issue is gradually being addressed through required test coverage

PR #6055, opened 2020-04-05
Adds datatype testing class, re #6054

chiatt:
Adds test fro string datatype validation re #6054


chiatt:
Tests pass locally, but never on Travis for some reason.

Issue #6094, opened 2020-04-10
Uploaded files are served directly from the web server without authentication


chiatt:
Uploaded files are served directly by the web server running Arches. If someone has the url to a file, they can access it without authentication. We should allow admins to configure arches so that file requests are proxied over a login protected view. The accepted answer of this post describes one way of doing this:

https://stackoverflow.com/questions/9454302/protecting-user-uploaded-files-django

Ideally the use of the proxy view would be optional because serving files directly from Apache or other web server is more performant than proxying requests. Also, in many cases, node and resource permissions are probably sufficient in preventing unauthorized people from getting file urls.

apeters:
fixed in the latest

PR #6113, opened 2020-04-15
Restricted file access re #6094

chiatt:
Saves proxy file url in tiles and checks nodegroup permissions against the nodegroup a file belongs to when accessing the proxy file url. re #6094


arches/app/datatypes/datatypes.py
```diff
@@ -1135,6 +1135,7 @@ def handle_request(self, current_tile, request, node):
                     if file_json["name"] == file_data.name and file_json["url"] is None:
                         file_json["file_id"] = str(file_model.pk)
                         file_json["url"] = str(file_model.path.url)
```
apeters, line 1137: this line is redundant

Issue #6257, opened 2020-06-04
CLI command to load via JSON-LD


azaroth42:
**Is your feature request related to a problem? Please describe.**

At the moment, JSON-LD documents can be loaded one at a time via the web API. For a few instances, or for making point updates this is fine and a good pattern.
However, if the user has command line access to the server, it is much more efficient and practical to use a CLI to load the data off of disk.

There should be a command that allows data to be loaded in this fashion.

**Describe the solution you'd like**

`manage.py load_jsonld` with options to configure which models and data  to load.

```
  -s SOURCE, --source SOURCE
                        the directory in which the data files are to be found
  -f, --force           reload records that already exist
  --nobig               Do not attempt to load records > 250kb
  -m MODEL, --model MODEL
                        the name of the model path to load (eg auction_of_lot)
  -b BLOCK, --block BLOCK
                        the name of the block in the model path to load (eg
                        00)
  --max MAX             Maximum number of records to load per model
  --fast FAST           Use bulk_save to store n records at a time
  -q, --quiet           Don't announce every record
  --skip SKIP           Number of records to skip
  --suffix SUFFIX       file suffix to load
```

**Describe alternatives you've considered**

This pattern has worked for me for multiple projects now, after several iterations.

PR coming!

azaroth42:
More verbose documentation:

To load all data from all models, one record at a time:

	$ manage.py load_jsonld -s <path-to-data>

Other options:

* Load a single model, by name (as above):

	$ ... -m <name-of-model>

* Load files that end in a particular suffix, rather than all files in the directories

	$ ... --suffix <suffix>

* Don't store/index one by one, instead every n records  [500 or 1000 are good numbers]

	$ ... --fast <n>

* Don't print out every record UUID, only timing data or warnings

	$ ... --quiet

* Force overwriting of records that already exist, rather than skipping

	$ ... --force

* Skip n records before starting to process

	$ ... --skip <n>

* Stop after processing n records

	$ ... --max <n>

* Don't load records larger than n kb in file size

	$ ... --toobig <n>

* Load only a subset of the data, by subdirectory block name

	$ ... --block <block>

For example:
	$ manage.py load_jsonld ... --block 00
would load only records in subdirectory 00

* Load a slice of the data, by set of blocks

	$ ... --block <n-blocks>,<max-blocks>

For example:
	$ manage.py load_jsonld ... --block 1,8
would load the first slice of 8 slices, thus blocks 00, 08, 10, 18, etc.	
[...]

PR #6258, opened 2020-06-04
add load_jsonld CLI command

azaroth42:
### Types of changes

### Description of Change

Enable command line loading of JSON-LD data in the same way as JSON and CSV.

### Issues Solved

#6257

### Checklist


apeters:
@azaroth42 do you have some data I can test this against?

azaroth42:
Here you go...

[test_load_cmd.zip](https://github.com/archesproject/arches/files/4897336/test_load_cmd.zip)

This is just some artificial data from the 6235 test.

To reproduce:
* Make a new project
* Uncompress the zip in the project root
* load the cidoc-crm ontology (pkg/ontology/cidoc_crm)
* load the model (pkg/models/)
* Then you can load the data:

> python ./manage.py load_jsonld -s pkg/data     

And you can experiment with the various options:

> python ./manage.py load_jsonld -s pkg/data --fast  6

This takes about 0.25 seconds, compared to 1.2 when saving each record individually, by batch save and index

> python ./manage.py load_jsonld -s pkg/data -m test   

This explicitly only loads the data in pkg/data/test ... if there were other directories, they'd be skipped

> python ./manage.py load_jsonld -s pkg/data --quiet

Doesn't announce every record

> python ./manage.py load_jsonld -s pkg/data --skip 1

Skips the first record  (e.g. if it's already loaded)

> python ./manage.py load_jsonld -s pkg/data  --max 1

Loads only the first record (e.g. for testing)

> python ./manage.py load_jsonld -s pkg/data --fast 6 --strip-issearchable

This uses my rewrite of get_documents_to_index to check node['issearchable'] and if it's false, don't even index it.
(This could be an option in core? Happy to do that refactor?)

azaroth42:
I accept the second point. I'll make the suffix a required default, and filter out anything starting with `.`

The first is a reporting question. I chose to have it report the total number of records that it saw, not the total that it tried to load. If you look in the system, it will have skipped the right number of records -- if you do `--skip 3` it'll say that it saw 6, and it will have loaded 3. So it does the right thing, just has (obviously) unintuitive feedback to the user.

The same would be true if you put in ignore-errors ... it'll report all of them, even if only one got loaded and the rest were garbage. The rationale for this is about timing. I have always wanted to know how long it spent divided by all the records it saw ... if it loaded 50% and the rest caused some ignored error, then the timing statistics would be double the actual amount of time spent for the number of records.

A solution could be to have two counts -- total seen and total loaded?

azaroth42:
Fixed the issues:
* every time it gets values from listdir() it filters out items that start with . or _
* Split the reporting into records seen and records loaded.
* Bonus: renamed force to ow/overwrite to be consistent with other commands

apeters:
if you try to run any command more than once it throws an error

```
File"/Users/alexei/Work/Projects/arches4/arches/management/commands/load_jsonld.py", line 241, in load_resources
    loaded += l
TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'
```

Issue #6292, opened 2020-06-16
Concept functions create too many searchengine instances


azaroth42:
**Describe the bug**

Each time a search engine is created it opens at least one connection to elasticsearch, and those connections take time to tear down when the search engine goes out of scope and is garbage collected. Thus, if too many search engines are created, the operating system runs out of file handles and blows away the offending processes.

The app.models.concepts function at the end of the file create search engines every time they are called in order to look up concepts by various properties. A single call to `get_preflabel_from_valueid` will create three search engine instances, as it called `get_preflabel_from_conceptid` and `get_concept_label_from_valueid`, each of which also create instances.

These functions get called during concept.from_rdf() parsing. So in bulk load of a model that has a lot of concepts, the number of search engine instances gets quickly out of control and the process gets killed.

Apart from rewriting the entire concept / rdm subsystem, some possible fixes:

* Pass a search engine instance through the datatype into these functions.  The issue is then the constructor for the datatype is inconsistent, as no other datatype needs a search engine instance.
* Rewrite the functions as nasty SQL queries against the concept and value tables in postgresql.
* Create a global scoped searchengine that these functions can reference.

**Expected behavior**

Arches should not create so many searchengine instances! :)

**Your Arches Information**

Master as of 2020-06-16

apeters:
added a few tweaks to what @azaroth42  submitted

PR #6293, opened 2020-06-16
pass searchengine instance into concept datatype

azaroth42:
### Types of changes

### Description of Change

Instantiate a global searchengine instance to be used from various deeply internal functions that are troublesome to pass an instance into. Previous implementation only partially solved the issue by passing into the concept datatypes, but it also is needed in the get_display_name function ... which can be called via get_resource_descriptor magic getter, and thus cannot have params passed in.

This is to prevent the need for the functions to create their own instances, which they do so quickly that the garbage collection can't get rid of them fast enough, and they eat up all the network file handles from the machine.

Running before this change I had about 500 connections between arches and elastic while loading data, that were constantly being spun up and spun down again. With this change I have 5 or less.

A better fix would be to reimplement the RDM of course :D

### Issues Solved

#6292


azaroth42:
Still generates lots of search engine instances in get_display_value(). Trying to fix that as well.

azaroth42:
Having it in a separate module resolved some of the test failures, but if there's a better way to do that, then great :)

Issue #6300, opened 2020-06-18
Branch Identification


azaroth42:
**Is your feature request related to a problem? Please describe.**

The system (sensibly) adds branch by copying the structure, rather than referencing it. This means that the branches can be customized after import, and don't get out of sync with the data if the branch definition changes.  The downside is that the model doesn't know that it has a copy of a branch, and there are many benefits to knowing that a branch in a model came from a particular branch definition:

* Code could list the uses of particular branches in models.  For example, basically every model has a Name branch and a Statement branch.  If either branch changed, it would be nice to be able to evaluate each usage to see if it should be updated to track the new structure or not. 
* Documentation can be constructed using these branch identifiers. For example, rather than cutting and pasting the documentation for the Name branch into every model (or worse, the Classification documentation into basically every semantic node), instead it could refer to the branch with its documentation. This doesn't obviate the need for custom documentation, but provides sensible defaults.
* It provides a way to track alignment with other definitions or conceptual models.  The Name branch in the A4S Person model /is/ the Name branch from ARM, which /is/ the Name branch from Linked Art. A single identifier for all of these would let us construct better references to further documentation that is externally maintained.
* It provides a hook for other application logic to be built from.  For example, if a Name advertises itself as being the linked-art-personal-name, then a renderer could have custom handling that makes use of that particular structure. Similarly, that structure could be associated with a schema that describes in machine understandable terms exactly what fields are expected, so that data can be validated using json-schema (or other) before coming into Arches. 
* It demonstrates the consistency of the modeling, both internally and with external specifications.  If a model has the same branch identifiers everywhere, then it is very consistent. And if those branch identifiers are well-known from external specifications, then it is provably externally consistent as well.
* It would validate the ARM approach of constructing a toolkit of models and branches. Usage of these branches would be obvious in the configurations of systems, and could then generate automatic reports (if the admin allowed it) that showed the benefit to the community.
* It allows ease of migration into other systems or incompatible future versions. Rather than having to process every node, the branches could be handled in a hierarchical fashion using these identifiers in the models.

(I'll stop there, I hope I've convinced you by now!)

**Describe the solution you'd like**

Each Model and Branch should have a field for an identifier separate from the name. This is to allow the names from external specifications to be used, and for the name to be at least somewhat human readable rather than a UUID.  So the Linked Art branch for Name could have a `name` field of `la-simple_name`, but a name of `Name`.  UI would be a text box, needs a field in the root configuration for saving the field.

[...]

Issue #6299, opened 2020-06-18
Node Identification separate from Naming


azaroth42:
**Is your feature request related to a problem? Please describe.**

The node name field in the graph designer has two conflicting uses:
* It is used in the UI for the name of the node in the tree structure, and in other documentation that could be generated from the structure. In this usage, it should be short and descriptive and can rely on its position in the hierarchy to be understood for context. Example:  Name Content
* ~~It is also used in the CSV import for aligning the incoming data against the nodes in the model. In this usage it should be unique amongst all the nodes in the model, and thus cannot be short or rely on the hierarchy for context. Example:  Production_Statement_Name_Content~~  This is actually not true.  Only node ids are using during import.  (Alexei)

These two conflicting uses result in confusion and additional work - when branches are imported, the value of the branch pattern is greatly reduced because the administrator has to go through and rename all of the nodes based on their position in the model if the import scenario is important.

**Describe the solution you'd like**

Instead, @habennin and @azaroth42 propose that there should be a separate node identifier with an automatic default to facilitate import, separate from the name of the node for UI and documentation.

The idea - Every node in a branch or model would have a separate, editable "ETL Identifier" string field that would be used for ETL loading instead of name. This would require:

* Add a new field to the UI and store the value, in the same way as the new description field per node.
* Swap the CSV import to looking at that field, rather than name.

Enhancements beyond the basic:

* The system could enforce that this field be unique per model and refuse to save with an appropriate error message when the constraint is broken.

* The system could automatically construct a default from the ontology and/or the branch.

When a branch is added into a model, then rename all of the imported node identifiers to prepend the node identifier of the current node.  e.g. if you have a production node in a physical thing model with the etl identifier of "production" and you add a name branch with identifier "name", and a child node of "name_content", then the resulting branch would be "production_name" that has a string which is "production_name_content".  This would follow the existing pattern from A4S and other projects.

It would be nice to have a reasonable default value generated:
When adding a node into a model or branch, set the default to the parent node's etl identifier plus `__` plus the ontology class. e.g. If you add p2_has_type of a E55_Type to the production node, it would be `production__E55_Type`.  This won't necessarily be unique, but would be easier to fix than to type out from nothing.

**Additional context**

This came up in the discussion about documenting Arches for Science and current practice of naming things, but we think it has a lot of merit for other projects as well.

Thoughts, @annabelleee?

annabelleee:
[...]

PR #6305, opened 2020-06-19
Add documentation/import fields to graph, node

azaroth42:
### Types of changes

### Description of Change

Add the following fields:

* import_id to node:  This is the name to use for CSV import and needs to be unique per model
* branch_id to node:  This is the identifier of the branch that was used to add the node to the current model
* branch_id to graph: This is the identifier of the branch to be copied when the branch is added to a model

It also exposes the slug field for models that was already present, just not visible.

### Issues Solved

#6299 #6300

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


azaroth42:
Closing as discussion resulted in it being too specific a solution. If there isn't any integrity constraint, then a micro-syntax in description is sufficient.

Issue #6332, opened 2020-06-25
Advanced Search facets should be ordered as per their cards


azaroth42:
**Is your feature request related to a problem? Please describe.**

When trying to perform an advanced search, it is frustrating to look through the potentially long list of cards which appear to be in a random order. 

**Describe the solution you'd like**

The facets should be ordered in the same order as the cards in the report, as this is what the user is familiar with and, one assumes, will put important information at the top.

**Describe alternatives you've considered**

Allow an admin to manually configure the sort order.  This seems useful, but would almost inevitably result in something very similar if not identical to the card order, just with extra time for everyone.

azaroth42:
Closing in favor of #6987

Issue #5433, opened 2019-10-22
Sort order for create new resource menu/screen should be configurable


azaroth42:
**Is your feature request related to a problem? Please describe.**

The order of the create new resource menu and equivalent /resource screen with tiles has an order, but I have no idea what that order is.

**Describe the solution you'd like**

It would be useful to at least alphabetize the order (similar to the order in the graph designer UI) and preferable would be to have an ordering function (preferably drag and drop rather than setting an integer, like in the RDM) that allowed a custom ordering based on the anticipated editorial workflow.

chiatt:
The 'new resource' menu is now listed alphabetically thanks to @azaroth42 commit in #6339.

azaroth42:
I think this could be left open for someone with time on their hands to implement a custom sortorder for them.

PR #6339, opened 2020-06-29
Alpha-sort models, cards in lists

azaroth42:
### Types of changes

### Description of Change

Sort `createable_resources` by model name, to improve UI of "create new resource" left hand menu, and `/resource` card list.

Sort models in advanced search facet list by name, and sort cards per model by name.

### Issues Solved

#5433 
Improves #6332 (though would still be much better if it was same order as card list)


azaroth42:
@chiatt True ... Is that a bug or a feature? :) Happy to make it case insensitive if that would have it able to be merged.

azaroth42:
Merge in master to ensure consistency, and made sort insensitive as per @chiatt

Issue #6101, opened 2020-04-13
Prevent users without view permissions from seeing resource instances in search


chiatt:
Users that do not have permissions to an instance should not see them:

- in the search results list
- on the search results map
- in related resources

apeters:
- I can still select "No Access" plus some other permission.  I think those should be mutually exclusive.
- After "Applying" permissions, the "Revert Permissions" button doesn't appear enabled (but I can still click on the button and then both buttons appear enabled).
- The "Allow Normal Access" button text is confusing.  Does this really mean "remove all access restrictions"?
- I would allow the table to fill the available height.
- I can't seem to apply permissions other than "No Access"

It's hard to review this much further until these issues are resolved

apeters:
I can still see restricted resources on the map and in related resources.

apeters:
fixed in the latest, there is a related ticket re #6398

PR #6384, opened 2020-07-09
Filters restricted instances from related resources and instance overlays, re #6101

chiatt:
Prevents restricted instances from appearing in related resources visualization and resource instance overlays. re #6101


apeters:
if the user permissions are set correctly, then this works as expected.
Making a ticket about the above issue, re #6398

Issue #6390, opened 2020-07-11
Resource-Instance node data not importing via csv


chiatt:
@annabelleee commented on [Thu Jul 09 2020](https://github.com/archesproject/arches-for-science-prj/issues/218)

I'm having an issue with the resource-instance csv import. I was initially able to reimport data that I had exported, but now, I can't. 

Currently, after entering data into the interface, the export spits out double quotes ("") like this:
`"[{""resourceId"": ""cef5a5b5-f1b2-4740-a72e-b30e2985b555"", ""ontologyProperty"": ""http://www.cidoc-crm.org/cidoc-crm/P65_shows_visual_item"", ""resourceXresourceId"": ""98e66526-c232-11ea-9f58-024e0d439fdb"", ""inverseOntologyProperty"": ""http://www.cidoc-crm.org/cidoc-crm/P65i_is_shown_by""}]"`

Once replacing the interior "" quotes with a ' single quote (because otherwise it stops the csv import process), everything imports EXCEPT the resource-instance node data and I get the following error in the resource import log:
`Thu Jul 09 15:37:21 2020 ERROR: datatype: ResourceInstanceListDataType value: [{'resourceId': 'cef5a5b5-f1b2-4740-a72e-b30e2985b555', 'ontologyProperty': 'http://www.cidoc-crm.org/cidoc-crm/P65_shows_visual_item', 'resourceXresourceId': '98e66526-c232-11ea-9f58-024e0d439fdb', 'inverseOntologyProperty': 'http://www.cidoc-crm.org/cidoc-crm/P65i_is_shown_by'}] on line 3 - Expecting property name enclosed in double quotes: line 1 column 3 (char 2) or is not a prefLabel in the given collection.`

This happens whether or not the resourceXresourceId value is populated or blank or even null.

This was happening for several data files, but I recreated some simple models and simple data to test this against. 
[SampleData.zip](https://github.com/archesproject/arches-for-science-prj/files/4899939/SampleData.zip)
This includes: Sample Image (model, mapping file and csv business data) and Sample Thing (model, mapping file and csv business data). Sample Image should be loaded first since Sample Thing references it. Sample Thing is where the resource-instance-list node exists.  

Data note: The Sample Thing instance, "Imported Item," was added after the export and was never entered into the interface, hence the lack of resourceXresourceID. The other two instances, "Annabel's Item" and "Lindsey's Item", were exported after being entered into the interface.

apeters:
[...]

PR #6391, opened 2020-07-11
Supports reimport of exported csv files re #6390

chiatt:
Supports reimport of exported csv files when csv contains resource instance relationship tile values re #6390


annabelleee:
@chiatt Works and is okay to merge.

Issue #5756, opened 2019-12-22
"Overlays" tab shows nothing in Map Layer Manager


whatisgalen:

robgaston:
@whatisgalen can you provide further details here? The tab will show nothing (by design) if no overlays have been added.

whatisgalen:
@robgaston I thought the hex bin was by default an overlay the user could pick? Also this is what I'm seeing (which, even if I don't have any overlays, still looks kind of buggy):

<img width="1645" alt="Screen Shot 2019-12-23 at 11 15 57 AM" src="https://user-images.githubusercontent.com/8366985/71335534-b79ad380-2575-11ea-8ca0-601a15b748e8.png">

robgaston:
@whatisgalen i think that’s the default behavior when no layers have been added; could be improved for sure, but i don’t think this is a bug per se

whatisgalen:
ok, I'll remove the "bug" label and refine the title -- thanks!

chiatt:
Works as expected

PR #6402, opened 2020-07-15
Moves no overlays message into trans tags, re #5756

chiatt:
Moves no overlays message into trans tags. re #5756


chiatt:
PR appears to be stuck on: ```Checking for ability to merge automatically…
Hang in there while we check the branch’s status.```

Issue #6406, opened 2020-07-15
Add command that exports all branches and resource models 


chiatt:

chiatt:
PR reviewed by @njkim

PR #6408, opened 2020-07-15
Adds command to export all models and branches, re #6406

chiatt:
Adds command to export all models and branches, re #6406


njkim:
The command works as expected

Issue #6392, opened 2020-07-13
CSV file import/export is not saving the file path correctly 


chiatt:
Export saves with the proxy url, not the url to the actual file location

chiatt:
Reviewed by @njkim before merging PR

PR #6423, opened 2020-07-16
FileList csv import bug fixes

chiatt:
During csv import of a filelist value, the private path, not the destination location is saved to the tile. On export, the human readable destination path is written to the csv file. Also ensures that the target tile is related to its corresponding file. re #6392

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -1214,8 +1214,17 @@ class FileListDataType(BaseDataType):
             tile_data.append(tile_file)
         return json.loads(json.dumps(tile_data))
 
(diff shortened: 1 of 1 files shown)


chiatt:
1. To QA: load this package alongside your existing package: [file_upload_test_pkg.zip](https://github.com/archesproject/arches/files/4934074/file_upload_test_pkg.zip)

2. Confirm that a file upload test instance was loaded with a tile containing 2 photos (in the report the url should look like `files/<uuid>` rather than the path to the uploads directory.

3. Finally, export the file upload test business data as csv and confirm that the path saved to csv is the path to the file uploads directory, not the `files/uuid` url.

PR #6436, opened 2020-07-22
Improves identification of file type in file viewer and sets default renderer on import

chiatt:
Improves identification of file type in file viewer and sets default renderer on import. re #6240

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -4,6 +4,7 @@ import decimal
 import base64
 import re
 import logging
+import os
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


chiatt:
To QA:
1. Load the AfS package and login as admin admin
2. Search for Digital Resources
3. Check the summary report of each Digital Resource and open the editor of any instance with files listed in the report
4. In the card tree select one of the files in the File card
5. Files with the bmp or pdf extension should open right away in the file viewer.
6. Files with the .txt extension should offer you a choice of file loaders
7. Files with the .dx extension should indicate that the format is unknown

Issue #5688, opened 2019-12-13
Graph designer error when changing node on model with instances


azaroth42:
**Describe the bug**

The graph designer UI allows settings to be configured on nodes that do not have instance data. For example, if I leave the model setting of a resource-instance node blank, I can then later come back and set it in the UI after adding instances of the model. 

However, current master generates the following cascade of errors:

```
Internal Server Error: /node/9519cb4f-b25b-11e9-8c7b-a4d18cec433a
Traceback (most recent call last):
  File "/Users/rsanderson/Development/getty/arches/current/arches/arches/app/views/graph.py", line 324, in post
    graph.save()
  File "/Users/rsanderson/Development/getty/arches/current/arches/arches/app/models/graph.py", line 340, in save
    self.validate()
  File "/Users/rsanderson/Development/getty/arches/current/arches/arches/app/models/graph.py", line 1297, in validate
    self.check_if_resource_is_editable()
  File "/Users/rsanderson/Development/getty/arches/current/arches/arches/app/models/graph.py", line 1278, in check_if_resource_is_editable
    raise GraphValidationError(
arches.app.models.graph.GraphValidationError: 'Your resource model: Physical Thing, already has instances saved.                                 You cannot modify a Resource Model with instances.'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/rsanderson/Development/getty/arches/current/ENV/lib/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/Users/rsanderson/Development/getty/arches/current/ENV/lib/python3.8/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/Users/rsanderson/Development/getty/arches/current/ENV/lib/python3.8/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/rsanderson/Development/getty/arches/current/ENV/lib/python3.8/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/Users/rsanderson/Development/getty/arches/current/ENV/lib/python3.8/site-packages/django/views/generic/base.py", line 97, in dispatch
    return handler(request, *args, **kwargs)
[...]

PR #6452, opened 2020-07-24
Prevents related instance datatype config changes after tile is saved

chiatt:
Prevents a user from changing the datatype configuration of a related instance node once a tile is saved to the corresponding node's nodegroup. re #5688


njkim:
works as expected

Issue #6464, opened 2020-07-28
Allow export of all instance data with one management command


chiatt:
Currently to export the business data of all of your models you have to do so one at a time and pass the graphid as an argument. We should allow a user to export data for all models in one command. This could also be used to support the 'update package' command.

azaroth42:
:100: this annoys me every time I have to run setup_db and not blow away data

chiatt:
Approved in PR #6468

PR #6468, opened 2020-07-29
Export data for all resource models

chiatt:
### Description of Change
Updates the export_business_data command to export data for all resource models. #6464
Also removes some cruft and prevents and error when saving a filelist node that has a None value. re #6392


chiatt:
To QA: `python manage.py packages -o export_business_data -f json -d ./`

Issue #6444, opened 2020-07-23
Update the Related Resource Card to support changes to the related instance datatype


chiatt:
Update the Related Resource Card to support changes to the related instance datatype. Related instances are no longer identified by their ID, but by a related instance object:
e.g.:
```
{
  inverseOntologyProperty: "http://www.cidoc-crm.org/cidoc-crm/P138_represents"
  ontologyProperty: "http://www.cidoc-crm.org/cidoc-crm/P138i_has_representation"
  resourceId: "4096eaae-4800-4c98-bacd-abbb68f85ad1"
  resourceXresourceId: "a8fac750-c619-11ea-840a-faffc210b420"
}
```

PR #6494, opened 2020-08-04
Updates the related resource map card to support datatype changes. re #6444

chiatt:
Updates the related resource map card to support datatype changes. re #6444
Also prevents and error in the FileList widget if a file does not have an extension.


chiatt:
To QA:

1. Load the ARFH package.
2. In the Information Resource Model, change the card for the Related Master Plan Zones to the Related Resources Map card
3. Set the `Select drawings map source` config property to `master-plan-zone`

Issue #6453, opened 2020-07-24
Relationship records cannot be deleted if their corresponding tile is deleted


chiatt:
If a user creates a tile with a resource instance datatype, the tile and relationship record are created. If the user then deletes the tile, the relationship records persist and the user cannot delete them in the Related Resources view in the editor.

apeters:
shouldn't this be handled by this in the `on_delete=models.CASCADE` setting in the `ResourceXResource` table?
```    
tileid = models.ForeignKey(
        "TileModel", db_column="tileid", blank=True, null=True, related_name="resxres_tile_id", on_delete=models.CASCADE,
)
```

chiatt:
@apeters It could be the index is persisting and that's why it still appears in the UI

azaroth42:
I can confirm this.  I had to do the same (or rather I deleted directly out of the index via the python elasticsearch library)

Issue #6421, opened 2020-07-16
Race condition? in delete with ResourceXResource


azaroth42:
**Describe the bug**

While deleting all the resources in the graph, I get this traceback:

```
All resources associated with the 'Physical Object' Resource Model will be removed. continue? [Y/n] Y
Traceback (most recent call last):
  File "manage.py", line 30, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/rsanderson/Development/getty/arches/current/ENV/lib/python3.8/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/Users/rsanderson/Development/getty/arches/current/ENV/lib/python3.8/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/rsanderson/Development/getty/arches/current/ENV/lib/python3.8/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/rsanderson/Development/getty/arches/current/ENV/lib/python3.8/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/Users/rsanderson/Development/getty/arches/current/arches/arches/management/commands/resources.py", line 49, in handle
    self.remove_resources(force=options["yes"], graphid=options["graph"])
  File "/Users/rsanderson/Development/getty/arches/current/arches/arches/management/commands/resources.py", line 71, in remove_resources
    graph.delete_instances(verbose=True)
  File "/Users/rsanderson/Development/getty/arches/current/arches/arches/app/models/graph.py", line 445, in delete_instances
    resource.delete()
  File "/Users/rsanderson/Development/getty/arches/current/arches/arches/app/models/resource.py", line 377, in delete
    self.save_edit(edit_type="delete", user=user, note=self.displayname)
  File "/Users/rsanderson/Development/getty/arches/current/arches/arches/app/models/resource.py", line 88, in displayname
    return self.get_descriptor("name")
  File "/Users/rsanderson/Development/getty/arches/current/arches/arches/app/models/resource.py", line 74, in get_descriptor
    return PrimaryDescriptorsFunction.get_primary_descriptor_from_nodes(self, functionConfig[0].config[descriptor])
[...]

PR #6497, opened 2020-08-05
Deletes indexed relationship when corresponding resource tile/instance is deleted

chiatt:
Deletes the indexed relationship in ElasticSearch if a resource instance or tile participating in the relationship is deleted. re #6453, #6421

### DIFF
--- a/arches/app/datatypes/base.py
+++ b/arches/app/datatypes/base.py
@@ -207,6 +207,13 @@ class BaseDataType(object):
         """
         pass
 
+    def post_tile_delete(self, tile, nodeid):
+        """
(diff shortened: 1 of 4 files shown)


chiatt:
To QA: 
1. Create an instance of a resource that has a related-instance-list node
2. Create two relationships
3. From the source instance, delete a tile participating in 1 relationship and then go to the related resources page and confirm the relationship is gone.
4. For the second relationship, open the resource editor of the target instance in another tab
5. Delete the source instance.
6. Going back to the editor for the target instance, confirm that no relationship exists with the deleted source instance.

Issue #6492, opened 2020-08-04
Error when searching for dates with advanced search


chiatt:
When searching on dates in advanced search the following py error occurs:
```
Traceback (most recent call last):
  File "/Users/cyrus/Documents/projects/GCI/arches4/arches/arches/app/search/search.py", line 134, in search
    ret = self.es.search(**kwargs)
  File "/Users/cyrus/Documents/projects/GCI/arches4/env/lib/python3.7/site-packages/elasticsearch/client/utils.py", line 92, in _wrapped
    return func(*args, params=params, headers=headers, **kwargs)
  File "/Users/cyrus/Documents/projects/GCI/arches4/env/lib/python3.7/site-packages/elasticsearch/client/__init__.py", line 1623, in search
    body=body,
  File "/Users/cyrus/Documents/projects/GCI/arches4/env/lib/python3.7/site-packages/elasticsearch/transport.py", line 362, in perform_request
    timeout=timeout,
  File "/Users/cyrus/Documents/projects/GCI/arches4/env/lib/python3.7/site-packages/elasticsearch/connection/http_urllib3.py", line 252, in perform_request
    self._raise_error(response.status, raw_data)
  File "/Users/cyrus/Documents/projects/GCI/arches4/env/lib/python3.7/site-packages/elasticsearch/connection/base.py", line 282, in _raise_error
    status_code, error_message, additional_info
elasticsearch.exceptions.RequestError: RequestError(400, 'search_phase_execution_exception', "failed to parse date field [2020-08-04T00:00:00] with format [-yyyy||yyyy||yyyy-MM||yyyy-MM-dd||yyyy-MM-dd HH:mm:ssZZZZZ||yyyy-MM-dd'T'HH:mm:ssZ||yyyy-MM-dd'T'HH:mm:ssZZZZZ||yyyy-MM-dd'T'HH:mm:ss.SSSZ||yyyy-MM-dd'T'HH:mm:ss.SSSZZZZZ]: [failed to parse date field [2020-08-04T00:00:00] with format [-yyyy||yyyy||yyyy-MM||yyyy-MM-dd||yyyy-MM-dd HH:mm:ssZZZZZ||yyyy-MM-dd'T'HH:mm:ssZ||yyyy-MM-dd'T'HH:mm:ssZZZZZ||yyyy-MM-dd'T'HH:mm:ss.SSSZ||yyyy-MM-dd'T'HH:mm:ss.SSSZZZZZ]]")
[...]

PR #6499, opened 2020-08-06
6492 advanced search datetime

njkim:
### Types of changes

### Description of Change

The time format to add ES search query for advanced search is modified to include the time zone to properly indexed for the search, and  also the support for the time format is added. re #6492 

### Issues Solved

Fixed the error in the advanced search for the card that is datetime data type.

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chiatt (commented):
@njkim I think what you've done is really good, but as you had observed - because the date format was hardcoded with the time in the date datatype search html, we had to always search with a time from advanced search (so if the value was `2000-01-01` advanced search would query `2000-01-01 00:00:00-07:00`. The problem with that is that the data was indexed without a time, so ES assumed the date was in UTC. That meant to search for a date, you had to make sure that the local time was adjusted to local time equivalent of UTC. 

To get around this problem it became clear that we needed to be able to configure the date format at the node/datatype rather than the widget. I went ahead and made that change. I think the date format belongs on the node because it affects how the data is stored on the node.

Because these changes are pretty significant, I'm going to pass this onto @apeters for review both of our changes.

chiatt:
To QA:
1. run migrations
2. create a two date nodes, one with a 'YYYY-MM-DD' format and the other with 'YYYY-MM-DD HH:mm:ssZ'. 
3. create some tile data for each node
4. in advanced search confirm that you can find your resource by searching on the tile data you created in step 2.

apeters (changes requested):
It looks like time is enabled in the search widget, but it is ignored and not sent to the sever.  We should either allow users to include time in their search, or remove the ability to specify time in the search widget.

Everything else looks good.!

apeters (approved):
this now works as expected

Issue #5967, opened 2020-03-06
Related Resource Card breaks on branches with multiple nodes


dwuthrich:
When trying to use the Related Resource Card to support a map-based selection of a related resource, that card breaks when the branch contains multiple nodes.  Works as expected when the branch contains a single node.

![Screen Shot 2020-03-06 at 11.23.27 AM.png](https://images.zenhubusercontent.com/56f1ad0a6caf56c9d38e1cbd/0fe72eb2-129f-46e9-83dd-24d6241710ba)

apeters:
confirmed that this is still an issue

chiatt:
This issue occurs when one of the additional nodes is a geojson-feature-collection node. To resolve this issue the card would need to be modified to support both geometry editing and related resource feature identification.

PR #6501, opened 2020-08-06
Adds support for the geometry editing in the related resource map card. re #6444

chiatt:
Adds support for the geometry editing in the related resource map card. re #6444, #5967


robgaston:
works as expected

PR #6503, opened 2020-08-07
6000 cbyrd update photo gallery card thumbnail container width on adjust

chrabyrd:
### Types of changes

### Description of Change
The photo gallery card thumbnail container was originally set using `position: fixed`, which completely removes the object from the document flow. This makes it impossible to calculate the correct width without some sort of JS listener. 

This has now been updated to use `position: sticky`, which when combined with a few modifications to parent elements gives the behavior we want.

UI changes seen here: https://recordit.co/3Aq3y73A3Q

### Issues Solved

#6000 

### Checklist

### Further comments

NOTE: The arrow button in the thumbnail preview UI are borked on master. Followup ticket #6504


chrabyrd:
Things still borked ( will file followups && post UI gifs on Monday ):

- Files that aren't photos pull the thumbnail bar from the bottom. Funky UI but nothing breaks.
- 500 error when trying to upload new photos in self-made resource. Upload works fine with already-created resource. May be user error.
- In the designer view, interacting with the `Add Photos` button, then clicking either `Upload Photographs` buttons will result in multiple OS upload prompts

chrabyrd:
Remaining funkiness: 

- 500 error when trying to upload new photos in self-made resource. Upload works fine with already-created resource. May be user error.
- In the designer view, interacting with the Add Photos button, then clicking either Upload Photographs buttons will result in multiple OS upload prompts
- drag && drop onto upload panel does nothing ( just like ze goggles! )

Issue #6507, opened 2020-08-07
Refactor setting of default ontology properties


chiatt:
Refactor setting of default ontology properties and resolve errors when graph does not have default properties defined.

chiatt:
Resolved in #6508

PR #6508, opened 2020-08-07
Resolves errors when saving and deleting tiles with related instance nodes.

chiatt:
Resolves errors when saving and deleting tiles with related instance nodes. re #6507


njkim:
works as expected

PR #6536, opened 2020-08-14
Adds missing card configs to the related resource map card. re #5967

chiatt:
Adds migration to update card configs to the related resource map card. This prevents an error when adding the related resource map card component to a card in the graph designer. re #5967


chiatt:
1. To QA, run migrations (`python manage.py migrate`)
2. In the graph designer change the card component of a card (any card) from the `Default Card` to the `Related Resource Map`.
3. Confirm that there are no javascript errors.

njkim:
Runs as expected without any error

Issue #6544, opened 2020-08-17
Set default ontology property on a resource relationship


chiatt:
Currently the default ontology is assigned on the client if it's available or on the server if the property returned from the client is an empty string. Instead of using an empty string (which could be an intended value), we should apply the defaults only if no value for the ontology property is returned from the client.

Issue #6526, opened 2020-08-13
Cannot exclude resource-instances from participating in RxR 


azaroth42:
**Describe the bug**

Every resource-instance tile creates a resource to resource relationship.  However there are situations where resource-instance relationships are not important enough to create such a link.

Example:

In an RDM-like environment, there are clearly important relationships such as narrower/broader, and the participation of an instance in a collection.  On the other hand, the concept of "preferred term"  (or "prefLabel") should not have a link to every single other concept in the system, just because they have a label of type "preferred term".  Similarly if languages are instances, then every instance will have rxr links to English (or any other such language).  These sorts of links just clutter up the system ... no one cares that every concept has a preferred term in english.

Or the relationship could be important to manage, but not display.  For example, the fact that every AAT concept is part of the AAT scheme is important to know ... but the AAT scheme instance shouldn't try to list every single concept under related resources.

**To Reproduce**
1. Make some instances that would be frequently referred to
2. make links to them
3. see a long long long list in the related resources UI

**Screenshots**

Note the two useful relationships in the screenshot below -- broader and has_member. The rest are worthless.

![Screen Shot 2020-08-13 at 10 44 22 AM](https://user-images.githubusercontent.com/871868/90148988-f8886800-dd51-11ea-8739-7302aa436940.png)

**Expected behavior**

One of the options for related resource configuration should be "do not create relationship"

PR #6545, opened 2020-08-17
6507 ontology prop defaults

chiatt:
Uses default ontology properties if none is specified by the client. re #6544


arches/app/datatypes/datatypes.py
```diff
@@ -1712,30 +1712,33 @@ def pre_tile_save(self, tile, nodeid):
...
-                    "relationshiptype": related_resource["ontologyProperty"],
-                    "inverserelationshiptype": related_resource["inverseOntologyProperty"],
                     "tileid_id": tile.pk,
                     "nodeid_id": nodeid,
                 }
-                if related_resource["ontologyProperty"] == "" or related_resource["inverseOntologyProperty"] == "":
-                    if models.ResourceInstance.objects.filter(pk=related_resource["resourceId"]).exists():
-                        target_graphid = str(models.ResourceInstance.objects.get(pk=related_resource["resourceId"]).graph_id)
+                if "ontologyProperty" not in related_resource or "inverseOntologyProperty" not in related_resource:
+                    if models.ResourceInstance.objects.filter(pk=related_resourceid).exists():
```
azaroth42, line 1724: :-1: to this check.  This would break a batch load with circular references.  Just load them and let it become eventually consistent.

azaroth42:
Seems like there should be a solution to this that takes in to account #6526 at the same time.

Issue #6524, opened 2020-08-12
Deleting child nodes doesn't affect the card designer view until refresh.


chrabyrd:
**Describe the bug**
In Arches Designer, when a child node is deleted the card remains unaffected until page refresh.

UI seen here: http://g.recordit.co/Qoeqj4RZhJ.gif

**To Reproduce**
1. Create a new resource model
2. Create top nodes. Add child nodes to them.
3. Navigate to the card designer view. Notice the number of nodes.
4. Navigate to the graph designer view, and remove a child node.
5. Navigate to the card designer view, and notice the node has not been deleted.

NOTE: As seen in the gif, any interaction with the deleted node throws a 500 error. This is expected. Backend error message is here: https://pastebin.com/mYjBdJ0J . Not putting elsewhere in ticket because it's expected.

**Screenshots**
UI seen here: http://g.recordit.co/Qoeqj4RZhJ.gif

**Expected behavior**
The card designer view should not show the deleted nodes.

**Your Arches Information**

*   Version used: SHA 9ea7a679b88e013f49af0bc80dab27a1755bf470
*   Operating System and version (desktop or mobile): macOS Catalina
*   Browser Name and version: Chrome 84.0.4147.105 (64-bit)
*   Link to your Arches Install (optional):

**Additional context**
UI seen here: http://g.recordit.co/Qoeqj4RZhJ.gif

**Ticket Background**
*   Found by: @chrabyrd

azaroth42:
Confirmed. Also, when creating new nodes they sometimes appear twice in the card list until refresh.

chrabyrd:
> Confirmed. Also, when creating new nodes they sometimes appear twice in the card list until refresh.

hmmm I can't seem to repro this @azaroth42 , if you can show how to reproduce this I'd be happy to take a look 👍

PR #6546, opened 2020-08-18
Deleting non-collector nodes now deletes widget, need to test for side effects

chrabyrd:
### Types of changes

### Description of Change

This changes card model behavior to remove orphaned widgets.

### Issues Solved

#6524 

### Checklist

### Further comments

Correct delete / display behavior occurs for collector and non-collector nodes. The grouping card appears to be borked on master again ( argh! ) so I want to fix that before further developing this branch. It is important to test the delete/display behavior for grouping cards.

### DIFF
--- a/arches/app/media/js/models/card.js
+++ b/arches/app/media/js/models/card.js
@@ -267,7 +267,6 @@ define([
         parseNodes: function() {
             var self = this;
             var attributes = this.sourceData;
-
             var nodeIds = [];
 
             var widgetNodeIds = ko.unwrap(this.get('widgets')).map(function(widget) {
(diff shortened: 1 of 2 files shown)


chrabyrd:
note: this is _not_ ready for review! The grouping card is borked on master and we need to fix that first so we cant test the grouping card widget interactions.

chrabyrd:
This card is ready for review. There is some minor funkiness: when deleting a node from a grouped card the grouping card's subtree doesn't show changes until clicked. _very_ minor nonbreaking error, but will file a followup

chrabyrd:
> seems like when a node is made into a card then the node isn't properly represented in the card tab
> 
> see: https://recordit.co/eOU028NhPz

This is now fixed! 🔧

PR #6556, opened 2020-08-19
6469 d3 upgrade

apeters:
re #6469


apeters:
@chiatt when you review this, can you also check that this fixes arches-for-science-prj/issues/#243

chiatt (approved):
Works as expected

Issue #6567, opened 2020-08-24
Unable to load reports for instances that do not have related resources


chiatt:
The following error occurs when loading reports for instances that do not have related resources:

![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/e8c5a1ea-f5ed-4c70-9264-856efc045a31)

chiatt:
Resolved in #6568

PR #6568, opened 2020-08-24
Fixes bug when rendering reports without related resources #6567

chiatt:
Fixes template error when rendering reports without related resources #6567


chrabyrd:
The documented error no longer occurs

Issue #6570, opened 2020-08-24
Tile data missing in card report for cards with cardinality n


chiatt:
Card report shows that a tile exists, but does not render the widget with its value. re arches/arches-for-science-prj#261

**This:**
![image](https://user-images.githubusercontent.com/1877663/91105646-720f3880-e625-11ea-9a88-16d3da7e8a79.png)

**Should look like this:**
![image](https://user-images.githubusercontent.com/1877663/91105608-59068780-e625-11ea-9b4a-70ad248a232d.png)

PR #6571, opened 2020-08-24
Fix issue with missing data in card reports. re #6570

chiatt:
Fixes issue with missing tile values in a card report in cards with cardinality n. re #6570


chrabyrd:
This fixes the issue! 🔧

Issue #6566, opened 2020-08-24
enable more customizable wastebin for workflows


whatisgalen:
The workflow step wastebin needs granularity to be able to specify (in the workflow params) whether to delete a generated tile for only that step, or a generated resource from only that step

created out of this project-specific ticket:
https://github.com/archesproject/arches-for-science-prj/issues/259

PR #6572, opened 2020-08-24
6566 customize workflow wastebin

whatisgalen:
### Types of changes

### Description of Change

This change enables more flexibility for the workflow developer to control what gets deleted if a workflow is abandoned via the "quit" button. By default, nothing gets deleted and any tiles or resource instances created from the workflow are retained. If the developer adds an object: `wastebin: {resourceid: null, description: 'an activity instance'}` to the list of params of a steps in a workflow, the resourceinstance whose resourceid is referenced by that step's params (on tile save) gets deleted on Quit. However given that there are more use cases than that, for example deleting select step-tiles, further customization on the workflow is necessary to achieve the desired behavior on Quit. Therefore this change enables granularity with respect to deleting specific tiles by way of declaring (for example) `wastebin: {tile: null, description: 'an activity timespan'}` for a particular workflow step params. Technical inline documentation is added to the relevant method in the NewTileStep viewmodel.

opened in tandem with this PR: https://github.com/archesproject/arches-for-science-prj/pull/265

### Issues Solved

#6566 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chiatt (approved):
Works as expected

Issue #6248, opened 2020-06-02
Allow a workflow developer to use alternate workflow buttons for a card


chiatt:
Workflow buttons should indicate that a user will advance upon saving a tile. This could be the labels on the card crud buttons e.g. 'Next' or 'Continue' instead of 'Add' See Dennis for more details.

PR #6586, opened 2020-08-27
6248 alt workflow buttons

whatisgalen:
### Types of changes

### Description of Change

This change introduces a 2nd case for loading buttons on a card. If the card has a property `altButtons` then those alternative buttons (presumably an array of objects with properties: label, onClick, classes, and visible) are loaded *only where that card is used in a specific workflow step which supplies such alternative buttons*. Otherwise the default buttons load. 

This was tested in workflows from Arches-For-Science, as well as resource editor (to ensure altButtons did not show up).
See this PR in the Arches-For-Science project for example buttons to test with: 
https://github.com/archesproject/arches-for-science-prj/pull/268

### Issues Solved

#6248 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/media/js/views/components/workflows/new-tile-step.js
```diff
@@ -35,6 +35,7 @@ define([
         this.alert = params.alert || ko.observable(null);
         this.complete = params.complete || ko.observable();
         this.completeOnSave = params.completeOnSave === false ? false : true;
+        this.altButtons =  params.altButtons || ko.observable(null);
```
robgaston, line 38: I think that `altButtons` should perhaps instead be two properties: one to hide the default buttons and one for additional buttons.  This would allow for reuse of the existing buttons alongside additional ones as well as make the templates slightly simpler

arches/app/templates/views/components/cards/default.htm
```diff
@@ -367,20 +367,30 @@ <h4 class="card-summary-name">
             <!-- /ko -->
             {% block form_buttons %}
             <div class="install-buttons">
+                <!-- ko if: !!ko.unwrap(card.altButtons) -->
+                    <!-- ko foreach: {data: card.altButtons, as: 'button'} -->
+                    <button data-bind="text: button.label, click: button.onClick($parent), attr: {'class': button.classes}, css: {disabled: (!$parent.card.isWritable && !self.preview)}, visible: button.visible($parent)" ></button>
```
robgaston, line 372: `css: {disabled: (!$parent.card.isWritable && !self.preview)}` seems problematic to me - I don't think we always want to disable additional buttons based on these conditions.  We should probably allow a user to define when a button should be disabled themselves.

I'm uncertain what this is: `visible: button.visible($parent)`; if we want to be able to control visibility of individual buttons, shouldn't this just be an observable?

robgaston (changes requested):

robgaston (changes requested):
I think that we should modify this bit to allow a developer to define when their button should be disabled: `css: {disabled: (!$parent.card.isWritable && !self.preview)}`

...perhaps something like: `css: {disabled: button.disabled}`

Issue #6398, opened 2020-07-14
Permissions: if a new user is added to the system, then they have access to resources that have been locked down


apeters:
new users added after permissions have been set on an instance receive full access to the instance but probably should have "no access"

![Screen Shot 2020-07-13 at 12.45.07 PM.png](https://images.zenhubusercontent.com/56f071cb6caf56c9d38e0f43/c3b101b6-4168-4f50-85bf-a1543f2bd9c6)

PR #6594, opened 2020-08-31
6398 instance permissions

apeters:
re #6398


apeters:
thanks for adding the indexing @chiatt !

PR #6609, opened 2020-09-02
Update migration to include changes to existing card configs. re #6574

chiatt:
Fixes migration to include changes to existing card configs rather than just to the default configs. re #6574


apeters (approved):
forward and backward migrations work as expected

PR #6612, opened 2020-09-05
Improve js package management

chiatt:
Makes arches a js dependency so that only project specific js file are in a project's package.json file.


chiatt:
The modules (there may also be others) are not downloaded when arches is added as a js dependency. I tested and confirmed they download properly when listed as a dependency in package.json rather than as a sub-dependency of arches.

- eonasdan-bootstrap-datetimepicker
- metismenu
- jquery-migrate

Issue #6613, opened 2020-09-05
enable customization for which card loads in NewTileStep and when


whatisgalen:
right now the logic for which cards follow this logic:
``ko.unwrap(this.resourceId) || ko.unwrap(params.graphid)``
thus if `this.resourceid` was set based on whatever the last step was, this would presume the same graph for the current step even if params.graphid is at odds with the corresponding graph of this.resourceid.

Generally speaking, if a workflow step could control which cards get seeded to the viewmodel, this would enable easier customization rather than having to abide by whatever case logic is hard-wired into NewTileStep.

Given a proposed solution of using an overwritable metthod like `seedCardGraphid`, it would require the cards not get loaded until after this overwritten method has been declared and called in an extending viewmodel. Thus, some control for timing is also needed.

PR #6614, opened 2020-09-05
6613 newtilestep card custom

whatisgalen:
Allow user to define the card used in a step regardless of the graph or resource instance defined in the previous step #6613


chiatt (changes requested):
@whatisgalen I'm holding off on reviewing this until archesproject/arches-for-science-prj#272 is completed

chiatt (approved):

PR #6619, opened 2020-09-09
Allow additional params to be passed into the search component

chiatt:
This is needed when it is instantiated outside of search, re archesproject/arches-for-science#274


chiatt:
@apeters This PR's branch needs to be used when testing  archesproject/arches-for-science-prj#274

apeters:
params are already being passed to the search filters, no need to add this code

Issue #6569, opened 2020-08-24
Hiding cards without tiles does not work in Tabbed Report


chiatt:
Widgets however do hide as expected

chiatt:
Resolved in #6879

PR #6620, opened 2020-09-10
6569 tabbed report hide tiles

whatisgalen:
### Types of changes

### Description of Change

This PR extends the sphere of the ``hideEmptyNodes`` observable into the card component so that cards without any tiles do not show up at all in the report. This also hides the ``<hr>``'s that would accompany each card (if that card is empty of tiles)

To test, navigate to a resource report with cards that do not have tiles, toggle the "Hide Null Values" switch to verify that empty Cards are indeed hidden entirely.
Test with any project.

### Issues Solved

#6569 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/templates/views/components/cards/default-card-report.htm
```diff
@@ -2,8 +2,8 @@
 {% load i18n %}
 
 <div data-bind="css: {'print-hide': !reportExpanded()}">
-    <!-- ko if: showHeaderLine--><hr class="rp-tile-separator"><!-- /ko -->
-    <div class="rp-card-section" data-bind="css: card.model.cssclass, visible: card.fullyProvisional() !== 'fullyprovisional'">
+    <!-- ko if: showHeaderLine && (hideEmptyNodes() === false || ko.unwrap(card.tiles).length > 0)--><hr class="rp-tile-separator"><!-- /ko -->
```
chrabyrd, line 5: `hideEmptyNodes() === false || ko.unwrap(card.tiles).length > 0`

This pattern persists throughout the change. Would it be more straightforward to alter/alter-and-rename the `hideEmptyNodes` val on the card component to use this pattern? That way the logic in the card report stays simple.

chrabyrd:
This appears to have some breaking behavior; it hides nodes behind semantic nodes and nodes grouped in other cards. To reproduce navigate to a Resource report that uses semantic nodes / nodes grouped in other cards and notice that their ( assumed valid ) children are not displayed.

chiatt:
Resolved in #6879

PR #6623, opened 2020-09-12
5032 prov edit

### LINKED ISSUE #5032: get_tile_data method doesn't return anything when there is more than 1 provisional edit
This will be an issue when calling functions (like the one on line 262 - https://github.com/archesproject/arches/blob/master/arches/app/datatypes/base.py#L262) expect a valid object.

https://github.com/archesproject/arches/blob/8acdc091f5e38d03c36c3140246d74377de12aff/arches/app/datatypes/base.py#L255-L260

I think if we're willing to return a provisional edit when there's only one, we should just return the first provisional edit no matter how many there are.

whatisgalen:
### Types of changes

### Description of Change

in the rare case that on tile save, more than one provisional edit exists for a given user *and* no authoritative data exists, an exception will be raised and logged that the provisional tile was invalid

### Issues Solved

fixes #5032 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/datatypes/base.py
+++ b/arches/app/datatypes/base.py
@@ -1,7 +1,7 @@
 import json, urllib
 from django.urls import reverse
 from arches.app.models import models
-from arches.app.models.tile import TileValidationError
+# from arches.app.models.tile import TileValidationError
 import logging
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


chiatt (commented):
It turns out multiple provisional edits can be put on a tile from collector. I've added logic for that and am giving to @apeters to review.

apeters (commented):
@chiatt when there is no data and no provisional edits then the system thows an exception because get_display_value expects an object.  Should we handle for this?

chiatt:
Having discussed this issue with @apeters we decided to hold off on additional work on this issue unless it becomes necessary in the future.

Issue #6632, opened 2020-09-18
Update resource-instance-select viewmodel so that it can be used select an instance outside of managing a node in an edit session


apeters:
Often times users want to select a resource instance for use in workflows or other contexts outside of managing node data.

Desired outcome:
- Update the component so that a node isn't required to instantiate the component
- Allow for graphids to be passed in directly to be used to inform the dropdown of what resources to display
- Hide the table of selected instances when managing a node is not desired

PR #6633, opened 2020-09-18
update resource instance component, re #6632

apeters:
re #6632

### DIFF
--- a/arches/app/media/js/viewmodels/resource-instance-select.js
+++ b/arches/app/media/js/viewmodels/resource-instance-select.js
@@ -1,13 +1,13 @@
 define([
     'knockout',
-    'jquery',
     'underscore',
     'viewmodels/widget',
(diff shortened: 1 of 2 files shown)


chiatt (changes requested):
@apeters I noticed that once you save a resource instance tile in graph with ontologies, you can no longer open the resource relationship to change or even see the ontology property. It looks like `graphIsSemantic` returns false on a saved tile, so these lines of the template are rendered: https://github.com/archesproject/arches/blob/dcdfd95dbe31d039e1e8e14197faea1937003f3c/arches/app/templates/views/components/widgets/resource-instance-select.htm#L76-L81

On master the user is able to change the ontology properties of a resource relationship after the initial save of a tile.

chiatt (approved):
Works as expected.

Issue #6130, opened 2020-04-23
Add a method to the tile proxy model to update values on a tile 


chiatt:
Add a method to the tile proxy model to update values on a tile creating the tile and it's parents if the tile to update does not exist.

ryan86:
Meets requirements.

PR #6636, opened 2020-09-21
Index resource when a tile is saved in the update_node_value api endpoint. re #6130 

chiatt:
Uses the Tile proxy model rather than the TileModel in the update_node_value endpoint to ensure that changes are indexed. re #6130


apeters (approved):
fixes the issue where tile updates wouldn't get re-indexed

PR #6641, opened 2020-09-23
Removes hard-coded filter names from filtersApplied computed. re #6542

chiatt:
Removes hard-coded filter names from filtersApplied computed. re #6542


njkim:
Works as expected

Issue #6647, opened 2020-09-24
Map doesn't support feature selection for non-arches layers


robgaston:
The map card viewmodel and popup template need some work to support generic layers.

PR #6648, opened 2020-09-24
allows for selection of non-arches layers in map card re: #6647

robgaston:
fixes map card viewmodel and template to support generic overlays


njkim:
Works as expected

Issue #3601, opened 2018-06-27
Remove all code that is now deprecated based on the new UI


apeters:
Remove all code related to old UI constructs:
- [x] Menus/forms - this will require a migration, also .json files contain references to forms (should those references be removed as well??)
- [ ] Permissions manger
- [ ] Graph manager
- [ ] Card manager
- [ ] Resource editor
- [ ] Report Manager
- [ ] all in-app help templates that are no longer used anywhere

mradamcox:
Work on this should also take into account urls.py.

PR #6658, opened 2020-09-28
Removes unused files re #3601

chiatt:
Removes unused files. re #3601


apeters (approved):
site still functions as normal

apeters:
fixes #3601

Issue #6659, opened 2020-09-29
Fix breaking error when loading popup of feature without permission information 


chiatt:
If a selected feature has no 'permission' data, then an error is thrown preventing the card from working properly. 
![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/d94fa56b-6290-4a11-bde0-df805b39cb37)

PR #6660, opened 2020-09-29
Moves permission logic from template to view model re #6659

chiatt:
Moves permission logic from template to view model and handles for error when permissions are not a JSON string. re #6659


chrabyrd:
had to change a call to ko.unwrap to unbreak the search page, but beyond that everything's looking and working well

chrabyrd (approved):

PR #6671, opened 2020-10-01
Removes check for absolute paths when loading related resources

chiatt:
Removes check for absolute path when loading related resources because relative paths are supported.


apeters (approved):
this looks good.  I'm so happy to be rid of the "isabs" check!

Issue #6473, opened 2020-07-30
Localize javascript files


chiatt:
Strings in JavaScript files also need to be localized
- localize client side files
    - mobile-survey-designer.js
    - map-editor.js
    - mobile-survey.js
    - resource-instance-select.js
    - workflow.js
    - more???

PR #6672, opened 2020-10-02
6473 localize js and .py files

apeters:
re #6473 

I think the best way to test this is just to confirm that the strings in the .js files have been moved over to using translated variables


chrabyrd:
Couldn't get this to repro. I can go through the pattern of translating the new strings and changing the app's language settings, but I can't get the end-user to see the updated strings. 

Current pattern is to change the map-tools strings (Add Line, Add Point, Add Polygon) and interact with the dropdown in Editor view with a map card. Happy to touch base synchronously or asynchronously to explain further

chrabyrd (changes requested):
Got this thing to successfully repro, but I'm noticing failing tests on Travis? If those are unrelated then this should be g2g

chrabyrd (approved):

chrabyrd:
ya tests are passing now and everything looks good 👍 .

Issue #6675, opened 2020-10-02
Export email link incorrect


csmith-he:
**Describe the bug**

The link within the email sent for a completed export contains a UUID (the export session UUID?) rather than a link to the ZIP file for download.

**To Reproduce**
1. Go to Search and zoom to a scale where you can see resource geometry on the map
2. Open Filter
3. Select 'Search by Map Extent' from the 'Select a filter' dropdown list
4. Close Filter and Open 'Search Export'
5. Retain all default values and enter your email address
6. Click Download button
7. Wait for email to be received and hover over the 'Download Now' link
4. See UUID rather than a valid link to a ZIP file

**Screenshots**
![image](https://user-images.githubusercontent.com/38533555/94945131-1ceefe00-04d2-11eb-81a7-aed61e76336e.png)

**Expected behavior**
The link would be to a ZIP file in the files folder for the Arches installation

**Your Arches Information**

*   Version used: stable 5.0.x
*   Operating System and version (desktop or mobile): Windows 10 Desktop
*   Browser Name and version: Chrome 85.0.4183.121 (Official Build) (64-bit)

**Additional context**
We have a fix in place in our application - this is us feeding the fix back to Arches Core now we know it works.

**Ticket Background**
*   Found by:  Historic England

csmith-he:
I'll send in a pull request for the fix next Monday

chiatt:
Resolved in #6677

PR #6677, opened 2020-10-05
Add email_link for href tag in export notify email

csmith-he:
As per bug #6675 , the e-mail template now uses email_link in the href tag so the URL where the export files are held can be accessed rather than just returning a UUID.

Passed unit tests run on local copy of code.


csmith-he:
@chiatt - bug fix for export email

Issue #6679, opened 2020-10-05
localize concept dropdowns so that the value displayed is the preflabel in the users chosen language 


apeters:
If someone saves a concept called "Concept 1" (in English) for a resource then show the Spanish version (if it exists) of the concept if the user is viewing the data in Spanish.

Always default to the preflabel.  EG, if a user has selected a label other than a preflabel, then still just show the preflabel in any other language.

apeters:
QA'd and approved by @chiatt

PR #6681, opened 2020-10-06
use localized concepts in the dropdown, re #6679

apeters:
re #6679


chiatt (approved):
Works as expected. The travis build is failing before tests are run, but tests pass locally. Also, there is a stickler error, but the formatting looks fine.

Issue #6685, opened 2020-10-07
number datatype node values show "None" when actually zero, hidden in report


whatisgalen:
**Describe the bug**

**To Reproduce**
Create a node in a graph that is number datatype. Save an instance where the node value is 0 (zero). Navigate to instance report and verify that the value is hidden altogether. Navigate to resource editor and verify that the node value shows "None" instead of 0 (zero).

**Screenshots**

![Screen Shot 2020-10-07 at 11 09 54 AM](https://user-images.githubusercontent.com/8366985/95370497-b3823d00-088d-11eb-9a35-685bc219710e.png)

**Expected behavior**

Expected behavior: node value should not be hidden in resource report (when "hide empty values" turned "on") *and* node value should show as "0" (zero) in resource editor, not "None"

**Your Arches Information**

*   Version used: master
*   Operating System and version (desktop or mobile): mac OS Catalina
*   Browser Name and version: mozilla firefox
*   Link to your Arches Install (optional):

**Additional context**

**Ticket Background**
*   Found by: @

PR #6694, opened 2020-10-08
Showing zero in report properly, re #6685

njkim:
### Types of changes

### Description of Change

It resolves the issue of consider 0 as null value in number report template, re #6685 
The fix will show zero in report instead of None and does not hide as null value when Hide_Null_Value = True

### Issues Solved

#6685 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chiatt (approved):
Works as expected

Issue #6693, opened 2020-10-08
if resource-instance name is number, then it comes across as a decimal


adamlodge:
I am linking one resource model to another resource model using the resource-instance datatype.  The name (per descriptor function) of the target model happens to be a number datatype.  The numbers that are the names are always integers, but the name is coming across as a decimal ie "2018.0". The correct display would be "2018".

Question:  Should we be able to configure number datatypes as integers, decimals, etc in the graph designer?

PR #6698, opened 2020-10-09
Add the logic to retain integer type, re #6693

njkim:
### Types of changes

### Description of Change

When an integer is entered to the number data type node, the type will remain the same, re #6693 
instead of being transformed into float

### Issues Solved

#6693 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chiatt (approved):
Works as expected

PR #6710, opened 2020-10-14
Related resource map card interface improvements. re #6574

chiatt:
Allows users to control visibility and configure the appearance of selection layer in the related resources map card. Users can also identify related resources using a map intersection with a geometry from a tile with a geojson feature.


apeters (approved):
functionally works as expected, but the UX could use a few improvements.  
EG: it would be nice to add the ability to intersect more than 1 feature.  Also, it would be nice if intersected features didn't clear already selected features.

Issue #6711, opened 2020-10-14
Improve UX in force-directed graph


chrabyrd:
**Describe the bug**

Our force-directed graph clumps new nodes in the upper-left hand corner, and cannot be 'zoomed out' from a scale of 1.0. This UX can be improved.

**To Reproduce**
1. Navigate to the search page
2. Choose a resource instance and press "Related Resources"
3. Interact with the graph

PR #6712, opened 2020-10-14
6711 cbyrd update graph ux

chrabyrd:
### Types of changes

### Description of Change

force-directed graph now has better UX. The double-scroll has been removed, the user can now 'zoom out' to a scale of 0.25, and nodes no longer clump in the upper-left corner. 

An issue remains, if the user selects 'Related Resources' from the search page before the application has completely loaded, some non-breaking incorrect node behavior persists.A good rule of thumb for testing this behavior is if the `Related Resources` link changes the cursor to a pointer. If it's not yet a pointer, the app has not completely loaded and we can observe the incorrect behavior.

### Issues Solved

#6711 

### Checklist


chiatt (approved):
Works as expected

Issue #6723, opened 2020-10-15
If a graph fails to be imported successfully, then it fails silently and the user isn't notified


apeters:
This occurs during package load as well as during graph import via the UI.

An easy way to test this is to load a graph that violates the sibling node naming issue assuming you're on master.

PR #6730, opened 2020-10-16
Inform user that graph failed to load

chiatt:
Informs a user that a graph failed to load and updates error message to notify a user that that graph import may have failed due to duplicate sibling node names or duplicate node names in a card. re #6723


njkim:
Works as expected

PR #6727, opened 2020-10-16
Add .gitignore template

chiatt:
Add a .gitignore template to the project template

PR #6734, opened 2020-10-19
Add migration for basemap and overlay configs

chiatt:
Add migration for basemap and overlay configs for map and related resource cards to support setting the default visible basemap and overlays in a card. re #6727


njkim:
works as expected

PR #6764, opened 2020-10-21
working load of Kibana objects during package load, re #6735

### LINKED ISSUE #6735: Add process to load and manage Kibana spaces and objects from within the package loading process
Including integrating within the package load process add management commands to manage Kibana spaces and objects

apeters:
fixes #6735 

To test this you should first update your Elasticsearch to version 7.9, then you should install Kibana version 7.9
Then, you'll need to edit the config/kibana.yml by adding these configs:
server.basePath: "/kibana"
server.rewriteBasePath: true
Then start Kibana

Then install the USDSSC package.  After the package loads you should be able to see the kibana dashboard in the site.

### DIFF
--- a/arches/app/views/api.py
+++ b/arches/app/views/api.py
@@ -111,6 +111,19 @@ class CouchdbProxy(ProtectedResourceView, ProxyView):
         return JSONResponse(_("Sync failed"), status=500)
 
 
+class KibanaProxy(ProxyView):
+    upstream = settings.KIBANA_URL
+
+    def dispatch(self, request, path):
+        try:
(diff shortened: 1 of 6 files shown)


chiatt (commented):
@apeters It seems like the changes to settings.py should also be reflected in the settings.py-tpl file.

apeters:
Thanks @chiatt I was going to do that, but just forgot!  Thanks for catching that!

chiatt (approved):
Works as expected

whatisgalen (approved):
works as expected

Issue #6765, opened 2020-10-21
Unable to add new nodes to a graph with associated tile data


chiatt:
When a user appends a node we run graph._update_node on every node to ensure graph integrity. Within this method we check if a node already has data under the assumption we would only be updating the appended node, not every node in the graph. So, if any node in the graph has data, the append fails. 

We should move the logic to check for data out of the update_node method and into graph.validate and only perform the check when a node's name is changing.

chiatt:
Fixed in #6779

PR #6766, opened 2020-10-21
Moves node.name validation logic to graph.validate, re #6765

chiatt:
Moves node.name validation logic to graph.validate to prevent error when appending new nodes to a graph, re #6765


apeters (approved):
checked all pathways in the code and everything looks solid.  Tried importing a graph with duplicate names as well and that fails to load as expected.

apeters (changes requested):
see below

apeters (changes requested):
There are 2 issues I found:
1. You can append the same branch several times without issue (violates sibling node constraint)
2. If you have 2 nodes that are parent and child.  They have the same name but the child node is in it's own card.  Then you change the child node to not be in it's own card.  The UI then hangs.

chiatt:
Closing because the commits in this PR are now also in #6779 which addresses the comments in @apeters review.

PR #6771, opened 2020-10-21
Strictly defined versions for metismenu, moment-timezone and datepicker

chiatt:
Strictly defined versions for metismenu, moment-timezone and datepicker


apeters (approved):
I tested this fix and it had all the correct package versions as specified

robgaston (approved):

Issue #6774, opened 2020-10-22
User cannot append the same branch several times without issue 


chiatt:
If a user tries to append the same branch several times the node name constraint is violated.

PR #6779, opened 2020-10-23
Resolve node name conflicts when appending branches

chiatt:
Resolves node name conflicts when appending branches and when moving a node into a different nodegroup.  #6774, #6765


chiatt:
Travis is taking a long time to build so I don't want to wait for the CI checks to pass. This now works as expected and tests pass locally.

PR #6780, opened 2020-10-23
Updates 5.1.x with latest changes in master

chiatt:
Adds support for Kibana dashboards
Improves resource instance permission UI
Removes unique constraint on function components


chiatt:
Travis seems to be stalled but tests pass locally.

PR #6784, opened 2020-10-23
handle for when Kibana is not running, re #6783

apeters:
re #6783 

to test, turn kibana off, then load the USDSSC package.  You should notice a warning that Kibana isn't running but the package load should proceed as usual.


adamlodge:
works as expected

Issue #6801, opened 2020-10-28
Cannot save changes to node with tile data


chiatt:
This restriction is should be applied only when the name is being changed, however, it is applied any time a node is saved.

chiatt:
Fixed in #6808

PR #6808, opened 2020-10-29
Throw error on node save when name is changed

chiatt:
Only prevents save of a node with data if the node's name is being changed.  re #6801


chiatt:
To QA just try to change a report template for a graph with resource instance data. It should save without a node constraint error.

apeters (approved):
fixes the error when saving a graph where no node names changed

Issue #6842, opened 2020-11-04
Concept datatype error messages are formatted for import 


chiatt:
Concept datatype error messages are formatted for import when they can appear UI or anywhere else where datatypes are used. This can be confusing for web users.

chiatt:
Resolved in #6844

PR #6844, opened 2020-11-05
Cleans up / localizes datatype validation error messages. re #6842

chiatt:
Refactors creation of most datatype error messages with a function on the base datatype and adds localization support for datatype error messages. re #6842


apeters (approved):
everything seems to work as expected and is a nice refinement

Issue #6807, opened 2020-10-28
Add setting to skip card constraint and required nodes check


chiatt:
If a user is confident that their data meets the necessary constraints on import, these checks can be skipped with a setting.

PR #6860, opened 2020-11-10
Add setting to skip validation and arches functions on import

chiatt:
Adds settings to skip validation and arches functions on import. re #6807


chiatt:
This tile validation bypass is addressed in https://github.com/archesproject/arches/pull/6948.

Issue #6859, opened 2020-11-09
Widget configs that store boolean values break dirty state comparison


chiatt:
In the configJSON computed in card-widget.js, a the check for a config key will evaluate falsy if the does not exist, but also key exists and the value for key is false. Instead it should check for the existence of the key rather than for falsiness.

chiatt:
Resolved in #6863

Issue #6839, opened 2020-11-03
Disable number and string widgets to support calculated fields


chiatt:

chiatt:
Resolved in #6863

Issue #6752, opened 2020-10-20
Boolean switch saves 'None' and 'true', but not 'false'


chiatt:
The boolean switch widget should support None, true, and false. The radio widget works as expected.

chiatt:
Resolved in #6863

PR #6863, opened 2020-11-11
Adds config to disable string and number widgets

chiatt:
Adds editable flag to number and string widgets. Fixes bug saving configs that store boolean values. re #6839, #6859, #6752


chiatt:
To QA: 
Add a new node group with a string, number, and boolean node.
In the card manager, confirm that you can toggle the disabled state of the text and number widgets.
For the boolean node, change the widget to the switch and confirm that you can change the default value from none, false and true.

Issue #6869, opened 2020-11-11
if nodes referenced in the primary descriptors function are null then the word "None" appears in the string


apeters:
by default we shouldn't show the word "None" and instead just replace with an empty string.  Also, if the entire function returns a blank string then we should return the string "Undefined"

apeters:
closed by #6871

PR #6871, opened 2020-11-11
6869 primarydescriptors function updates

apeters:
re #6869


adamlodge:
detects if inputs to the descriptor function are null and handles appropriately

Issue #6809, opened 2020-10-29
Defer resource indexing until after package load


chiatt:

chiatt:
Resolved in #6874

PR #6874, opened 2020-11-13
6809 defer index on package load

apeters:
re #6809

### DIFF
--- a/arches/app/models/resource.py
+++ b/arches/app/models/resource.py
@@ -113,6 +113,7 @@ class Resource(models.ResourceInstance):
             raise ModelInactiveError(message)
         request = kwargs.pop("request", None)
         user = kwargs.pop("user", None)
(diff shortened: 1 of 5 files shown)


chiatt (commented):
Maybe the language of the command argument should be something other than 'prevent_index'? Prevent may imply that indexing will not occur as part of the command when in fact it still does. 

Also, perhaps we should extend deferring the index for concepts as well if we are going to reindex the entire project anyway.

apeters:
@chiatt I think there are actually a few scenarios:

doing a package load in which indexing is deferred until after all business data is loaded, but that method makes calls to individually load business data and that command doesn't actually ever index data (hence prevent_index) it's the calling method that eventually indexes data.  

So deferring indexing in the context of package load makes sense, but I think in the context of the loading business data method I think there are a few different options
- preventing indexing entirely
- deferring indexing
- indexing during save (what we do now).

Issue #6880, opened 2020-11-16
Remove the error messages printed when loading system settings.


chiatt:
The error messages in the handled import exceptions in settings.py should be removed because they imply that there is a problem when Arches is working without issue.

https://github.com/archesproject/arches/blob/b2dc29b5ff9a8041ce8494a59cd60f9959886f8a/arches/settings.py#L670-L679

chiatt:
Resolved in #6881

PR #6881, opened 2020-11-16
Removes settings local error messages. re #6880

chiatt:
Removes settings local error messages. re #6880


apeters (approved):
looks good to me

PR #6892, opened 2020-11-20
6809 defer index on package load

apeters:
re #6809


apeters:
to test, load a package (not USD) once with the standard arguments and observe that the concepts and resources only get indexed at the very end (You can use the Elasticsearch head extension to confirm this)

Next load the package again but add the flag `-di false` (to set deferred indexing to false) and observe that the index starts populating during the load.

chiatt (approved):
Works as expected

Issue #6903, opened 2020-11-24
json output of a resource instance fails when the root node of a tile is a semantic node and the data was created via the UI


apeters:
data created upon import is ok because semantic nodes are included (incorrectly?) with imported data

See below: the screen shot is what is in the database, the json output is the representation of that data
the "Semantic node" is a card with a semantic node as the root node in the nodegroup (the "testing" value was imported)

![Screen Shot 2020-11-24 at 3 34 01 PM](https://user-images.githubusercontent.com/823773/100163419-7e946e80-2e6a-11eb-8ce9-5297311a8daa.png)

```
{
    "displaydescription": "undefined",
    "displayname": "undefined",
    "graph_id": "55f0a07a-2e97-11eb-a031-acde48001122",
    "legacyid": "5d756a1c-e7cd-46ec-a036-f0bcdf38a82f",
    "map_popup": "undefined",
    "resource": {
        "Non Semantic Node": [
            {
                "Number Node": "26",
                "Semantic Node": "NON_DATA_COLLECTING_NODE"
            },
            {
                "Number Node": "3"
            }
        ],
        "Semantic node": {
            "String Node": "testing"
        }
    },
    "resourceinstanceid": "5d756a1c-e7cd-46ec-a036-f0bcdf38a82f"
}
```

apeters:
closed by #6904

PR #6904, opened 2020-11-24
fix bug, where semantic nodegroup root nodes won't render a tile in t…

apeters:
re #6903


chrabyrd:
( In my mind ) new logic splice shoulda come with a test case, but we as an org haven't landed on a firm testing pattern so meh I'm good for now 👍

Issue #6913, opened 2020-12-02
Map filter geometry is not properly buffered using GDAL v3


chiatt:
Buffering geometries within the map filter can cause an error or incorrect results if GDAL v3 is used.

chiatt:
Resolved in #6914

Issue #6850, opened 2020-11-06
Implement a consistent logging pattern across Arches


chiatt:
Currently arches still relies on print statements in many parts of the application. We should standardize logging across the application (especially in models and views) to give implementors more insight into issues in Arches.

chiatt:
This has largely been completed. Rather than addressing this as a single issue print statements will be changed to log statements as the occur in PRs

PR #6914, opened 2020-12-02
Buffers map filter geom with postgis rather than the geo api, re #6913

chiatt:
Buffers map filter geom with postgis rather than the geo api and replaces print statement with log statement, re #6913, #6850


robgaston:
looks good!

Issue #6924, opened 2020-12-05
improve the way custom indexes are used in the system


apeters:
add a reindex method, and only reindex after all managed resources have been indexed

chiatt:
These changes have been merged into 5.1.x

PR #6925, opened 2020-12-05
add better feed back during data indexing operations, seperate indexing of resources from custom indexes

apeters:
re #6924


chiatt (approved):
Works as expected

PR #6928, opened 2020-12-07
Files changed for app specific email config

csmith-he:
Changes as detailed in Arches ticket 6673 for stable/5.1.x branch

- Replaces hard-coded 'from' email address in models.py with email address set in settings.DEFAULT_FROM_EMAIL
- Replaces hard-coded application name with the application name set in settings.APP_NAME
- Replaces "You're" in subject line for user update email with "Your"
- Adds DEFAULT_FROM_EMAIL to settings.py and settings.py-tpl


chiatt (approved):
Works as expected. Thanks @csmith-he!

Issue #6941, opened 2020-12-11
allow users to hide the status bar during indexing of resources


apeters:
We need to be able to do this because when running an index operation in celery, celery errors when the status bar is printed to the screen

chiatt:
Addressed in  #6942

PR #6942, opened 2020-12-11
add a command paramter to hide status bar output during indexing operations

apeters:
re #6941


apeters:
test this by calling 
this should hide the status bars
`python manage.py es reindex_database -q  `

chiatt (commented):
Works as expected

Issue #6650, opened 2020-09-25
import_business_data bypasses cardinality constraints


whatisgalen:
**Describe the bug**

if a card in a resource model has a cardinality of 1 such that only 1 tile can exist for that card, this constraint gets enforced on the front end but it does not get enforced when a user uses a command-line command such as import_business_data where overwrite is set to "append". This assumes the business data being imported appends a tile to a node on a resource instance where a tile already exists.

**To Reproduce**
1. create a graph with a card with a cardinality of 1
2. create a csv file to import data for that card
3. run the import command more than once and notice that multiple tiles can be saved to the card with cardinality 1

PR #6945, opened 2020-12-14
Supplement cardinality constraint error feedback with card/tile details

chiatt:
Provide constraint violation details so that user can identify the problematic tiles/card re #6650


apeters (approved):
I observed additional feedback as intended

PR #6948, opened 2020-12-15
Adds settings to allow tile validation to be bi-passed

chiatt:
Adds settings to allow tile validation to be bi-passed, re #6807


apeters (approved):
this looks good to me and works as expected

Issue #6908, opened 2020-11-26
Allow users to get geojson back from a search export query


chiatt:
Arches should allow users to define a search query and get geojson in the response to support clients such as QGIS.

chiatt:
Resolved in #6908

PR #6961, opened 2020-12-17
Fixes geojson search endpoint url when localization in applied. re #6908

chiatt:
Geojson search endpoint URL properly formed even if a language identifier is specified in the URL. re #6908


njkim:
works as expected

Issue #6969, opened 2020-12-21
Develop alternate workflow tabbed template


chrabyrd:
This is to catch changes to arches-core for https://github.com/archesproject/arches-for-science-prj/issues/288

PR #6970, opened 2020-12-24
tabbed workflows

chrabyrd:
### Types of changes

### Description of Change

Workflows can now accept a `tabbedWorkflow` param, which changes the UI to use tabbed workflows. Currently we're just splitting logic in the workflow template but if this becomes more complex we should properly break out the components.

Button logic and 'Guidance box' logic will be included in followup tickets.

Screen Shot 2020-12-23 at 3 59 01 PM

### Issues Solved

#6969 

### Checklist


chiatt (approved):
Works as expected

Issue #6977, opened 2021-01-04
Widget Manager hidden from view in Card Designer


lindsgant:
**Describe the bug**

In the Card Designer of our [DISCO implementation](http://54.183.150.116/graph_designer/0b9235d9-ca85-11e9-9fa2-a4d18cec433a), we are unable to see the Widget Manager without expanding our browser page. The width of the Arches page appears to exceed the browser width. The only way to view this Widget Manager is to expand the browser screen, sometimes across two monitors, as a scroller for page width is not available.

Please see recording of error here: https://recordit.co/El9HgyPKqs

Appears to be similar to open issue: https://github.com/archesproject/arches/issues/6563

**To Reproduce**

1. Go to http://54.183.150.116/graph_designer/0b9235d9-ca85-11e9-9fa2-a4d18cec433a
2. Click on Arches Designer (linked above)
3. Select Cards
4. Scroll down to Scale of Project node underneath Scale of Project to view radio dials
4. Expand screen to the right in order to see Widget Manager
5. Watch error recording at: https://recordit.co/El9HgyPKqs

**Screenshots**

https://recordit.co/El9HgyPKqs

**Expected behavior**
Widget Manager should be in view from Card Designer without scrolling or expanding screen. This may simply be an HTML or CSS fix.

**Your Arches Information**

*   Version used: 5.1
*   Operating System and version (desktop or mobile): Windows and macOS (desktop)
*   Browser Name and version:  Chrome 87.0.4280.88 (Official Build) (x86_64) and Firefox 84.0.1 (64-bit)
*   Link to your Arches Install (optional): http://54.183.150.116/graph_designer/0b9235d9-ca85-11e9-9fa2-a4d18cec433a

**Ticket Background**
*   Found by: George Bruseker; filed by @lindsgant

chiatt:
I suspect the issue is caused by the concept-radio-select widget. We should probably modify that so that it wraps the options rather than pushing out the width of the card.

chiatt:
re #6725

chiatt:
Hey @lindsgant, I think I've got this fixed now. When you get a chance, can you pull the latest from stable/5.1.x and confirm?

lindsgant:
[...]

Issue #6725, opened 2020-10-16
Setting domain values to domain-radio-widget on card designer makes other configs fall off the page


adamlodge:
Hopefully this image is illustrative.  I can demonstrate if helpful.
![Screen Shot 2020-10-15 at 8.26.45 PM.png](https://images.zenhubusercontent.com/56f150f76caf56c9d38e16f8/d7d2a877-516d-4a8f-8918-3a0d9036b436)

chiatt:
Fixed in #6996

PR #6996, opened 2021-01-08
Fixes width of radio and checkbox widgets 

chiatt:
Fixes width of radio and checkbox widgets, re #6977, #6725


njkim:
works as expected

PR #7047, opened 2021-01-27
181 2 alignment of null values

apeters:
re fargeo/usdssc#182


apeters:
all tests passed locally

Issue #7088, opened 2021-02-05
Error saving tiles in a new project


chiatt:
@chiatt commented on [Thu Feb 04 2021](https://github.com/archesproject/arches-for-science-prj/issues/322)

To reproduce, create a new project, run setup_db and start the server.

Go to the system settings and change any existing tile and you should see the following error:
![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/d7a0894d-60d0-4a49-ba13-1e567b93d54d)

This occurs on both 5.1.x and 5.2.x.

chiatt:
Resolved in #7091

PR #7091, opened 2021-02-05
fix tile cardinality validation bug, re #7088

apeters:
re #7088


ryan86:
Code works but tests failing.

Issue #7095, opened 2021-02-07
CSV uploaded images do not appear in image header report template


chiatt:
If a user uploads images from csv, they will appear in the body of a report, but not in the report header.

chiatt:
Resolved in #7097

PR #7097, opened 2021-02-08
Fix missing images from csv import in report header

chiatt:
Fixes bug preventing images from appearing in the report header when loaded during CSV import. re #7095


chiatt:
To QA, just add the following model and business data to a package, load the package and confirm that 3 images show up in the report header: 
[ImageTesting.zip](https://github.com/archesproject/arches/files/5945637/ImageTesting.zip)

ryan86:
Meets requirements

Issue #7099, opened 2021-02-08
Relations import ignores resource instance graphid fields


chiatt:
When importing a relations file the `to` and `from` graphids are not applied to the ResourceXResource instance. The load will run without error, but the relationships will not appear in the report's related resources section.

chiatt:
Fixed in #7100

PR #7100, opened 2021-02-08
Add to and from graphids to relations import. re #7099

chiatt:
Add to and from graphids to relations import. re #7099


chiatt:
Merging without a review because it's a small change and the fix was confirmed by @annabelleee

Issue #7120, opened 2021-02-17
Allow users to specify a cluster distance below 1 pixel


chiatt:
This will allow users better control over their clustering distance and improve clustering performance for some projects.

chiatt:
Resolved in #7121

PR #7121, opened 2021-02-17
Allow float values for cluster distance, re #7120

chiatt:
Allow float values for cluster distance, re #7120


robgaston (commented):
looks good!

robgaston (approved):

PR #7124, opened 2021-02-18
adds coordinate editor feature to map card and widget

robgaston:


chiatt (approved):
Crud operations, coordinate system, vertex reordering all work as expected.

Issue #7142, opened 2021-02-25
Error when saving files with same name using file_list datatype


chiatt:
If a user tries to save the same file twice (or save a file with the same name) they get an error 'file already exists'. This is because in the datatype we use ```models.File.objects.get_or_create(pk=file["file_id"], path=file_path)```.  Rather than 'get' a 'create' is called because in some cases django has created a new file name which is assigned to the file_path variable.

chiatt:
Resolved in #7144

PR #7144, opened 2021-02-25
Fix error when saving files with the same name. re #7142

chiatt:
Prevent error when using get_or_create to save files that already exist.


njkim (approved):
Works as intended

Issue #7167, opened 2021-03-05
Missing 'nodeid' property in geojson api response


chiatt:
Features returned from the geojson api include a nodeid property. The problem is that sometimes the `i` in nodeid is uppercase and sometimes it is lowercase depending on the source of the geometry.

chiatt:
Resolved in #7169

PR #7168, opened 2021-03-05
Ensures nodeid is in feature.properties, re #7167

chiatt:
Ensures `nodeid` is in feature.properties regardless whether `nodeId` already exists, re #7167


chiatt:
Closing the PR because the CI build keeps failing during the python dependency installation.

PR #7169, opened 2021-03-05
Ensures nodeid in geojson api's feature.properties, re #7167

chiatt:
Ensures nodeid is in feature.properties regardless whether nodeId already exists, re #7167


chiatt:
To QA:
1.  Unzip and load this Resource Model: 
[Robot.json.zip](https://github.com/archesproject/arches/files/6094509/Robot.json.zip)

2. Create some geometries in the map card
3. Then check that nodeid is lower case in api response: 
     http://localhost:8000/geojson?nodeid=88c0c3fc-7e1c-11eb-b36c-acde48001122

apeters (approved):
this fixes the isses

Issue #7178, opened 2021-03-06
Dirty state indicated in related resources map card within workflow 


chiatt:
When some widget values are changes in the related resources map card within workflow, the dirty state is not always activated.

chiatt:
Resolved in #7179

PR #7179, opened 2021-03-06
Ensures dirty state is activated in new tile step, re #7178

chiatt:
Ensures dirty state is activated in new tile step, re #7178


chiatt:
To QA:
1.  Unzip and load this Resource Model: 
[Robot.json.zip](https://github.com/archesproject/arches/files/6094509/Robot.json.zip)

2. Create some geometries in the map card
3. Then check that nodeid is lower case in api response: 
     http://localhost:8000/geojson?nodeid=88c0c3fc-7e1c-11eb-b36c-acde48001122

njkim (approved):
Work as expected

Issue #7108, opened 2021-02-10
Update manifest manager main menu


chiatt:
The manifest manager main menu should be updated in accordance with the latest mockups: https://www.dropbox.com/home/Arches4/Mockups/AfS?preview=Manifest+Manager.pdf

chiatt:
Resolved in #7206

PR #7198, opened 2021-03-11
Update manifest start page

chiatt:
Updates manifest start page to reflect mockups and renames Manifest Manager, Image Service Manager, re #7108


njkim (approved):
Works as expected

Issue #7136, opened 2021-02-24
Update "new-multi-tile" workflow step to support the new workflow logic


njkim:

chrabyrd:
fixed in 7137

PR #7199, opened 2021-03-12
Make Component-based-step handle tile-based components

chrabyrd:
### Issues Solved

#7136 

To test: use this data in the arches-HER consultation-workflow.js file: https://pastebin.com/D7nfVuwz


chiatt (approved):
Resolves #7136, Stickler error is an error in the linter rather than the linted code.

PR #7205, opened 2021-03-16
Handle Windows .astimezone bug

jkemper-pca:
*Recreating the PR after correctly linking my GitHub account to the commits at Adam Cox's suggestion*

There is a bug in python when calling `.astimezone()` on a Windows machine for `datetime` values before 1970 (start of epoch). [Here](https://bugs.python.org/issue36759) is the python bug post about it.

### Types of changes

### Description of Change
I've added a check for dates when they are being converted to first see if the date falls under the error specifications and, if so, it simply skips the timezone conversion. This isn't supposed to be a permanent fix and when python fixes the bug it will no longer be needed. The reason I thought the change was worth doing now though is that when I'm trying to import data through CSV files using the CLI there is no way to get dates before 1970 to import correctly and I don't know how long before python fixes this bug.

### Issues Solved

#

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -341,12 +341,14 @@ class DateDataType(BaseDataType):
                     # should never happen but don't want a infinite loop
                     raise Exception("Backup timezone conversion failed: no matching year found")
             return new_year
+
         def is_same_calendar(year1, year2):
             year1_weekday_1 = datetime.strptime(str(year1) + "-01-01", "%Y-%m-%d").strftime("%w")
             year1_weekday_2 = datetime.strptime(str(year1) + "-03-01", "%Y-%m-%d").strftime("%w")
             year2_weekday_1 = datetime.strptime(str(year2) + "-01-01", "%Y-%m-%d").strftime("%w")
             year2_weekday_2 = datetime.strptime(str(year2) + "-03-01", "%Y-%m-%d").strftime("%w")
             return (year1_weekday_1 == year2_weekday_1) and (year1_weekday_2 == year2_weekday_2)
+
         converted_dt = dt.replace(year=same_calendar(dt.year)).astimezone().replace(year=dt.year)
         return converted_dt
 
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -337,16 +337,15 @@ class DateDataType(BaseDataType):
             new_year = 1971
             while not is_same_calendar(year, new_year):
                 new_year += 1
-                if new_year > 2020:
-                    # should never happen but don't want a infinite loop
+                if new_year > 2020: # should never happen but don't want a infinite loop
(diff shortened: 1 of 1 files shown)


chiatt:
Hi @jkemper-pca, thanks for this PR. Before we review it, would you mind updating your branch with the latest from master and pushing again? I merged some changes into master late yesterday that include a fix to the CI build which includes a run of our unit tests.

arches/app/datatypes/datatypes.py
```diff
@@ -320,10 +321,14 @@ def transform_value_for_tile(self, value, **kwargs):
             value = value[0]
         valid_date_format, valid = self.get_valid_date_format(value)
         if valid:
-            value = datetime.strptime(value, valid_date_format).astimezone().isoformat(timespec="milliseconds")
+            v = datetime.strptime(value, valid_date_format)
         else:
             v = datetime.strptime(value, settings.DATE_IMPORT_EXPORT_FORMAT)
-            value = v.astimezone().isoformat(timespec="milliseconds")
+        if (not sys.platform.startswith("win")) or (
```
chiatt, line 327: It would be better to use a try/except block here instead of checking the OS because if a fix is ever implemented in Python, we won't have to remember to come back and change this code to get things working properly again.

arches/app/datatypes/datatypes.py
```diff
@@ -320,10 +321,14 @@ def transform_value_for_tile(self, value, **kwargs):
...
+            v >= datetime.fromtimestamp(0)
+        ):  # The .astimezone() function throws an error on Windows for dates before 1970
+            v = v.astimezone()
```
chiatt, line 330: Unfortunately we can't skip astimezone because it is necessary to properly format the tilevalue to ensure that the value works properly with search. Also, depending on how the incoming data is formatted, the data will fail without calling astimezone. Instead of skipping astimezone, what is needed is equivalent logic.

chiatt (changes requested):
Thanks again @jkemper-pca for this PR. I've added a couple of comments. I think the astimezone method will have to be replaced rather than bypassed to prevent causing other problems. I've created a ticket for this fix. https://github.com/archesproject/arches/issues/7217

chiatt (commented):
Thanks @jkemper-pca for your latest commits. Sorry I've been very busy, I just wanted to let you know that I've looked at the changes you've made. They look good, but I need to do some testing before merging. I'll try to get to it soon.

chiatt (approved):
This is a clever solution and it works great in most cases. There can be an issue with governments changing their daylight savings policy over time that might affect the timezone calculation. If a date and time prior to 1970 is imported into arches and then the timezone is calculated for a year after 1970, the timezone policy could be different between those years and may end up offsetting the time by an hour or so. 

Nevertheless, most dates should be unaffected, and this is a big improvement over not being able to import dates prior to 1970 at all on a Windows server.

PR #7206, opened 2021-03-16
Manifest manager start page improvements

chiatt:
Makes entire start page tab content a dropzone and changes tab label cursor to pointer. re #7108


njkim (approved):
Works as expected.
The user experiences are much improved.

Issue #7171, opened 2021-03-05
Create api endpoint to support existing object validation (resources, tiles, etc...)


apeters:
we'd like to be able to validate objects that already exist in the system (resources, tiles, etc...). 
This could be used in the future in a post data load step.

azaroth42:
Improvement request: The invalid error message when the search doesn't match is a little too specific that the resource doesn't exist ... it might exist, just not currently match the search criteria.

For example:

* I created a concept resource "Primary Name" and set the search for name types.
* I created a Person with a Name and set the type of name to the Primary Name resource.
* I validated and made certain it was valid.
* I then added a different classification which made Primary Name not valid for that tile
* Now validation generates the error message:

> Resource id: 606359f5-4bab-470f-b97c-e63214aeca27 is not in the system. This relationship will be added once resource 606359f5-4bab-470f-b97c-e63214aeca27 is loaded.

(Whereas 606... does exist, it's not just no longer valid for this tile any more)

azaroth42:
When verbose is false, the validation seems to return true even when the resource is not valid (as per above)

chiatt:
@azaroth42 It looks like these issues are addressed in #7328

PR #7211, opened 2021-03-17
initial commit on an api to validate resource instance and tile objects in the system

apeters:
re #7171

example requests to test:
http://localhost:8000/validate/resource/{resource id}
http://localhost:8000/validate/tile/{tile id}

you can throw in a "raise" statement in any datatype "validate" function to see an example of output when valid is false


chiatt (approved):
Resolves #7171

Issue #7210, opened 2021-03-17
arches-project create fails because arches branch in package.json is incorrect


chiatt:
The branch in 5.2.x should point to the stable/5.2.x branch.

azaroth42:
Thanks Cyrus!

chiatt:
Fixed in #7212

PR #7212, opened 2021-03-17
Set proper branch to package.json when version is a beta or '.x' branch

chiatt:
Assigns the proper branch name to package.json depending on whether the branch is a final, beta (.x branch), or alpha (master) branch, re #7210


apeters (approved):
this fixed the issue

Issue #7214, opened 2021-03-17
Remove check for request in tile.check_for_missing_nodes


chiatt:
Admins can now control if the required node check is enforced with the BYPASS_REQUIRED_VALUE_TILE_VALIDATION setting, so checking for a request object is no longer necessary

chiatt:
Fixed in #7215

PR #7215, opened 2021-03-17
Check for missing nodes from web and cli, re #7214

chiatt:
Removes 'request' argument from tile.check_for_missing_nodes because admins can control if the check is run or not using the `BYPASS_REQUIRED_VALUE_TILE_VALIDATION` setting.


apeters (approved):
this refactor makes sense

Issue #7224, opened 2021-03-19
url datatype


adamlodge:
Add url datatype to core Arches per https://github.com/thegetty/arches-urldatatype

chiatt:
Completed in #7226

PR #7226, opened 2021-03-19
7224_url_datatype

apeters:
re #7224


adamlodge:
It works in the brach.

Issue #7227, opened 2021-03-19
Prevent steps from getting unnecessarily skipped


chiatt:
Some step in the arches-her workflows are getting unexpectedly passed over without allowing the user to complete them.

chiatt:
Fixed in #7228

PR #7228, opened 2021-03-20
Prevents non-required steps from getting skipped

chiatt:
Prevents non-required steps from getting skipped due to workflow.next getting called twice - once when a user clicks `Save and Continue`, and a second time during the subscription on complete, re #7227


chrabyrd (approved):
works as expected!

Issue #7234, opened 2021-03-23
core changes to accommodate thematic report templates


chrabyrd:
related to https://github.com/archesproject/arches-for-science-prj/issues/329

PR #7235, opened 2021-03-23
add changes to support thematic report templating #7234

chrabyrd:
small changes to support https://github.com/archesproject/arches-for-science-prj/pull/332

#7234


chiatt (approved):
Nice improvement to support custom report development. Given that a footer block has been added here, I wonder if we should consider moving the [report.htm footer](https://github.com/archesproject/arches/blob/f4215fd165bf9386b3f6429a58f1a66c5df8f6d9/arches/app/templates/views/resource/report.htm#L38)?

Issue #7237, opened 2021-03-23
concept-select incorrectly persists most recently set value


chrabyrd:

chrabyrd:
fixed in https://github.com/archesproject/arches/pull/7259

PR #7238, opened 2021-03-23
fix concept select persistence

chrabyrd:
This fixes the issue with concept-select-widget incorrectly setting a defaultValue on value select.

To reproduce, interact with a concept-select-widget and notice how it now behaves as it should.

#7237


chrabyrd:
This was fixed in https://github.com/archesproject/arches/pull/7259

Issue #7252, opened 2021-03-25
API to output resource instances as json-ld fails in some instances


apeters:
**Describe the bug**
Export of Resource Instances in json-ld format can fail occasionally with a NoneType is not subscriptable error.

This may be due to the change we made to graphs that allow for data collecting nodes to have child nodes as well.

chiatt:
Resolved in #7264

PR #7254, opened 2021-03-26
allow the domain node to be in a parent tile, re #7252

apeters:
re #7252


apeters:
Here's how to test this:

- Create a new graph.
- Add a non-literal data collecting node (concept for example)
- Add a child node (any datatype is fine) to this node and put it in it's own card
- Create a new resource instance of this graph and populate all nodes
- view the json-ld output of this instance in the browser (http://localhost:8000/resources/{resource instance uuid}?indent=4)
- Observe the output 
- View the output again, but without the applied fix
- Notice that a 500 error is thrown

chiatt (approved):
Resolves #7252

Issue #7247, opened 2021-03-24
Photo workbench panel shows all card nodes, even if they are configured as invisible in the card designer


dwuthrich:
In the photo gallery card, the edit panel shows all nodes in the card even if a node's visibility parameter has been configured as invisible.

This behavior can be seen in the Arches-HER package: add an image to the photos branch on the Consultations site visit card and compare to how the card is configured in Arches Designer

chiatt:
Fixed in #7258

PR #7258, opened 2021-03-27
In the photo gallery card show only widgets configured as visible, re #7247

chiatt:
In the photo gallery card show only widgets configured as visible, re #7247


njkim (approved):
fixed the issue of showing non-visible widget

Issue #7240, opened 2021-03-24
multi-tile configuration of component-based-step doesn't progress on save after edit


chrabyrd:
In a workflow, the multi-tile configuration of component-based-step doesn't properly progress the workflow && empty the form if the tiles have been edited and resaved. This can cause issues if the user then interacts with the form.

PR #7259, opened 2021-03-27
component-based-step  logic rewrite

chrabyrd:
Major changes to how workflows function. Should now have correct dirty logic and autoAdvance logic. Can also now handle grouping card components. 

To test, navigate the `Consultation` arches-her workflow and the AfS `Sample` workflow. That should cover all current cases ( no tile, single tile-grouping, single tile-non-grouping, and multi-tile) . 

At the time of writing these have been tested with no adverse effects.

resolves: #7240 #7244 

also allows for true async save/complete logic


chiatt (approved):
resolves #7244 and #7240

Issue #7267, opened 2021-03-25
Workflow Multi-tile step shows all nodes, even hidden nodes


dwuthrich:
In the Site visit workflow, the site visit attendees step allows the user to add information about people who attended a site visit.

After entering and saving attendee data, Arches shows the user's input as a tile on the right side of the workflow step.  As currently implemented, the widget shows the user's input AND ANY ADDITIONAL nodes on the card, even those configured as invisible.

What should happen: the widget should only display node values that have been configured at "visible" in the card designer

chiatt:
Resolved in #7273

PR #7273, opened 2021-03-30
Show only visible widgets in multi-tile workflow step

chiatt:
Show only visible widgets in multi-tile workflow step, re #7267


njkim (approved):
Fixed the issue

PR #7276, opened 2021-04-03
Remove final step cruft and support component based final step

chiatt:
Remove final step cruft and support component based final step, re archesproject/arches-her#578


chrabyrd (approved):
lgtm! We should come to a consensus on arrow functions as a team.

Issue #7290, opened 2021-04-07
Deleting a resource does not delete that resource from a custom index


apeters:
During deletion of a resource, delete on custom indexes are missing.  This causes custom indexes to not reflect changes when resources are deleted.

chiatt:
Resolved in #7325

PR #7293, opened 2021-04-08
when deleting resources make sure documents in custom indexes are deleted as well, re #7290

apeters:
re #7290


chiatt (approved):
@adamlodge tested and confirmed it works as expected

Issue #7251, opened 2021-03-25
Update workflow buttons: hide "previous step" button from 1st workflow step, "next step" from last workflow step


dwuthrich:
We should implement logic that:

> precludes showing the "previous step" button on the first workflow step

> hides the "next step" button from the last workflow step

> includes a "edit this resource" button on the last workflow step.  Pressing this button should open the instance being created (when the workflow creates data for a single instance).  If multiple instances were created during the workflow, pressing the button should show a listing of the instances created (perhaps as hyperlinks) and open the appropriate instance in Arches' resource editor when selected 

> Include a "complete this workflow" button on the last workflow step.  Pressing this button should return the user to the workflows page.

chiatt:
Resolved in #7297

PR #7297, opened 2021-04-08
Hide prev and next step buttons on 1st and last steps

chiatt:
Hide the previous and next step buttons on 1st and last steps, respectively. re #7251


njkim (approved):
works as described

Issue #7298, opened 2021-04-08
Add button to return to workflow init page in final steps


chiatt:
per mockup: https://www.dropbox.com/home/Arches4/Mockups/Consultations?preview=final-step-summary.pdf

chiatt:
This button should appear in all steps once the workflow is completed

chiatt:
Resolved in #7301

PR #7301, opened 2021-04-08
Add button to allow user to navigate back to init workflow page

chiatt:
Add button to allow user to navigate back to init workflow page, #7298


njkim (approved):
Works as described

Issue #7307, opened 2021-04-12
Some workflows create multiple resources


chrabyrd:
This is reproducible in the consultations workflow in arches-HER, but is indicative of a deeper issue that lives in core. 

If the nodegroup_id param of a workflow step isn't the top-level node of the resource, 2 resources are created. One containing the tile data from the workflow step; and a blank, unrelated resource containing the parent node structure.

We are getting this extra tile/resource creation without any workflow interaction (eg the work saving a step multiple times). We should leverage this 'free' behavior from the arches system. 

The currently-proposed solution is to dig into how to relate these two tiles on creation or save via resource_id.

chiatt:
Resolved in #7308

PR #7308, opened 2021-04-12
Some workflows create multiple resources

chrabyrd:
### Types of changes

### Description of Change

Reproducible in the consultation workflow. Instead of creating many resources, it should now create 2. The 2nd resource is due to the final step of the workflow, and will be refactored in a later ticket.

### Issues Solved

#7307


chiatt (approved):
Resolves #7307

Issue #7283, opened 2021-04-06
Workflow step information box settings should persist for every instance of the workflow


chiatt:
If a user disables the appearance of the information box in a workflow step, that choice should apply to each future instance of that workflow as well as the current instance. Currently it only applies to the current workflow instance.

chrabyrd:
fixed in #7310

PR #7310, opened 2021-04-13
persistent infobox status, #7283

njkim:
### Types of changes

### Description of Change

Update the infobox setting to remember user's selection, #7283
Once a user choose to display or hide, the selection is saved for the next new session.

### Issues Solved

#7283

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/viewmodels/workflow-step.js
+++ b/arches/app/media/js/viewmodels/workflow-step.js
@@ -134,7 +134,6 @@ define([
                     heading: config.informationboxdata['heading'],
                     text: config.informationboxdata['text'],
                 })
-                console.log(self.informationBoxData())
             }
         };
         
--- a/arches/app/media/js/viewmodels/workflow-step.js
+++ b/arches/app/media/js/viewmodels/workflow-step.js
@@ -126,7 +126,7 @@ define([
             /* cached informationBox logic */ 
             if (config.informationboxdata) {
                 var isHidden = true;
-                if (self.getInformationBoxHiddenStateFromLocalStorage()){
+                if (self.getInformationBoxHiddenStateFromLocalStorage() !== undefined){
                     isHidden = self.getInformationBoxHiddenStateFromLocalStorage();
                 }
                 self.informationBoxData({
@@ -134,6 +134,7 @@ define([
                     heading: config.informationboxdata['heading'],
                     text: config.informationboxdata['text'],
                 })
+                console.log(self.informationBoxData())
             }
         };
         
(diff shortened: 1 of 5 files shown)


arches/app/media/js/viewmodels/workflow-step.js
```diff
@@ -209,16 +209,47 @@ define([
...
 
-        this.getInformationBoxHiddenStateFromLocalStorage = function() {
-            return self.getFromLocalStorage('informationBoxHidden')
+        this.getInformationBoxDisplayedStateFromLocalStorage = function() {
+            return self.getMetadataFromLocalStorage('informationBoxDisplayed')
+        };
+
+        this.getMetadataFromLocalStorage = function(key) {
+            var workflowsMetadataLocalStorageData = JSON.parse(localStorage.getItem('workflow-metadata')) || {};
+            var pathArray = window.location.pathname.split('/');
```
chrabyrd, line 226: workflow is a param of workflow-step, and workflowName is on workflow. Can we avoid parsing the location path to get the workflow name?

chrabyrd (changes requested):

arches/app/media/js/viewmodels/workflow-step.js
chiatt: @chrabyrd It looks like it may be parsing the slug from the path rather than the workflow name. The advantage to the slug is that it enforces uniqueness. If the slug was accessible in the step, I think that would be ideal - no parsing needed. Of course, in all likelihood the name would also be unique as well, just not guaranteed (as far as I know).

chrabyrd: If we need it to have a unique constraint I'd say let's enforce it explicitly at the workflow level and pass it down. I think basing logic off of pathname in this instance is a bit of an anti-pattern, as we don't have that interaction anywhere in workflows, and it's straightforward to use alternative logic.

chiatt: I agree. If the `slug` was available on `config.workflow` then that would be the way to go. However, if using the slug is much trouble at all, it's probably fine to use the workflow name given that having 2 workflows with the same name would be really unlikely.

arches/app/media/js/viewmodels/workflow-step.js
```diff
@@ -209,16 +199,62 @@ define([
...
+                    isDisplayed = self.getInformationBoxDisplayedStateFromLocalStorage();
+                }
+                self.informationBoxData({
+                    displayed: isDisplayed,
+                    heading: config.informationboxdata['heading'],
+                    text: config.informationboxdata['text'],
+                })
+            }
+        }
+        this.setMetadataToLocalStorage = function(key, value) {
```
chrabyrd, line 240: This is fantastic code overall.

My one nit is that I think it's living in the wrong place.

Instead of having a listener in `workflow-step` for `workflow.workflowName`,  and then having a function to set/get localStorage that knows data about the workflow ( small red flag ), why not have this logic live in the workflow?

If it lives in the workflow we can remove the observable subscription in workflow-step. Workflow can then pass the hidden/displayed status to the workflow-step on creation as an observable, which the `workflow-step` could then manipulate. The workflow has a subscription on this observable to set to localStorage.

That keeps the `workflow-step` from knowing anything about `workflow` , and removes the need for a workflowName subscription. It still uses the good persistence logic you've written it just removes some maybe extra stuff.

Also this is just my opinion. If you feel strongly I'm happy to discuss alternatives~!

arches/app/media/js/viewmodels/workflow.js
```diff
@@ -37,9 +37,16 @@ define([
...
+
+        this.nameReady = ko.computed(function(){
+            return self.ready() && self.workflowName();
+        })
+
+        this.nameReady.subscribe(function(ready) {
+            if (ready){
+                var components = _.unique(self.steps.map(function(step) {return step.component;}));
+                require(components, function() { self.initialize(); });    
+            }
```
chrabyrd, line 49: Love the code overall, this is my last and only nit, and it's more about the architecture present around the code written here.

`ready` is a remnant of v1 workflows, and will likely be factored out in the near future. Instead the component will just function off `loading`.  It's only ever called in the `${project}-workflow.js` file and it's easy to change the code involving it to use different logic. It's counter-intuitive to have both `loading` and `ready` logic in the same component and is potentially asking for trouble down the line.

Just talking about future direction, not asking for those changes in this ticket :) 

Following this logic, would it be possible to drop the `ready` nomenclature ? Sommit like:

```
        this.ready = ko.observable(false);
        
        this.workflowName = ko.observable();
        this.workflowName.subscribe(function(workflowName) {
            if (workflowName && self.ready()) {
                var components = _.unique(self.steps.map(function(step) {return step.component;}));
                require(components, function() { self.initialize(); });
            }
        });

```

Line 1 on `workflow.htm` will need to be changed to accommodate the new code introduced here, be it yours or mine. Something like `ready() && workflowName()`

As always if you disagree or feel it's better to just push this out the door, I'm always open to discussion and opposing opinions 👍

njkim: good suggestion. fixed accordingly.

chrabyrd:
@njkim 

overall lgtm! Had to make a small change to `workflow-step` L211 to keep workflow steps where we're not passing in `informationboxdata` from breaking. If you can verify on your end that workflows are still working correctly, and that this change works for you as well, I'm happy to merge in 😄

chrabyrd (approved):
lgtm! touched base with @njkim over slack and the changes are working for him as well.

Issue #7303, opened 2021-04-09
In label based graph endpoint cardinaltiy n cards sometimes return an object rather than array


chiatt:
The label based graph endpoint `resources/[uuid]?format=json` returns an array of tiles if a card has a cardinality of n and there is more than 1 tile for the card. However, if there is only one tile, the endpoint returns a tile object rather than array with 1 tile. This can make programming against the api a bit tricky because the developer does not know whether the value will be an object or an array. To resolve this issue, the value for a cardinality n card should always be an array regardless of the number of tiles.

chrabyrd:
fixed in #7315

PR #7315, opened 2021-04-16
Put cardinality n tiles in an array, even if there is only one tile

chiatt:
Put cardinality n tiles in an array, even if there is only one tile, re #7303.


njkim (approved):
works as described

chrabyrd (approved):
lgtm and all tests pass. @chiatt last look-over before merge? :)

Issue #7332, opened 2021-04-20
CSV load fails when loading data with child tiles


chiatt:
When loading a csv file with child tiles, the csv loader can end up in an infinite recursion loop due to recent changes in Tile.get_blank_tile method.

chiatt:
Resolved in #7333

PR #7333, opened 2021-04-20
Fix check for whether child tile creation is needed

chiatt:
Fixes check for whether child tile creation is needed, re #7332

Prior to this [refactor](https://github.com/archesproject/arches/commit/250b7bc78dd8458cf0dbee5e66e8e9da39175757#diff-009b1fa3c3618e758589d21b4512631be7da49ff01e8b422e78cfa4ad78f255a), get blank tile would return a parent tile with an empty data object,`{}`. Because that was fixed, checking whether child tiles needed to be created within csv import no longer worked. This PR corrects the conditional in csvfile.py.

### DIFF
--- a/arches/app/utils/data_management/resources/formats/csvfile.py
+++ b/arches/app/utils/data_management/resources/formats/csvfile.py
@@ -1,4 +1,3 @@
-from pprint import pprint as pp
 import csv
 import pickle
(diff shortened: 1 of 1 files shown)


ryan86 (approved):
Resolves #7332

Issue #7334, opened 2021-04-20
The related resource section of the default report shows erroneous relations in the section that reference the same type as the subject of the report


apeters:
See the image below.
You can in the section entitled "Ruscha Negative" that the same resources are referenced in the above sections.  The references in "Ruscha Negative" are incorrect.

![Screen Shot 2021-04-20 at 3.29.50 PM.png](https://images.zenhubusercontent.com/56f071cb6caf56c9d38e0f43/192d0238-6a08-4159-8a17-83ff7eb0f61b)

chiatt:
Resolved in #7372

PR #6590, opened 2020-08-28
Paginate async loaded related resources in reports

whatisgalen:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

To Test:
Go to the report of a resource instance that has at least 6 related resource instances of a single resource model, click "Load more"
Use any Project / package

### Issues Solved

#6481 

### Checklist

-   [ ] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

PR #7335, opened 2021-04-21
7334 fix related resource report section

apeters:
re #7334


apeters:
it looks like this bug was introduced in re #6590

apeters:
To test chose any report that has related resources.  Notice that before the fix, resources would erroneously show up in the same related resource section that happened to be the type of the resource itself.  See images below for explaination.

Before: 
Screen Shot 2021-04-20 at 5 22 00 PM

After:
Screen Shot 2021-04-20 at 5 21 31 PM

chiatt (approved):
Resolves #7334

Issue #7341, opened 2021-04-22
Some concepts no longer displaying in report


whatisgalen:
**Describe the bug**

On some reports, tiles with concept datatype data are not displaying values and instead the default report shows "None". 
On investigating the issue, it appeared that the logic to create a name_lookup for concept values was wrapped in a subscription that may not fire in a report. 
Source: https://github.com/archesproject/arches/blob/stable/5.2.x/arches/app/media/js/viewmodels/concept-select.js#L57-L75
Latest relevant PR: https://github.com/archesproject/arches/pull/6964

**To Reproduce**

**Screenshots**

**Expected behavior**

**Your Arches Information**

*   Version used: stable/5.2.x
*   Operating System and version (desktop or mobile):
*   Browser Name and version:
*   Link to your Arches Install (optional):

**Additional context**

**Ticket Background**
*   Found by: @

whatisgalen:
closing as presumed fixed

PR #7342, opened 2021-04-22
places setName logic into own fn, calls once, called from this.value …

whatisgalen:
…subscript, re #7341

### Types of changes

### Description of Change

places logic to set `Name_lookup` dict and set display name into own function, calls it once in the vm but also gets called on change from the subscription for `this.value`

### Issues Solved

#7341 

### Checklist

#### Ticket Background
*   Sponsored by: Los Angeles City Planning - Historic Places LA (https://github.com/CityOfLosAngeles)
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/viewmodels/concept-select.js
+++ b/arches/app/media/js/viewmodels/concept-select.js
@@ -54,7 +54,7 @@ define([
             return displayVal;
         });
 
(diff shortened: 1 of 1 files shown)


chiatt (approved):
Resolves #7341

PR #7347, opened 2021-04-23
Removes unused graph manager files

chiatt:
Removes unused graph manager files and associated css, re #3601


apeters (approved):
it's nice to remove more old files

Issue #7353, opened 2021-04-27
Create Cytoscape.js binding handler for use with new related resources graph


robgaston:
See more on Cytoscape.js here: https://js.cytoscape.org/#getting-started/including-cytoscape.js

In order to take advantage of the API  provided by Cytoscape.js for visualizing and analyzing graphs, we will need a knockout binding handler to support creating Cytoscape.js visualizations, analogous to what we do with Mapbox GL JS.

PR #7354, opened 2021-04-27
adds cytoscape binding handler

robgaston:
Adds the Cytoscape.js library to the JS dependencies and creates a binding handler to support creating Cytoscape.js graph visualizations

re: #7353


chiatt (approved):
Installs cytoscape and adds binding

Issue #7322, opened 2021-04-19
info button not working in new related resource creation component 


chiatt:
If a user clicks the `i` button on an instance in the resource instance an error will prevent the summary report from displaying.

Issue #7359, opened 2021-04-28
Summary report display for the resource instance type is broken


apeters:
**Describe the bug**
Clicking on the small "i" button on a related resource instance widget causes the page to hang

![Screen Shot 2021-04-27 at 10 44 16 PM](https://user-images.githubusercontent.com/823773/116352981-e00a9180-a7aa-11eb-90b9-4803d2d9fb9e.png)

It looks like this was caused by the work done in re #7262, specifically ["_load_resource_data"](https://github.com/archesproject/arches/blob/master/arches/app/views/resource.py#L702) can return either a context object or a JSONResponse object.  It's the return of the JSONResponse object that causes the issue.

chrabyrd:
looks to be a duplicate of #7322 , closing

PR #7361, opened 2021-04-28
Fix the resource report logic when returning json, # 7322

njkim:
### Types of changes

### Description of Change

Fix the logic to properly return json when query string includes json=True, #7322
Also, fix the issue with related resource info button not working
 
### Issues Solved

#7322

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/views/resource.py
+++ b/arches/app/views/resource.py
@@ -698,7 +698,7 @@ class ResourceReportView(MapBaseManagerView):
         graph = Graph.objects.get(graphid=resource.graph_id)
         templates = models.ReportTemplate.objects.all()
 
-        context = (
+        response = (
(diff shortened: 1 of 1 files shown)


apeters:
I just made a ticket that was a duplicate of this (er #7359).  Looking at this fix, I still think this won't work in all cases.  
If you look at the code, `_load_basic_data` returns a context object while `_load_resource_data` returns a response object.  Those should return the same object type (either context object or response object).

chiatt (approved):
Resolves #7322. @apeters Nice catch regarding the return type. @njkim addressed this so both methods now return a response.

Issue #7364, opened 2021-04-29
Improve and harden the way the display name for related resources is derived


apeters:
I recently ran into an issue where the display name for related resources weren't being populated correctly from our disambiguated graph api which was being used to power a custom report.  Whereas the default report was correct.
This was ultimately due to an issue where the "resourceXresourceId" property on the datatype wasn't being populated but the related resource id was.  To harden this code, we shouldn't rely on the using the "resourceXresourceId" property but instead use the "resourceId" property to lookup the related resource name.

apeters:
to test this you can use the disambiguated graph api.
confirm that related resource values show the name of the resource.  
if you want to test the issue that i was seeing, find a specific tile that has a related resource and remove the "resourceXresourceId" value from tile.data.  Without the fix, the output from the api will deliver an empty string for the related resource.

PR #7367, opened 2021-04-29
improve get_display_value method of resource instance datatype, re #7364

apeters:
re #7364


apeters:
to test this you can use the disambiguated graph api.
confirm that related resource values show the name of the resource.
if you want to test the issue that i was seeing, find a specific tile that has a related resource and remove the "resourceXresourceId" value from tile.data. Without the fix, the output from the api will deliver an empty string for the related resource.

chiatt (approved):
Resolves #7364

PR #7389, opened 2021-05-04
7364 harden ri display name code, bug fix

apeters:
re #7364


apeters:
to test this bug fix, create a new Instrument resource in AFS and just save the name.  Previously this was broken because there is a hidden resource-instance node being saved with a "None" value in that card.  This should fix that.

chrabyrd (approved):
works as expected!

Issue #7395, opened 2021-05-07
Add support for Sass preprocessing


robgaston:
adding support for Sass preprocessing is a first step towards improving our stylesheets... see here: https://sass-lang.com/documentation

we will need to implement this as part of our static files pipeline using supported Django apps/modules

chiatt:
Resolved in #7429

PR #7396, opened 2021-05-07
adds support for Sass preprocessing

robgaston:
re: #7395

This PR provides support for Sass preprocessing in Arches and converts the arches.css stylesheet to use Sass so that we can take advantage of the features it provides when improving these stylesheets. 

A couple of changes will need to be considered for projects taking on this change:
- there are two new python dependencies in requirements.txt that will need to be installed into your virtual environment
- the `compressor` application will need to be included in the `INSTALLED_APPS` setting (it is there in core arches settings, but may be overwritten in the project settings and therefore need to be updated)
- the `STATIC_ROOT` setting should point to the project root in development, so if this is configured for production in your project settings you will need to account for that
- with `STATIC_ROOT` pointing to the project root folder, `compressor` will create compiled files in a folder in the root called `CACHE`, which should be added to the project's .gitignore file to prevent committing these files

@aj-he @phudson-he - I couldn't help myself, so I took the liberty of implementing Sass preprocessing after our conversation this morning (or yesterday afternoon).  We can talk about how to integrate this into the work you are doing once it is reviewed and merged in.

### DIFF
--- a/arches/app/media/css/arches.scss
+++ b/arches/app/media/css/arches.scss
@@ -1,0 +1,13889 @@
+@import url(fonts/openSans.css);
+@import url(../packages/bootstrap/dist/css/bootstrap.min.css);
+@import url(../packages/eonasdan-bootstrap-datetimepicker/build/css/bootstrap-datetimepicker.min.css);
+@import url(../packages/bootstrap-colorpicker/dist/css/bootstrap-colorpicker.min.css);
+@import url(nifty.min.css);
+@import url(../packages/font-awesome/css/font-awesome.min.css);
+@import url(../packages/ionicons/css/ionicons.min.css);
+@import url(../packages/lt-themify-icons/themify-icons.css);
+@import url(../packages/chosen-js/chosen.css);
+@import url(../packages/select2/select2.css);
+@import url(../packages/select2/select2-bootstrap.css);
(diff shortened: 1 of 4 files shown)


apeters (approved):
this works well.  We'll need to remind people that they will need to update their settings

chiatt:
On Ubuntu Focal I was unable to install the python dependencies because of missing headers. Running the following cleared the  issue:
```
apt update
apt install python3-dev build-essential
pip install django_compressor==2.2
pip install django-libsass==0
```

dwuthrich:
@apeters @mradamcox Does this support for sass require documentation beyond this ticket?

mradamcox:
@dwuthrich as far as I can tell, any brand new Arches 5.2.x projects will have all of the proper default settings, so I don't _think_ any extra documentation will be necessary. As for upgraded projects from 5.1, we tend to have notes like this only in the release notes (which should get linked in the docs, if they aren't already).

dwuthrich:
@mradamcox Thanks for clarifying the proper way to document this enhancement

mradamcox:
@dwuthrich Was just setting a project up with apache and found that there actually is a new requirement that apache has write access to the static files directory (which is where the CACHE directory gets created). I've added a note here https://github.com/archesproject/arches-docs/commit/ec4b26583566648c9364342cf14eeaf057e31306 for now.

Issue #7410, opened 2021-05-10
Allow resource Instance widget TermFilter configuration to apply even when term is empty


chiatt:
The resource instance widget can be configured with a TermFilter that takes an advanced search query to control the available options in the widget's dropdown. Currently the TermFilter is applied only after a user starts writing a term, but in some cases it's needed even when the term is an empty string.

chiatt:
Resolved in #7411

PR #7411, opened 2021-05-10
Apply term filter even if term is an empty string.

chiatt:
Apply term filter even if term is an empty string. re #7410


njkim (approved):
Works as expected

Issue #7405, opened 2021-05-09
load_package defer_indexing doesn't actually defer indexing of resources in bulk


whatisgalen:
**User Story**

when using the `bulk` flag, resources/tiles imported as part of the broader `load_package` operation get indexed at the time of ingestion/save even if `defer_indexing` is `True`, instead of deferring the indexing of these imported business data until the end of the parent operation

**Acceptance Criteria**
-   [ ] <FILL IN>
-   [ ] <FILL IN>
-   [ ] Unit Tests passed
-   [ ] Documentation

**Definition of Done**

**Ticket Background**
*   Sponsored by: 
*   Suggested by: @ 
*   Designed by: @

whatisgalen:
closing as after multiple tests/scenarios, deferring indexing while using the bulk flag performed only as fast as when indexing was not deferred, not faster, therefore for now it seems indexing should always happen in the `Resource.bulk_save` method.

aj-he:
@whatisgalen I disagree that the indexing should always happen in `Resource.bulk_save` as `--bulk_load` is an option for package loading and package loading defers indexing to the end. So you end up indexing twice if you package load and use the `--bulk_load` option.

Plus, my fix which allows bulk_save to be used for appending after the initial resource creation (e.g. you have two CSV files one creates resources, the other adds more branch data), will mean that I'll be forced to sit through 2 indexing operations without the choice to prevent indexing.

PR #7415, opened 2021-05-11
7405 defer resource indexing

whatisgalen:
### Types of changes

### Description of Change

Note that while there are a lot of commits on this PR, for convenience this branch was branched off of this: https://github.com/archesproject/arches/pull/7057
but nearly all changes in `csvfile.py` were reverted to reflect what's currently in `stable/5.2.x`.

The spirit of this PR is really to defer indexing of resources and their tiles during the `import_business_data` function. 
Because `defer_indexing` is `True` by default, running `load_package` will cause a reindex at the end anyway. Thus, if the indexing of newly-imported resource instance data can be prevented, it will effectively only be deferred until the end. Because `bulk_load` aims to speed things up by using bulk methods, one of the core changes here is to completely prevent indexing during `Resource` model's `bulk_save` method. Thus, for a `prevent_indexing` param to be passed in a way that makes sense, it will take the value of `bulk_load` or a named kwarg `prevent_indexing`, whichever evaluates to `True`. Another small change (read: fix) is to only reindex at the end of package load if celery is *not* running, as it doesn't make sense to do a synchronous reindexing while async import_business_data tasks might be ongoing. Creating said followup celery/async reindexing task is noted in a comment as a `TODO`.

The changes herein address the standalone `import_business_data` arches command slightly differently from logic tied up in `load_package`.

### Issues Solved

addresses #7405 only

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @ 

### Further comments

created https://github.com/archesproject/arches/issues/7416 as a result of work done for this PR


whatisgalen:
closing in favor of https://github.com/archesproject/arches/pull/7458

Issue #7360, opened 2021-04-28
Create new related resources graph visualization component


robgaston:
We need the scaffolding for a component that reworks/replaces our existing related resources component.

For this ticket, we should create the basic component files extending our "workbench" layout (as shown in the mockup below), stub out the tabs as mocked up, and stub out the visualization element using Cytoscape.js.

This component should be API driven for all data retrieval, expecting to be instantiated with a resource ID.

[Full Mockup](https://app.zenhub.com/files/49451058/dfff9fbb-b13d-4d59-9290-90469dc09d37/download)

chiatt:
Remaining work on this issue will be addressed in followup tickets.

PR #7418, opened 2021-05-11
Creates new related resources graph component

robgaston:
re: #7360

This PR creates the basic new related resources graph component using Cytoscape.js

Major functionality remains to be developed in forthcoming tickets

At the moment, the component has been implemented on the search page, but there are other places in the application where the old visualization remains which will need to be updated (tickets for this work will also be coming)


robgaston:
@chiatt at long last, I think that this PR is ready for your review.  Thanks!

chiatt (approved):
Works as expected

PR #7429, opened 2021-05-13
updates sass with the latest CSS

robgaston:
re: #7395


chiatt (approved):
Resolves issue with styles in graph designer

PR #7435, opened 2021-05-17
Remove update of '_tileData' causing widget to lose focus

chiatt:
Remove update of '_tileData' causing widget to lose focus, re #7352


chrabyrd:
Unfortunately this breaks workflow dirty state. Reproducible by attempting to navigate `Sample Taking Workflow` in AfS.

chrabyrd (changes requested):
see previous comment

chrabyrd (approved):
lgtm

PR #7452, opened 2021-05-21
Add capability to flag lock in the workflow step, afs #368

njkim:
### Types of changes

### Description of Change

Add locked flag in the workflow step, re: archesproject/arches-for-science#368
When other steps needs to lock a previous step to prevent for a user to change it once the resource instance is created, the `locked` flag is added so that it can be used to disable selection widget, etc.

### Issues Solved

archesproject/arches-for-science#368

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/viewmodels/workflow-step.js
+++ b/arches/app/media/js/viewmodels/workflow-step.js
@@ -37,6 +37,11 @@ define([
         this.postSaveCallback = ko.observable();
 
         this.externalStepData = {};
+        this.locked = ko.observable();
+
+        this.setLock = function(stepName, locked) {
+            config.workflow.setLock(stepName, locked);
+        };
 
         var externalStepSourceData = ko.unwrap(config.externalstepdata) || {};
         Object.keys(externalStepSourceData).forEach(function(key) {
--- a/arches/app/media/js/viewmodels/workflow-step.js
+++ b/arches/app/media/js/viewmodels/workflow-step.js
@@ -237,7 +237,9 @@ define([
         }
 
         this.setSourceStepLock = function(locked) {
-            config.workflow.setLock(self.externalStepData.sourcesteptolock.stepName, locked);
+            self.lockableExternalSteps().forEach(function(step){
+                config.workflow.setLock(step, locked);
+            })
         };
 
         _.extend(this, config);
--- a/arches/app/media/js/viewmodels/workflow-step.js
+++ b/arches/app/media/js/viewmodels/workflow-step.js
@@ -38,6 +38,7 @@ define([
 
         this.externalStepData = {};
         this.locked = ko.observable(false);
+        this.lockableExternalSteps = config.lockableExternalSteps || []
 
         var externalStepSourceData = ko.unwrap(config.externalstepdata) || {};
         Object.keys(externalStepSourceData).forEach(function(key) {
@@ -238,8 +239,12 @@ define([
         }
 
         this.lockExternalStep = function(step, locked) {
+            if (self.lockableExternalSteps.indexOf(step) > -1){
                 config.workflow.toggleStepLockedState(step, locked);
-        };
+            } else {
+                throw new Error("The step, " + step + ", cannot be locked")
+            }
+        }
 
         _.extend(this, config);
 
--- a/arches/app/media/js/viewmodels/workflow-step.js
+++ b/arches/app/media/js/viewmodels/workflow-step.js
@@ -43,6 +43,10 @@ define([
             config.workflow.setLock(stepName, locked);
         };
 
+        this.locked.subscribe(function(val){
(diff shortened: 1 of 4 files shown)


arches/app/media/js/viewmodels/workflow-step.js
```diff
@@ -37,6 +37,15 @@ define([
         this.postSaveCallback = ko.observable();
 
         this.externalStepData = {};
+        this.locked = ko.observable();
+
+        this.setLock = function(stepName, locked) {
```
chrabyrd, line 42: small nit: can this function be moved to below the initialize method, with the other function declarations?

arches/app/media/js/viewmodels/workflow-step.js
```diff
@@ -182,7 +196,7 @@ define([
         this.getFromLocalStorage = function(key) {
             var allStepsLocalStorageData = JSON.parse(localStorage.getItem(STEPS_LABEL)) || {};
 
-            if (allStepsLocalStorageData[self.id()]) {
+            if (allStepsLocalStorageData[self.id()] && allStepsLocalStorageData[self.id()][key]) {
```
chrabyrd, line 199: This needs to be refactored to return falsey values. eg if someone stored int 0, it wouldn't be returned here.

arches/app/media/js/viewmodels/workflow-step.js
chrabyrd, line 42: A few thing here:

- Can this function declaration be moved to below initialize where the other func declarations are?
- As written, this function offers too much flexibility down the stack. Can it be refactored to not take any step name, and instead rely on a `config` value?
- Can the func name be changed to something more explicit? `setLock` sounds localized on this level and this deals with external steps

arches/app/media/js/views/components/workflows/component-based-step.js
```diff
@@ -586,6 +588,9 @@ define([
     function viewModel(params) {
         var self = this;
 
+        this.locked = params.locked;
+        this.setLock = params.setLock;
```
chrabyrd, line 592: Instead of setting/passing down `setLock`, can this be refactored to use a subscription that manipulates `params.setLock`?

arches/app/media/js/views/components/workflows/component-based-step.js
```diff
@@ -547,13 +547,15 @@ define([
...
-    function WorkflowComponentAbstract(componentData, previouslyPersistedComponentData, externalStepData, resourceId, title, complete, saving) {
+    function WorkflowComponentAbstract(componentData, previouslyPersistedComponentData, externalStepData, resourceId, title, complete, saving, locked, setLock) {
         var self = this;
 
         this.saving = saving;
         this.complete = complete;
         this.resourceId = resourceId;
         this.componentData = componentData;
+        this.locked = locked;
+        this.setLock = setLock;
```
chrabyrd, line 558: Instead of passing down `setLock` here, can this be refactored into an observable, something like `shouldLockExternalStep`. that way in the viewModel (~L592) we can subscribe to the observable to manipulate params.

chrabyrd (changes requested):
Overall looks pretty good. There's some minor refactoring requested, please reach out if you'd like to chat!

arches/app/media/js/viewmodels/workflow-step.js
```diff
@@ -227,6 +236,10 @@ define([
             }
         }
 
+        this.setSourceStepLock = function(locked) {
+            config.workflow.setLock(self.externalStepData.sourcesteptolock.stepName, locked);
```
chrabyrd, line 240: ah, one small thing. Otherwise everything looks good!

`externalStepData` is used for passing data in one direction from one step to another.
Can the config key be something separate? Something like `lockableExternalSteps` ?

chrabyrd (changes requested):

arches/app/media/js/viewmodels/workflow-step.js
```diff
@@ -227,6 +236,12 @@ define([
             }
         }
 
+        this.setSourceStepLock = function(locked) {
+            self.lockableExternalSteps().forEach(function(step){
```
chrabyrd, line 240: I see `self.lockableExternalSteps` being invoked here, but I'm not seeing where it's declared on this level? Can we do something similar to `externalStepData`, where we set the value here based on the config?

chrabyrd (changes requested):
overall looks good! Mostly nits and a small refactor, please lemme know if you have any questions!

chrabyrd (changes requested):
overall looks good! Just gotta tie up some of the lockableExternalStep logic :)

chrabyrd (approved):
lgtm!

Issue #7437, opened 2021-05-18
pathing to the uploaded images used in saved search is broken when Arches is hosted under a subpath


apeters:
when Arches is hosted under a stubpath (www.mysite.com/arches/myarchessite/....) then the images users upload for saved searches are incorrectly pathed.

chiatt:
Resolved in #7453

PR #7453, opened 2021-05-24
7437 url pathing with subpaths

apeters:
re #7437

### DIFF
--- a/arches/app/media/js/reports/image.js
+++ b/arches/app/media/js/reports/image.js
@@ -28,7 +28,7 @@ define([
                                     _.contains(nodes, key)
                                 ) {
                                     imgs.push({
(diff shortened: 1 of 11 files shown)


apeters:
To test this setup a single project and host it under the root domain and also under a subpath.
When hosting arches under a sub path, set the FORCE_SCRIPT_NAME setting to the subpath name (use a beginning and trailing slash)

You need to test several areas in Arches.  It's simplest to create a simple graph that uses a filelist datatype.  Also setup the report for that graph with the "image header" report type.
Check the following things.
1. Images in reports and editor render correctly
2. Report header renders correctly
3. Adding an image to a concept renders correctly
4. Anywhere else that you can think of that uses user uploaded images

Issue #7416, opened 2021-05-11
celery package load flow should include tasks for load relations, indexing


whatisgalen:
**User Story**

When celery is used, `import_business_data` gets called asynchronously while `import_business_data_relations` and `reindex_database` get called synchronously even though they both depend on `import_business_data` having completed. These two operations should be moved into async celery tasks. 

**Acceptance Criteria**
-   [ ] <FILL IN>
-   [ ] <FILL IN>
-   [ ] Unit Tests passed
-   [ ] Documentation

**Definition of Done**

**Ticket Background**
*   Sponsored by: 
*   Suggested by: @ 
*   Designed by: @

PR #7458, opened 2021-05-25
7405 defer resource indexing 2

whatisgalen:
### Types of changes

### Description of Change

redo of this PR sans commits from other branch
https://github.com/archesproject/arches/pull/7415

### Issues Solved

#7405 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -1172,17 +1172,20 @@ class FileListDataType(BaseDataType):
 
     def append_to_document(self, document, nodevalue, nodeid, tile, provisional=False):
         try:
-            document["strings"].extend(
-                [{"string": f["name"], "nodegroup_id": tile.nodegroup_id, "provisional": provisional} for f in tile.data[str(nodeid)]]
-            )
-        except KeyError:
-            for pe in list(tile.provisionaledits.values()):
-                document["strings"].extend(
-                    [{"string": f["name"], "nodegroup_id": tile.nodegroup_id, "provisional": provisional} for f in pe["value"][nodeid]]
-                )
+            for f in tile.data[str(nodeid)]:
+                val = {"string": f["name"], "nodegroup_id": tile.nodegroup_id, "provisional": provisional}
+                document["strings"].append(val)
+        except KeyError as e:
+            for k, pe in tile.provisionaledits.items():
+                for f in pe["value"][nodeid]:
+                    val = {"string": f["name"], "nodegroup_id": tile.nodegroup_id, "provisional": provisional}
+                    document["strings"].append(val)
 
     def get_search_terms(self, nodevalue, nodeid):
-        terms = [file_obj["name"] for file_obj in nodevalue if file_obj["name"] is not None]
+        terms = []
+        for file_obj in nodevalue:
+            if file_obj["name"] is not None:
+                terms.append(file_obj["name"])
 
         return terms
 
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -1172,14 +1172,14 @@ class FileListDataType(BaseDataType):
 
     def append_to_document(self, document, nodevalue, nodeid, tile, provisional=False):
         try:
-            document["strings"].extend([
-                {"string": f["name"], "nodegroup_id": tile.nodegroup_id, "provisional": provisional} for f in tile.data[str(nodeid)]
-            ])
+            document["strings"].extend(
+                [{"string": f["name"], "nodegroup_id": tile.nodegroup_id, "provisional": provisional} for f in tile.data[str(nodeid)]]
+            )
         except KeyError:
             for pe in list(tile.provisionaledits.values()):
-                document["strings"].extend([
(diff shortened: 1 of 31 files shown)


apeters (changes requested):
A few items here:
1. we should remove the "flat" argument to the "bulk_save" method as it's not actually used anywhere
2. Did we want to keep the display of the time to bulk create resources?  If so then we should uncomment those lines, otherwise we should just remove those lines altogether
3. On line 929 we should remove the check to see if the celery worker is running.  
4. I think the option to bulk load shouldn't override the "prevent_indexing" setting.  That is, if I select "prevent_indexing=True" and "bulk_load=True" then the data should NOT be indexed.  It would be up to the user to index at a late time.

whatisgalen:
1. I've removed flat from the bulk_save method and its usage in csfile.py
2. I've removed the print statements and time delta calcs from bulk_save -- I think it makes more sense to just time import_business_data than to time the specific method it ends up using
3. ln 929 in packages.py is actually needed because we don't want to index the database if celery *is* running (see this issue: https://github.com/archesproject/arches/issues/7416 )
4. I've added `prevent_indexing` to ln 929 of packages.py -- is there anywhere in particular bulk_load would override prevent_indexing? Right now `prevent_indexing` is getting passed to `import_business_data` (when called for resource/tile biz data) as `prevent_indexing or bulk_load` so indexing will be prevented if either are true. Later on, if the bulk_load was true but prevent_indexing was not (in the context of packages.py), a reindex will happen.

apeters (commented):
I don't think the "prevent_indexing" property should depend on whether "bulk_load" is true or not.  See lines 286, 715, 720, 726

I don't think you need lines 288-290 .  This would redundantly index the db twice if "prevent_indexing" is False and "bulk_load" is True.

Also, I don't think we should rely on whether the `celery_worker_running` property is False on line 931.  Also, maybe we should prevent users from selecting "defer_indexing" = True and "prevent_indexing" = True.  Then we can remove the check for "prevent_indexing = False" on line 931.

whatisgalen:
How would 288-290 index the db twice? That code block is for the management command "import_business_data". The Load_package management command calls the method `self.import_business_data`, which is also called by the management command of the same name but obviously distinct from it. 

288-290 is for the case where the user calls the management command to import a single business_data file and --bulk would call the bulk method, speeding it up (more so if indexing is deferred). 

Why do you think we shouldn't rely on `celery_worker_running` on 931? If that value is `True`, then we'd be indexing the database at the same time that data is still being imported, which means only some of the data will ultimately get indexed as part of package_load. 

I think you're right that `defer_indexing` and `prevent_indexing` have some overlap at certain times and can make things a little confusing. I would argue that if `bulk_load` is a proxy for "do a more efficient/faster load", then maybe deferred indexing is implied and we don't need to offer that as an option at all. In fact I'm inclined to just take `deferred_indexing` out of the equation altogether and then if the user adds `--bulk` flag, all indexing gets deferred by default, separately the user can add the `prevent_indexing` which would override indexing, deferred or otherwise.

apeters:
> How would 288-290 index the db twice? That code block is for the management command "import_business_data". The Load_package management command calls the method self.import_business_data, which is also called by the management command of the same name but obviously distinct from it.

> 288-290 is for the case where the user calls the management command to import a single business_data file and --bulk would call the bulk method, speeding it up (more so if indexing is deferred).

You're right, it won't... my bad!  But....  as proposed, it will reindex the ENTIRE database.  I think we should just undo all those edits to the function "import business data".  As it stands it will already bulk index the data imported in the most efficient manner available and only the resources just loaded.  No need to delay that until after the call completes.

> Why do you think we shouldn't rely on celery_worker_running on 931? If that value is True, then we'd be indexing the database at the same time that data is still being imported, which means only some of the data will ultimately get indexed as part of package_load.

Actually what I think needs to happen is that we should remove the option to pass "prevent_indexing" when loading a package.  We should either defer indexing or index while loading.  The exception is if you bulk load then we shouldn't defer indexing even if defer_indexing is true, because in that case the bulk indexing we employ should be relatively fast.

Also, I noticed that the task to load business data never passes the "bulk" flag to the method [here](https://github.com/archesproject/arches/blob/7405_defer_resource_indexing_2/arches/app/tasks.py#L102-L107).  That seems like a problem.

This is quite complicated, so it might warrant a phone call.

apeters (changes requested):
After further review, I am going to suggest we do the following:

1. let's remove all the code related to list comprehensions.  All that should belong in another branch related to optimizations/speedup **(lines 288-289, 301-303, 318-319, and 331-333 in resource.py and _all_ edits to datatypes.py).**
2. let's make another ticket for deferring indexing during business data import, and then let's remove the code related to that from this PR **(lines 278-286, 296, 298-309 in packages.py, and _all_ the edits to es.py, index_database.py, and tasks.py)**.  

This PR then will directly address the ticket it's linked to (#7405)

whatisgalen:
closing as much of the original work on this branch has been split up across other branches, the original ticket has been closed given that deferral of bulk indexing does not increase performance of `import_business_data`

Issue #7459, opened 2021-05-26
unique together constraint missing from node and nodegroup id


apeters:

chiatt:
Resolved in #7462

PR #7460, opened 2021-05-26
7459_unique_together_constraint

apeters:
re #7459


apeters:
this is a messy PR that I'm going to close in favor of something cleaner

PR #7462, opened 2021-05-26
7459_fix_unique_together_bug

apeters:
re #7459

### DIFF
--- a/arches/app/models/models.py
+++ b/arches/app/models/models.py
@@ -537,7 +537,7 @@ class Node(models.Model):
         managed = True
         db_table = "nodes"
         constraints = [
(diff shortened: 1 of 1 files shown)


apeters:
before this fix, if you ran "migrate" you might have noticed that django would throw this warning

```
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

if you ran makemigrations then you'll notice that the system thinks that the unique constraint on "name" and "nodegroup" on the nodes table has not been applied (but it was in the migration `6700_unique_nodename`).  This PR fixes this.

to test run migrate on a fully migrated db and notice the warning.  after the fix, the warning should be gone.

chiatt (approved):
Resolves #7459

Issue #7362, opened 2021-04-28
Unsaved change in a workflow step is lost but the step remains dirty


njkim:
If a user leaves a step with some unsaved changes and comes back,
the unsaved change is lost but the step remains dirty and cannot be saved or undone.

chiatt:
Resolved in #7465

PR #7465, opened 2021-05-27
Resets dirty state when revisiting a step

chiatt:
Reset dirty state when revisiting a step. re #7362


njkim (approved):
Works as expected

Issue #7467, opened 2021-05-27
Update streets layer


chiatt:
Update the streets source and layer with a more recent version of mapbox streets.

apeters:
fixed by #7468

PR #7468, opened 2021-05-28
Update street layer in dml

chiatt:
Updates the street layer and source so that when a user creates a new project or rebuilds their database, the new layer will load., re #7467


apeters (approved):
works as expected

PR #6676, opened 2020-10-05
added arches json support to resources put api

aj-he:
## Description

**Enhancement related to #6638** 

Resources API put method now supports arches json format as well as arches json-ld format.

## Checks
- [x] Run unit tests
- [x] Documentation already suggests json format is supported.
- [x] Tested on internal application at v5.0

Issue #6638, opened 2020-09-22
Provide support for Arches JSON format for the API resources PUT method


aj-he:
**Description**

Provide support for Arches JSON format for the API resources PUT method. Currently it only appears to support JSON-LD while the GET method support both JSON and JSON-LD.

**Comment**
We have developed this for our v5.0.0 implementation and will be looking to submit a PR to stable/5.0.x on completion of testing.

chiatt:
Thanks @AJ-HistoricEngland, looking forward to your PR.

aj-he:
This is now including POST verb and targeting 5.2.x. PR incoming from @khodgkinson-he

PR #7470, opened 2021-05-28
6638 api arches json put post

khodgkinson-he:
## Description

**Enhancement related to #6638**

Resources API put and post methods now supports arches json format as well as arches json-ld format. 

GET and DELETE already handle all formats

This supersedes PR #6676 submitted previously.

## Checks
- [x] Run unit tests
- [x] Documentation already suggests json format is supported.
- [x] Tested on internal application at v5.2.x

### DIFF
--- a/arches/app/views/api.py
+++ b/arches/app/views/api.py
@@ -774,26 +774,57 @@ class Resources(APIBase):
             indent = int(request.POST.get("indent", None))
         except Exception:
             indent = None
+        allowed_formats = ["arches-json", "json-ld"]
+        format = request.GET.get("format", "json-ld")
+        if format not in allowed_formats:
+            return JSONResponse(status=406, reason="incorrect format specified, only %s formats allowed" % allowed_formats)            
 
         try:
             if user_can_edit_resource(user=request.user, resourceid=resourceid):
-                data = JSONDeserializer().deserialize(request.body)
-                reader = JsonLdReader()
-                if slug is not None:
-                    graphid = models.GraphModel.objects.get(slug=slug).pk
-                reader.read_resource(data, graphid=graphid)
-                if reader.errors:
-                    response = []
-                    for value in reader.errors.values():
-                        response.append(value.message)
-                    return JSONResponse({"error": response}, indent=indent, status=400)
-                else:
-                    response = []
-                    for resource in reader.resources:
-                        with transaction.atomic():
-                            resource.save(request=request)
-                        response.append(JSONDeserializer().deserialize(self.get(request, resource.resourceinstanceid).content))
-                    return JSONResponse(response, indent=indent, status=201)
+                if format == "json-ld":
+                    data = JSONDeserializer().deserialize(request.body)
+                    reader = JsonLdReader()
+                    if slug is not None:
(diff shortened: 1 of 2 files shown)


apeters (changes requested):
POSTing should always make a new resource 
- Currently if an arches-json object is POSTed that object is updated.   
- POSTing should ignore any resource instance id's in the url or in the data being POSTed

PUTing should always update an existing resource
- IF a resource id is supplied in the url it should match the resource ids in the body of the request
- Resource id's in the request body take precedence over the id supplied in the url
- PUTing a resource should act like a POST if the supplied resource instance id doesn't exist in the system

In either case the POSTed/PUTed resource should be returned from the request

khodgkinson-he:
@apeters 

Hi Alexei!  ^_^  

Just to check my understanding here…

- 	PUT : With no matching resourceInstanceId, to act like POST. 
- 	POST : To ignore any resource instance id's in the url or in the data.

This will preclude any loading of records with pre-specified resourceInstanceIds through the resources API.

Though I confess to having some difficulties with loading json-ld formated resources back into Arches, stepping through the code it seems that PUT (with no matching resourceInstanceId) is not coded to act like POST, but rather preserves the ResourceInstanceId supplied and uses it in JsonLdReader.read_resource to set self.resource.pk = resourceid (and hence ResourceInstanceId is set). 

Am I mistaken?  ^_^

Many thanks

Kingsley.

chrisdolby-HE:
@apeters @chiatt Kingsley has this as a task in our current sprint to close - are there any additional changes you'd like him to make, do you need more time to review?

apeters:
@chrisdolby-HE @khodgkinson-he @chiatt 
I'd still like to a bit more time to respond to @khodgkinson-he comment above (some new work may come of that).  I'll try to get to that today.

apeters:
@khodgkinson-he 
Hi Kingsley,
I just took a closer look at the code you mention and you are correct that currently:

- PUTting a json-ld doc will esentially create a document with the given resource id (regardless of whether it existed in the system or not).
- POSTing a json-ld doc will ignore any supplied resource id and will always create a new resource

Because of this I suppose we should follow that pattern with your updates.  Can you make that change?
Also, could you add a test that includes tile data to confirm that tile data is persisted as well?

Cheers,
Alexei

chrisdolby-HE:
@apeters Thanks very much for the speedy follow-up. Kingsley is implementing as requested.

khodgkinson-he:
@apeters 

Hi Alexei, ^_^

  Sure thing.. Those changes requested are in for you now, sir!  ^_^

  Any problems, please let me know..

                       many thanks

                                  Kingsley.

khodgkinson-he:
@apeters 

Hi Alexei, 

  I've included a supplementary unit test to increase coverage..

  Please let me know if there is anything else you need from me on this PR..  ^_^

                                    many thanks

                                               Kingsley.

apeters:
@khodgkinson-he 
So sorry for the delay on this, but could you merge the latest from stable/5.2.x into your branch.  There are some changes in 5.2.x that I need to be able to test this adequately.
Thanks,
Alexei

Issue #7128, opened 2021-02-18
Allow the resource-instance datatype to filter records by a user supplied search string


apeters:
Currently users are only allowed to select types of resources to display in the resource-instance widget dropdown.  Some users, though, would like to be able to further refine the resource instances displayed in the dropdown by supplying a search string from the search page.

chiatt:
Resolved in #7474

PR #7474, opened 2021-05-29
7128 resource instance filter

apeters:
re #7128


chiatt (approved):
Worked as expected for me and the migration ran successfully forward and backward.

azaroth42:
Presentation out of the way, will review tomorrow. Does it fix the branch issue? Also, the %trans for the label on the default value for the card?

azaroth42:
> django.template.exceptions.TemplateSyntaxError: 'compress' is not a registered tag library. [...]

I did a new `pip install -r arches/install/requirements.txt` which found the compressor library ... but something else needs to happen as well?

robgaston:
@azaroth42 see the description of this PR for changes that will need to be applied at the project level for the latest (this will, of course, be included in the next release notes)

azaroth42:
INSTALLED_APPS was the problem. Thanks @robgaston!

azaroth42:
Issues, as discussed on previous call but not filed specifically until now:
* https://github.com/archesproject/arches/issues/7493 
* https://github.com/archesproject/arches/issues/7494

azaroth42 (changes requested):
See: #7493 #7494

apeters:
@azaroth42 can you check the latest fixes for #7493, and #7494

azaroth42:
Yup, confirmed these solve the two issues :) Thanks @apeters !

apeters:
FYI, I'm not going to fix the stickler errors as I think the whitespace improves readability.

Issue #7387, opened 2021-05-03
Workflow save buttons are confusing


dwuthrich:
After completing a workflow, a user may go back to a previous step and update their input.

When this occurs the form is dirty and it then shows the previous/undo/save&continue buttons in the bottom right corner of the screen.  However, the form also shows the "Complete Workflow" button as enabled in the upper right corner of the form (see image below)

A user would naturally conclude that pressing the complete worklfow button would automatically save the edits on the current form.  However, this is not the case.  Pressing the complete workflow button finishes the workflow without saving the current form's edits.

Either the complete workflow button should save the current state of the form as part of the commit process (preferred solution) or it should be disabled, forcing the user to click the "save and continue" button![Screen Shot 2021-05-03 at 2.08.12 PM.png](https://images.zenhubusercontent.com/56f1ad0a6caf56c9d38e1cbd/d47646c0-2cf0-43c4-a6e9-8cff5008a301).

chiatt:
Resolved in #7477

PR #7477, opened 2021-06-01
Save dirty step when a user clicks 'Complete Workflow'

chiatt:
Save dirty workflow step when a user clicks 'Complete Workflow', re #7387


chrabyrd (commented):
This works exactly as it should. However is it good UE to have something implicitly save?

chrabyrd (approved):
lgtm!

changed the signposting for better ux

PR #7481, opened 2021-06-02
Uses disabled param if passed to the resource-instance-select component.

chiatt:
Use disabled observable if true, re archesproject/arches-her#645


chiatt:
Closing because this functionality is already in #7452

Issue #7482, opened 2021-06-02
load_package using bulk errors if the package contains data to be appended


aj-he:
**Describe the bug**

When loading a package with many data files that append data onto the same resources, the bulk_create call will error with a key conflict when loading data for existing resources.

**To Reproduce**
1. have a package with two or more CSVs, where then are loading different branches
2. run the load_package command with --bulk_load e.g. `python manage.py packages -o load_package --source data\pkg --bulk_load -y
3. first file will load fine but the second will error when the first batch tries to save.

```
2021-05-14 12:08:37,208 arches.app.utils.data_management.resources.formats.csvfile ERROR    Traceback (most recent call last):

2021-05-14 12:08:37,208 arches.app.utils.data_management.resources.formats.csvfile ERROR      File "c:\development\project\arches\arches\app\utils\data_management\resources\formats\csvfile.py", line 859, in import_business_data
    prevent_indexing,

2021-05-14 12:08:37,209 arches.app.utils.data_management.resources.formats.csvfile ERROR      File "c:\development\project\arches\arches\app\utils\data_management\resources\formats\csvfile.py", line 375, in save_resource
    Resource.bulk_save(resources=resources)

2021-05-14 12:08:37,209 arches.app.utils.data_management.resources.formats.csvfile ERROR      File "c:\development\project\arches\arches\app\models\resource.py", line 194, in bulk_save
    Resource.objects.bulk_create(resources)

2021-05-14 12:08:37,209 arches.app.utils.data_management.resources.formats.csvfile ERROR      File "C:\development\project\env\lib\site-packages\django\db\models\manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)

2021-05-14 12:08:37,209 arches.app.utils.data_management.resources.formats.csvfile ERROR      File "C:\development\project\env\lib\site-packages\django\db\models\query.py", line 468, in bulk_create
    self._batched_insert(objs_with_pk, fields, batch_size, ignore_conflicts=ignore_conflicts)

2021-05-14 12:08:37,210 arches.app.utils.data_management.resources.formats.csvfile ERROR      File "C:\development\project\env\lib\site-packages\django\db\models\query.py", line 1204, in _batched_insert
    ignore_conflicts=ignore_conflicts,

2021-05-14 12:08:37,210 arches.app.utils.data_management.resources.formats.csvfile ERROR      File "C:\development\project\env\lib\site-packages\django\db\models\query.py", line 1186, in _insert
[...]

PR #7483, opened 2021-06-02
fix for when using bulk_load when appending

aj-he:
### Types of changes

### Description of Change

Bulk loading fails when data in a package is being appended to an existing resource.

### Issues Solved

#7482

### Checklist

#### Ticket Background
*   Found by: @aj-he  
*   Tested by: @aj-he 
*   Designed by: @


aj-he:
@whatisgalen as discuss in issue #7482 this has cross over with your work on CSV loading. Could you review or pass to someone authorised to? Thanks

aj-he:
@whatisgalen is there any movement on this bug fix as we are needing it for bulk data loading?

arches/app/models/resource.py
```diff
@@ -189,22 +189,31 @@ def bulk_save(resources):
             resource.tiles = resource.get_flattened_tiles()
             tiles.extend(resource.tiles)
 
-        # need to save the models first before getting the documents for index
+        # need to handle if the bulk load is appending tiles to existing resources/
+        existing_resources = Resource.objects.filter(resourceinstanceid__in=[resource.resourceinstanceid for resource in resources])
```
whatisgalen, line 193: On line 193 I would suggest the following changes for performance:
``existing_resources = set(Resource.objects.filter(resourceinstanceid__in=[resource.resourceinstanceid for resource in resources]).values_list('resourceinstanceid', flat=True))``
This would obviate the need for the work on line 193

arches/app/models/resource.py
```diff
@@ -189,22 +189,31 @@ def bulk_save(resources):
...
+        existing_resources_ids = [existing_resource.resourceinstanceid for existing_resource in existing_resources]
+        resources_to_create = [resource for resource in resources if resource.resourceinstanceid not in existing_resources_ids]
```
whatisgalen, line 195: on line 195 I would suggest as followup to my comment above:
``resources_to_create = [resource for resource in resources if resource.resourceinstanceid not in existing_resources_ids]``

whatisgalen:
I'm wondering if it would be justified to actually pass in an ``overwrite`` kwarg to Resource proxy's ``bulk_save`` method here. I don't think checking for whether resources exist is necessary a slow operation, but done thousands of times it might shave off a bit of time to simply do one flow if ``overwrite == 'append'`` vs ``overwrite == 'overwrite'``

arches/app/models/resource.py
```diff
@@ -189,22 +189,31 @@ def bulk_save(resources):
...
-        print(f"Time to bulk create tiles and resources: {datetime.timedelta(seconds=time() - start)}")
+        logger.info(f"Time to bulk save tiles and resources: {datetime.timedelta(seconds=time() - start)}")
 
         start = time()
-        for resource in resources:
+        for resource in resources_to_create:
             resource.save_edit(edit_type="create")
 
-        resources[0].tiles[0].save_edit(note=f"Bulk created: {len(tiles)} for {len(resources)} resources.", edit_type="bulk_create")
+        try:
```
whatisgalen, line 207: Just out of curiosity what prompted you to put this into a try/except?

arches/app/models/resource.py
```diff
@@ -189,22 +189,31 @@ def bulk_save(resources):
...
+            resources[0].tiles[0].save_edit(note=f"Bulk created: {len(tiles)} for {len(resources)} resources.", edit_type="bulk_create")
+        except:
+            pass
 
-        print("Time to save resource edits: %s" % datetime.timedelta(seconds=time() - start))
+        logger.info("Time to save resource edits: %s" % datetime.timedelta(seconds=time() - start))
+
+        # refetch new AND updated resources and index
+        for resource in Resource.objects.filter(resourceinstanceid__in=[resource.resourceinstanceid for resource in resources]):
```
whatisgalen, line 215: suggested edit for line 215:
``for resource in Resource.objects.filter(resourceinstance__in=resources):``

arches/app/models/resource.py
```diff
@@ -189,22 +189,31 @@ def bulk_save(resources):
             resource.tiles = resource.get_flattened_tiles()
             tiles.extend(resource.tiles)
 
-        # need to save the models first before getting the documents for index
+        # need to handle if the bulk load is appending tiles to existing resources/
+        existing_resources = Resource.objects.filter(resourceinstanceid__in=[resource.resourceinstanceid for resource in resources])
+        existing_resources_ids = [existing_resource.resourceinstanceid for existing_resource in existing_resources]
```
whatisgalen, line 194: here on line 194 I would suggest the following change to a ``set`` comprehension for performance:
``existing_resources_ids = {existing_resource.resourceinstanceid for existing_resource in existing_resources}``

arches/app/models/resource.py
```diff
@@ -189,22 +189,31 @@ def bulk_save(resources):
...
-        Resource.objects.bulk_create(resources)
+        Resource.objects.bulk_create(resources_to_create)
         TileModel.objects.bulk_create(tiles)
 
-        print(f"Time to bulk create tiles and resources: {datetime.timedelta(seconds=time() - start)}")
+        logger.info(f"Time to bulk save tiles and resources: {datetime.timedelta(seconds=time() - start)}")
 
         start = time()
-        for resource in resources:
+        for resource in resources_to_create:
```
whatisgalen, line 204: we may also want to save an edit for resources being appended to where`` edit_type="append"`` or something.
This could be done where we iterate through ``existing_resources`` this for loop through something like this:
``for resource in existing_resources:

whatisgalen:
@aj-he when you get a chance, take a look at this diff for reference of the changes I'm suggesting, interested to hear your feedback:
https://github.com/archesproject/arches/pull/7560/files

arches/app/models/resource.py
aj-he: I'll change to a `set` and test

arches/app/models/resource.py
aj-he: I've change changed this and am testing.

arches/app/models/resource.py
aj-he: I was not happy with blindly calling `save_edit` on a hardcoded array without null testing.  In .NET I would use `resources[0]?.tiles[0]?.save_edit()` which tests for an object before trying to access an attribute or function. If it did crash out then it would miss the indexing stage.

I'm wondering if the resource save_edit a note should be included to state that it is a bulk operation. Is there any inherent reason why the bulk_create edit record is on a single tile?

arches/app/models/resource.py
aj-he: I've changed this and am testing

aj-he:
OK, instances where I used `Resource.objects.filter(resourceinstance__in=resources)` failed as it wouldn't resolve to a field. I tried `resource__in` and `resource_instance` but none worked. Looking elsewhere in the code I could see that  `TileModel.objects.filter(resourceinstance__in=resources)` worked but that made sense as there is a relation field in `TileModel`.

This kind of makes sense as why would you query for an array of objects you already have right!? So I have reverted to the original code for lines `193` and `215`.

arches/app/models/resource.py
aj-he: this failed testing so reverted - see comment below

arches/app/models/resource.py
aj-he: this failed testing so reverted - see comment below

arches/app/models/resource.py
aj-he: passed testing so commited

arches/app/models/resource.py
aj-he: This makes sense so I have added to the branch.

aj-he:
With regards to the overwrite/append option, it makes sense I guess to avoid unnecessary processing but shouldn't the default be `"append"` as it appears to be the case for the `BusinessDataImporter`?.

When I looked into this further in `packages.py` the default would appear to be `""` and forces an error in `import_business_data` should the user not pass in "append" or "overwrite".

I can see inconsistencies in a few places such as when running via `load_package` there is no overwrite kwarg available and `True` is passed when calling `import_business_data` (lines 709 and 716). This never errors because the load logic only ever tests if `overwrite == ""`  or `overwrite =="overwrite"`. So, "append" is not tested for anywhere!

It feels like overwrite in `packages.py` needs to be made consistent and validate that the values must be `"append"` or `"overwrite"`.

I think that `"append"` should be the default throughout the load process due to its non-destructive nature.

aj-he:
@whatisgalen could we look at getting this included as we are going to be loading good chunks of data soon.  I am currently having to run on a branch with the fix in.

Also, re the prevent/defer index parameter question for `Resource.bulk_save` - if reindex_db at the end of a `load_package` call is not going to be looked at soon, can we include this? `Resource.save` has an `index` optional parameter thta stops indexing... could we just make bulk_save match that pattern? I can provide my tested implementation of this if you like.  I have added my timings below based on loading into an empty instance (I know we have discussed the differences of a DB with data already in)

| Step | Start time| Finish time | Duration|
|-------|-------|--------|---------|
| Package load **(no prevent indexing on bulk_save)**| 12:54 | 17:20 | 4hr 26mins |
| Final package reindex_db | 17:20 | 19:13 | 1hr 53 mins |
| ----------------------------------------------------------|-------| **Total** | **6hr 17mins** |
| Package load **(prevent indexing on bulk_save)**| 17:22 | 19:27 | 2hr 05mins |
| Final package reindex_db | 19:27 | 21:19 | 1hr 52 mins |
| ----------------------------------------------------------|-------| **Total** | **3hr 57mins** |

aj-he:
@whatisgalen @apeters Hi both... can this reviewed some more please? I've added Alexei to see if this needs to be reviewed by someone else.... not sure of Galen's work pressures? Thanks

arches/app/models/resource.py
```diff
@@ -193,24 +193,38 @@ def bulk_save(resources, transaction_id=None):
             resource.tiles = resource.get_flattened_tiles()
             tiles.extend(resource.tiles)
 
-        # need to save the models first before getting the documents for index
+        # need to handle if the bulk load is appending tiles to existing resources/
+        existing_resources = Resource.objects.filter(resourceinstanceid__in=[resource.resourceinstanceid for resource in resources])
```
apeters, line 197: @aj-he I would suggest using a "values_list" here: https://docs.djangoproject.com/en/3.2/ref/models/querysets/#values-list
to get a list of existing_resource_ids directly.
Something like this:

```
existing_resources_ids = Resource.objects.filter(resourceinstanceid__in=[resource.resourceinstanceid for resource in resources]).values_list('resourceinstanceid', flat=True)
```

Which should be able to replace lines 197 and 198 and reduce the overhead of creating model instances

apeters (changes requested):

apeters:
Also, what are the commands that you're using to test this?

arches/app/models/resource.py
aj-he: @apeters thanks, I've used the `value_list` function to bypass returning resources from the DB and can now split the in mem resources list between new and existing.

aj-he:
@apeters I have run tests loading a multi-part set of CSVs using the `load_package --bulk_load` and `import_business_data -ow append --bulk_load`.

... multiple runs of this with successive parts of the model in each CSV
```
python manage.py packages -o import_business_data -s "C:\path_to_source_file\my_CSV.csv" -c "C:\path_to_source_file\my_CSV.mapping" -ow "append" --bulk_load
```
and a packge containing the same files
```
python manage.py packages -o load_package --source "C:\path_to\pkg" -y --bulk_load
```

aj-he:
@apeters any other thoughts on my answers to your comments?

apeters:
@aj-he nope, everything looks good.  I'll get around to QA'ing this PR this week.

arches/app/models/resource.py
```diff
@@ -193,24 +193,41 @@ def bulk_save(resources, transaction_id=None):
             resource.tiles = resource.get_flattened_tiles()
             tiles.extend(resource.tiles)
 
-        # need to save the models first before getting the documents for index
+        # need to handle if the bulk load is appending tiles to existing resources/
+        existing_resources_ids = Resource.objects.filter(
+            resourceinstanceid__in=[resource.resourceinstanceid for resource in resources]
+        ).values_list("resourceinstanceid", flat=True)
+
+        existing_resources = [resource for resource in resources if resource.resourceinstanceid in existing_resources_ids]
```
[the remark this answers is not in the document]
whatisgalen: @aj-he seems like these are worthy changes to get into the dev/6.0.x, want to push across the finish line?

aj-he:
Closing as functionality is being built into the Relational Views/ETL manager

Issue #7485, opened 2021-06-03
indexing optimizations


whatisgalen:
Methods related to indexing don't always use the most performant code. This Issue seeks to make performance enhancements mostly through implementation of dict and list comprehensions. This issue also seeks to implement dict and list comprehensions throughout core arches where appropriate, including Resource and Tile proxy models, concepts, Card and Graph.

**User Story**

**Acceptance Criteria**
-   [ ] <FILL IN>
-   [ ] <FILL IN>
-   [ ] Unit Tests passed
-   [ ] Documentation

**Definition of Done**

**Ticket Background**
*   Sponsored by: 
*   Suggested by: @ 
*   Designed by: @

chiatt:
Closing re #7487

PR #7487, opened 2021-06-03
7485 indexing optimizations

whatisgalen:
### Types of changes

### Description of Change

also implements list comprehensions in methods not related to indexing for efficiency

### Issues Solved

#7485 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chiatt (commented):
I've indicated a few changes in line that should be kept because, whether or not they improve performance, they tighten up the code and make it more readable.

chiatt:
Some of the changes here look really good, they but touch on critical parts of Arches that do not have complete unit test coverage. Merging this PR would require significant time to test both to be certain that there is no change in behavior as well as whether some of these changes would result in a performance improvement. List comprehensions can be faster, but that is not always the case. In some cases they can be slower. It would be difficult to tell without extensive testing which really improve performance. Running tests across all of the changes didn't show a significant performance improvement. Nevertheless, there are some nice changes here. I've indicated inline a few really good uses of list comprehensions that improve readability. 

For the time being I'm going to close this PR, but I think it would be nice to have a new or updated PR with just the changes I've indicated.

Issue #7491, opened 2021-06-03
further leverage deferred indexing in package_load and import_business_data


whatisgalen:
**User Story**

**Acceptance Criteria**
-   [ ] <FILL IN>
-   [ ] <FILL IN>
-   [ ] Unit Tests passed
-   [ ] Documentation

**Definition of Done**

**Ticket Background**
*   Sponsored by: 
*   Suggested by: @ 
*   Designed by: @

PR #7492, opened 2021-06-03
7491 leverage deferred indexing

whatisgalen:
### Types of changes

### Issues Solved

#7491 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


apeters (changes requested):
An error is thrown when loading a package with celerey is running.
The issue is that "load_business_data" is called from the task, and that call doesn't include the "config_file" path, which is assumed 

https://github.com/archesproject/arches/blob/57c31156062158b7936e2ab9a73241155d48dba1/arches/management/commands/packages.py#L306  

Previously, the mapping file was inferred if not supplied (see importer.py).  We should do the same when calling "import_bussiness_data"

Issue #7496, opened 2021-06-03
node_value endpoint fails for some datatypes


chiatt:
The concept, concept-list, and url datatypes fail when `transform_value_for_tile` is called in the node_value endpoint

apeters:
closed by #7497

PR #7497, opened 2021-06-03
Add missing **kwarg passed from the node_value endpoint

chiatt:
Add missing **kwarg arg passed from the node_value endpoint, re #7496.


apeters (approved):
changes look good

Issue #7495, opened 2021-05-28
Consultations Workflow: Creating new Application Area from Related Resources widget can result in 2 instances


dwuthrich:
When creating a new app area from the related application area widget in the first step of the workflow, the order of input operations matters.

If the user enters the geometry of the application area first, then adds a value for the app area name Arches creates 2 application area instances, one with a geometry (but no name) and a second with a name (but no geometry).

If the user enters the name of the application area first followed by the geometry, then Arches creates a single instance with both name and geometry, as expected.

Not sure what happens if the user creates a different tile (e.g.: not the geometry card) before entering a name

chiatt:
It turns out this occurs with other resource models as well. If the user edits a card other than the top card, an instance is created. If they then edit the top card, a second instance is created. Subsequent card edits will continue to modify the instance created in the top card while the first instance that was created will only have a tile from the first edit.

chiatt:
Resolved in #7498

PR #7498, opened 2021-06-04
Rename new resource instance component and ensure all tiles created with it have the same resourceid

chiatt:
Renames new-resource-instance.js to related-instance-creator.js and ensure all tiles created with it have the same resourceid, re #7495


njkim (approved):
Works as described

PR #7504, opened 2021-06-07
Silky hack

chrabyrd:


chiatt (approved):
Changes improving performance by reducing postgres queries.

Issue #7515, opened 2021-06-15
Add the ability for developers to easily get an access token via a get request.  


apeters:
During development it's often necessary to get an access token when writing custom scripts against the api, for example.
It would be nice to be able to easily get a token using a simple get request and passing in credentials via the url.  

Because passing credentials via a url is security risk, it should only be allowed when the application is in DEBUG=True mode.

PR #7516, opened 2021-06-15
add simple way to get oauth token for developers only, re #7515

apeters:
re #7515


apeters:
to test, go to the url
http://localhost:8000/auth/get_dev_token?username=admin&password=admin
set DEBUG=False to confirm that the page is now not accessible

chiatt (approved):
Works as expected for me.

robgaston (approved):
I reviewed these changes and they look good - might be improved by not using a POST inside the view logic, but not really a big concern since this is only for developers and should never be used in production.

Issue #7521, opened 2021-06-18
Add "responsive" extension to the datatables library to better support presentation of tile data in tables


dwuthrich:
We use the datatable library to present data in tabular form.  The library provides some nice usability features such as table filtering, sorting by column values, etc.  See the active consultations table view (in Arches-HER for an example of how we use it).

The datatable library includes an extension called "responsive" (https://datatables.net/extensions/responsive/) that supports automatically hiding columns based on screen size parameters (to better support responsive UIs).  This feature also supports showing the hidden columns by clicking on the associated row, which is a nice way to progressively show data in a table with many columns.

We should include the responsive extension to allow us to better display tile data in tables, particularly for our reports

chiatt:
I looked into this and it appears that we have already implemented this in the edit history page.

In edit-history there is a configuration object, `resourceTableConfig`, for the datatable defined here: https://github.com/archesproject/arches/blob/1937859e3213e5f564236cac1256a2efb742df42/arches/app/media/js/views/edit-history.js#L9-L31

In this case `responsive` is set to `true` in the config.

This config is then passed in the datatable binding in the template here:
https://github.com/archesproject/arches/blob/1937859e3213e5f564236cac1256a2efb742df42/arches/app/templates/views/edit-history.htm#L13

This can be applied in a similar manner to any table so long as the `bindings/datatable` binding is added to the require statement of the component's js file like it is here: https://github.com/archesproject/arches/blob/a52b7f0df39c5425203ee5244f7141e1daefadde/arches/app/media/js/views/edit-history.js#L3

PR #7524, opened 2021-06-21
Require the datatable binding only where used

chiatt:
Requires the datatable binding only where it is used, re #7521


ryan86:
Works as expected

Issue #7526, opened 2021-06-22
The undo dirty state doesn't update when coming back to a step when using the NonTileBasedComponent in the component based step


apeters:
@apeters commented on [Fri Jun 18 2021](https://github.com/archesproject/arches-for-science-prj/issues/379)

To reproduce (using AFS):
Enter data for the select-phys-thing-step (first step in the Upload Dataset workflow), then go the next step.
Then go back to the select-phys-thing-step, make a change and hit the undo button.
Notice that the value is correctly undone, but the undo button persists.

chiatt:
Resolved in #7527

PR #7527, opened 2021-06-22
fix bug where undo is incorrectly displayed, re #7526

apeters:
re #7526


chrabyrd (changes requested):
These changes do not allow a step that hasn't been saved yet to clear properly.

see https://recordit.co/Ddw1ZZSk4X

apeters:
I think this is a workflow config issue.  If you add the param: `'renderContext': 'workflow'` then this should work

chrabyrd:
> I think this is a workflow config issue. If you add the param: `'renderContext': 'workflow'` then this should work

Yep! That was it, merging now

chrabyrd (approved):

Issue #7537, opened 2021-06-26
Modify eslintrc to allow for ES6 syntax


chiatt:

chiatt:
Resolved in #7544

PR #7544, opened 2021-06-30
Modify .eslintrc file to permit ES 6 syntax

chiatt:
Modify .eslintrc file to permit ES 6 syntax, re #7537


robgaston (approved):
looks good!

Issue #7542, opened 2021-06-29
Can't load json-ld document into Arches after exporting from Arches


apeters:
To reproduce the issue:

1. Create a model with card of cardinality "n" (can be semantic node)
2. Create a sub-card of that card of cardinality "1" (can be semantic node)
3. To that card add a resource-instance node 
4. Create an instance of that model and add two instances of the card
5. Each instance of the resource-instance node should reference the same instance
6. Output the resource instance as json-ld
7. Import that same instance
8. Observe that the import fails with the error `Multiple possible branches and no @type given`

The output of the json-ld might look something like this:
```
{
    "@id": "http://localhost:8000/resources/8758c97c-58e9-480b-b419-cdd41c95432b",
    "@type": "http://www.cidoc-crm.org/cidoc-crm/E29_Design_or_Procedure",
    "http://www.cidoc-crm.org/cidoc-crm/P102_has_title": [
        {
            "@id": "http://localhost:8000/tile/e365f3bd-38fc-4802-b35e-198bde193b25/node/5ce54492-d90c-11eb-97df-acde48001122",
            "@type": "http://www.cidoc-crm.org/cidoc-crm/E35_Title",
            "http://www.cidoc-crm.org/cidoc-crm/P103_was_intended_for": {
                "@id": "http://localhost:8000/tile/e365f3bd-38fc-4802-b35e-198bde193b25/node/788f744c-d90c-11eb-97df-acde48001122",
                "@type": "http://www.cidoc-crm.org/cidoc-crm/E99_Product_Type",
                "http://www.cidoc-crm.org/cidoc-crm/P129i_is_subject_of": {
                    "@id": "http://localhost:8000/resources/5a485b0c-91f8-4ced-8eec-3b137b99bba8"
                }
            }
        },
        {
            "@id": "http://localhost:8000/tile/297625fc-a1ea-47e7-ba44-d1a22ec28553/node/5ce54492-d90c-11eb-97df-acde48001122",
            "@type": "http://www.cidoc-crm.org/cidoc-crm/E35_Title",
            "http://www.cidoc-crm.org/cidoc-crm/P103_was_intended_for": {
                "@id": "http://localhost:8000/tile/297625fc-a1ea-47e7-ba44-d1a22ec28553/node/788f744c-d90c-11eb-97df-acde48001122",
                "@type": "http://www.cidoc-crm.org/cidoc-crm/E99_Product_Type",
[...]

PR #7546, opened 2021-06-30
potential fix for jsonld import bug, re #7542

apeters:
re #7542


arches/app/utils/data_management/resources/formats/rdffile.py
```diff
@@ -429,8 +431,21 @@ def is_concept_node(self, uri):
                 return True
         return False
 
+    def find_local_references(self, jsonld_document):
+        if "@id" in jsonld_document and "@type" in jsonld_document:
```
azaroth42, line 435: suggest a further check that it is NOT a tile reference, as tiles can only exist in one place in the tree.

apeters: what would that check look like?  I'm unsure of how to check for that.

apeters: I assume you mean references like these
```
"http://localhost:8000/tile/d83757c8-203e-4755-b1fb-45e260105d78/node/e2f83f46-bacf-11ea-81b2-3af9d3b32b71"
```

apeters: To me, the added complexity of implementing a regex to filter out those tile references doesn't seem to offset the additional entries in the cache.  I think I'm just going to leave it as it for now.

chiatt (approved):
Resolves #7542

Issue #7503, opened 2021-06-07
related-resources-manager.htm template has absolute path and causing 500 error


EnisBerk:
**Describe the bug**

related-resources-manager.htm template is not working and causing 500 error. Problem is that an image file's path is hard coded and starting with a slash '/'.

**To Reproduce**

0. Requires localization with more than 1 language
1. Go to localhost.com:8000/en/search?paging-filter=1&tiles=true
2. Set STATIC_URL to something other than '/'
3. Check console and server logs for error

**Screenshots**

```
[07/Jun/2021 07:27:56] "GET /en/search/resources HTTP/1.1" 200 4839
Attempted access to '/img/Workflow.png' denied.
Internal Server Error: /en/templates/views/resource/related-resources/related-resources-manager.htm
Traceback (most recent call last):
  File "/root/venvs/archesenv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/root/venvs/archesenv/lib/python3.8/site-packages/django/core/handlers/base.py", line 124, in _get_response
    raise ValueError(
ValueError: The view arches.app.views.main.templates didn't return an HttpResponse object. It returned None instead.
[07/Jun/2021 07:27:57] "GET /en/templates/views/resource/related-resources/related-resources-manager.htm?_dc=5.1.4 HTTP/1.1" 500 79506
```
**Expected behavior**

Removing slash from the path fixes the problem.
[line1](https://github.com/archesproject/arches/blob/3a8f8764a5f4a9f340a986c59a31b083354e0158/arches/app/templates/views/resource/related-resources/related-resources-manager.htm#L9)
[line2 ](https://github.com/archesproject/arches/blob/3a8f8764a5f4a9f340a986c59a31b083354e0158/arches/app/templates/views/resource/related-resources/related-resources-manager.htm#L311)

**Your Arches Information**

*   Version used: 5.1.4 (also checked 5.2.x)
*   Browser Name and version: Firefox

EnisBerk:
Related to https://github.com/archesproject/arches/issues/4278

chiatt:
Hey @EnisBerk, thanks for submitting this. @robgaston has submitted a fix in `stable/5.2.x`. We were not able to reproduce the issue, but the change was easy to make, so hopefully it clears up the error for you.

EnisBerk:
[...]

PR #7547, opened 2021-06-30
Fixes image paths with leading slashes in templates

robgaston:
we were using the `static` template tag with image paths that included leading slashes, which was causing problems for applications with certain Django configurations.

@chiatt do you understand the conditions to recreate the original poster's issue?  I was able to confirm that the new paths work, but I was not able to reproduce the original issue just by changing my Django settings.

re: #7503


chiatt (approved):
@robgaston I was unable to reproduce the error following the instructions in the ticket, but maybe the issue is OS dependent (I tested on an aws instance). In either case, your changes should fix the problem.

Issue #7555, opened 2021-07-01
blank tile properties should all be assigned in one method


whatisgalen:
**User Story**

Currently the Tile Proxy method ``get_blank_tile`` calls a helper method ``get_blank_tile_from_nodegroup_id`` to actually get a blank tile, but then assigns the properties tiles and tileid outside of this helper method. It makes more sense to assign these properties in the helper method that way a developer knows they will get a fully complete blank tile if only calling ``get_blank_tile_from_nodegroup_id`` (which would be done so as to get a flattened tile, as implemented here: https://github.com/archesproject/arches/pull/7473/files)

**Acceptance Criteria**
-   [ ] <FILL IN>
-   [ ] <FILL IN>
-   [ ] Unit Tests passed
-   [ ] Documentation

**Definition of Done**

**Ticket Background**
*   Sponsored by: 
*   Suggested by: @ 
*   Designed by: @

PR #7557, opened 2021-07-01
moves assignment of tileid and tiles list for blank tile into get_bla…

whatisgalen:
…nk_tile_from_nodegroup_id method, implements dict comp speedup re #7555

### Types of changes

### Issues Solved

solves #7555 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chiatt (commented):
At first glance this appears like a practical and harmless refactor,  but I looked at this PR with @apeters and he noticed that it actually changes what gets returned by `get_blank_tile`.  This is because `get_blank_tile_from_nodegroup_id` gets called on lines 552 and 555. In both cases the nodeid would get set to `None` on the newly created tile when that was not happening previously. It is difficult to say what the outcome of this change would be, but it very well could introduce other issues.

Issue #7558, opened 2021-07-01
Implement transaction ID as part of save via bulk load & API


robgaston:
In #7444 we will create a "transaction ID" which can be used to group edits in the edit log.  This will be useful for grouping edits from bulk edit operations such as bulk data loads or workflows.

there are two places in the code where `edit_log` records are created, where we will need to optionally accept transaction ids: 
- for tiles: https://github.com/archesproject/arches/blob/871d03f4a4e337a686bce89cf61cddec9fd546c0/arches/app/models/resource.py#L92-L104
- for resources instances: https://github.com/archesproject/arches/blob/871d03f4a4e337a686bce89cf61cddec9fd546c0/arches/app/models/tile.py#L109-L143

We should pass transaction IDs into bulk load operations here...: https://github.com/archesproject/arches/blob/871d03f4a4e337a686bce89cf61cddec9fd546c0/arches/management/commands/packages.py#L1017-L1027
...and in web based calls to save tiles in these two locations:
- https://github.com/archesproject/arches/blob/871d03f4a4e337a686bce89cf61cddec9fd546c0/arches/app/views/tile.py#L90-L167
- https://github.com/archesproject/arches/blob/871d03f4a4e337a686bce89cf61cddec9fd546c0/arches/app/views/api.py#L1118-L1123

blocked by #7444

chiatt:
Resolved in #7563

PR #7563, opened 2021-07-02
7558 implement transactionid for loading

aarongundel:
### Types of changes

### Description of Change
Implements the transaction id feature in the resource creation process

### Issues Solved

#7558 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/views/tile.py
```diff
@@ -88,6 +88,9 @@ def handle_save_error(self, e, tile_id="", title=_("Saving tile failed"), messag
         return JSONErrorResponse(_(title), _(str(message)), {"message": message, "title": title})
 
     def post(self, request):
+        # TODO: accept UUID, don't generate here.
+        transaction_id = uuid.uuid1()
```
robgaston, line 92: I think that here we can simply do something like this:
```transaction_id = request.POST.get("transaction_id", uuid.uuid1())```
in order to retrieve the transaction id from the request where needed.

robgaston (changes requested):
This looks good, but I think that we could complete the work on the tile API portion by implementing the change that I suggested to line 92 of arches/app/views/tile.py.  Could you  please take a look?

robgaston (approved):
looks good!

Issue #7566, opened 2021-07-06
get_pref_label Error in concept widget


chiatt:
Values passed to get_pref_label endpoint are sometimes ko observables causing error.

![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/562d2f04-67ca-4227-8464-930f1830e237)

chrabyrd:
solved in #7567

PR #7567, opened 2021-07-06
Ensure values passed in requests are valid

chiatt:
Ensure values passed in requests are valid, re #7566


chrabyrd (approved):
works as expected 👍

Issue #7571, opened 2021-07-06
Text widget enabled when user doesn't have card write permissions


chiatt:
![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/f9f8eb80-3986-4927-b79c-6f113ec43b5f)

PR #7573, opened 2021-07-07
Adds computed property for text/number

aarongundel:
determines whether field is disabled by checking for disabled and uneditable properties

### Types of changes

### Description of Change
When a user has read only permissions, the input for a field should be disabled.  Currently there's a conflict with the uneditable property.  Now there's a computed property that checks both of them (uneditable and disabled).  Naming is hard - I called new computed property "disable" (open to better ideas!).  At least now it's the only thing that needs to be checked.

### Issues Solved

#7571 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/views/components/widgets/number.js
+++ b/arches/app/media/js/views/components/widgets/number.js
@@ -19,7 +19,7 @@ define(['knockout', 'underscore', 'viewmodels/widget', 'bindings/formattedNumber
         var self = this;
 
         this.disable = ko.computed(() => {
-            if (typeof self.disabled == 'function') {
+            if (ko.isObservable(self.disabled)) {
                 return self.uneditable() || self.disabled();
             }
             else {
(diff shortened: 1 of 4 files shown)


arches/app/media/js/views/components/widgets/number.js
```diff
@@ -18,6 +18,15 @@ define(['knockout', 'underscore', 'viewmodels/widget', 'bindings/formattedNumber
 
         var self = this;
 
+        this.disable = ko.computed(() => {
+            if (ko.isObservable(self.disabled)) {
```
robgaston, line 22: This is functional, but I think that we should default to `uneditable` when `disabled` is unavailable, ie something like this:
```
let disable = self.uneditable();
if (ko.isObservable(self.disabled)) {
    disable = disable || self.disabled();
}
return disable;
```

robgaston (changes requested):
could you please take a look at my requested change and update both viewmodels (`number.js` and `text.js`) to use the value in `uneditable` by default (rather than hardcoding `true` as the default)?

robgaston (approved):
looks good!

Issue #7584, opened 2021-07-09
Tile api post endpoint fails when saving file_list datatype


chiatt:
The tile api post endpoint expects data to be on request.body rather than request.POST. To pass the request along to the TileView post method, data on request.body is moved to request.POST. When this happens, request.FILES is no longer readable and the request fails.

chiatt:
Resolved in #7585

PR #7585, opened 2021-07-09
Prevent error when saving data to tile API with data on POST or FILES

chiatt:
Prevents error when saving data to the tile API with data on request.POST or request.FILES request objects, re #7584


apeters (approved):
When we document the Tile endpoint, we should clarify that when using a javascript FormData() object you need to put the stringified tile object onto a FormData key called "data".  When not using a FormData() object the tile object should just be stringified to the request body attribute.

Issue #7314, opened 2021-04-15
Disambiguated graph endpoints should provide more information for complex datatypes


apeters:
Currently the disambiguated graph endpoint leverages the "get_display_value" method on datatypes to present the values for resource instance node values.  For most datatype this is fine, but for the resource-instance datatype it would be nice to have bit more information about the related resource then just the name.  In that case it would be nice to have the resource id of the related resource as well so that links could be made to that resource.
If we added a "to_json" method to our datatypes which could in most instances just inherit from "get_display_value" then the disambiguated graph endpoint could leverage that method.

so instead of returning this:
```
{
    "Person": "Ed Norton"
}
```

it could return this:
```
{
    "Person": {
          "resourceName": "Ed Norton",
          "resourceId": "d5fa6d9b-fe36-454a-9d45-6f549697f9e8",
          "ontologyProperty": "",
          "inverseOntologyProperty": "",
      }
}
```

This would of course be backward incompatible with what we currently have.

PR #7589, opened 2021-07-12
update disambiguated json output, version label based graph

apeters:
re #7314

### DIFF
--- a/arches/app/utils/label_based_graph_v2.py
+++ b/arches/app/utils/label_based_graph_v2.py
@@ -4,7 +4,6 @@ from arches.app.models import models
 RESOURCE_ID_KEY = "@resource_id"
 NODE_ID_KEY = "@node_id"
 TILE_ID_KEY = "@tile_id"
-VALUE_KEY = "@value"
 
 NON_DATA_COLLECTING_NODE = "NON_DATA_COLLECTING_NODE"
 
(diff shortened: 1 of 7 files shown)


arches/app/utils/label_based_graph_v2.py
```diff
@@ -0,0 +1,268 @@
+from arches.app.datatypes.datatypes import DataTypeFactory
```
chrabyrd, line 1: The only difference between this and V1 is the changing the `get_display_value` call to `to_json`, yeah? Should we be duplicating everything, or should we work versioning logic into the current endpoint? Or maybe follow an inheritance pattern and overwrite the one method?

chrabyrd (commented):
overall looks good! Only thing I really have is that the tests from v1 should make it over to v2

chiatt (commented):
I reviewed the output of each datatype and they all make sense to me. The one exception maybe, as we discussed earlier, is the 'selected' property on the domain and domain list datatypes. It seems unnecessary, but I'm not sure if it should be filtered out here or whether it ought to be removed from the config itself.

chiatt (approved):

chrabyrd (changes requested):
just wanna make sure we don't accidentally merge before getting the v1 tests in ;-)

apeters:
Here's a model of all datatypes to test with.

[_All Datatypes.json.txt](https://github.com/archesproject/arches/files/6864502/_All.Datatypes.json.txt)

apeters:
@chiatt, @chrabyrd, because the timeing is critical, maybe we should write a ticket specifically for crating tests for the v2 version of label_based_graph, so we can merge this in sooner rather than later.

chrabyrd (approved):
@apeters agreed! It might also be a good exercise to get peoples' feet wet with testing patterns.

chiatt:
@apeters & @chrabyrd I just noticed your latest comments. This afternoon I added a new file for the v2 tests in this PR: https://github.com/archesproject/arches/pull/7636

PR #7595, opened 2021-07-13
allows for transaction rollback

### LINKED ISSUE #7588: Create a utility module/method for rolling back a transaction by ID
Now that we have a transaction ID stored for each record in the edit log, we should create a new module in [utils](https://github.com/archesproject/arches/tree/master/arches/app/utils) that for managing transactions (ie `transaction.py`) and add a method there that will support rolling back a transaction by ID.

Initially we should only be concerned with handling updates and creates (not deletions, which currently cannot be grouped by transaction ID).  This method should:
- get all edit log entries for a given transaction ID
- group entries by timestamp and order them reverse chronologically to ensure that the oldest entries are processed last
- iterate over entries and "undo" all creates (ie delete the tile/resource instance) and updates (update the tile's data based on the old content stored in the entry)

We should also wrap this method in a management command that can be called with a transaction id supplied as a required argument.  For this, we can also create a transaction module in [commands](https://github.com/archesproject/arches/tree/master/arches/management/commands) to store our new "rollback" management command.

We should add a new view and URL endpoint as well to support rolling back a transaction by an ID.  This could be a new view module (transaction.py or similar).  In this view, we should respect user permissions to ensure that the user has the proper permissions to rollback this transaction as we do in the view to handle posted tile data.

aarongundel:
* Adds transaction rollback api endpoint
* Adds CLI management command for transaction rollback
* Adds utility function for transaction rollback

### Description of Change

Adds utility functions, cli, and view/endpoint for rollback of a transaction, given a specific transaction ID

### Issues Solved

resolves #7588 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/utils/transaction.py
+++ b/arches/app/utils/transaction.py
@@ -1,4 +1,6 @@
-from arches.app.models.models import EditLog, TileModel, ResourceInstance
+from arches.app.models.resource import Resource
+from arches.app.models.tile import Tile
+from arches.app.models.models import EditLog
 from django.db import DatabaseError
 
 
@@ -9,11 +11,11 @@ def reverse_edit_log_entries(transaction_id):
     try:
         for edit_log in transaction_changes:
             if edit_log.edittype == "create":
-                number_of_db_changes += ResourceInstance.objects.filter(resourceinstanceid=edit_log.resourceinstanceid).delete()[0]
+                number_of_db_changes += Resource.objects.filter(resourceinstanceid=edit_log.resourceinstanceid).delete()[0]
             elif edit_log.edittype == "tile create":
-                number_of_db_changes += TileModel.objects.filter(tileid=edit_log.tileinstanceid).delete()[0]
+                number_of_db_changes += Tile.objects.filter(tileid=edit_log.tileinstanceid).delete()[0]
             elif edit_log.edittype == "tile edit":
-                number_of_db_changes += TileModel.objects.filter(tileid=edit_log.tileinstanceid).update(data=edit_log.oldvalue)
+                number_of_db_changes += Tile.objects.filter(tileid=edit_log.tileinstanceid).update(data=edit_log.oldvalue)
     except DatabaseError:
         print("Error connecting to database")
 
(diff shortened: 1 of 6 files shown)


arches/app/utils/transaction.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+# Given a transaction ID, reverse (delete or update) tiles and resources created/updated during the transaction
+def reverse_edit_log_entries(transaction_id):
+    transaction_changes = EditLog.objects.filter(transactionid=transaction_id).order_by("-timestamp").all()
+    number_of_db_changes = 0
+    try:
+        for edit_log in transaction_changes:
+            if edit_log.edittype == "create":
+                number_of_db_changes += ResourceInstance.objects.filter(resourceinstanceid=edit_log.resourceinstanceid).delete()[0]
```
robgaston, line 12: we should use the proxy models for deleting and saving [resources](https://github.com/archesproject/arches/blob/22fbc5bd534b99a874bc9835e501cdf572742524/arches/app/models/resource.py#L58) and [tiles](https://github.com/archesproject/arches/blob/22fbc5bd534b99a874bc9835e501cdf572742524/arches/app/models/tile.py#L47) here, as that will also manage ES indexes, etc.

robgaston (changes requested):
please see my comment about using proxy models for resources and tiles in the reverse logic.  thanks!

arches/app/utils/transaction.py
```diff
@@ -0,0 +1,22 @@
...
+    try:
+        for edit_log in transaction_changes:
+            if edit_log.edittype == "create":
+                number_of_db_changes += Resource.objects.filter(resourceinstanceid=edit_log.resourceinstanceid).delete()[0]
+            elif edit_log.edittype == "tile create":
+                number_of_db_changes += Tile.objects.filter(tileid=edit_log.tileinstanceid).delete()[0]
+            elif edit_log.edittype == "tile edit":
+                number_of_db_changes += Tile.objects.filter(tileid=edit_log.tileinstanceid).update(data=edit_log.oldvalue)
+    except DatabaseError:
+        print("Error connecting to database")
```
chiatt, line 20: Because this gets called in the view (rather than just being in the management command) we should probably use the logger rather than a print statement. re #6850. Also, maybe this should be wrapped in an atomic transaction?

robgaston (approved):

Issue #7587, opened 2021-07-12
Delete workflow should use the transactionid to identify objects for deletion


chiatt:
When a user deletes a workflow, the workflow should delete any resource instance or tile referenced in the edit log that shares the current workflowid as its transactionid. This functionality should be in addition to the current 'wastebin' functionality until workflow steps in Arches-HER that use the wastebin can be migrated to this new process.

chiatt:
Resolved in #7599

PR #7599, opened 2021-07-14
core changes to allow for workflow transaction deletion

aarongundel:
### Types of changes

### Description of Change

This change depends on PR #7595 and is branched off it.
Core changes to allow for workflow transaction deletion. 
* Adds transaction ID to tile viewmodel.  
* Adds workflow ID to component based step.
* Adds the transaction reversal endpoint to quitWorkflow in workflow.js

### Issues Solved

#7587 - makes it possible.  But individual workflows will need to be modified to pass the transaction ID/workflow ID over.

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/viewmodels/workflow.js
+++ b/arches/app/media/js/viewmodels/workflow.js
@@ -385,7 +385,7 @@ define([
                     function(){
                         resourcesToDelete.forEach(function(resource){deleteObject('resource', resource.resourceid);});
                         tilesToDelete.forEach(function(tile){deleteObject('tile', tile.tile);});
-                        self.reverseWorkflowTransactions()
+                        self.reverseWorkflowTransactions();
                         window.location.href = self.quitUrl;
                     }
                 )
--- a/arches/app/media/js/viewmodels/workflow.js
+++ b/arches/app/media/js/viewmodels/workflow.js
@@ -303,7 +303,7 @@ define([
 
         this.reverseWorkflowTransactions = function() {
             $.ajax({
-                type: "GET",
+                type: "POST",
                 url: arches.urls.transaction_reverse(self.id()),
                 context: self,
             });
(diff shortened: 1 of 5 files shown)


arches/app/media/js/viewmodels/workflow.js
```diff
@@ -301,6 +301,14 @@ define([
             });
         };
 
+        this.reverseWorkflowTransactions = function() {
+            $.ajax({
```
chrabyrd, line 305: Overall everything looks good, just one nit:

we're using a `GET` here to roll back the DB, yeah? Can this be a POST instead?

chrabyrd (changes requested):

arches/app/media/js/viewmodels/workflow.js
aarongundel: Good idea - done.

aarongundel:
Also fixed merge conflict

arches/app/media/js/viewmodels/workflow.js
```diff
@@ -301,6 +301,14 @@ define([
...
+                type: "POST",
+                url: arches.urls.transaction_reverse(self.id()),
+                context: self,
```
chrabyrd, line 308: Do we need to pass the entire workflow as context in this call?

chrabyrd (changes requested):

arches/app/media/js/viewmodels/workflow.js
aarongundel: I don't think so since there's no callback on this POST anyway.  Will remove.

chrabyrd (approved):
works as expected

PR #7602, opened 2021-07-14
graph api endpoint modification

### LINKED ISSUE #7598: Modify the Graphs api endpoint to return just graph information
Currently the 'Graphs' api endpoint returns the complete graph including all permitted cards. The resource-instance-select.js viewmodel uses this endpoint to retrieve basic graph information, but the inclusion of the card information is not necessary. We should add a boolean parameter (e.g. `cards=false`) to that endpoint to prevent collection of card information. 

This should speed up the loading of the resource-instance widget dropdown, so we should implement this parameter in `resource-instance-select.js`: https://github.com/archesproject/arches/blob/22fbc5bd534b99a874bc9835e501cdf572742524/arches/app/media/js/viewmodels/resource-instance-select.js#L75

aarongundel:
### Types of changes

### Description of Change
Adds a "cards" parameter to the graphs API endpoint.  If it's false, the endpoint won't load cards.  In my testing, the performance improvement when using the cards parameter is noticeable and measureable (without setting "cards" to false it was taking a little more than double the time - this will vary by resource and number of cards, most likely)

### Issues Solved

resolves #7598 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/views/api.py
+++ b/arches/app/views/api.py
@@ -517,7 +517,7 @@ class Graphs(APIBase):
 
             return JSONResponse({"datatypes": datatypes, "cards": permitted_cards, "graph": graph, "cardwidgets": cardwidgets})
         else:
-            return JSONResponse({"datatypes": datatypes, "graph": graph})
+            return JSONResponse({"graph": graph})
 
 
 @method_decorator(csrf_exempt, name="dispatch")
--- a/arches/app/views/api.py
+++ b/arches/app/views/api.py
@@ -484,24 +484,38 @@ class MVT(APIBase):
 @method_decorator(csrf_exempt, name="dispatch")
 class Graphs(APIBase):
     def get(self, request, graph_id=None):
+        cards_querystring = request.GET.get("cards", None)
+        
+        if(cards_querystring == 'false'):
+            get_cards = False
+        else:
+            get_cards = True
+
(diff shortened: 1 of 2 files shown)


arches/app/views/api.py
```diff
@@ -484,24 +484,40 @@ def get(self, request, nodeid, zoom, x, y):
...
+                widget
+                for widgets in [card.cardxnodexwidget_set.order_by("sortorder").all() for card in permitted_cards]
+                for widget in widgets
+            ]
+
+            permitted_cards = JSONSerializer().serializeToPython(permitted_cards, sort_keys=False, exclude=["is_editable"])
+
+            return JSONResponse({"datatypes": datatypes, "cards": permitted_cards, "graph": graph, "cardwidgets": cardwidgets})
+        else:
+            return JSONResponse({"datatypes": datatypes, "graph": graph})
```
robgaston, line 520: I think that we can also exclude the datatypes here and move the call to retrieve them into to `get_cards` conditional block, since they should only really be useful in conjunction with cards & widgets

robgaston (changes requested):
please see my comment about datatypes in the graphs API

robgaston (approved):
thanks!

PR #7603, opened 2021-07-15
boolean datatype rework

### LINKED ISSUE #7581: the "get_display_value" method of the BooleanDataType returns None when the value is "False"
the method should return "False" instead of None

aarongundel:
### Types of changes

### Description of Change

Changes return from get_display_value for the boolean data type to return "True" or "False" or null if no value is specified.  Commented code can be used instead when v2 of the api is released.

### Issues Solved

closes #7581 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


robgaston (approved):
looks good - thanks!

Issue #7605, opened 2021-07-16
Tile save with an annotation (and GeoJson) node fails intermittently.


apeters:
This is caused by and ES mapping exception
```
elasticsearch.exceptions.RequestError: RequestError(400, 'illegal_argument_exception', 
'mapper [tiles.data.08b43d14-e510-11eb-aed6-acde48001122.features.geometry.coordinates] 
cannot be changed from type [long] to [float]')
```

This can be fixed by adding an explicit mapping in the datatype for annotation (and probably GeoJson as well)

chiatt:
Resolved in #7607

PR #7606, opened 2021-07-16
es mapping error with annotation and geojson

apeters:
re #7605


apeters:
closing because of invalid base

PR #7612, opened 2021-07-16
required content in workflow step

### LINKED ISSUE #7604: Required content in a workflow step
Much like a 'required' node in a card, a workflow step (one that does not leverage a card) may require user input to allow a user to save and proceed. To enforce this, it seems we would need a boolean on the WorkflowStep that would allow a developer to control when all data requirements are satisfied. We could then disable the save button or throw an alert on save informing the user of what needs to be completed to save the step. 

re archesproject/arches-for-science#422

aarongundel:
### Types of changes

### Description of Change

Allows for use of the alertviewmodel at the step level.  

### Issues Solved

fixes #7604 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/views/components/workflows/component-based-step.js
+++ b/arches/app/media/js/views/components/workflows/component-based-step.js
@@ -558,9 +558,11 @@ define([
     };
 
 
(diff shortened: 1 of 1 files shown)


chrabyrd (approved):
lgtm!

PR #7614, opened 2021-07-19
Fixes edit log issue on resource creation through API

### LINKED ISSUE #7611: No edit log record when resource instance created through the api
Creating a collection instance using the Project workflow in AfS results in edit log records for tiles created for the new collection. However, there is no 'create' record for the instance itself. This presents a problem if we want to rollback a transaction - we can delete all of the tiles of an instance, but not the instance itself. The objective of this ticket is to identify why the 'create' edit log record for the instance is not getting created, and to ensure that it is not possible to create a new instance without also creating a corresponding edit record.

aarongundel:
### Types of changes

### Description of Change

API is using the Tile proxy model to do a save; in some instances it does not create a log entry when new resource instance is created to support a new tile.  Adds functionality to save the new resource entry.

### Issues Solved

closes #7611 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/models/tile.py
+++ b/arches/app/models/tile.py
@@ -117,7 +117,24 @@ class Tile(models.TileModel):
         oldprovisionalvalue=None,
         provisional_edit_log_details=None,
         transaction_id=None,
+        new_resource_created=False
     ):
+        if(new_resource_created):
+            timestamp = datetime.datetime.now()
+            resource_edit = EditLog()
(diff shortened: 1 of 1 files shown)


chiatt (approved):
Resolves #7611

Issue #7629, opened 2021-07-22
Core changes to support afs #453


chrabyrd:

chiatt:
Resolved in #7630

PR #7630, opened 2021-07-22
update editor iiif-workbench tab to have classname #7629

chrabyrd:
### Types of changes

### Description of Change

adds classname to to iiif-workbench editor tab, to give the ability to hide it at the project level

### Issues Solved

#7629 

### Checklist

### DIFF
--- a/arches/app/media/js/views/components/iiif-annotation.js
+++ b/arches/app/media/js/views/components/iiif-annotation.js
@@ -29,6 +29,8 @@ define([
         this.fillOpacity = ko.observable(0.2);
         this.showStylingTools = ko.observable(false);
 
+        this.hideEditorTab = params.hideEditorTab || ko.observable(false);
+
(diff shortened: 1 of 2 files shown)


chiatt (changes requested):
I'm a little worried that if somebody notices a class that is used in an arches core template that doesn't exist in arches.scss they might assume that it is cruft and remove it. As an alternative, I'd propose we add a visible binding like this: 
```
<div class="workbench-card-sidebar-tab" data-bind="click: function() {
    toggleTab('editor');
}, css: {
    'active': activeTab() === 'editor', 'disabled': !canvas() || expandGallery(),
}, visible: showEditTab">
``` 
Then you would just have to add `showEditTab = true` to `iiif-annotation.js` and to `analysis-area-annotation-step.js` to control the tab display.

chiatt (approved):
Works as expected

Issue #7343, opened 2021-04-23
Disambiguated (Label-based) graph needs full test coverage.


chrabyrd:
Current the `from_resources` method is uncovered. We should change this as soon as we're able.

PR #7638, opened 2021-07-25
Update tests for label based graph api endpoint beta version

chiatt:
Updates tests for label based graph api endpoint beta version, re #7314


chrabyrd (approved):
these look good! At some point v1 sank to 94% coverage but these tests match those so we're g2g!

have already filed a ticket here: https://github.com/archesproject/arches/issues/7343

PR #7639, opened 2021-07-26
Placeholder 06 28 21

whatisgalen:
### Types of changes

### Description of Change

DO NOT MERGE

### Issues Solved

#

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chiatt:
Hey @whatisgalen, is this PR still needed?

whatisgalen:
> Hey @whatisgalen, is this PR still needed?

This PR is purely for ease of seeing the diff between my branch and dev branch

arches/app/models/resource.py
```diff
@@ -190,15 +184,36 @@ def bulk_save(resources, flat=False):
...
 
-        for resource in resources:
-            resource.save_edit(edit_type="create")
+        resources_to_update = [resource for resource in resources if resource.resourceinstanceid in existing_resources_ids]
+        resources_to_create = [resource for resource in resources if resource.resourceinstanceid not in existing_resources_ids]
 
-        resources[0].tiles[0].save_edit(note=f"Bulk created: {len(tiles)} for {len(resources)} resources.", edit_type="bulk_create")
+
+        Resource.objects.bulk_create(resources_to_create)
+        TileModel.objects.bulk_create(tiles)
```
aj-he, line 196: @whatisgalen in the local code we are using we found that we needed to include the following to ensure that actions in the pre_tile_save were actioned...
```py
# need to run logic in pre_tile_save fro the datatypes otherwise resource relationships and file aren't loaded correctly.
# Needs to be some kind of bulk_pre_tile_save
    for tile in tiles:
        for nodeid in tile.data.keys():
            try:
                node_datatype = datatype_factory.get_instance(node_datatypes[nodeid])
                node_datatype.pre_tile_save(tile, nodeid)
            except:
                pass
```

arches/app/models/resource.py
```diff
@@ -190,15 +184,36 @@ def bulk_save(resources, flat=False):
...
+                resource.tiles[0].save_edit(
+                    note=f"Bulk created: {len(resource.tiles)} for resourceid {resource.resourceinstanceid} ",
+                    edit_type="bulk_create",
+                    transaction_id=transaction_id,
+                )
+            except:
+                pass
+
+        # refetch new AND updated resources and index
+        for resource in Resource.objects.filter(resourceinstanceid__in=[resource.resourceinstanceid for resource in resources]):
```
aj-he, line 215: @whatisgalen I also found that indexing was problematic for updated resources. Because we are in the middle of a transaction, requesting all the resources from the DB here brings back the pre-transaction state.

I didn't get around to fixing this, I just added my own prevent_indexing clause so that all of it was indexed at the end.

I think that we'll need to take the input resources list and - for the existing resources that are being updated - merge in the tiles from the database before indexing.

whatisgalen: interesting, couldn't we just change the argument `fetchTiles=True` ? I don't know what the performance implications would be but it would do what you're saying it needs to do (i.e. re-query the db for all tiles belonging to an existing resource)

whatisgalen: like so:
`# refetch new AND updated resources and index
        fetchTiles = len(resources_to_update) > 0
        for resource in Resource.objects.filter(resourceinstanceid__in=[resource.resourceinstanceid for resource in resources]):
            document, terms = resource.get_documents_to_index(
                fetchTiles=fetchTiles, datatype_factory=datatype_factory, node_datatypes=node_datatypes
            )`

whatisgalen: actually I think this is slightly optimal
```
# refetch new AND updated resources and index
        for resource in Resource.objects.filter(resourceinstanceid__in=[resource.resourceinstanceid for resource in resources]):
            fetchTiles = resource.resourceinstanceid in existing_resources_ids
            document, terms = resource.get_documents_to_index(
                fetchTiles=fetchTiles, datatype_factory=datatype_factory, node_datatypes=node_datatypes
            )
            
```

aj-he: I had not spotted that fetchTiles option. If it does fetch the existing tiles in addition to the new tiles then that should work.

PR #7666, opened 2021-08-03
resource instance list fix

### LINKED ISSUE #7662: In resource editor resource-instance-list widget remains disabled
In the resource editor the resource-instance-list widget (and most likely the resource-instance widget as well) can remain disabled preventing users from entering data. The requests for data to initialize the widget seem fine and there are no apparent errors. This can be reproduced by loading an AfS Physical Thing in the resource editor and selecting the `In Collection or Set` card.

The user should be able to interact with the widget as soon as it is initialized.

aarongundel:
### Types of changes

### Description of Change

fixes issue where the first load of the resource instance select list wouldn't be enabled in the create new resource form.  The issue is that sometimes undefined, and false || undefined is undefined, which is invalid with the select2 binding's strict disabled comparison.

### Issues Solved

resolves #7662 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/bindings/select2-query.js
+++ b/arches/app/media/js/bindings/select2-query.js
@@ -52,7 +52,7 @@ define([
             if (ko.unwrap(select2Config.disabled)) {
                 $(el).select2("disable");
                 select2Config.disabled.subscribe(function(val){
-                    if (val === false) {
+                    if (!val) {
                         $(el).select2("enable");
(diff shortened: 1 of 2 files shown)


chiatt (approved):
Resolves #7662

Issue #7668, opened 2021-08-04
Have the tile (and abstract) model return a Promise for use with async/await


apeters:
it would be nice if the base absytract model and the tile model returned a promise from their save/read/delete methods.

chiatt:
Resolved in #7670

PR #7670, opened 2021-08-04
7668 update models to return promises

apeters:
re #7668

I'm not sure how to test this as we currently don't expect a return from ajax calls when using the abstract.js model.
This shouldn't affect any currently running code because of that.


chiatt (approved):
Works as expected

Issue #7669, opened 2021-08-04
"tiles" argument in search_results view doesn't get handled properly


whatisgalen:
**User Story**

when the ``search_results`` url gets called with a request, the request object has a parameter called ``tiles`` which is either true or false. When this parameter gets checked in the Django view, it looks like this:
``if request.GET.get("tiles", None) is not None:``
The problem here is that **any** value, whether ``"false"`` or ``"true"`` will make this statement evaluate to ``True``. It would be better and more logical if the value for ``"tiles"`` on the request object was evaluated as a boolean and proceed from there.

**Acceptance Criteria**
-   [ ] <FILL IN>
-   [ ] <FILL IN>
-   [ ] Unit Tests passed
-   [ ] Documentation

**Definition of Done**

**Ticket Background**
*   Sponsored by: 
*   Suggested by: @ 
*   Designed by: @

chiatt:
resolved in #7671

PR #7671, opened 2021-08-04
loads json value from request object, converts to python to determine…

whatisgalen:
… whether to load tiles in response re #7669

### Types of changes

### Description of Change

This PR changes how the ``"tiles"`` kwarg gets handled in the search view search_results method; with these changes, tiles would not be included in the query/search results unless the value for the kwarg is explicitly ``True`` or ``"true"``.

### Issues Solved

#7669 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chiatt (approved):
Resolves #7669

Issue #7674, opened 2021-08-05
User edit permissions not passed to search components


chiatt:

chiatt:
Resolved in #7676

Issue #7021, opened 2021-01-18
Edit button missing from map bubble


adamlodge:
![Screen Shot 2021-01-18 at 11.01.10 AM.png](https://images.zenhubusercontent.com/56f150f76caf56c9d38e16f8/aa2f4c7d-595e-4415-8b98-fa07a1386de5)

PR #7676, opened 2021-08-05
Pass permissions info to search components

chiatt:
Passes whether users can read and edit resources down to search components and fixes logic to display edit link in the map popup. re #7674, #7021


apeters (approved):
these changes look good and restore the ability for users with the appropriate privileges to see the edit link in search results and map popup.

PR #7690, opened 2021-08-10
Changes to support archesproject/arches-for-science-prj#516

chiatt:
Remove footer in summary step and allow initialization of abstract-summary-report without the CardViewModel to support 
 archesproject/arches-for-science#516.


chrabyrd (approved):
made a small change to require logic, but overall lgtm

PR #7692, opened 2021-08-10
Allow resource to load in editor

chiatt:
Allows a resource to load in editor by preventing it from hanging if CardViewModel is not available in the resource-report-abstract component, re #7697


njkim (approved):
Works as expected

Issue #7709, opened 2021-08-16
Error loading the map filter when search aggregation results are not available


chiatt:
![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/e9ef39e1-5c83-4b72-a3aa-562ff263c041)https://github.com/archesproject/arches/pull/7702/files

PR #7710, opened 2021-08-16
Prevent error in loading map-filter before results are available 

chiatt:
Prevents error when iterating over aggregated search results before they are available. re #7709


chrabyrd (approved):
lgtm!

PR #7716, opened 2021-08-17
resource indices persist

### LINKED ISSUE #7714: Resource indexes persist after workflow is deleted
Upon deleting a workflow all resources created in the process of the workflow are deleted. However, the indexed documents for those resources persist and remain visible in search.

aarongundel:
### Types of changes

### Description of Change

Instead of calling the queryset delete method (which only calls delete on the queryset itself, not on the proxy models) use the individual delete method on each returned proxy object - django does the cast

### Issues Solved

fixes #7714 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chiatt (approved):
Resolves #7714

Issue #7717, opened 2021-08-18
Advanced search fails when card contains related instance node


chiatt:
![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/34f2cfcc-653b-4914-9042-a68acf303354)

apeters:
fixed by #7718

PR #7718, opened 2021-08-18
Use synchronous require import to prevent Advanced Search error

chiatt:
Use a synchronous require import to avoid error related to a circular import in the related-instance-select viewmodel. re #7717

For docs on synchronous require:
https://github.com/requirejs/requirejs/wiki/Differences-between-the-simplified-CommonJS-wrapper-and-standard-AMD-define#notes-about-synchronous-require


apeters (approved):
this fixes the issue

Issue #7719, opened 2021-08-18
saveOnQuit not available in custom workflow steps


chiatt:

chiatt:
Resolved in #7720

PR #7720, opened 2021-08-18
Allow custom steps to override step saveOnQuit method

chiatt:
Update step saveOnQuit when step component initializes, re #7719


njkim (approved):
works as expected

PR #7739, opened 2021-08-24
moves parent tile save behavior to card viewmodel

### LINKED ISSUE #7738: Move parent tile save behavior in card components to card viewmodel
**Is your feature request related to a problem? Please describe.**
When interacting with the card tree in custom reports, there is no way to create/save the parent tile object for cards that are nested underneath a tile.  

**Describe the solution you'd like**
Move the save functionality from the card component to the card viewmodel

**Describe alternatives you've considered**
* Leaving the save function on the component itself and trying to interact with it seems too messy.  Custom reports don't have or need access to the card component.  Having the component call the card to create the parent tile for nesting seems straightforward.

**Additional context**

aarongundel:
### Types of changes

### Description of Change

Moves parent tile save behavior to card viewmodel instead of card component

### Issues Solved

resolves #7738

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/viewmodels/card-component.js
+++ b/arches/app/media/js/viewmodels/card-component.js
@@ -205,12 +205,14 @@ define([
             });
         };
         
-        this.createParentAndChild = function(parenttile, childcard) {
-            if (parenttile.tileid === "") {
-                var callback = function(){childcard.selected(true);};
-                parenttile.save(function() {
-                    return;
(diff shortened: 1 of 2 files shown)


robgaston (approved):
looks good - thanks!

Issue #7748, opened 2021-08-25
Grouping card produces a new resource instance each time a user makes an edit


chiatt:
Each time a user saves a new edit to a node in the grouping card, a new resource instance is created containing just a single tile with that edit, rather than updating the tile of the original instance the user intended to edit.

chiatt:
Resolved in #7750

PR #7750, opened 2021-08-25
Ensure all tiles save to the same resource instance. re #7748

chiatt:
The grouping card was using a different resourceinstanceid for each tile save. This fix ensures all tiles save to the same resource instance. re #7748


njkim (approved):
Fixes the issue

Issue #7681, opened 2021-08-06
Cytoscape graph won't load unless all of the related resources exist in the system


apeters:
If you have a resource instance that references other instances that don't yet exist in the system, but others do, then the cytoscape graph won't load at all.

I would expect to be able to visualize all related resources that exist in the system regardless of whether other related resources aren't yet loaded.

This can happen during a long data load operation where you want to check functionality, or if you happen to make a subset of a larger dataset and haven't all the related resources (as happened in my case).

robgaston:
@apeters do you have an example dataset (sounds like you do) that would make recreating this easy?

apeters:
@robgaston If you load the `arches_provenance` repo package  and then load data into it from the `pir_data_subset` repo.  That should give you plenty of examples.

to load the data you can run this command
`python manage.py load_jsonld -s pir_data_subset/data/`

chiatt:
Resolved in #7772

apeters:
thanks @chiatt and @robgaston for QA'ing this!

PR #7772, opened 2021-09-01
make sure cytoscape graph has matching nodes for every edge, re #7681

apeters:
re #7681


apeters:
@robgaston I'm not sure how to test this, but what you'll need is a reference in the resource_x_resource table that includes a reference to a resource that not loaded into the system

robgaston (approved):
looks good - I made some minor changes to prevent duplication of nodes, otherwise was all good.

Issue #7779, opened 2021-09-01
Error when rolling back tile edits in transaction.py


chiatt:
Rolling back a tile edit fails in transaction.py because the Tile proxy model does not have an 'update' method.

chiatt:
Fixed in #7780

PR #7780, opened 2021-09-01
Use save rather than update to rollback tile edits, re #7779

chiatt:
Prevents error due to missing `update` method on the Tile proxy model. Also, `save` will reindex the resource after the tile is updated., re #7779


robgaston (approved):
looks good - thanks!

Issue #6653, opened 2020-09-26
If importing a csv file that references existing resources and exceptions are thrown then those resources are deleted


apeters:
**Describe the bug**
If importing a csv file that references existing resources and exceptions are thrown then those existingresources are deleted from the system 

**To Reproduce**
1. create a graph that has a card with a unique value constraint 
2. create an instance of the resource with a value that would employ the unique constraint
3. create a csv file that references that resource and would add another value to the card with the "unique" constraint
4. import the csv
5. go to the resource that you created in step 2 and notice that it doesn't exist anymore

This happens because an exception is thrown during the tile.save operation that is caught by the csv importer.
The importer incorrectly assumes that the resource being imported is new and tries to clean up after itself by deleting the resource.

PR #7788, opened 2021-09-03
Add try/except in unique node constraint to check if resource already…

ryan86:
… exists, re #6653

### Types of changes

### Issues Solved

#

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/utils/data_management/resources/formats/csvfile.py
+++ b/arches/app/utils/data_management/resources/formats/csvfile.py
@@ -427,7 +427,17 @@ class CsvReader(Reader):
                         e,
                     )
                     errors.append({"type": "WARNING", "message": msg})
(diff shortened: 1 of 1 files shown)


chiatt:
Hey @apeters, I just wanted to make sure this is still on your radar. It would be great to have this fix in the 6.0 release if we can.

chiatt:
@apeters This makes sense. We just need to confirm that it **does not** exist in the database before we save. That way we can confirm that it is indeed 'new'. If it does not exist in the database, and then we save and we end up with a resource that is saves but errors on index, then we need to delete the resource.

apeters:
@chiatt I agree we should confirm that first.

Issue #7813, opened 2021-09-09
Add error property to allow for workflow step component alerts


chiatt:
Add an error property to the workflow abstract component to allow workflow step components to display alerts without advancing steps. re archesproject/arches-for-science#577

chiatt:
Fixed in #7815

PR #7815, opened 2021-09-09
Resolve componentBasedStep after step component sets error to true

chiatt:
Resolve componentBasedStep after step component indicates error is true, re #7813. Supports allowing alert messages to be displayed on outer save. QA using the instrument-info step in the upload dataset workflow on this PR: https://github.com/archesproject/arches-for-science-prj/pull/579/files


chrabyrd (approved):
works as expected

Issue #7820, opened 2021-09-14
Add ability to sort graph data on export


apeters:
Exporting graphs in a consistent manner would be helpful when having to compare changes to graph structure between edits to the graph.  This will allow diffs to be more easily understood and make it easier to understand changes.

chiatt:
Fixed in #7821

PR #7821, opened 2021-09-14
make sure graph data are sorted before export, re #7820

apeters:
re #7820


chiatt:
@apeters Nice that you tackled this. Is master where you want it to go rather than 5.2?

chiatt (approved):
Works well for me

Issue #7723, opened 2021-08-19
New option to export data as HTML documents


aj-he:
Users often request the ability to export resources from a search in a format more like a report (PDF etc.). Currently the only option is to export it in a data set or data feed format.

This ticket proposes an export option to generate an collection of html documents containing one or more resources and allow developers to create formatted template for each model.

The export function should work using the standard export functionality.

**User Story**

As an Arches user I need to be able to export my searches into easy to read and well presented documents so that my customers have an easily accessible way to read the information without needing to process it further.

**Acceptance Criteria**
-   [ ] Users must be able to export search results as an html document
-   [ ] HTML templates must be easily written by developers
-   [ ] the templates can be added easily to a project
-   [ ] the foreground and background downloads must have a suitable threshold based on performance
-   [ ] any resources for models without a template must be ignored.
-   [ ] the HTML option is hidden if there are no valid templates.
-   [ ] an example template is provided as part of a new project.
-   [ ] Unit Tests passed
-   [ ] Documentation

**Definition of Done**

**Ticket Background**
*   Sponsored by: Historic England
*   Suggested by: @aj-he 
*   Designed by: @aj-he

chiatt:
Resolved in #7825

PR #7825, opened 2021-09-17
7723 html export option

aj-he:
### Types of changes

### Description of Change

Provides a new export option to output search results into a formatted HTML document. The HTML documents are django templates, each written for a specific graph and named `<graphid>.htm`. The templates are passed a list of disambiguated resource objects for that model, which are then used to render the document and passed back to the export mechanism.

An example template is included when a project is created. A developer can use this as a starting point to build templates for models in their project.

The example template is embedded with a CSS framework that allows for offline viewing of the HTML file. The framework provides a responsive experience to help support different device screen sizes, and supports text resizing and spacing for accessibility.

### Issues Solved

#7723 

### Checklist

-   [x] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Tested by: @aj-he 
*   Designed by: @aj-he

### Further comments

Documentation for the proposed changes can be found in the Historic England arches-docs fork here: https://github.com/HistoricEngland/arches-docs/blob/master_html_export_dev_doc/docs/create-html-export-templates.txt


aj-he:
@dwuthrich @chiatt @robgaston as discussed

chiatt:
Thanks @aj-he, I'll have a look at this on Monday

arches/app/utils/data_management/resources/formats/htmlfile.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @staticmethod
+    @lru_cache(maxsize=None)
+    def has_html_template(graphid=None):
+        """
+        Check an html template for the graph is available
+        """
+        result = False
+        try:
+            temp = get_template(f"html_export/{graphid}.htm")
```
chiatt, line 43: Can you use os.path here rather than writing out the path? Also, it is possible that there is a template with the extension `.html` rather than `.htm`. It might be good to handle for that with something like: 
```
        for extension in ('htm', 'html'):
            try:
                path = os.path.join("html_export", str(graphid))
                get_template(f"{path}.{extension}")
                result = True
            except TemplateDoesNotExist as e:
                pass
```

arches/app/utils/data_management/resources/formats/htmlfile.py
```diff
@@ -0,0 +1,139 @@
...
+            result = True
+        except TemplateDoesNotExist as e:
+            pass
+        return result
+
+    def load_html_template(self, graphid=None):
+        template = None
+        try:
+            template = get_template(f"html_export/{graphid}.htm")
```
chiatt, line 52: Similar concern here as with line 43. That could be handled with something like: 
```        
        for extension in ('htm', 'html'):
            try:
                path = os.path.join("html_export", str(graphid))
                template = get_template(f"{path}.{extension}")
                return template
            except TemplateDoesNotExist as e:
                pass

File: arches/app/templates/javascript.htm [hunk 3]
```diff
@@ -380,6 +380,9 @@
             },{% endfor %}{% endautoescape %}],
             userEmail: '{{ user.email }}',
             celeryRunning: '{{celery_running}}',
+            export_html_templates: [{% autoescape off %}{% for template in export_html_templates %}
```
chiatt, line 383: Could the key `export_html_template` be changed to `exportHtmlTemplate` just to keep with camel case on the JavaScript side? Line 22 in `views/components/search/search-export.js` would need to be updated to reflect the change.

chiatt (changes requested):
I think the example template in core (`arches/app/templates/html_export/example-000000-0000-0000-0000-0000000.html`) can be removed. It seems sufficient to have it in the project template.

arches/install/arches-templates/project_name/templates/html_export/example-000000-0000-0000-0000-0000001.htm
```diff
@@ -0,0 +1,90 @@
...
+                     </div>
+                     {% for src in resource_data|val_from_key:"External Cross References" %}
+                     <div class="rrow">
+                        <div class="rcell" data-title="Number">{{ src|val_from_key:"External Cross Reference Number" }}</div>
+                        <div class="rcell" data-title="Description">
+                           {% if src|has_key:"External Cross Reference Notes" %}
+                           {{ src|val_from_key:"External Cross Reference Notes"|val_from_key:"External Cross Reference Description" }}
+                           {% endif %}
+                           </div>
+                        <div class="rcell" data-title="Source">{{ src|val_from_key:"External Cross Reference Source"] }}</div>
```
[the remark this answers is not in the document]
aj-he: Not sure how that passed testing! Fixed this and tweaked my example comment blocks.

arches/app/utils/data_management/resources/formats/htmlfile.py
aj-he: Done

arches/app/utils/data_management/resources/formats/htmlfile.py
aj-he: Agreed. I originally used html and then changed it to htm in order to match the file extension used for all the other templates.

arches/app/templates/javascript.htm
aj-he: Was in PEP8 mode! Corrected.

chiatt:
@aj-he This looks really good and I think it's a pretty cool feature. Another issue I noticed this morning, though, is that you are adding on the resource metadata [here](https://github.com/HistoricEngland/arches/blob/ac9a2866caf73cd45e64d355b072aa470d8c32c0/arches/app/utils/data_management/resources/formats/htmlfile.py#L94-L99). In the latest in 5.2.x, that is now included in the resource.to_json method [here](https://github.com/archesproject/arches/blob/02eec472eaf089cc96a63d2055275707182e2074/arches/app/utils/label_based_graph_v2.py#L184-L192), so you don't have to do that anymore. Once you pull from 5.2.x into your branch, you'll need to adjust for that change, otherwise your templates will likely break. 

One other concern is that the data used for the HTML template is pulled from the `compact` version of resource's to_json method. This would limit a developer's ability to create very rich templates because the uncompacted version includes information such as the resourceids of related resources. I don't think it will take long before developers want the uncompacted data and switching now would be a lot easier than switching after a release.

One way to solve this would be to store template information in the project's settings.py file rather than looking for graphids in the html_export directory. That way parameters for the template could be included in the settings and the developer could specify if they wanted to use the compacted or complete data.  

However, it is nice how you've implemented things not to require configuration, so another solution would be simply to switch from the compact version to the complete ('beta') version and just forego the chance to use the compacted data. To make that change you'd have to switch compact parameter to `True` in the call to resource.to_json and add the parameter `version='beta'`. 

I think switching to using the uncompacted version of the to_json method would be the easiest solution, but it depends on how complex your templates are because they will need to be updated a bit. If they are pretty complex, it might be easier just to get the template information from settings.py. Perhaps there is another solution that I haven't considered though. What do you think?

aj-he:
@chiatt is there a performance hit from not using the compact version? I'd like not to have another setting if possible.

I'll look at the `to_json` output and adjust accordingly. You say the `version='beta'` ... how long is it likely to stay beta?

chiatt:
@aj-he I spoke with @robgaston and @apeters and the both think it's okay to stick with the compact version for now. The compact 'beta' version has pretty much everything that would be needed for creating html export templates. 

The beta version probably won't be beta for very long. We'll probably have a major version release early next year in which beta will become the standard.

aj-he:
@chiatt sorry this has taken a while to get back to.

So, I have made a minor modification to the `HtmlWriter` [here](https://github.com/HistoricEngland/arches/blob/e8dfd2ea1d0811d0864dd04ddfbeb313f34389ae/arches/app/utils/data_management/resources/formats/htmlfile.py#L87) that assumes it is v2, and then tests for compliant on LN97. If not compliant then it assumes v1 is being used and builds the metadata wrapper.

When trying v2 I did note that the values are now passed in a `@displayValue` key. There is nothing that breaks in the code but any templates that were written for v1 will need to be changed.

I have also added a new filter to handle json values in the context, such as those provided by the url datatype.

Issue #7836, opened 2021-09-22
enable map filter in search to display total mappable result set from query


whatisgalen:
**User Story**

**Acceptance Criteria**
-   [ ] <FILL IN>
-   [ ] <FILL IN>
-   [ ] Unit Tests passed
-   [ ] Documentation

**Definition of Done**

**Ticket Background**
*   Sponsored by: 
*   Suggested by: @ 
*   Designed by: @

PR #7841, opened 2021-09-23
7836 search map

whatisgalen:
### Types of changes

### Description of Change

This PR untethers the map-filter search component in the main search page so that the map can be populated with the total result set, not just those on the current page (as indicated in the search-results component). The map-filter component has been changed to manage its own request to the `search_results` view using an argument called `points_only`. The view has been changed accordingly to handle such a request. The view has also been changed to minimize the number of `scrolls of the search results as each scroll has a fixed cost.

### Issues Solved

#7836 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/views/components/search/map-filter.js
+++ b/arches/app/media/js/views/components/search/map-filter.js
@@ -559,16 +559,70 @@ define([
             },
 
             updateResults: function() {
-                if (!!this.searchResults.results){
-                    this.searchAggregations({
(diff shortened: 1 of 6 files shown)


whatisgalen:
I have added a change which will let the package developer choose to turn this functionality on via an argument in settings which gets interrogated in the map-filter search component vm

PR #7848, opened 2021-09-25
Install Rust to satisfy dependencies for fabric.

cscawley:
Building the image for getty/arches:5.0 fails in its current state. A dependency of fabric (cryptography) now requires Rust to be installed and available. Otherwise cryptography fails installation.

### Types of changes

### Issues Solved

#

### Checklist

-   [x] Dockerfile builds successfully

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -100,6 +100,9 @@ RUN mv ${WHEELS}/entrypoint.sh entrypoint.sh
 
 RUN python3.8 -m venv ENV \
     && . ENV/bin/activate \
+    && pip install --upgrade pip \
+    && pip install rust \
+    && pip install setuptools_rust \
     && pip install requests \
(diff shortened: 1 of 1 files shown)


chiatt:
Thanks @cscawley for your contribution. I'm actually not sure if we actually need `fabric` anymore.
@robgaston & @apeters do you think we could remove `fabric` from `requirements_dev.txt`. It might be time for `sauceclient` to go as well.

robgaston:
@chiatt I think that you are correct that we no longer use fabric, at least not in core arches.  If a project still uses fabric, I think that they can define that dependency local to their project.  So, yes, I think that we should prefer to just remove fabric as a dependency, with thanks to @cscawley for surfacing this issue.

chiatt:
To be resolved under #7852

Issue #7842, opened 2021-09-23
Create a custom Django Field to replace TextFields that need to be localized


apeters:
re #6956

apeters:
closed by #7863

PR #7851, opened 2021-09-29
localizable text fields

apeters:
re #7842


arches/app/models/fields/i18n.py
```diff
@@ -0,0 +1,114 @@
...
+        return I18n_String(value)
+
+    def get_prep_value(self, value):
+        print(type(value))
+        print(f"in get_prep_value, value={value}")
+        """
+        If the value was set to a string, then check to see if it's 
+        a json object like {"en": "boat", "es": "barco"}, or just a simple string like "boat".
+        If it's a json object then use the I18n_String.as_sql method to insert it directly to the database.
+        If it's just a simple string then use the I18n_String.as_sql method is used to update one language value
```
chiatt, line 109: Remove `is used`

chiatt (changes requested):
Looks good - just needs a small edit on line 109 of i18n.py

Issue #7854, opened 2021-10-04
Load annotation json with single quotes from csv


chiatt:
If annotation json data is formatted in a csv file with single quotes rather than double quotes, the import will fail.

chiatt:
Fixed in #7855

PR #7855, opened 2021-10-04
Convert json to dict regardless of single or double quotes

chiatt:
Fixes error when annotation json in a csv import file is written with single rather than double quotes re #7854

This can be tested on AfS with this csv and mapping file and the following command:
[Phys_thing_test.zip](https://github.com/archesproject/arches/files/7280462/Phys_thing_test.zip)

```
python3 manage.py packages -o import_business_data -s 'Physical Thing_test.csv' -c 'Physical Thing_test.mapping' -ow 'overwrite'
```


njkim (approved):
Resolve the issue #7854

Issue #7737, opened 2021-08-24
Provide ability to order CSV, SHP and GeoJSON export 


csmith-he:
**User Story**

A boolean setting will be provided to system administrators in the settings.py file to allow the ordering of nodes in the CSV, SHP and GeoJSON format exports to follow the ordering seen in the resource instance editor (i.e. card order and the order of nodes within those cards).

**Acceptance Criteria**
-   [ ] If setting is True then CSV output data order follows the card order of the resource model as far as the model's sort order settings allow.
-   [ ] If setting is True then SHP output data order follows the card order of the resource model as far as the model's sort order settings allow.
 -   [ ] If setting is True then GeoJSON output data order follows the card order of the resource model as far as the model's sort order settings allow.
-   [ ] Unit Tests passed
-   [ ] Documentation

**Ticket Background**

This functionality has been requested as part of the Keystone project but may be of use to the Arches community as a whole.

*   Sponsored by: Historic England
*   Suggested by: @csmith-he 
*   Designed by: @csmith-he

chiatt:
@mradamcox This is actually resolved in #7867

PR #7867, opened 2021-10-08
7737 Option to Apply Card Order to Data Export

csmith-he:
Fulfils requirement from #7737 to allow application admin to enforce an ordering on the data exports based on the order of nodes in cards for a given resource model.  This was requested for the Keystone project.  Tested in Historic England Keystone instances.

NB: Concatenation issue for nodes with the same node name in CSV and GeoJSON exports.  This is, however, no different from out of the box core functionality so is not a bug introduced by this functionality.  Issue is known and relates to #6299 

@chiatt I'm sending this PR in as I don't know what's been decided for the issue described above.

### DIFF
--- a/arches/app/search/search_export.py
+++ b/arches/app/search/search_export.py
@@ -62,8 +62,7 @@ class SearchResultsExporter(object):
 
     def insert_subcard_below_parent_card(self, main_card_list, sub_card_list):
         for sub_card in sub_card_list:
-            nodegroup_obj = models.NodeGroup.objects.get(nodegroupid=sub_card.nodegroup_id)
-            parent_obj = nodegroup_obj.parentnodegroup_id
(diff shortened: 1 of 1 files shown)


chiatt:
@csmith-he, @apeters and I discussed this issue and concluded that we can add an export name column to the node model. Given the timing of the 6.0 release we'll first add a column to support the fix in v6 #7874 and then a fix for the bug #7873 will be available in the 6.1 patch release.

csmith-he:
@chiatt All requested changes made.  Sort ordering still working on my local copy so - once the checks have completed - we should be good to go.

chiatt:
Thanks for making those changes @csmith-he. I think this is a nice improvement. Maybe card order should be the default?

Issue #7783, opened 2021-09-03
Resource models should support being "published"


robgaston:
Arches should include in the designer a way to "publish" resource models and track "publishing events" alongside resource models going forward.  These should include enough information for developers to reliably depend on a particular published version of a model in developing a project.  Packages should be able to specify their dependent published models on load such that alerts could be generated by the system if dependent models were modified.

From RDM requirements:

> To support the RDM's requirement for managing graph model content, the Arches Designer needs to be updated to support “freezing” of a model shape. This will allow Arches package developers
> to introduce models that require branches of a particular shape. Once an Arches model has been
> designed, a user will need to “un-freeze” the model to modify it and “publish” it before it can support
> data entry tasks again.
> 
> Introducing the concept of publishing Arches models will:
> - Substantially decrease the time it takes to load Arches Editor and Arches Designer pages,
> making Arches appear much faster to users
> - Allow developers to be certain that models and model branches conform to required shapes
> so that they can implement functions, plugins, and workflows in an Arches package
> - Make it possible to warn users when they are attempting to alter a model (or portion of a
> model) required by an Arches package
> 
> As the RDM will be developed using Arches’ semantic modeling capabilities, and because the RDM
> models will be required to include a core set of attributes, it will be essential to implement a method
> for publishing models.
> 
> Publishing models means that we will be able to cache model designs. Model caches mean that
> Arches will only need to query its database once (for example, the first time any user enters Arches
> editor for the model) instead of every time every user wishes to create or edit a resource instance.
>

chrabyrd:
solved in #7934

PR #7934, opened 2021-10-27
Adds graph publications

chrabyrd:
### Types of changes

### Description of Change

This removes the `isactive` column from graphs and replaces it with a `publication`  column. Adds graph_publications table

### Issues Solved

#7783 

### Checklist

### DIFF
--- a/arches/app/media/js/views/graph-designer.js
+++ b/arches/app/media/js/views/graph-designer.js
@@ -100,14 +100,13 @@ define([
                     complete: function(response, status) {
                         if (status === 'success') {
                             viewModel.isGraphPublished(true);
-                            viewModel.graphPublicationNotes(null);
                             viewModel.alert(new AlertViewModel('ep-alert-blue', response.responseJSON.title, response.responseJSON.message));
                         }
                         else {
-                            viewModel.graphPublicationNotes(null);
                             viewModel.alert(new JsonErrorAlertViewModel('ep-alert-red', response.responseJSON));
                         }
-
+                        
+                        viewModel.graphPublicationNotes(null);
                         viewModel.shouldShowPublishModal(false);
                         viewModel.loading(false);
                     }


arches/app/models/resource.py
```diff
@@ -695,9 +689,9 @@ def is_uuid(value_to_test):
         return False
 
 
-class ModelInactiveError(Exception):
+class PublishedModelError(Exception):
```
robgaston, line 692: I think that what we actually want here is an `UnpublishedModelError` -- that is, we want to disallow certain actions against unpublished models (specifically, where we were previously doing so for "inactive" models)

robgaston (changes requested):
@chrabyrd functionally this looks good to me.

it does seem that we are not enforcing graph publications throughout the application as we previously were w/ the `active` flag.  Was it your idea that this would be handled in another ticket?  I think that we are close in this PR, but we just need to adjust some conditional logic (see my notes about `UnpublishedModelError`).

arches/app/models/resource.py
chrabyrd: ah -- after giving it some thought I believe we now need both. If a model is `published`, we want to disallow changes to the model, hence `PublishedModelError`. We also want to disallow certain actions against 'inactive'/`unpublished` models, which means we'd also need a `UnpublishedModelError`. Then we need to be careful about how we're enforcing published state across the app.

chrabyrd:
> @chrabyrd functionally this looks good to me.
> 
> it does seem that we are not enforcing graph publications throughout the application as we previously were w/ the `active` flag. Was it your idea that this would be handled in another ticket? I think that we are close in this PR, but we just need to adjust some conditional logic (see my notes about `UnpublishedModelError`).

@robgaston thanks! Please see the response to your comment; I'm happy to build both errors into this branch if you think that's also a good idea. I've created #8082 to encapsulate enforcing the `published` changes.

arches/app/media/js/views/graph-designer.js
```diff
@@ -71,6 +85,62 @@ define([
                 window.open(arches.urls.export_mapping_file(viewModel.graph.graphid()), '_blank');
             };
 
+            viewModel.openPublishModal = function() {
```
robgaston, line 88: @chrabyrd we should replace these methods (`openPublishModal` and `closeModal`) with an observable and a `visible` binding on the `.modal-background` element.  We should always avoid doing DOM manipulation in viewmodels.

robgaston: now that I think about it... it might be fine (and maybe even nice) to use an observable boolean to drive the `visible` binding but retain these methods and modify them to simply toggle that observable value

robgaston (approved):

Issue #7935, opened 2021-10-28
Date validation fails if imported date is a year before 1000 CE


chiatt:
Date validation fails if imported date is a year before 1000 CE because Python can't match the year with YYYY format which requires 4 characters.

chiatt:
Resolved in #7937

PR #7937, opened 2021-10-28
Save years before 1000 CE from csv import 

chiatt:
### Types of changes

### Description of Change
Save years before 1000 CE from csv import by padding year string with zeros to match Python's YYYY format (eg year 200 becomes 0200 and year 12 becomes 0012). This solution does not apply to years BCE. 

### Issues Solved

#7935

Model and data for QA:
[date_testing.zip](https://github.com/archesproject/arches/files/7435436/date_testing.zip)

Command for QA:
`python3 manage.py packages -o import_business_data -s 'date_testing/Robot.csv' -c 'date_testing/Robot.mapping' -ow 'overwrite' -g 93d791e8-3781-11ec-8ea7-0242ac1a0007`

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -359,8 +359,8 @@ class DateDataType(BaseDataType):
         if value is not None:
             if type(value) == list:
                 value = value[0]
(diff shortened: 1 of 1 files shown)


apeters (approved):
this works to allow dates less than 4 digits to be entered via csv import

Issue #7933, opened 2021-10-27
Automatically add nonexistent languages when importing business data from CSV


aarongundel:
**Is your feature request related to a problem? Please describe.**
Upon detecting that a language doesn't exist in the new language table on CSV import,
1) automatically add the new language, prompting for necessary fields
2) after adding, finish importing the new business data

PR #7931, opened 2021-10-27
csv import & export for i8ln

aarongundel:
IMPORTANT:  DO NOT MERGE before #7930 and #7902

### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

Allows for import & export of internationalized business data in CSV format.

### Issues Solved

#7915 

### Checklist

-   [x] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

PR #7930, opened 2021-10-27
Test fixes for i8ln business data

aarongundel:
IMPORTANT:  DO NOT MERGE until #7902 is merged

### Types of changes

-   [x] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

Started as simple fixes for string based tests, ended up Fixing all tests for i8ln business data

### Issues Solved

#7905 

### Checklist

-   [ ] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

PR #7902, opened 2021-10-20
ES changes for i8ln changes to business data

aarongundel:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

Adds ES fixes, now data is indexed and searchable.  Also adds a language filter to demonstrate functionality.  Fixes some broken unit tests - most of the remaining broken tests are with import/export functionality.

### Issues Solved

#

### Checklist

-   [ ] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

PR #7938, opened 2021-10-28
adds nonexistent languages on import

aarongundel:
IMPORTANT:  Do not merge until #7902, #7930, #7931 are merged

### Types of changes

### Description of Change

Adds a check for nonexistent languages in csv files.  If there are new nonexistent languages, prompt the user for more information.  

### Issues Solved

#7933 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


apeters (changes requested):
When importing a csv I get into an infinite loop that keeps asking me to load the same language over and over.  I have 2 columns that both reference the same new language which is causing the issue.
Also, I noticed that the system takes in whatever the user types for language directionality when adding a new language.  This should be changed so that the system defines those keywords.

apeters (approved):
this works as expected

apeters (commented):
This works for the csv format, but what about other formats like json-ld, arches-json, etc...?

aarongundel:
@apeters we should probably make additional tickets for that, since this one specifically addressed https://github.com/archesproject/arches/issues/7933

Issue #7616, opened 2021-07-20
Disambiguated graph returns an object instead of a list for nested cards with cardinality 'n'.


apeters:
This only is an issue when the value of the node is null/None.  

To reproduce, make a resource graph with a nested card with Cardinality 'n'.
1. Create an instance of that graph but leave the card in question blank.
2. Visualise the instance as json and notice that the "@value" of that card is an object
3. Now add data to that card
4. Now notice that the "@value" of that card is a list

In the below example 'Child Card' is a card with cardinality 'n'.  I would expect it to be an empty array instead of an object with `@value` = null.

```
{
    "displaydescription": "undefined",
    "displayname": "undefined",
    "graph_id": "44b5dbb4-e8f9-11eb-89f0-acde48001122",
    "legacyid": null,
    "map_popup": "undefined",
    "resource": {
        "Name Node": {
            "@node_id": "9487e268-e8f9-11eb-89f0-acde48001122",
            "@tile_id": "30184358-a330-431b-95e4-f8dfd8a965e1",
            "@value": "TEST"
        },
        "String Node": {
            "@node_id": "59d0f920-e8f9-11eb-89f0-acde48001122",
            "@tile_id": "5e41aafa-f622-49cd-a6a6-31d8094ca13d",
            "@value": "NON_DATA_COLLECTING_NODE",
            "Child Card": {
                "@node_id": "dfef281a-e8f9-11eb-89f0-acde48001122",
                "@tile_id": "5e41aafa-f622-49cd-a6a6-31d8094ca13d",
                "@value": null
            },
            "NUMBER Node": {
                "@node_id": "6b2891be-e8fa-11eb-89f0-acde48001122",
                "@tile_id": "5e41aafa-f622-49cd-a6a6-31d8094ca13d",
                "@value": "1"
            }
        }
    },
    "resourceinstanceid": "342ec86c-e725-452f-ab46-0ec5150a989d"
}
```

chiatt:
@apeters I think this is resolved, right?

chiatt:
Confirmed that this is still a bug in the beta endpoint

apeters:
@chrabyrd could you take a look at this?  @aarongundel just ran into this issue today and it would be nice to get this resolved soon.

chrabyrd:
@apeters would this be considered fixed?

chiatt:
Oops! I should have closed this when I merged #7939.

PR #7939, opened 2021-10-28
Updates the label-based-graph to associate parent nodegroups #7616

chrabyrd:
### Types of changes

### Description of Change

this updates the disambiguated graphs to only associate nodes with a `parentnodegroup` lineage

### Issues Solved

#7616 
#7977

### Checklist

### DIFF
--- a/arches/app/utils/label_based_graph.py
+++ b/arches/app/utils/label_based_graph.py
@@ -232,7 +232,20 @@ class LabelBasedGraph(object):
     def _build_graph(
         cls, input_node, input_tile, parent_tree, node_ids_to_tiles_reference, nodegroup_cardinality_reference, node_cache, datatype_factory
     ):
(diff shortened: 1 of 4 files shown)


apeters (commented):
I don't see this as being fixed when I tested locally.  @chiatt can you verify that this fixes the issue?

chrabyrd:
@apeters when able, can we connect so you can show me how it fails? This is passed reproduction attempts locally

chiatt (approved):
This looks really good - resolves #7616 and was successfully tested where the endpoint is applied in AfS and HER custom reports and workflows. Generally it also resolves #7977, but in the case where a semantic node is a child of another semantic node, the child will get the same tileid as the parent even though there is no child tile (see Production_Statement_name in screenshot). This is not a serious problem, so I don't think it should hold up this PR, but I have reopened #7977 so it can be addressed later. 

image

Issue #7374, opened 2021-04-30
Add missing unregister / list operations to report management command


bferguso:
The report.py management command accepts operations to unregister, list and update a report, however the methods have not been implemented. This causes an error when trying to run any of those operations:

```AttributeError: 'Command' object has no attribute 'list'```

**To Reproduce**
1. Activate the venv and go to the arches project directory and activate : 
`# source ENV/bin/activate && cd arches`
2. Run the report management commands to list, unregister or update a report
`# python manage.py report [list | unregister | update]` or
`# python manage.py report unregister -n '<report componentname>'` or
`# python manage.py report update -s '<report json config file>'`
3. See error as above

**Expected behavior**
The management function should list, unregister or update the report.

**Arches Information**

*   Version used: 5.1.3
*   Operating System and version (desktop or mobile): Linux CentOS 7
*   Browser Name and version: n/a

**Ticket Background**
*   Found by: @bferguso

chiatt:
Resolved in #7942

PR #7942, opened 2021-10-30
7374 report add missing commands

coherit:
Adds functionality for update, unregister, and list report templates

### Description of Change

The report management command had handles for four operations: register, update, unregister, and list, but only the register function had been written. This PR adds the missing update, unregister, and list functions, following the same pattern of similar management commands (e.g. widget).

### Issues Solved

#7374 

### Checklist

-   [X ] Unit tests pass locally with my changes

#### Ticket Background
*   Sponsored by: Self funded
*   Found by: @bferguso
*   Tested by: @coherit
*   Designed by: @coherit

### DIFF
--- a/arches/__init__.py
+++ b/arches/__init__.py
@@ -6,7 +6,7 @@ try:
 except ModuleNotFoundError as e:
     print(e)
 
-VERSION = (6, 0, 0, "final", 0)  # VERSION[3] options = "alpha", "beta", "rc", or "final"
+VERSION = (6, 0, 0, "beta", 0)  # VERSION[3] options = "alpha", "beta", "rc", or "final"
 
 __version__ = get_version(VERSION)
 
--- a/arches/management/commands/report.py
+++ b/arches/management/commands/report.py
@@ -18,6 +18,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 import os
 import uuid
+import imp
 from arches.management.commands import utils
 from arches.app.models import models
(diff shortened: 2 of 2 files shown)


arches/__init__.py
```diff
@@ -6,7 +6,7 @@
 except ModuleNotFoundError as e:
     print(e)
 
-VERSION = (6, 0, 0, "beta", 0)  # VERSION[3] options = "alpha", "beta", "rc", or "final"
+VERSION = (6, 0, 0, "final", 0)  # VERSION[3] options = "alpha", "beta", "rc", or "final"
```
chiatt, line 9: Could you change the version back to "beta"?

arches/management/commands/report.py
```diff
@@ -18,6 +18,7 @@
 
 import os
 import uuid
+import imp
```
chiatt, line 21: I don't see the `imp` package used. Could you remove that import statement?

chiatt (changes requested):
Thanks @coherit for adding in these commands. This looks really good. Just a couple of really minor changes to make and this will be ready to merge.

arches/__init__.py
coherit: done

arches/management/commands/report.py
coherit: yep - done

chiatt (approved):
Looks good. Thanks @coherit!

Issue #7943, opened 2021-11-01
File viewer: handle for cases when file object has no associated file


chiatt:
@chiatt commented on [Mon Nov 01 2021](https://github.com/archesproject/arches-for-science-prj/issues/750)

If for some reason a file object has been created, but there is no path to the uploaded file in the object, the file viewer will break without explanation. We need to gracefully handle for this case so that the user either upload a file to the object or delete the invalid tile.

chiatt:
Resolved in #7944

PR #7944, opened 2021-11-01
Handle for file tiles with no file name in the file value object, re #7943

chiatt:
Handles for cases when a file tile is created from import or the api and for some reason the file object has no name and extension.


njkim (approved):
Resolved the issue #7943

Issue #7955, opened 2021-11-05
GeoJSON api fails to serve data loaded via CSV


aj-he:
**Describe the bug**

The geojson api fails where geometry data has been loaded via import_business_data. This is because features don't have the `nodeId` property added to them through this process, unlike when drawn in the geojson-feature-collection widget.

Therefore, when the `nodeId` value is checked at the point below, it errors and the feature is excluded.
https://github.com/archesproject/arches/blob/33c3de8fea01041e889ed350d7b92c0a577bd1db/arches/app/views/api.py#L362-L363

The `nodeId` key should be added as part of the GeoJSON datatype `transform_value_for_tile` function.

**To Reproduce**
1. Load geometry data via `import_business_data` that includes geometries.
2. Try and view that data in a map client using the geojson api

**Screenshots**

**Expected behavior**
Data loaded via `import_busniess_data should be visible via the geojson api

**Your Arches Information**

*   Version used: 6.0.0
*   Operating System and version (desktop or mobile):
*   Browser Name and version:
*   Link to your Arches Install (optional):

**Additional context**

**Ticket Background**
*   Found by: @aj-he

aj-he:
@chiatt please assign me as I'll submitting a fix shortly.

robgaston:
@aj-he I think we don't want to require the `nodeId` property, so I think that the proposed fix ("The nodeId key should be added as part of the GeoJSON datatype transform_value_for_tile function.") here is not what we want.  Rather, we should adjust the GeoJSON API endpoint to not throw an error where that property is missing.  The goal of the lines you show above is to simply remove that property where it is present in the geojson features.

PR #7956, opened 2021-11-05
7955 geojson load prop bug

aj-he:
### Types of changes

### Description of Change

Fixes issue with imported geojson features missing the `nodeId` property 

### Issues Solved

#7955

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @aj-he 
*   Tested by: @ 
*   Designed by: @


aj-he:
@robgaston can you take a look as blocking koop testing for aher

robgaston:
@aj-he I think that we should rework this so that rather than adding the node ids on load, we simply fix the GeoJSON API so that they are not required, as we are simply trying to remove them from the payload in the lines that you specified in the issue (see my comment here: https://github.com/archesproject/arches/issues/7955#issuecomment-962033320)

ultimately, we don't want to require these node ids to be included as properties because they can always be regenerated in code.  The goal is that features loaded must only be valid GeoJSON features, with no special arches requirements.

I think that we can simply change that conditional in api.py into a try/except to get the endpoint working.

aj-he:
@robgaston That's easy enough. My original solution was to check for the `nodeId` key rather than test the value in the GeoJSON api.

Because the property is there in data I have digitised, the assumption is that it is required. Therefore, I adjusted my solution accordingly.

If it is not needed then it should be removed when saving for the same reason you gave?

robgaston:
@aj-he removing those on save would probably be tidier and more proper.  they are used though (and generated when not available) in the web front end, so it would probably be best for that interface to also be responsible for cleaning up it's mess (rather than building that into tile save).  I'd say, let's open a separate ticket to clean that up, as they are innocuous (ie not causing an error or related to the error in #7955) but messy and worse potentially confusing.  I've created this ticket here: #7957

aj-he:
@robgaston great, sounds like a plan

aj-he:
@robgaston Ok, this should be in order now. Could you review?

Could you also look at the koop service PR https://github.com/archesproject/arches-koop/pull/13, which we have been testing alongside this today. Seems to work ok on our small Aircraft dataset.

We can then get going with some ESRI add-in testing of the bigger datasets on Monday/Tuesday.

robgaston (approved):
looks good - thanks!

Issue #7784, opened 2021-09-03
Cache serialized graph proxy model for resource models on publish


robgaston:
When Arches models are published, we should be able to cache the complete & serialized graph data during that publish event.

Based on these requirements, a suitable caching/indexing method for these data should be determined and implemented.

When available, all code that relies on serialized graph data should use the cached versions.

Already published models should be cached on package load, and there should be a management command to cache all published graphs on demand.

Since user permissions are applied outside of the graph design/publishing process (ie they are usually applied as part of a project, not package development), they should be cached separately (see: #7785).

chiatt:
Hey @robgaston About a year ago I wrote a simple command for managing the graph cache.
arches/arches/management/commands/cache.py

It's actually all ready to be implemented in package load: https://github.com/archesproject/arches/blob/e768cd0a533df4ac6b563381fb579a73fc33fa79/arches/management/commands/packages.py#L810-L812

`cache_graphs` just needs to be called after all graph elements (functions, cards, etc..) have been successfully loaded.

PR #7962, opened 2021-11-09
Cache serialized graphs

chrabyrd:
### Types of changes

### Description of Change

Caches graphs, and leverages cached graphs across the codebase. Remove Graph.isactive in favor of Graph.publication

### Issues Solved

#7784

### Checklist


arches/app/views/base.py
```diff
@@ -96,7 +96,7 @@ def get_context_data(self, **kwargs):
         geom_datatypes = [d.pk for d in models.DDataType.objects.filter(isgeometric=True)]
         geom_nodes = models.Node.objects.filter(
             graph__isresource=True,
-            # graph__publication=True,
+            graph__publication=True,
```
robgaston, line 99: same issue here as in advanced_search.py.  we should use the QuerySet API field lookup `isnull`: https://docs.djangoproject.com/en/2.2/ref/models/querysets/#std:fieldlookup-isnull

arches/app/views/resource.py
```diff
@@ -24,6 +24,7 @@
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
 from django.contrib.auth.models import User, Group, Permission
 from django.db import transaction
+from django.db.models.signals import pre_save
```
robgaston, line 27: this import appears to be unused and can probably be removed.

arches/app/views/resource.py
```diff
@@ -217,19 +209,34 @@ def get(
...
+
+        serialized_graph = None
+        if graph.publication and graph.publication.serialized_graph:
+            serialized_graph = graph.publication.serialized_graph
+
+        if serialized_graph:
+            serialized_cards = serialized_graph["cards"]
+            cardwidgets = [
+                widget
+                for widget in models.CardXNodeXWidget.objects.filter(
```
robgaston, line 221: I think that here again we can avoid an extra query to the database if we just create a list of models from the serialized dicts (as in graph.py)

robgaston (changes requested):
@chrabyrd this is looking good for the most part.  I did a pass over the code and added some comments/change requests.

I updated the migration ordering, but I'm sure that will have to change as more migrations will likely be added in 6.x

I also noticed that one of the tests (`test_api_resources_archesjson `) is failing as well, but I haven't looked deeper into that yet.  please take a look at that when you get a chance.

I am going to do some manual testing of the application, but I suspect some of the changes that I requested above will fix some issues that may appear.

Thanks!

arches/app/models/graph.py
```diff
@@ -1524,8 +1580,9 @@ def publish(self, notes=None):
         Adds a row to the GraphPublication table
         Assigns GraphPublication id to Graph
         """
-        publication = models.GraphPublication(
+        publication = models.GraphPublication.objects.create(
             graph=self,
+            serialized_graph=JSONSerializer().serialize(self, force_recalculation=True),
```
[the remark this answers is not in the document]
chrabyrd: just double-checking that this is a non-issue?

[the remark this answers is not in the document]
robgaston: yeah, assuming my understanding above is correct, i don't think any action here is required. thanks!

arches/app/models/graph.py
```diff
@@ -114,22 +116,53 @@ def check_default_configs(default_configs, configs):
...
+                has_deferred_args = False
+                for arg in args:
+                    if type(arg) == Deferred:
+                        has_deferred_args = True
+
+                #  accessing the graph publication while deferring args results in a recursive loop
+                if not has_deferred_args and self.publication and self.publication.serialized_graph:
+                    self.serialized_graph = self.serialize()  # reads from graph_publication table and returns serialized graph as dict
+
+                    nodes = models.Node.objects.filter(pk__in=[node_dict["nodeid"] for node_dict in self.serialized_graph["nodes"]])
```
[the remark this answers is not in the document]
chrabyrd: cool, done. It was a bit more involved without filtering args out of the serialized dict, but it saves us some backend calls so ++

arches/app/models/graph.py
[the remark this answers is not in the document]
chrabyrd: cool yep!

arches/app/models/migrations/7787_relational_data_model.py
```diff
@@ -4,7 +4,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ("models", "7128_resource_instance_filter"),
+        ("models", "7784_add_serialized_graph_to_graph_publications"),
```
robgaston, line 7: I don't think that we want to change this migration, because it is going in 6.1, so this migration will already be executed well before `7784_add_serialized_graph_to_graph_publications` -- if you need this locally and already migrated, you'll need to reverse migrate and then re-run migrations

robgaston (changes requested):
@chrabyrd I noticed an issue with migrations in this PR that I wanted to get out ahead of -- I need to review the rest of the changes still, but wanted to point this out.  thanks!

arches/app/models/migrations/7787_relational_data_model.py
chrabyrd: ah -- fair and smart

chrabyrd:
@robgaston cool, all issues have been addressed. We should probably still follow through with https://github.com/archesproject/arches/issues/8082 after this is merged, with an updated scope. Should likely add both error classes as we were discussing, and ensure the app works as intended throughout. I'll update the scope of the ticket

Issue #7968, opened 2021-11-10
Loading spinner hangs for some steps


chiatt:
In some steps the loading spinner hangs despite the fact that the step is completed and step data has been saved.

chiatt:
resolved in #7970

PR #7970, opened 2021-11-10
Ensure loading spinner does not hang after saving some workflow steps. re #7968

chiatt:
Set complete to false to ensure complete subscription fires. re #7968


chrabyrd (approved):
lgtm!

Issue #7971, opened 2021-11-10
can't save config properties related to widgets and cards


apeters:
it looks like properties that are not localizable can't be saved

apeters:
closed by #7972

PR #7972, opened 2021-11-10
fix save of non localized config properties, re #7971

apeters:
re #7971


ryan86:
Seems to work as expected

Issue #7974, opened 2021-11-12
Reduce queries in get_blank_tile method


chiatt:
The get_blank_tile method in the Tile proxy models queries for the parentnodegroup_id more than once. It could be refactored to make that query just one time.

chiatt:
resolved in #7975

PR #7975, opened 2021-11-12
Reduce queries in Tile.get_blank_tile method

chiatt:
### Types of changes

### Description of Change
Refactors the get_blank_tile Tile proxy model method to reuse parentnodegroup_id reducing db queries

### Issues Solved
#7974

### DIFF
--- a/arches/app/models/tile.py
+++ b/arches/app/models/tile.py
@@ -579,12 +579,10 @@ class Tile(models.TileModel):
 
     @staticmethod
     def get_blank_tile(nodeid, resourceid=None):
-        node = models.Node.objects.filter(pk=nodeid).select_related('nodegroup')[0]
+        node = models.Node.objects.filter(pk=nodeid).select_related("nodegroup")[0]
         parentnodegroup_id = node.nodegroup.parentnodegroup_id
         if parentnodegroup_id is not None:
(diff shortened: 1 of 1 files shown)


arches/app/models/tile.py
```diff
@@ -579,16 +579,15 @@ def is_blank(self):
 
     @staticmethod
     def get_blank_tile(nodeid, resourceid=None):
-        parent_nodegroup = None
         node = models.Node.objects.get(pk=nodeid)
```
apeters, line 582: should we be using `prefetch_related('nodegroup')` here?

apeters (commented):
see my line item comment

arches/app/models/tile.py
chiatt: That's a good suggestion, but prefetch_related is a method on a QuerySet. When we call `objects.get` we don't get a QuerySet, but rather the Node object itself which doesn't have the `prefetch_related` method. 

We could instead do something like:
```models.Node.objects.filter(pk=nodeid).prefetch_related('nodegroup')[0]```

chiatt: I've updated the PR to use `select_related`.  It turns out `prefetch_related` would still used 2 queries joining the data with python. `select_related` uses a single query with a JOIN.

apeters (approved):
looks good to me

PR #7982, opened 2021-11-16
Fix minor in workflows to support HER workflow migration

chiatt:
Fixes undefined variables and uses resourceid from step params passed from earlier step.

### DIFF
--- a/arches/app/media/js/views/components/workflows/final-step.js
+++ b/arches/app/media/js/views/components/workflows/final-step.js
@@ -6,15 +6,7 @@ define([
         this.urls = arches.urls;
         this.loading = ko.observable(true);
         this.workflows = ko.observableArray();
-        try {
-            this.resourceid = ko.unwrap(params.workflow.resourceId);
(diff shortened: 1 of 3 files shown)


arches/app/media/js/views/components/workflows/final-step.js
```diff
@@ -6,15 +6,7 @@ define([
...
-        try {
-            this.resourceid = ko.unwrap(params.workflow.resourceId);
-        } catch(e) {
-            try {
-                this.resourceid = ko.unwrap(params.form.resourceId);
-            } catch(e) {
-                // pass
-            }
-        }
+        this.resourceid = params.resourceid;
```
njkim, line 9: This change seems to break some of the afs workflows.

chiatt: We'll have to adjust the AfS steps - maybe after I finish this ticket: https://github.com/archesproject/arches/issues/7983

PR #7984, opened 2021-11-17
Refine email message and log when email exception

RobOatesHistoricEngland:
### Types of changes

### Description of Change
Refinement of email notification so that the message reads correctly irrespective of whether settings.ADMINS is set, or not. Provide logging when the email fails to send.

### Checklist

-   [X] Unit tests pass locally with my changes

#### Ticket Background
*   Sponsored by: 
*   Designed by: @

### DIFF
--- a/arches/app/views/user.py
+++ b/arches/app/views/user.py
@@ -35,6 +35,10 @@ from arches.app.views.base import BaseManagerView
 from arches.app.utils.forms import ArchesUserProfileForm
 from arches.app.utils.response import JSONResponse
 from arches.app.utils.permission_backend import user_is_resource_reviewer
+import logging
+
+
+logger = logging.getLogger(__name__)
 
 
 class UserManagerView(BaseManagerView):
@@ -174,11 +178,12 @@ class UserManagerView(BaseManagerView):
                         + settings.APP_NAME
                         + " profile was just changed.  If this was unexpected, please contact your "
                         + settings.APP_NAME
-                        + " administrator at %s." % (admin_info)
+                        + " administrator"
+                        + (" at %s." % (admin_info) if (admin_info != "") else ".")
                     )
                     user.email_user(_("Your " + settings.APP_NAME + " Profile Has Changed"), message)
                 except Exception as e:
-                    print(e)
+                    logger.error("Error sending email", exc_info=True)
                 request.user = user
             context["form"] = form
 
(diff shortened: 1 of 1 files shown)


aj-he:
@chiatt I've added you as reviewer as a first step. Please change reviewer as you see fit.

chiatt:
Thanks @aj-he, I'll check it out.

arches/app/views/user.py
```diff
@@ -174,11 +178,12 @@ def post(self, request):
                         + settings.APP_NAME
                         + " profile was just changed.  If this was unexpected, please contact your "
                         + settings.APP_NAME
-                        + " administrator at %s." % (admin_info)
+                        + " administrator"
+                        + (" at %s." % (admin_info) if (admin_info != "") else ".")
```
chiatt, line 182: @RobOatesHistoricEngland, could you change this to use `format` instead of the old string formatting? We are trying to move to more consistent string formatting syntax throughout Arches. I think the line would look like: ```(" at {0}".format(admin_info) if admin_info != "" else ".")```

arches/app/views/user.py
```diff
@@ -174,11 +178,12 @@ def post(self, request):
...
                     )
                     user.email_user(_("Your " + settings.APP_NAME + " Profile Has Changed"), message)
                 except Exception as e:
```
chiatt, line 185: Since `print(e)` was deleted, the `as e` can be removed from this line.

chiatt (changes requested):
Thanks @RobOatesHistoricEngland - this is a nice improvement. I have just a couple of really minor changes requested inline.

chiatt (changes requested):
Thanks @RobOatesHistoricEngland for making those changes. This is a nicely composed solution, but unfortunately you cannot use [f-strings within a translation method](https://code.djangoproject.com/ticket/30109) ```_()```.  In fact, all that you can pass into the translation method are strings and tags, so the ternary within the translation method won't work either. 

Verbose as it is, something like this should work:
```
admin_info = settings.ADMINS[0][1] if settings.ADMINS else None
if admin_info:
    message = _(
        "Your {0} profile was just changed.  If this was unexpected, please contact your administrator at {1}."
    ).format(settings.APP_NAME, admin_info)
else:
    message = _(
        "Your {0} profile was just changed.  If this was unexpected, please contact your administrator."
).format(settings.APP_NAME)
user.email_user(_("Your {0} Profile Has Changed").format(settings.APP_NAME), message)
```

chiatt (approved):
Thanks @RobOatesHistoricEngland, looks great!

Issue #7983, opened 2021-11-16
Make final step more generic


chiatt:
Currently the final step references Arches HER. This needs to be removed or replaced with something more generic.

chiatt:
Fixed in #7986

PR #7986, opened 2021-11-17
Replace final step with generic step, re #7983

chiatt:
### Description of Change
Remove HER specific code from default final step, re #7983


njkim (approved):
Looks good.

Issue #7987, opened 2021-11-17
Fix undefined variables in workflows


chiatt:

chiatt:
Resolved in #7988

PR #7988, opened 2021-11-17
Fix undefined variables in workflow component abstract, re #7987

chiatt:
Temporary fixes for undefined variables until multi-tile logic workflow component abstract is updated. re #7987


njkim (approved):
Fixes some issue multi-tile workflow steps.

Issue #7992, opened 2021-11-21
Url Datatype fails if value is None


chiatt:
If a tile is saved with the URL datatype and no value is saved to the URL node, the tile will fail validation. The URL datatype should allow for `None`.

chiatt:
Resolved in #7993

PR #7993, opened 2021-11-21
Handle for null value in URL datatype, re #7992

chiatt:
### Types of changes

### Description of Change
Prevents Attribute error when saving URL nodes with no value

### Issues Solved
#7992


aarongundel (approved):
works as advertised

Issue #7805, opened 2021-09-08
Localize the domain list widget and datatype configs


apeters:
All configurable widget properties should be localized:
- placeholder
- default value

All configurable datatype properties should be localized:
- options

apeters:
fixed in the latest

PR #7995, opened 2021-11-22
Add migration for domain datatype, re #7805

ryan86:
### Types of changes

### Issues Solved

#

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


apeters (changes requested):
when you first save a card, `[object Object]` shows up as the "placeholder" value

it looks like when saving the node when changing the datatype, it returns an object for the widget that has the placeholder as an object

apeters:
the reverse migration fails with this error:

```
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/Users/alexei/Work/Projects/arches/arches4/env/lib/python3.9/site-packages/django/db/backends/utils.py", line 82, in _execute
    return self.cursor.execute(sql)
django.db.utils.ProgrammingError: operator does not exist: jsonb ||
LINE 9: ..._extract_path(defaultconfig, 'placeholder', 'en'))::jsonb ||
                                                                     ^
HINT:  No operator matches the given name and argument type. You might need to add an explicit type cast.

```

apeters (approved):
looks good to me

Issue #8010, opened 2021-12-03
Create a Setting to Completely Disable Search Exports for Anonymous Users


csmith-he:
**User Story**

Create a setting that allows an application administrator to completely disable search exports for anonymous users.

**Acceptance Criteria**
-   [ ] There is a setting in settings.py that disables anonymous exports.
-   [ ] The setting by default allows exports for anonymous users.
-   [ ] When accessing search page as an anonymous user, the export button and panel **should not** be visible/accessible if the setting is set to True
-   [ ] When logged in with a named user or the setting is set to False, the export button and panel **should** be visible/accessible.
-   [ ] The export API endpoints do not allow exports to occur if unauthenticated
-   [ ] Unit Tests passed
-   [ ] Documentation

**Ticket Background**
*   Sponsored by: Historic England
*   Suggested by: @ Historic England (as part of the Keystone project)
*   Designed by: @ @csmith-he

chiatt:
It probably would be better to make this a list of groups, rather than just the guest user.

PR #8011, opened 2021-12-06
#8010 Disable Export Setting for Anonymous Users

csmith-he:
Provides functionality described in #8010 

If the new setting is set to True and the user is anonymous i.e. not signed in with an application account then the Export tab on the Search page is hidden.  If the user tries to export via the export endpoint then a 403 error is returned.

The setting is False by default so as not to alter core code behaviour on install.

This was functionality requested as part of the Keystone project but seemed like it could have a wider use for the Arches community.

It's been tested by Historic England.


chiatt:
Hey @csmith-he, I think this is a nice feature, but that rather than a boolean, the setting should be a list of groups so that admin's have more control over who can have download access. An admin may not want Crowdsource Editors to have download access either. I think that would be a relatively simple change.

Also, it looks like this branch was made from a branch other than 6.0.x because there are some unrelated commits. I suspect these commits are awaiting reviews in other PRs. If that's the case, we could hold off on merging this in until those commits are merged. If not, it would be better to cherrypick just your commits related to search export into a designated branch and create a new PR.

csmith-he:
@chiatt Yup, I used the wrong base branch, sorry about that.

I'll have a chat with @aj-he about the change you've suggested as I can see how that would be a more flexible approach.  I'll close the PR and will get back to you with amendments once they're done

chiatt:
Thanks @csmith-he - Just to clarify my proposal, the list would include all groupids of groups permitted to use search export. So, by default it would include all groups, and to exclude the anonymous user, you'd just have to remove the `Guest` group. Then in your application logic, you'd just have to check if a user belongs to any groups in the list.

Issue #7785, opened 2021-09-03
Cache user permissions on resource models


robgaston:
Arches should cache resource model user permissions when they are updated in a way that will allow for quick lookup when users are interacting with business data.  This cache will need to be refreshed whenever user permissions are updated as well as when graphs are published & on package load.

Throughout the application, where resource model permissions are applied, the cached objects should be used where available.

From the RDM requirements:

> 
> ...[Implement] the ability to cache user permissions on Arches
> components such as the Arches Designer, Arches Editor, Search, and report pages. This will
> significantly improve the speed of the Arches application for users, and will reduce the number of
> system queries required to establish whether a user has permission to an Arches application feature
> (further improving system performance).
> 
> Similar to model caches, user permission caches will be refreshed every time a model is re-published
> or a system administrator changes the access permissions for a user or user group. This will be
> particularly important for version 2 of the RDM, as users will be requesting access to concepts and
> collections that will be managed just like any other Arches resource model.

PR #8019, opened 2021-12-10
7785 cbyrd cache user permissions

chrabyrd:
### Types of changes

### Description of Change

Caches user permissions

### Issues Solved

#7785 

### To Test

1. Create a user with all privileges, but is not a superuser.
2. Ensure you are running the `silk` DB profiling plugin
3. Compare page load times and DB queries across the application between this branch and master
4. Ensure permissions filtering still works via search
5. Ensure cache busting works via:

   - interaction with the modal in admin templates
   - adding a node to a graph
   - changing node permissions via the permissions manager in the Graph Designer

       Note: There should be a difference in the number of DB queries in a page load with a busted cache vs a non-busted cache.

### Checklist


arches/app/utils/permission_backend.py
```diff
@@ -293,7 +310,10 @@ def check_resource_instance_permissions(user, resourceid, permission):
     try:
         resource = ResourceInstance.objects.get(resourceinstanceid=resourceid)
         result["resource"] = resource
-        all_perms = get_perms(user, resource)
+
+        permissions_checker = CachedObjectPermissionChecker(user, ResourceInstance)
+        all_perms = permissions_checker.get_perms(resource)
```
chiatt, line 315: I've found that this can be stale in some cases allowing access to a restricted instance. To reproduce, create a new instance with admin by adding a single tile. Then 'Manage Permissions' > 'Configure Access to this Instance'. Next paste the new instance's url into another browser in which a different user is logged in and notice that the resource editor loads the instance. It should return a 403 instead. 

Maybe the permissions cache should be cleared when resource instance permissions are updated, or maybe it's worth considering not using the cache here?

chiatt (changes requested):
This looks really good and I've verified the reduction in postgres queries is huge (thousands sometimes in a single request). There are just a couple of comments that I've made inline that are worth some attention before merging.

arches/app/utils/permission_backend.py
robgaston: I think it would be fine to bypass the cache in the case of resource instance permissions for now, and try to address resource instance permission caching as part of #8021

chrabyrd: added a comment in #8021

robgaston (approved):
looks good! thanks!

Issue #7787, opened 2021-09-03
Develop functions for managing and synchronizing Arches graph data with relational database entities


robgaston:
Arches should provide (database) functions for generating and populating a schema of relational database entities representing Arches resource models and the cards of which they are comprised.

A proof-of-concept was done of this as a python script, but database functions would be preferred as a way of quickly generating and populating these entities using only postgresql.

These entities might be tables, or they might instead be views.  The impact of these approaches on performance and maintenance should be considered here.

This will support automatic indexing of business data via connectors between postgresql and elasticsearch, other SQL-based integration points, and possible enhancements to data import, load and bulk-update.

chiatt:
Resolved in #8036

PR #8036, opened 2021-12-17
adds migration for relational data model functions

robgaston:
re: #7787 

This PR adds the basic functions used to generate relational schema/views for arches resource models.

the following SQL query can be used after the migration is executed to generate schema for all active resource models:
```SQL
select __arches_create_resource_model_views(graphid)
from graphs
where isresource = true
    and isactive = true
    and name != 'Arches System Settings';
```

The schema for any resource model should include writable (insert, update & delete) views for instances as well as all nodegroups.  These views should be queryable for read & write operations.

This feature should be considered provisional and only announced in release notes when it is considered complete.  These functions will evolve over time and the interface may change between now and feature completion.

@chiatt I've assigned you for review (thanks!), but it may be helpful to get other team members to review/test this as well -- what do you think?

### DIFF
--- a/arches/app/models/migrations/7787_relational_data_model.py
+++ b/arches/app/models/migrations/7787_relational_data_model.py
@@ -44,7 +44,7 @@ class Migration(migrations.Migration):
             $$ language sql strict immutable;
 
             create or replace function __arches_get_node_value_sql(
-                node record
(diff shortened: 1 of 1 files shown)


chiatt (approved):
Looks good to me!

Issue #8047, opened 2021-12-24
after_update_all not triggered in node_value api post


aj-he:
When changing a value using the node_value api endpoint, the `tile.after_update_all()` is not called to triggered. This means that the geojson_geometries table is not updated if a geojson-feature-collection node is updated.

chiatt:
Fixed in #8048 and cherry picked into dev/6.0.x

PR #8048, opened 2021-12-24
runs after_update_all in Tile.update_node_value

aj-he:
Ensures that the `Tile.update_node_value()` method executes `after_update_all()` on the target tile so that any functionality tied to that gets run.

addresses #8047


chiatt (approved):
Nice catch @aj-he

Issue #8015, opened 2021-07-29
Remove RTF formatting from CSV exports


rlwilliamss:
**Issue Description:** (priority 1 enhancement) - Export Search Results - csv files.  The description text in some of the Records is showing html/RTF formatting. Exports should have all RTF formatting removed but basic line breaks should remain to provide a separation of paragraphs/headers.

![image](https://user-images.githubusercontent.com/78143228/127497767-c30f4588-8840-45b7-a712-85f70dcb65a3.png)

PR #8053, opened 2021-12-29
Remove tag csv search export, #8015 

njkim:
### Types of changes

### Description of Change

Allow the users to choose not to retain html tags when exporting search result in csv format, #8015  

### Issues Solved

#8015 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/search/search_export.py
+++ b/arches/app/search/search_export.py
@@ -36,7 +36,6 @@ from arches.app.utils.geo_utils import GeoUtils
 from arches.app.utils.response import JSONResponse
 import arches.app.utils.zip as zip_utils
 from arches.app.views import search as SearchView
-from arches.app.models.system_settings import settings
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


chiatt (changes requested):
Generally this looks really good, but I think it would be better to move this logic from `to_csv` into the string datatypes `transform_export_values` method. That way we ensure the logic is only applied to strings and doesn't have any unintended consequences on other datatypes. I wouldn't anticipate any issues with core datatypes, but custom datatypes could have tags in them that need to be preserved. Also, putting the logic in `transform_export_values` would apply the feature to other export formats as well, and I think that would be desirable.

chiatt:
I think the need for this to be in `transform_export_values` is even more important now that export is being added to the bulk data manager.

chiatt:
Closing this because I think we will need a datatype specific approach which would be quite different from the solution here

Issue #7846, opened 2021-09-24
Create a way for a .po file to update dynamic labels in the system


apeters:
Create a management command that will update all dynamic labels in the system from a given .po file.

apeters:
fixed by #8057

Issue #7845, opened 2021-09-24
Create a way for a .po file to be generated from dynamic labels in the system 


apeters:
We need to create a management function that will introspect all models looking for localizable fields and then generating a .po file from those entries.

If a .po file already exists, then the command should ask the user if it's ok to overwrite the file.

apeters:
fixed by #8057

PR #7953, opened 2021-11-04
jsonld import export

aarongundel:
IMPORTANT:  Do not merge until #7905 is merged.

### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

i8ln business data import/export

### Issues Solved

#7940 

### Checklist

-   [x] Unit tests pass locally with my changes
-   [x] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

PR #8057, opened 2021-12-30
write and load PO files

aarongundel:
### Types of changes

### Description of Change

Allows writing and loading PO files for graph i8ln

### Issues Solved

#7845 #7846 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


apeters (changes requested):
I tested with a very simple graph, and noticed that some of the settings in the settings graph didn't get an entry in the .po file
```
Image
Map
Project map extent
```
It's because an exception is throw and the label is never processed.
See my inline comment...

Could you also rename files called i8ln.py to i18n.py?

arches/app/utils/i8ln.py
```diff
@@ -0,0 +1,172 @@
...
+
+        self.pofile.save()
+
+    def populate_from_card_x_node_x_widget(self, queryset):
+        for row in queryset:
+            try:
+                i18n_properties = row.config["i18n_properties"]
+                for prop in i18n_properties:
+                    self.append(row.config[prop], "cards_x_nodes_x_widgets")
+                self.append(row.label, "cards_x_nodes_x_widgets")
```
apeters, line 81: if a KeyError is triggered then this row won't be run

aarongundel: good catch; I've added a new test to check for this and modified accordingly.

aarongundel:
CI build appears to be failing because of the api_tests group - this is addressed in the [jsonld](https://github.com/archesproject/arches/pull/7953) ticket.  That should probably be merged first to address the build failure.

arches/app/utils/i18n.py
```diff
@@ -0,0 +1,175 @@
...
+            try:
+                i18n_properties = row.config["i18n_properties"]
+                for prop in i18n_properties:
+                    try:
+                        self.append(row.config[prop], "cards_x_nodes_x_widgets")
+                    except KeyError:
+                        pass
+                self.append(row.label, "cards_x_nodes_x_widgets")
+            except KeyError:
+                pass
```
apeters, line 86: this still fails for me.  line 84 doesn't always get run because of a KeyError triggered on line 78.
Maybe put line 84 ahead of 78.

apeters (changes requested):

aarongundel:
Got it - this should do it.  Added associated test, too.

arches/app/utils/i18n.py
```diff
@@ -0,0 +1,175 @@
...
+        if origin is None:
+            occurrences = []
+        else:
+            occurrences = [(origin, "1")]
+
+        msgid = cell[self.id_language]
+        if self.target_language in cell:
+            msgstr = cell[self.target_language]
+        else:
+            msgstr = ""
```
apeters, line 112: I wonder if this should be rewritten as 
```
try:
    msgstr = cell[self.target_language]
except KeyError:
    msgstr = ""
```

arches/app/utils/i18n.py
```diff
@@ -0,0 +1,179 @@
...
+                    if entry is not None and entry.msgstr != "":
+                        row.config[prop][self.target_language] = entry.msgstr
+                    else:
+                        row.config[prop].pop(self.target_language, None)
+
+                self.update_i18n_string_cell(row.label)
+
+                row.save()
+            except KeyError:
+                pass
```
apeters, line 159: the ArchesPOLoader fails in the same way as the ArchesPOWriter did.
If line 147 throws a KeyError then line 155 and 157 won't run when they should.
I think if you move 155 outside the try block and also add a row.save() after it then it should fix the problem.

apeters (changes requested):
We should also create a unit test to catch this.

apeters (approved):
looks good to me.

Issue #8062, opened 2022-01-05
RDM is broken because the sparql endpoint provider value is localized


apeters:
The RDM can't be accessed because the sparql endpoint value is localized when it wasn't before.  This is because some settings are stored in the database as string data which is now all localized.
The issue appears on line 75 of concept.py

What's strange is that other values that are stored as stings don't seem to have this issue.

PR #8063, opened 2022-01-05
fix sparql endpoint to work with i18n, re #8062

apeters:
re #8062


apeters:
to test this go to the RDM and select any concept.  
It should be broken before the fix is applied.

aarongundel:
works for me.

aarongundel (approved):

PR #8092, opened 2022-01-15
8076 fix related resource broken config

apeters:
to test this create 2 models (A and B)
Model A has a resource instance node that references Model B
Then delete Model B
Go back to Model A and select the resource instance node.  Notice that the UI is broken and can not be recovered from.
Apply the fix and notice that you can now reassign the graph reference.


apeters:
closing in favor of https://github.com/archesproject/arches/pull/8093

PR #8093, opened 2022-01-15
fix broken reference issue with resource instance select widget, re #…

apeters:
to test this create 2 models (A and B)
Model A has a resource instance node that references Model B
Then delete Model B
Go back to Model A and select the resource instance node. Notice that the UI is broken and can not be recovered from.
Apply the fix and notice that you can now reassign the graph reference.


chrabyrd (approved):
works as expected!

Issue #8094, opened 2022-01-15
Node-value node mis-configuration can lead to unhelpful error message and inability to save a tile


apeters:
When a node-value node is in a card that is referenced in the resource descriptor function and the node-value node becomes mis-configured then tiles associated with that nodegroup can't be saved and throw an unhelpful error message.

This most likely happens because the node the node-value node references no longer exists, which is easy to do during graph development.

PR #8095, opened 2022-01-15
fix issue with node-value node mis-configuration, re #8094

apeters:
To test this create a card with 2 string nodes and 1 node-value node.
The node-value node should reference one of the string nodes.
Configure the graph's resource descriptor function to utilize this card and the string node not being referenced by the node-value node.
After the graph is saved, delete the string node that is referenced by the node value node.
Create an instance of the resource and try and save the card by filling out the string value and saving.

Before the fix an unhelpful error message is displayed.
After the fix a much more helpful error message is displyed

re #8094


chrabyrd (approved):
works as intended!

PR #8104, opened 2022-01-20
first attempt at lazy loading portions of knockout components

### LINKED ISSUE #8097: Lazy loading of front end scripts for designer and editor
**Describe the bug**

In the designer and editor, many scripts are loaded up front.  Lighthouse (used in chrome for performance diagnostics) identifies many of these scripts as unused.  

The feasibility of this remains to be seen without a larger rewrite, but the potential gains are significant.  This ticket is related to #8091 

**Your Arches Information**

*   Version used: 6.0.x
*   Operating System and version (desktop or mobile):
*   Browser Name and version:
*   Link to your Arches Install (optional):

**Additional context**

**Ticket Background**
*   Found by: @

aarongundel:
### Types of changes

### Description of Change

Lazy loads certain components - mapboxgl, codemirror, file viewers, and ckeditor.  Needs thorough testing.  Many components cannot be lazy-loaded without major architectural changes to arches.  The lowest hanging fruit relates to lazy loading the dependencies of functions that do not need to return values - and that is what this PR attempts to address.  Examples of that include knockout bindings, as the initialize function does not return values needed to move forward.  The most significant piece of this PR is probably file bindings - which will also affect AFS, so there will be a subsequent PR for AFS to take advantage of this PR, and they should be merged at the same time.  It is also worth noting that as a result of the file bindings change, there are changes to how file bindings get loaded.  I only know of AFS being affected by this, but it's possible other client projects might be affected - especially if they are making heavy use of custom file viewers outside of the file viewer component (AFS uses renderers outside of the file viewer component and is thus affected).  May be worth breaking this up, but I'll leave it whole for now and break up later if desired.  This does not completely resolve #8097, but starts us down the road to resolving it - there are more steps that can be taken in the future. 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


aarongundel:
I've linked to the respective AFS issue in comments above - see [afs 885](https://github.com/archesproject/arches-for-science-prj/issues/885).  Requesting a review from @robgaston and @chrabyrd - since feedback from both of you would be valuable here, especially as it relates to patterns and the viability of this solution.  I see a pattern emerging within knockout bindings.   I do have some concerns about race conditions (outside of the bindings, which I don't think will be affected), but I have not seen them play out yet.  Feel free to hit me up if you're unsure of how to work this PR.

arches/app/media/js/views/components/search/map-filter.js
```diff
@@ -9,18 +9,18 @@ define([
...
-    'codemirror/mode/javascript/javascript'
-], function($, _, arches, ko, BaseFilter, MapComponentViewModel, binFeatureCollection, mapStyles, turf, geohash, mapboxgl, MapboxDraw, geojsonExtent, uuid, geojsonhint) {
+], function($, _, arches, ko, BaseFilter, MapComponentViewModel, binFeatureCollection, mapStyles, turf, geohash,  geojsonExtent, uuid, geojsonhint) {
     var componentName = 'map-filter';
     return ko.components.register(componentName, {
         viewModel: BaseFilter.extend({
             initialize: function(options) {
                 var self = this;
+                self.mapboxgl = require('mapbox-gl');
+                self.MapboxDraw = require('mapbox-gl-draw');
```
chrabyrd, line 22: do these need to be hung on `self`?

arches/app/templates/javascript.htm
```diff
@@ -494,9 +494,12 @@
                 'text': '{{ renderer.title }}',
                 'type': '{{ renderer.type }}',
                 'ext': '{{ renderer.ext }}',
+                'paths': '{{ renderer.paths }}',
```
chrabyrd, line 497: why is `paths` being added?

arches/app/templates/javascript.htm
```diff
@@ -483,8 +483,8 @@
          return;
     });
 
-    define('file-renderers', [{% for renderer in app_settings.RENDERERS %}'{{ renderer.component }}',{% endfor %}], function () {
-        return [{% autoescape off %}{% for renderer in app_settings.RENDERERS %}
+    define('file-renderers', function () {
+        const renderers = [{% autoescape off %}{% for renderer in app_settings.RENDERERS %}
```
chrabyrd, line 487: nit: no sense in declaring a variable to just return it. Inline?

chrabyrd (commented):
Overall looks good! There's some stickler stuff that should be cleared, and I've left a few comments as well. Also I'm gonna circle in @robgaston just to get his blessing as well 👍

arches/app/templates/javascript.htm
aarongundel: good catch; this is cruft from an earlier thought about isolating renderers to specific paths.  Turns out - that's hard and not as useful.

arches/app/templates/javascript.htm
aarongundel: originally I was doing some additional work here, but removed that with the cruft above.  Changed.

arches/app/media/js/viewmodels/map-editor.js
```diff
@@ -5,19 +5,22 @@ define([
...
     'views/components/map',
     'views/components/cards/select-feature-layers',
     'text!templates/views/components/cards/map-popup.htm'
-], function(arches, $, _, ko, koMapping, uuid, MapboxDraw, geojsonExtent, geojsonhint, toGeoJSON, proj4, MapComponentViewModel, selectFeatureLayersFactory, popupTemplate) {
+], function(arches, $, _, ko, koMapping, uuid, geojsonExtent, geojsonhint, toGeoJSON, proj4, MapComponentViewModel, selectFeatureLayersFactory, popupTemplate) {
     var viewModel = function(params) {
         var self = this;
         var padding = 40;
         var drawFeatures;
+        self.MapboxDraw = undefined;
```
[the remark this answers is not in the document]
aarongundel: No; no need for self - I'll fix that.  In re: to your latter question - unless I'm misunderstanding you, I can't change the interior of the viewmodel constructor function to the mapbox-gl-require block because the require is asynchronous - so when the template is renderered, mapbox-gl-draw is not ready yet, so none of the needed observables are setup on the viewmodel and the template loading process errors out.  Happy to discuss this further and come up with an alternative loading method.

arches/app/media/js/viewmodels/map.js
```diff
@@ -381,82 +380,88 @@ define([
...
-            ko.applyBindingsToDescendants(
-                self.getPopupData(feature),
-                self.popup._content
-            );
-            if (map.getStyle() && feature.id) map.setFeatureState(feature, { selected: true });
-            self.popup.on('close', function() {
-                if (map.getStyle() && feature.id) map.setFeatureState(feature, { selected: false });
-                self.popup = undefined;
-            });
+            require(['mapbox-gl'], function(mapboxgl) {
```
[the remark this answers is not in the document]
aarongundel: Yeah; we could always load these when the viewmodel is executed.  This was one my early attempts to load this dynamically; I think it probably makes more sense to put it at the beginning of the vm.

arches/app/media/js/views/components/cards/map.js
```diff
@@ -4,8 +4,7 @@ define([
     'viewmodels/card-component',
     'viewmodels/map-editor',
     'bindings/chosen',
-    'bindings/codemirror',
-    'codemirror/mode/javascript/javascript'
```
robgaston, line 8: I think that removing this prevents the syntax highlighting that we use in this interface.  Is that something that you considered, or did you believe that this module was unused?

arches/app/media/js/views/components/datatypes/geojson-feature-collection.js
```diff
@@ -6,7 +6,6 @@ define([
     'bindings/color-picker',
     'bindings/mapbox-gl',
     'bindings/codemirror',
-    'codemirror/mode/javascript/javascript',
```
robgaston, line 9: same comment here as in `map.js`

arches/app/media/js/views/components/geocoders/base-geocoder.js
```diff
@@ -10,6 +10,12 @@ function (ko, mapboxgl, arches) {
      */
     var BaseGeocoderViewModel = function(params) {
         var self = this;
+        let mapboxgl;
+
+        require(['mapbox-gl'], (mapbox) => {
```
robgaston, line 15: this is probably safe from race conditions given the way the module is used internal to this file (it is only called when self.map is available, which means this module has been loaded), so I think in this case it's fine.

arches/app/media/js/views/components/widgets/map.js
```diff
@@ -7,7 +7,6 @@ define([
     'viewmodels/map-editor',
     'bindings/chosen',
     'bindings/codemirror',
-    'codemirror/mode/javascript/javascript',
```
robgaston, line 10: same question here as in `cards/map.js`

arches/app/media/js/views/map-layer-manager.js
```diff
@@ -12,7 +12,6 @@ define([
     'arches',
     'bindings/mapbox-gl',
     'bindings/codemirror',
-    'codemirror/mode/javascript/javascript',
```
robgaston, line 15: same question here as in `cards/map.js`

robgaston (changes requested):
@aarongundel looks good, but a few comments came up that I'd like you to take a look at.  should be pretty straightforward to resolve, I think.  thanks!

arches/app/media/js/views/components/cards/map.js
robgaston: makes sense; I thought that might be the case but wanted to make sure, as I think that was the only reason these were ever included.  thanks!

robgaston (approved):
thanks!

Issue #8058, opened 2022-01-03
Allow user to tab through workflow elements in efficient order


chiatt:
In order to streamline workflow data entry, a user should be able to tab through workflow buttons to minimize mouse usage.

For custom steps this will require some custom html changes, re archesproject/arches-for-science-prj#862

chiatt:
Closing per #8109

PR #8109, opened 2022-01-22
Enable tab focus through the workflow buttons

njkim:
### Types of changes

### Description of Change

Add tab index to the workflow buttons to let users to navigate through the buttons using tab, #8058 

### Issues Solved

#8058 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/css/arches.scss
+++ b/arches/app/media/css/arches.scss
@@ -762,6 +762,11 @@ input[type="checkbox"] {
     padding: 8px 15px;
 }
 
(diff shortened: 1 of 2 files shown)


chiatt:
@njkim I think one thing that we will need to do as part of this is to indicate to the user that the workflow buttons are in focus. Currently that's not the case.

chiatt:
@dwuthrich has already resolved this in the latest

Issue #8110, opened 2022-01-25
Nullguard datatype lookup in Resource.get_documents_to_index


chrabyrd:
If a node has been deleted from a Model, but references to the node still exist in Tile data, the Resource will fail to index. The failure stems from attempting to lookup the node's datatype.

chiatt:
Resolved in #8118

PR #8118, opened 2022-01-27
Cherry pick 8111 (nullguard resource index datatype lookup) into 6.0.x , re #8110

chrabyrd:
### Types of changes

### Description of Change

Pulls the changes from #8111 , which is 5.1.x based, into 6.1.x

### Issues Solved

#8110


chiatt (approved):
Resolves the issue, but only if the tile value of the missing node is [], {}, "", or None.

Issue #7838, opened 2021-09-22
Add a parameter to the label based graph to filter hidden nodes and nodegroups 


chiatt:
Currently the label based graph provides data for all nodes and nodegroups. This allows data to be seen by users that would otherwise be hidden in reports and the editor based on widget and card visibility. The label based graph should take a parameter indicating whether the 'hidden' nodes and nodegroups be filtered from the returned data.

chiatt:
Alternatively, a 'hidden' property could be sent with the payload. A parameter could still be added to the url to prevent that property from appearing in the response.

chiatt:
Fixed in #8121

PR #8121, opened 2022-01-28
filter hidden nodes

aarongundel:
### Types of changes

### Description of Change

Adds new feature to filter nodes based on card "visible" attribute.  

### Issues Solved

#7838 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/models/resource.py
+++ b/arches/app/models/resource.py
@@ -604,7 +604,7 @@ class Resource(models.ResourceInstance):
 
         return JSONSerializer().serializeToPython(ret)
 
-    def to_json(self, compact=True, hide_empty_nodes=False, user=None, perm=None, version=None):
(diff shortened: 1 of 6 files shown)


aarongundel:
@chiatt I added this to the bulkdisambiguatedresourceinstance endpoint under the hidden=false parameter.  Let me know if there are other places I should add this.

chiatt (changes requested):
This looks really good, but `resource.to_json` is called in the api.Resources.get method as well. The hidden property will need to be passed in there too.

aarongundel:
Thanks for pointing that out @chiatt - I've corrected that and simplified some variable parsing logic.

chiatt (approved):
Looks good to me!

PR #8158, opened 2022-02-12
7804_7805_i18n_domain_and_domainlist options

apeters:


aarongundel:
I've found a strange issue with the internationalized domain values in the graph designer.  It's hard to consistently reproduce, but this works most of the time.  

1.  Create a new graph.  
2. Provide a name for the top node and add a child.
3. Provide a name for the child and type it as a **domain value list**.  Add some values.  Save.
4. Create a new child node off the top node.
5. Provide a name for the child node and change the type to **domain value**.  Add some values.  **Do not save.**
6. Switch the language. 
7. The graph will explode in the designer.  For some reason, dictionaries get improperly nested like so: `{en: {en: "name", es: "nombre"}, es: "nombre"}` and this causes the error.

apeters:
@aarongundel I wasn't able to reproduce the error.  Maybe you didn't reverse migrate before you pulled?
If you run this, this might fix the issue for you.
```
python manage.py migrate models 6458_language
python manage.py migrate
```

aarongundel (changes requested):
Needs merge conflict fix.  I think it should be as simple as removing the offending file.

Issue #8162, opened 2022-02-15
Provide map source and map layer group permissions using Django admin back-office


RobOatesHistoricEngland:
**User Story**
Provide the ability to restrict access to maps layers and/or map sources for specific user groups.

Map layers have an attribute `Addtomap` which when set forces the map layer to be loaded. This behaviour needs to be honoured as a runtime error will occur otherwise.

A check is made for superuser privilege. If logged in as superuser, then all sources and layers are loaded in their entirety completely circumventing the user/group permissions.

Map layers are loaded if they are not excluded based on the user/group permissions set, but also if the associated map source is also not excluded. Most of the time a map layer is associated with a single map source, but this is not always the case. Sometimes a layer can be associated with several sources. For the layer to be included, not only must the layer not be excluded but all sources must also be not excluded. In the situation where a layer has two sources defined - if one source is not excluded but the other is excluded then the layer and its two sources are excluded completely. There is no way to know whether the layer would operate correctly if one of the two sources was omitted.

To recap, if a layer is excluded, none of it's sources are added. However, if the sources that make up that layer are later used in a different layer, then the sources will get added. If you specifically wanted to exclude a source, then any layer that was linked to that source would be excluded and therefore the source would also be excluded. The exception to this rule is for any map layer that has the `Addtomap` flag set, in which case the layer and associated sources are added regardless of any specific exclusions set at the layer or source level.

A future enhancement will be to provide this functionality via the Arches UI, rather than Django admin.

Process
**Acceptance Criteria**
-   [ ] Map layer permissions available in Django admin (No access to map layer)
-   [ ] Map source permissions available in Django admin (No access to map source)
-   [ ] Default behaviour is that all maps sources/layers are available to all. This is an opt-out permission not opt-in
-   [ ] May layers marked `Addtomap` will always be available
-   [ ] Superusers have access to all map sources/layers
-   [ ] Map layers/sources are excluded based on the specific user group exclusions set at the layer/source level as detailed above

**Definition of Done**
Defined groups with restricted access to maps layers/sources are not presented with the layers/sources within Arches.  The exception is for those users who have superuser status - where all layers/sources are provided, or where the map layer is mandatory based on the `Addtomap` property set.

**Ticket Background**
*   Sponsored by: Historic England
*   Suggested by: @HistoricEngland
*   Designed by: @RobOatesHistoricEngland

PR #8163, opened 2022-02-15
Map source and map layer permissions

RobOatesHistoricEngland:
### Types of changes

### Description of Change
Allow user/group based restrictions for map sources and map layers. See [issue 8162](https://github.com/archesproject/arches/issues/8162)  for more information.

### Issues Solved

#

### Checklist

#### Ticket Background
*   Sponsored by: @HistoricEngland
*   Found by: @ 


aj-he:
@robgaston assigned a review to you but by all means reallocate.

robgaston:
This change looks mostly good, but I question the assignment of permissions to map sources.  I think that all applying permissions to map sources really achieves is the concealing the source definitions themselves, but this comes at the price of some brittleness as you must apply permissions to both layers and sources and therefore they might be out of sync.  If we really just want to control the set of overlays that a user sees in the front end, then I think we can achieve that by just applying permissions to `map_layers`.

My initial take is that the correct way to secure sources is on the service end.  Applying permissions to map sources could be deceptive, as users might assume that actually secures the backend service when, in fact, it just controls what source definitions are supplied to the front end, and a user who knew the service endpoint could still access those data unless the service itself was secured in some way.

Thoughts?

PR #8173, opened 2022-02-22
graph publication column save type should be json instead of string

### LINKED ISSUE #8172: Cached graph is saved to JSONField as a string, should be dictionary/object
We save a serialized version of the graph models to `graph_publications.serialized_graph`, which is a JSONField.  Currently, this is stored as a string, which causes some issues with how we handle the serialized graph (see the add resource page).  It also would be more useful/usable in the database if stored as an object.

When we assign this column, it should be using a dictionary so that we store a proper JSON object.

aarongundel:
### Types of changes

### Description of Change

Fix for issue where a string based index was expected, but a string object was received.  This happened because the graph was saved to the db as a string instead of a JSON structure

### Issues Solved

Fixes #8172 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/models/graph.py
+++ b/arches/app/models/graph.py
@@ -1596,14 +1596,14 @@ class Graph(models.GraphModel):
                     notes=notes,
                 )
                 publication.save()
-            except Exception:
-                raise UnpublishedModelError("Could not save graph publication")
+            except Exception as e:
+                raise UnpublishedModelError(e)
 
             try:
                 self.publication = publication
                 self.save(validate=False)
-            except Exception:
-                raise UnpublishedModelError("Could not save graph with new publication")
+            except Exception as e:
+                raise UnpublishedModelError(e)
 
     def unpublish(self):
         """
--- a/arches/app/models/graph.py
+++ b/arches/app/models/graph.py
@@ -1307,13 +1307,13 @@ class Graph(models.GraphModel):
         exclude = [] if exclude is None else exclude
 
         if self.publication and not force_recalculation:
-            deserialized_graph = self.publication.serialized_graph
+            serialized_graph = self.publication.serialized_graph
 
             for key in exclude:
-                if deserialized_graph.get(key) is not None:  # explicit None comparison so falsey values will still return
-                    deserialized_graph[key] = None
+                if serialized_graph.get(key) is not None:  # explicit None comparison so falsey values will still return
+                    serialized_graph[key] = None
 
-            return deserialized_graph
+            return serialized_graph
         else:
             ret = JSONSerializer().handle_model(self, fields, exclude)
             ret["root"] = self.root
@@ -1377,17 +1377,14 @@ class Graph(models.GraphModel):
                 unpermitted_edits = []
                 db_nodes = models.Node.objects.filter(graph=self)
                 for db_node in db_nodes:
-                    try:
-                        unpermitted_node_edits = find_unpermitted_edits(
-                            db_node,
-                            self.nodes[db_node.nodeid],
-                            ["name", "issearchable", "ontologyclass", "description", "isrequired", "fieldname", "exportable"],
-                            "node",
-                        )
-                        if unpermitted_node_edits is not None:
-                            unpermitted_edits.append(unpermitted_node_edits)
-                    except KeyError:
-                        pass
+                    unpermitted_node_edits = find_unpermitted_edits(
+                        db_node,
+                        self.nodes[db_node.nodeid],
+                        ["name", "issearchable", "ontologyclass", "description", "isrequired", "fieldname", "exportable"],
+                        "node",
+                    )
+                    if unpermitted_node_edits is not None:
+                        unpermitted_edits.append(unpermitted_node_edits)
                 db_graph = Graph.objects.get(pk=self.graphid)
                 unpermitted_graph_edits = find_unpermitted_edits(
                     db_graph,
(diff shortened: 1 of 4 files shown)


aarongundel:
Note that in order to test you should roll back to 8085 and then roll forward - this will re-save the cached items as json structured items instead of strings.

aarongundel:
@robgaston requested more eyes since he's occupied with training until tomorrow

arches/app/models/graph.py
```diff
@@ -1307,7 +1307,7 @@ def serialize(self, fields=None, exclude=None, force_recalculation=False):
         exclude = [] if exclude is None else exclude
 
         if self.publication and not force_recalculation:
-            deserialized_graph = JSONDeserializer().deserialize(self.publication.serialized_graph)
+            deserialized_graph = self.publication.serialized_graph
```
chrabyrd, line 1310: nit: can `deserialized_graph` be changed to something else?  `deserialized_graph = ...serialized_graph` is confusing

arches/app/models/graph.py
```diff
@@ -1586,15 +1589,18 @@ def publish(self, notes=None):
...
+                notes=notes,
+            )
+            publication.save()
 
-        self.publication = publication
-        self.save()
+            self.publication = publication
+            self.save()
+        except Exception:
+            raise UnpublishedModelError("Could not save graph publication ")
```
chrabyrd, line 1603: can this error be more descriptive, can we leverage the exception?

arches/app/models/graph.py
```diff
@@ -1586,15 +1589,18 @@ def publish(self, notes=None):
         Adds a row to the GraphPublication table
         Assigns GraphPublication id to Graph
         """
-        publication = models.GraphPublication.objects.create(
-            graph=self,
-            serialized_graph=JSONSerializer().serialize(self, force_recalculation=True),
-            notes=notes,
-        )
-        publication.save()
+        try:
```
chrabyrd, line 1592: when will this fail? is it possible / worth it to replace the try/except with a check of some kind?

arches/app/models/migrations/7784_add_serialized_graph_to_graph_publications.py
```diff
@@ -14,7 +14,9 @@ class Migration(migrations.Migration):
     def forwards_add_serialized_graph_column_data(apps, schema_editor):
         for graph in Graph.objects.all():
             if graph.publication:
-                graph.publication.serialized_graph = JSONSerializer().serialize(graph, force_recalculation=True)
+                graph.publication.serialized_graph = JSONDeserializer().deserialize(
```
chrabyrd, line 17: is there any point to saving a deserialized serialized graph? Can the original graph just be used?

chrabyrd (commented):
overall looks good just some minor questions 👍

arches/app/models/graph.py
```diff
@@ -1377,14 +1377,17 @@ def find_unpermitted_edits(obj_a, obj_b, ignore_list, obj_type):
...
+                    try:
+                        unpermitted_node_edits = find_unpermitted_edits(
+                            db_node,
+                            self.nodes[db_node.nodeid],
+                            ["name", "issearchable", "ontologyclass", "description", "isrequired", "fieldname", "exportable"],
+                            "node",
+                        )
+                        if unpermitted_node_edits is not None:
+                            unpermitted_edits.append(unpermitted_node_edits)
+                    except KeyError:
```
chrabyrd, line 1389: what causes this to error?

arches/app/models/graph.py
aarongundel: yep it is.

arches/app/models/migrations/7784_add_serialized_graph_to_graph_publications.py
aarongundel: No; the original graph isn't compatible with the JsonField type and blows up.

chrabyrd: cool just wanted to check 👍

arches/app/models/graph.py
aarongundel: This was actually the result of a corrupt graph I had - where there was no db_node.nodeid in self.nodes.  I'll remove this; as it should never happen.

arches/app/models/graph.py
aarongundel: not sure how much more descriptive it can be; happy to get your thoughts on this.

chrabyrd: oh sure, instead of that sentence can pass the declared Exception? `except Exception as e:`

robgaston (changes requested):
this looks good to me, I think we just need to resolve the merge conflicts in graph.py before this will be ready to go

arches/app/models/graph.py
```diff
@@ -1586,15 +1586,22 @@ def publish(self, notes=None):
...
+        with transaction.atomic():
+            try:
+                publication = models.GraphPublication.objects.create(
+                    graph=self,
+                    serialized_graph=JSONDeserializer().deserialize(JSONSerializer().serialize(self, force_recalculation=True)),
+                    notes=notes,
+                )
+                publication.save()
+            except Exception:
+                raise UnpublishedModelError("Could not save graph publication")
```
chrabyrd, line 1598: Instead of raising the string, can we pass the exception here?

arches/app/models/graph.py
```diff
@@ -1586,15 +1586,22 @@ def publish(self, notes=None):
...
 
-        self.publication = publication
-        self.save()
+            try:
+                self.publication = publication
+                self.save()
+            except Exception:
+                raise UnpublishedModelError("Could not save graph with new publication")
```
chrabyrd, line 1604: Instead of raising the string, can we pass the exception here?

chrabyrd (commented):
Looks great just a minor nit 👍

arches/app/models/graph.py
aarongundel: good feedback, done and done.

robgaston (approved):
looks good

PR #8193, opened 2022-03-01
Hide button with HIDE_RESOURCE_COPY_BUTTON setting

RobOatesHistoricEngland:
### Types of changes

### Description of Change
Enable the Copy Resource button to be removed from the UI if there is an issue using this functionality.

### Issues Solved
Prevent copying of resources when the copy function is not working properly.

### Checklist

#### Ticket Background
*   Sponsored by: @HistoricEngland
*   Found by: 

### DIFF
--- a/arches/app/templates/navbar/resource-manage-menu.htm
+++ b/arches/app/templates/navbar/resource-manage-menu.htm
@@ -9,6 +9,7 @@
         <div class="panel-body">
             <ul class="list-group ep-menu-list editor-tools">
 
+                <!-- ko ifnot: ("{{app_settings.HIDE_RESOURCE_COPY_BUTTON}}" === "True") -->
                 <!-- Menu Item -->
                 <li id="settings-manager" class="edit-menu-item" data-bind="css: {'disabled': !resourceId()}">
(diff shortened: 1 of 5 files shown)


chiatt (commented):
Though I can see the value in this, we want to avoid adding more settings to Arches if there is a reasonable alternative to doing so. In this case I would say that hiding the copy button could be handled by adding some css to a project's `project.css` file - maybe something like:

```
.list-group.ep-menu-list.editor-tools > li:first-child {
    display: none;
}
```

aj-he:
@chiatt I think that there is a functionality consideration here too, whereby the admin may not want resources to be copyable?

Perhaps this needs to be a configuration of the graph?

While the function extensions aren't able to differentiate between copy and non-copy operations, and thus adjust their behaviour, we will need to include this in any arches deployment that has a custom function.

chiatt:
I'm going to close this because the copy resource button is going to be hidden in #8445 and won't be available until #8435 is resolved. Once #8435 is fixed, hiding the copy button won't be necessary because developers can pass the copy context to their functions re #8339

PR #8191, opened 2022-02-28
publication id preservation

aarongundel:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

Changes resource instance model to preserve the graph publication ID on creation or update of a resource instance or its tiles.  

### Issues Solved

Fixes #8186

### Checklist

-   [ ] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

PR #8206, opened 2022-03-03
8186 publication id preservation

chrabyrd:
This is meant to supercede: https://github.com/archesproject/arches/pull/8191

### DIFF
--- a/tests/models/graph_tests.py
+++ b/tests/models/graph_tests.py
@@ -156,10 +156,6 @@ class GraphTests(ArchesTestCase):
         }
         models.Edge.objects.create(**edges_dict).save()
 
-    @classmethod
-    def tearDownClass(cls):
-        cls.deleteGraph("2f7f8e40-adbc-11e6-ac7f-14109fd34195")
-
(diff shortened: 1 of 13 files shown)


tests/models/graph_tests.py
```diff
@@ -156,10 +156,6 @@ def setUpClass(cls):
         }
         models.Edge.objects.create(**edges_dict).save()
 
-    @classmethod
-    def tearDownClass(cls):
-        cls.deleteGraph("2f7f8e40-adbc-11e6-ac7f-14109fd34195")
```
aarongundel, line 161: This is interesting.  This is why the tests were blowing up for me originally, because teardown was failing right here.  This makes sense though; seems like you've chosen to just let the test suite tear down the DB than fiddling with deleting this graph.

aarongundel (commented):
This looks good to me; but I'm biased since I wrote part of it, so I'll save the approval for @robgaston

robgaston (approved):
looks good.  thanks!

Issue #8009, opened 2021-12-02
Create a new ETL plugin


njkim:
Create a new ETL plugin that allows the users to  import / export though the arches ui.
based on the mockup below.

[https://app.zenhub.com/files/49451058/70450544-b396-4816-a19f-7b4d3213787e/download](https://app.zenhub.com/files/49451058/70450544-b396-4816-a19f-7b4d3213787e/download)

PR #8209, opened 2022-03-04
Allow user to load etl modules from a package, re #8009

chiatt:
Allow users to load etl modules from a package and creates new staging tables, re #8009


njkim (approved):
It looks great

PR #8232, opened 2022-03-10
8231_i81n_dir_switch

apeters:


apeters:
To test, add a language that renders in rtl (Arabic, Hebrew, etc...).  Select the language and notice the UI switches directionality.

aarongundel (approved):
works for me!

Issue #8247, opened 2022-03-14
Provide ability to override default primary descriptors function


bferguso:
**User Story**
Developers can implement custom functions for Primary Display descriptors to allow for application-specific logic to generate the primary descriptors. Function definition would follow the existing pattern in Arches, and functions providing that service would need to extend a base primary_descriptors class.

**Acceptance Criteria**
-   [x] Base class created
-   [x] Existing primary_descriptors class updated
-   [x] Sample implementation of an alternate function
-   [x] Unit Tests passed
-   [ ] Documentation

**Definition of Done**
- Default `primarydescriptors` configuration is compatible with existing implementation or data migration provided
- Custom `primarydescriptors` functions can be implemented and configured within the current Functions pattern
- Different `primarydescriptors` functions can be assigned to different graphs
- Remove `primarydescriptors` implementation-specific logic from function consumers

**Ticket Background**
*   Sponsored by: BC Government
*   Suggested by: @bferguso
*   Designed by: @bferguso
*   Initial discussion: https://community.archesproject.org/t/custom-primary-display-name-functions/1552

aj-he:
@bferguso @chiatt Can an example of creating a custom primary descriptor function be added to arches-docs under the Types of Extension section?

https://arches.readthedocs.io/en/latest/creating-extensions/#types-of-extensions

chiatt:
Resolved in #8256

PR #8256, opened 2022-03-15
8247 Allow for custom primary descriptors functions

bferguso:
### Types of changes

### Description of Change

Removes the hard-coded primary descriptors function module and class name from the arches core and embeds that data in the `functions.defaultconfig` and `functions_x_graphs.config` database column so that custom `primarydescriptors` functions can be implemented.

### Issues Solved

Removes limitations of the built-in primary descriptors function by allowing custom `primarydescriptor` functions to be configured

### Issues to still be resolved
- [x] Fix workaround in `app/views/resource.py` [lines 180-197] and `app/views/graph.py` [lines 177-193] to make function.functiontype available in the knockout model
- [x] Determine how to handle the updated `primarydescriptors` config JSON structure. \*
- [ ] Determine what to do with messages in `.../resources/formats/csvfile.py` that are tightly tied to existing PrimaryDescriptorsFunction implementation \*
- [ ] Determine what to do with logic in `card-tree.js`, `graph-tree.js` and `node-form.js` that is tightly tied to existing PrimaryDescriptorsFunction implementation \*
- [ ] Define/implement behaviour if multiple `primarydescriptors` functions are defined for a graph

\* See _Further comments_ section below

### Checklist

#### Ticket Background
*   Sponsored by: Self Funded
*   Found by: @bferguso
*   Tested by: @bferguso
*   Designed by: @bferguso

### Further comments
Issue: #8247 

#### Changes to existing `functions.defaultconfig` and `functions_x_graphs.config` JSON structure
The current proposal to update the JSON structure will cause issues for projects updating to this change. This was done to allow the module and class name to be embedded in the function's configuration. Possible solutions:
1. provide a model migration to update the configuration, or 
2. make the new logic backwards compatible.

#### Messages and formatting tightly tied to current implementation of `PrimaryDescriptorsFunction` logic
The current logic to show what nodes are used in the name, description and popup_menu primary descriptors is only valid with the current implementation of that function.  Custom function implementations will likely not use the same logic to determine which nodes are part of those descriptors, making the current logic invalid. Some possible solutions:
1. Bypass the logic that is tightly tied to the default PrimaryDescriptorsFunction implementation if it is not used. This would effectively eliminate the csvfile messages and GraphEditor formatting that indicates what nodes are used in the descriptors. (Probably not ideal)
2. Embed this information in the defaultconfig JSON so that it becomes generic for all functions independent of the `primarydescriptor`'s implementation
3. Add another function to the AbstractPrimaryDescriptorsFunction interface that would provide this functionality


arches/app/media/js/views/graph/graph-designer/graph-tree.js
```diff
@@ -93,35 +94,24 @@ define([
...
+                return null;
+
+            ['name', 'description'].forEach(function (descriptor) {
+                try {
+                    primaryDescriptorNodes[pdFunction()['config']['descriptor_types'][descriptor]['nodegroup_id']] = descriptor;
+                } catch (e) { } // Descriptor doesn't exist so ignore the exception
+            });
+
+            [node].concat(!!node['childNodes']() ? node['childNodes']() : [])
+                .find(nodeToCheck => !!(descriptorType = primaryDescriptorNodes[nodeToCheck['id']]));
```
bferguso, line 112: The `children` attribute didn't seem to exist. I've changed it to childNodes, which I think was the intent. I believe this was to highlight the parent node to make it easier to find the display node, however it can cause the tree root node to highlight which is a bit confusing. The node-forms.jsFuncNode has the same logic.

I'm wondering if the parents of the display nodes should be formatted slightly differently than the nodes that actually participate in the name/description function, or the parents not be highlighted at all.

chiatt: By display node, are you referring to the selected node? If that's the case aren't the parents of the selected node already styled differently (light grey as opposed to dark blue)?

chiatt: Ah, that's right. I think your change makes sense. 

As far as changing the highlighting of parents, I'm not sure what the right answer would be. Others would have to weigh in (maybe @dwuthrich, @robgaston, @apeters, @aj-he). The intent is to show editors where to find the cards that need to be completed in order to make the resource instance recognizable. For deeply nested branches the descriptor cards can be buried quite deep, so highlighting of the parents is intended to light the way. I think the assumption was made that users would understand that the lowest highlighted card is the target, but I can see how that might not be clear to people unfamiliar with the system.  

The drawback with highlighting parents differently is that we already have a color in the resource editor for cards that are provisional and another color for cards that have data, but are unsaved. If we add more colors, we're going to need a legend (maybe we are already at that point). On the other hand, if we were to remove the highlighting, the user would have to expand all of the cards in order to find the descriptor card(s). Would symbols rather than colors be better? I'm not sure.

In either case, we should probably create a different issue for this question.

bferguso: Created the issue for discussion here: #8260

arches/app/media/js/views/graph/graph-designer/card-tree.js
```diff
@@ -360,27 +361,23 @@ define([
...
             selection: selection,
             filter: filter,
-            isFuncNode: function() {
-                var appFuncs = null, appFuncDesc = false, appFuncName = false, nodegroupId = null;
-                if(params.card && this.appliedFunctions()) {
+            isFuncNode: function () {
+                var nodegroupId = null, pdFunction = this.primaryDescriptorFunction;
+
+                // params.card always seems to be undefined...
+                if (params.card && pdFunction) {
```
bferguso, line 368: params.card seems to always be undefined. It's not passed as a parameter to the CardTreeViewModel. Not sure if/where this is supposed to happen.

bferguso:
@chiatt - I'm wondering about moving the the unresolved issues in the description to another issue so I can finish the work in this PR. The changes made will still work wth the built-in descriptors function, and I can propose a solution to the rest separately.

If that makes sense, can you let me know if JSON configuration changes like this are typically handled as a DML migration? If so I'll put one in and this PR should be good for review.

The only stickler exception is the naming convention for the map_popup identifier in primary-descriptors.js, but that flows through to the JSON function config so I think it should be disregarded.

chiatt:
@bferguso That sounds like a good plan to me. You're right, typically we would handle the JSON configuration changes with a DML migration, but hat can easily be handled as a separate issue.

bferguso:
I've added the migration, but the dependencies may need to be updated if others need to go in first.

chiatt:
Thanks @bferguso, that's great. I'll test this out tomorrow.

arches/app/models/resource.py
```diff
@@ -71,11 +71,13 @@ def __init__(self, *args, **kwargs):
...
 
     def get_descriptor(self, descriptor):
-        module = importlib.import_module("arches.app.functions.primary_descriptors")
-        PrimaryDescriptorsFunction = getattr(module, "PrimaryDescriptorsFunction")()
-        functionConfig = models.FunctionXGraph.objects.filter(graph_id=self.graph_id, function__functiontype="primarydescriptors")
-        if len(functionConfig) == 1:
-            return PrimaryDescriptorsFunction.get_primary_descriptor_from_nodes(self, functionConfig[0].config[descriptor])
+        function_config = models.FunctionXGraph.objects.filter(graph_id=self.graph_id, function__functiontype="primarydescriptors")
+        if len(function_config) == 1:
+            module = importlib.import_module(function_config[0].config["module"])
```
bferguso, line 76: This could have originally been:
```
            func = function_config[0].function.get_class_module()()
            return func.get_primary_descriptor_from_nodes(
                 self, function_config[0].config["descriptor_types"][descriptor]
            )
```
I'm not sure if this was previously hard-coded for performance issues. the config.module & config.classname JSON attributes could potentially be collapsed into packagename if the above is too heavy. Could be something like this:
```
            module = importlib.import_module(function_config[0].config["package"]+"."+function_config[0].modulename)
            primary_descriptors_function = getattr(module, function_config[0].classname)()
            return primary_descriptors_function.get_primary_descriptor_from_nodes(
                self, function_config[0].config["descriptor_types"][descriptor]
            )
```
Same holds true for api.py

arches/app/media/js/views/graph/graph-designer/card-tree.js
chiatt: I agree - there doesn't look like a case in which params.card has a value. It must have been possible in 2019 (https://github.com/archesproject/arches/commit/82c9e88b9c5e15459910d6a94d6b32557a855ad1).

arches/app/models/resource.py
chiatt: I took another look at this and realized that the original code is using the function object's get_class_module and immediately creating and instance of the module. In this way the module location and class name don't come from the config, but rather the classname and modulename columns of the Function model. This is also the pattern that is used with Functions that are not primary descriptors. 

I think, however, the config would still have to come from the FunctionXGraphs instance. So something like this:

```
def get_descriptor(self, descriptor):
    function = models.Function.objects.filter(graphmodel=self.graph_id, functiontype="primarydescriptors")
    if len(function) == 1:
        module = function[0].get_class_module()()
        config = models.FunctionXGraph.objects.get(graph_id=self.graph_id, function=function[0].functionid).config
        return module.get_primary_descriptor_from_nodes(self, config["descriptor_types"][descriptor])
    else:
        return "undefined"
```

chiatt (changes requested):
This is really great work @bferguso. I went through the process of loading my own custom descriptor function and it worked like a charm. I have just a couple of changes that hopefully aren't too much trouble. One is adding a reverse migration (this is super helpful for developers when switching between branches with different migrations). The other is using the `get_class_module` method to get a module instance rather than using the module and classname in the function config.

chiatt:
@bferguso I updated master and there is a new migration `7874_node_alias`, so you will need to make that dependency in your migration.

arches/app/models/resource.py
bferguso: @chiatt - How about this to take advantage of the single database call and still using the get_class_module() on the related function?
```
    def get_descriptor(self, descriptor):
        graph_function = models.FunctionXGraph.objects.filter(
            graph_id=self.graph_id, function__functiontype="primarydescriptors"
        ).select_related("function")
        if len(graph_function) == 1:
            module = graph_function[0].function.get_class_module()()
            return module.get_primary_descriptor_from_nodes(
                self, graph_function[0].config["descriptor_types"][descriptor]
            )
        else:
            return "undefined"
```
I totally agree about getting rid of the module and class name in the config, it was bugging me too. This gets hit a lot when performing data imports so it might be worth keeping the number of DB calls down.

bferguso:
@chiatt - Thanks for letting me know about the new migration. 

I've made the changes as discussed and added the reverse migration. I incorporated the call to `get_class_module()` and removed the redundant config sections, but left the single DB call using the FunctionsXGraphs class. If you'd prefer I split it into two DB calls I'll do that, but thought the single call would be more efficient.

chiatt (approved):

chiatt:
Thanks very much @bferguso for your contribution. This PR looks great, and I'm sure this is going to be a very helpful feature.

PR #8261, opened 2022-03-16
webpack search and graph designer functionality

aarongundel:
### Types of changes

### Description of Change

Gets search page (mostly) working and starts getting the graph designer working (more to be done there)

### Issues Solved

#

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/frontend/webpack/webpack.common.js
+++ b/arches/app/frontend/webpack/webpack.common.js
@@ -23,15 +23,31 @@ const buildJavascriptFilepathLookup = function(path, outerAcc) {
     }, outerAcc);
 };
 
-const entryPointConfiguration = buildJavascriptFilepathLookup('../media/js', {});
+const buildTemplateFilePathLookup = function(path, outerAcc) {
+    return fs.readdirSync(path).reduce((acc, name) => {
+        if (fs.lstatSync(path + '/' + name).isDirectory() ) {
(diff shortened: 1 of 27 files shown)


arches/app/frontend/webpack/webpack.common.js
```diff
@@ -209,8 +221,8 @@ module.exports = {
                 type: 'javascript/auto',
             },
             {
-                test: /\.htm$/i,
-                loader: "html-loader",
+                test: /\.html?$/i,
+                use: ["html-loader"],
```
chrabyrd, line 225: what's up with this?

chrabyrd (approved):
overall looks good just 1 question

arches/app/frontend/webpack/webpack.common.js
aarongundel: It was just a test of html loader functionality; but it allows for .htm or .html extensions to be used.  There is a .html extension - registration/base.html, so this should probably stay - or the file extension should be changed.  I am also not sure how this will play out with arches subprojects in the future, so the lookback may be needed.

chrabyrd: great! ty

PR #8295, opened 2022-03-23
Handle tile data in workflows

chrabyrd:
### Types of changes

### Issues Solved

#8293

### DIFF
--- a/arches/app/views/api.py
+++ b/arches/app/views/api.py
@@ -1482,7 +1482,6 @@ class NodeGroup(APIBase):
         return JSONResponse(nodegroup, status=200)
 
 
-
 @method_decorator(csrf_exempt, name="dispatch")
 class Node(APIBase):
     def get(self, request, nodeid=None):
--- a/arches/app/views/api.py
+++ b/arches/app/views/api.py
@@ -1458,6 +1458,32 @@ class Tile(APIBase):
 
 
 @method_decorator(csrf_exempt, name="dispatch")
+class NodeGroup(APIBase):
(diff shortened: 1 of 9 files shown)


robgaston (approved):
These changes look good to me with the understanding that this is an incremental step towards improving workflows that doesn't break backwards compatibility.

robgaston:
@chiatt could you please review these changes as well since you know the domain a bit better here?

arches/app/views/api.py
```diff
@@ -1457,6 +1457,31 @@ def post(self, request, tileid):
...
+        except ValueError as e:
+            del params["nodegroupid"]
+
+        try:
+            nodegroup = models.NodeGroup.objects.get(pk=params["nodegroupid"])
+            permitted_nodegroups = [nodegroup.pk for nodegroup in get_nodegroups_by_perm(user, perms)]
+        except Exception as e:
+            return JSONResponse(str(e), status=404)
+
+        if not nodegroup or not nodegroup.pk in permitted_nodegroups:
```
chiatt, line 1479: Isn't the result the same if it's written like this? I find the syntax clearer.
```not nodegroup or nodegroup.pk not in permitted_nodegroups```

chiatt (changes requested):
This looks great. There are just a few minor Stickler errors. Could you resolve them real quick (unless there is a reason not to that I'm overlooking)?

arches/app/views/api.py
chrabyrd: 👍

chiatt (approved):

Issue #8294, opened 2022-03-23
Persist legacy id transferring data from staging to resource instance table


chiatt:
When we move data from staging to the resource instance table, the legacyid is replaced with the resourceid. Instead the legacyid should be saved as it is defined in the staging table.

chiatt:
Resolved in #8296

PR #8296, opened 2022-03-23
Fix legacyid from staging table, re #8294

chiatt:
Modifies __arches_staging_to_tile(load_id uuid) so that legacyids are preserved from the staging table to the resource_instances table.


njkim (approved):
Works great.

PR #8302, opened 2022-03-24
Fix bug in relate resources map popup, re archesproject/arches_her#800

chiatt:
Fixes a bug in the relate resources map card preventing the creation of resource relationships using the map popup.


njkim:
Works great

Issue #8300, opened 2021-09-30
No vertical scroll bar on Map Layer Manager page


rlwilliamss:
**Describe the issue** - On the Map Layer Manager page, users are unable to reach the icon fields or models that are not visible as there is no horizontal scroll bar - users using smaller displays are unable to reach the layer icon selection fields.

On larger displays, you can scroll within the icon selection field to search for desired icons, but a horizontal scroll bar is needed so that users can scroll to view the rest of the models. The screenshot shows the page display on a smaller (laptop) screen.

Steps to reproduce
1) Open the Map Layer Manager on a laptop display

2) There is no vertical scroll bar - models not displayed on the left cannot be reached, and icon selection fields cannot be reached

![image](https://user-images.githubusercontent.com/78143228/135480034-c3a016ed-fdc6-4a6b-83a9-096e425b802d.png)

Arches version 5.2

chiatt:
Fixed in #8308

PR #8308, opened 2022-03-29
Fix map layer manager scrolling, re #8300

chiatt:
Allows map layer manager layer list and main panel to scroll vertically and removes some css cruft. re #8300.


njkim:
works great

Issue #8311, opened 2022-03-30
Prevent error when icon class is missing


chiatt:
![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/6d7f650c-f569-45b7-85e3-ca71d8a0464a)

chiatt:
Fixed in #8311

PR #8312, opened 2022-03-30
Prevent error if iconclass not available, re #8311

chiatt:
Prevent error if iconclass not available, re #8311


njkim:
Fixes the issue, works great.

Issue #5307, opened 2019-09-25
Improve UX for overlapping geometries on the search map


adamlodge:
If two or more resource instance geometries overlap at a given location, only one of the resource's attributes are identifiable on click at that overlapping location.  Typically GIS interfaces support some way to acknowledge that more than one feature exists at that location and let the use pick which one is of interest.  Would be nice if Arches does the same on the search map.

chiatt:
Resolved in #8038

Issue #4983, opened 2019-07-08
New map component: allow for custom overlay popups


robgaston:
The new map component should support custom overlay popups (either with html content or a url in the feature properties) as we do with the existing map widget.

bferguso:
@chiatt  - Has anyone taken this on? I've taken a cut at this for one of the BC Gov requirements. If nobody has taken it on I'll create a PR and submit what I have, otherwise I'll stand down if there is already a proposed/desired solution.

chiatt:
@bferguso - Nobody has started on this. It would be great to see what you've got.

PR #8330, opened 2022-04-05
4983 - First cut of custom map popup overlays

bferguso:
### Types of changes

### Description of Change
Provide ability to show custom text in the Map Popup overlays. Provides the ability to:
1. implement custom isFeatureClickable() function
2. provide custom popup template
3. format arbitrary map features into the required `displayname` and `map_popup` attributes shown on in the overlay

### Issues Solved

#4983 

### Checklist

#### Ticket Background
*   Sponsored by: Self Funded
*   Found by: @robgaston 
*   Tested by: @ 
*   Designed by: @bferguso 

### Further comments
Originally planned to implement this change by letting the templates do the heavy lifting, however once integrating #5307, multi-feature-popup, ended up following the primarydescriptors pattern and made the implementation more data-centric, allowing the source feature properties to be translated into `displayname` and `map_popup` observables consumed by the default template.


bferguso:
I'm wondering if the default popup UI should be extended as part of this to be able to display > 4 lines for the map_popup primarydescriptor text. With the ability to customize the popup text for arbitrary features, it is likely that the map_popup can be longer than what is currently allowed for.

chiatt (changes requested):
@bferguso I discussed this with @robgaston and we both agree this is a really nice change. However, we are trying to limit the introduction of new settings when reasonably possible. @robgaston noticed that you could remove the MAP_POPUP_CONFIG_PROVIDER setting by overriding map-popup-provider.js and/or map-popup.htm in your project. As long as the relative path in core matches the path in arches, the project file will be used instead of the core file. I think this would simplify things quite a bit and still be pretty easy for users to implement.

chiatt:
I can see how it might be nice to show more than four lines in the default popup, but a height limit might still be useful to allow a vertical scrollbar to appear if a description happened to be really long. What do you think @dwuthrich

bferguso:
@chiatt - Not introducing more config makes sense. I had originally done that so the overriding javascript could reference the base map-popup-provider.js but the default implementation is very simple so I totally agree. I'll remove all the config parts and resubmit.

chiatt (approved):

chiatt:
Thanks very much @bferguso, this looks great. I spoke with @dwuthrich and @robgaston regarding the description height and the consensus was just to leave the default as is and let project implementer's customize it if they need to.

bferguso:
@chiatt - That makes sense, thanks for letting me know.

Issue #8171, opened 2022-02-21
Provide a way for functions to know if a resource copy operation is happening to help logic


aj-he:
**User Story**

Some function logic may need to behave differently when tile saves occur as part of a resource copy operation. There needs to be a way for functions to know if a resource copy operation is in progress.

A simple suggestion is to use a `is_copy_operation` kwarg that is part of the `resource.save` call (defaulting False). This can be passed through to the `tile.save` and followup save methods that call the functions. 

**Acceptance Criteria**
-   [ ] The function has method of checking for a copy operation
-   [ ] The default state is that it is not a copy operation so as to not affect existing functionality.
-   [ ] Unit Tests passed
-   [ ] Documentation

**Definition of Done**

**Ticket Background**
*   Sponsored by: 
*   Suggested by: @aj-he 
*   Designed by: @aj-he

chiatt:
re #6862

chiatt:
Resolved in #8339

PR #8339, opened 2022-04-06
Pass copy context to functions, re #8171

chiatt:
### Types of changes

### Description of Change
Passes copy context to functions on save so that functions can behave accordingly

### Issues Solved
 re #8171

### DIFF
--- a/arches/app/functions/primary_descriptors.py
+++ b/arches/app/functions/primary_descriptors.py
@@ -1,8 +1,11 @@
+import logging
 import uuid
 from arches.app.functions.base import BaseFunction
 from arches.app.models import models
 from arches.app.datatypes.datatypes import DataTypeFactory
 from django.utils.translation import ugettext as _
+
+logger = logging.getLogger(__name__)
 
 
 class AbstractPrimaryDescriptorsFunction(BaseFunction):
@@ -47,8 +50,8 @@ class PrimaryDescriptorsFunction(AbstractPrimaryDescriptorsFunction):
                             if value is None:
                                 value = ""
                             config["string_template"] = config["string_template"].replace("<%s>" % node.name, str(value))
-        except ValueError as e:
-            print(e, "invalid nodegroupid participating in descriptor function.")
+        except ValueError:
+            logger.error(_("Invalid nodegroupid, {0}, participating in descriptor function.").format(config["nodegroup_id"]))
         if config["string_template"].strip() == "":
             config["string_template"] = _("Undefined")
         return config["string_template"]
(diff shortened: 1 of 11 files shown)


njkim (approved):
Passing context parameter works fine and copying resources also works great.

arches/app/models/resource.py
```diff
@@ -228,7 +226,7 @@ def bulk_save(resources, transaction_id=None):
         se.bulk_index(documents)
         se.bulk_index(term_list)
 
-    def index(self):
+    def index(self, context=None):
```
apeters, line 229: It would be nice to add a docstring to methods that now take in a context parameter describing what can be passed in and maybe what it's used for.
EG:
```
Keyword Arguments:
context -- defaults to False, set to True to return the raw object saved in the db 
```

arches/app/views/resource.py
```diff
@@ -546,8 +542,6 @@ def get(self, request, resourceid=None, view_template="views/resource/edit-log.h
 
             return render(request, view_template, context)
 
-        return HttpResponseNotFound()
```
apeters, line 549: Shouldn't we keep this as a failsafe return?

arches/app/functions/primary_descriptors.py
```diff
@@ -1,18 +1,20 @@
...
 
 class AbstractPrimaryDescriptorsFunction(BaseFunction):
-    def get_primary_descriptor_from_nodes(self, resource, config):
+    def get_primary_descriptor_from_nodes(self, resource, config, context=None):
         pass
 
 
 class PrimaryDescriptorsFunction(AbstractPrimaryDescriptorsFunction):
-    def get_primary_descriptor_from_nodes(self, resource, config):
+    def get_primary_descriptor_from_nodes(self, resource, config, context=None):
```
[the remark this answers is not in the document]
chiatt: That's right, but I think we need to keep it in the function definition because it is now possible to override this function with a custom primary descriptor function. When the `get_primary_descriptor_from_nodes` function gets called in resource.py and api.py we pass in the context. This allows for the possibility that someone may want to pass in the context into their own custom function.

arches/app/models/resource.py
chiatt: I've added some docstrings for `context`

apeters (approved):
This works as expected.  Now that we have this functionality, it might be nice to use it when making copies of resources, by appending `_copy` to then end of the copied resource display name.

chiatt:
I think that's a good idea @apeters. I've created a ticket for it:
re #[Append '_copy' to a resource model's display name when it is copied.](https://github.com/archesproject/arches/issues/8399)

Issue #8280, opened 2022-03-18
Add branch csv etl module


chiatt:
Add a new etl module that loads the branch csv format.

chiatt:
Resolved in #8362

PR #8362, opened 2022-04-15
Update branch csv importer to import files, #8280 

njkim:
Update branch csv importer to import files, #8280


chiatt (approved):
Looks great. The only issue I found is when a file is missing, that error message isn't passed to the load status, but I've created separate ticket for that #8380.

Issue #8364, opened 2022-04-22
Popup displayed when map in drawing mode (search or edit)


bferguso:
**Describe the bug**
Map popup is displayed when drawing features (either filter geometries or resource geometries) when a point lays on an existing geometry.

**To Reproduce**
1. On the map search page zoom in to a resource with a geometry displayed on the map
2. Click on the "Filter" tab on the right
3. Select "Draw a polygon" in the "Select a filter" pick list (any drawing mode will do)
5. Click inside the displayed geometry
6. Note that the popup displays even though in a drawing mode

**Screenshots**
<img width="1920" alt="popup_while_drawing" src="https://user-images.githubusercontent.com/6750642/164748249-3b8057aa-ad2c-42a3-856a-c563066f0994.png">

**Expected behavior**
Map popups not be displayed when in a drawing mode (either in search or when editing/creating a resource geometry)

**Your Arches Information**

*   Version used: master branch
*   Operating System and version (desktop or mobile): OSX
*   Browser Name and version: Firefox Developer 100.0b7
*   Link to your Arches Install (optional):

**Additional context**
I'll be submitting a PR with a proposed fix shortly

**Ticket Background**
*   Found by: @bferguso

chiatt:
Resolved in #8365

PR #8365, opened 2022-04-22
8364 - Do not display map popup when in draw mode

bferguso:
### Types of changes

### Description of Change
Modify the map popup provider so that map popups are not displayed when in a draw mode. Needed to align the name of the `drawMode` ko observable in `viewmodels/map-filter.js` and `views/components/search/map-filter.js` with the similar observable called `selectedTool` in the `viewmodels/map-editor.js` so that the tool context can be passed into the `utils/map-popup-provider.js` in both contexts.

### Issues Solved
#8364 Popup displayed when map in drawing mode

### Checklist

#### Ticket Background
*   Sponsored by: Self Funded
*   Found by: @bferguso
*   Tested by: @bferguso
*   Designed by: @bferguso

### Further comments
n/a

### DIFF
--- a/arches/app/media/js/utils/map-popup-provider.js
+++ b/arches/app/media/js/utils/map-popup-provider.js
@@ -10,8 +10,10 @@ define(['arches',
          * @param feature Map feature to check
          * @returns <code>true</code> if the feature can be clicked, otherwise <code>false</code>
          */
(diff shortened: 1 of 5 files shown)


chiatt (approved):
Thanks @bferguso, this looks great.

Issue #7371, opened 2021-04-29
Add function to track when resources are edited


apeters:
The main use case for this function is to be able to track alternate identifiers for resource instances in a dedicated lookup table.  This is to help facilitate data import for large (or small) datasets where only alternate ids are know ahead of time.  This is a first step towards multiple identifiers for an instance.

This should be a function that calls out to another function that can take the data and persist it wherever the developer wants. 
- Should have the UUID of the resource (and the tile?) and the value
- Should deal with (at least) strings and the URIs from the url datatype
- Doesn’t need to /do/ anything with it – that can be on users for now, but gives opportunity for experimenting without making a hard decision without evidence or design

chiatt:
@apeters Do you think this will change how we manage the legacyid?

apeters:
This is really meant to be open-ended.   It's simply a proxy to another module.

chiatt:
@apeters Is this feature still in progress?

PR #8385, opened 2022-04-28
7371 function to track resource edits

apeters:
re #7371


chiatt:
With the introduction of the bulk data manager (aka etl manager) and import via relational views, there are now ways to bypass this function (which was written before those two features were developed). At this point it seems like safer solution that would capture all edits would be to use an SQL approach rather than an arches function.

Issue #8389, opened 2022-04-29
Allow card to override default map-popup


chiatt:
The map-popup-provider uses the default `/components/map-popup.htm` template. Project developers can override this file or the map-popup template, but cannot have different popups for different cards. This presents a problem for the related-resource-map card which uses a different popup than the default.

The related-resource-map card (or any custom project card) needs to be able to use its own popup.

robgaston:
fixed in #8390

PR #8390, opened 2022-04-29
Allow card to override default popup, re #8389

chiatt:
### Types of changes

### Description of Change
Allows a card to define its own primary descriptor template. If a card has this.popupTemplate object defined the map will use it instead of the popup defined in the map-popup-provider. 

Example of Related Resource Map card using its own popup:
https://github.com/archesproject/arches/blob/83c785d51c16071ba212eea4899db3c189a9f172/arches/app/media/js/views/components/cards/related-resources-map.js#L11-L12
https://github.com/archesproject/arches/blob/83c785d51c16071ba212eea4899db3c189a9f172/arches/app/media/js/views/components/cards/related-resources-map.js#L281

### Issues Solved
#8389

### DIFF
--- a/arches/app/media/js/viewmodels/map-editor.js
+++ b/arches/app/media/js/viewmodels/map-editor.js
@@ -11,8 +11,7 @@ define([
     'proj4',
     'views/components/map',
     'views/components/cards/select-feature-layers',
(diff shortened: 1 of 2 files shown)


robgaston (approved):
this looks good to me - thanks!

PR #8411, opened 2022-05-11
fixes for asynchronous iiiflayer vs synchronous side by side layer

aarongundel:
### Types of changes

### Description of Change

The iiif layer adds its layers to the map asynchronously.  The side by side layer is expecting its layers to be _fully rendered_ before it actually uses them in comparison.  As a result, this PR attempts to leverage the "load" event of a leaflet layer (in which all tiles are fully rendered) in order to do the leaflet side by side operations.  Additionally, the zoom to bounds function is modified and only executed on manifest change.  

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/views/components/iiif-viewer.js
+++ b/arches/app/media/js/views/components/iiif-viewer.js
@@ -525,15 +525,24 @@ define([
             self.greyscale(false);
         };
 
+        const zoomToBounds = (map, layer) => {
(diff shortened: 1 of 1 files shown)


chiatt (approved):
Tested it pretty thoroughly with some large files and it seems really stable.

Issue #8415, opened 2022-05-11
Displayname not available in related resources


chiatt:
When related resources are requested, the displayname of the related instances is not available causing errors in the browser. To reproduce error, click on the related resources link in search on an instance with relationships. You should see the following error:

![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/e86b2f92-e654-4ca1-8c62-c680368ba6dc)

chiatt:
Fixed in #8416

PR #8416, opened 2022-05-11
Adds missing displayname to related resources request, re #8415

chiatt:
Ensures displayname is added to the resource as a string rather than a method.


njkim (approved):
Works great

Issue #8424, opened 2022-05-12
Hard coded clear-user-permission-cache URL invalid if context root isn't /


bferguso:
**Describe the bug**
The hard coded `/clear-user-permission-cache/` URL isn't correct if the root URL isn't /. Our implementation runs behind a reverse proxy which requires a prefix to the context root (eg arches.urls.root = '/int/arches-fossils/') so the hard coded URL generates a 404.

**To Reproduce**
Steps to reproduce the behavior (admin):
1. Arches install must have a context root other than '/'
2. Navigate to `<arches context root>/admin/auth/user/7/change/`
3. Press the Clear User Permission cache button
4. Note the 404 in the console and the modal window doesn't disappear

Steps to reproduce the behavior (graph designer):
1. Arches install must have a context root other than '/'
2. Edit a graph in the Arches Designer -> Resource Models
3. Select the Permissions tab above the tree
4. Select a node in the tree
5. Select a group in the "Set permissions for:" select
6. Click the Apply Permissions button
7. Note the 404 in the console and the modal window doesn't disappear

**Expected behavior**
The URLs should be relative to the context root, not assume there is no

**Your Arches Information**

*   Version used: dev/6.1.x
*   Operating System and version (desktop or mobile): OSX
*   Browser Name and version: Firefox Developer
*   Link to your Arches Install (optional):

**Additional context**
The hard coded URLs should be updated in the `app/templates/admin/change_form.html` and the `app/media/js/views/graph/permission-manager/permission-settings-form.js` files.

**Ticket Background**
*   Found by: @bferguso

apeters:
closed by #8424

PR #8425, opened 2022-05-12
8424 hard coded permissions url

bferguso:
### Types of changes

### Description of Change

Add context root to clear-user-permission-cache URLs

### Issues Solved

#8424

### Checklist

#### Ticket Background
*   Sponsored by: Self Funded
*   Found by: @bferguso
*   Tested by: @bferguso
*   Designed by: @bferguso

### DIFF
--- a/arches/app/media/js/views/graph/permission-manager/permission-settings-form.js
+++ b/arches/app/media/js/views/graph/permission-manager/permission-settings-form.js
@@ -141,7 +141,7 @@ define([
         clearUserPermissionCache: function() {
             return $.ajax({
                 type: 'POST',
-                url: '/clear-user-permission-cache/',
+                url: arches.urls.root + 'clear-user-permission-cache/',
             });
         }
     });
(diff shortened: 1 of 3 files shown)


apeters (changes requested):
@bferguso, I think the way to do this properly is to use django's `reverse` url functionality (see [https://docs.djangoproject.com/en/4.0/ref/urlresolvers/#reverse](https://docs.djangoproject.com/en/4.0/ref/urlresolvers/#reverse)).
You can use that directly from within the .html template like so
```request.open("POST", "{% url 'clear_user_permission_cache' %}", true);```

To use that in the .js file you'll need to create a new entry in `javascript.htm` under the `arches.urls` section (lines ~170-280).  See the entries there for examples of how to do that.  Then use that new value directly in `persmission-settings-form.js` like you used `arches.urls.root`.

bferguso:
> @bferguso, I think the way to do this properly is to use django's `reverse` url functionality (see https://docs.djangoproject.com/en/4.0/ref/urlresolvers/#reverse). You can use that directly from within the .html template like so `request.open("POST", "{% url 'clear_user_permission_cache' %}", true);`
> 
> To use that in the .js file you'll need to create a new entry in `javascript.htm` under the `arches.urls` section (lines ~170-280). See the entries there for examples of how to do that. Then use that new value directly in `persmission-settings-form.js` like you used `arches.urls.root`.

@apeters  - That's a much better solution - thanks for pointing me in the right direction.

apeters (approved):
works as expected

Issue #8431, opened 2022-05-14
[ETL manager] Create resource_x_resource record for resource-instance datatype


njkim:
A resource_x_resource record should be created during etl transformation when tile-data moved from the staging table to the tile table for the resource instance and resource instance list datatype.

chiatt:
Resolved in #8432

PR #8432, opened 2022-05-14
Create resource x record during ETL, #8431

njkim:
### Types of changes

### Description of Change

Create resource_x_resource record for resource instance datatype during ETL using db functions, re #8431

### Issues Solved

#8431

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chiatt (commented):
I'm wondering if this shouldn't be moved into its own function to reduce the length of this function and possibly make this new logic available to the relational schema. We should check in with @robgaston on this.

robgaston:
worth noting we've got a database function to refresh these records on a per tile basis called `__arches_refresh_tile_resource_relationships`:
https://github.com/archesproject/arches/blob/ef95ef3b08204b33d2ee6aff457f38c2d295a89b/arches/app/models/migrations/7787_relational_data_model.py#L654-L699

chiatt:
Good to know - thanks @robgaston

chiatt (approved):
Looks good to me!

Issue #6959, opened 2020-12-17
Improve labels under Manage dropdown in Arches Designer Front page


adamlodge:
IMHO: Export, Clone, and Delete should say: "Export Model", "Clone Model", "Delete Model" respectively.

This is in order to mitigate the possibility of confusing exporting, cloning, or deleting instances versus the model.

![Screen Shot 2020-12-17 at 11.00.02 AM.png](https://images.zenhubusercontent.com/56f150f76caf56c9d38e16f8/73e94fb0-90b4-4cb0-bd4f-291f7c17aace)

chiatt:
Fixed in #8446

PR #8446, opened 2022-05-18
Make button label for graph operations more explicit, re #6959

chiatt:
Makes button label for graph operations more explicit, re #6959


njkim:
Looks Good!

njkim (approved):
Looks Good!

Issue #8435, opened 2022-05-16
In branch csv importer notify user of file to be loaded before executing load


chiatt:
Users should see metadata information about the file(s) they are loading before executing the etl process.

chiatt:
Resolved in #8453

PR #8453, opened 2022-05-19
Allow user review files before confirming load #8435 

chiatt:
### Description of Change
Allows a user to review the files they are about to load before starting etl process. 

### Issues Solved

#8435


njkim (approved):
Works great

Issue #8451, opened 2022-05-19
url datatype widget erroring when setting url label


aj-he:
**Describe the bug**

URL datatype widget expecting ko observable for the `url_label` property but seems to standard var.

**To Reproduce**
1. populate a url node with just the url and save
2. edit the node by adding a label.
3. save button fails to appear.
4. check dev tools

**Screenshots**

![image](https://user-images.githubusercontent.com/25295526/169305267-6b9ce983-1df6-48a1-a82f-e9330d52ef9d.png)

**Expected behavior**
URL label should set dirty state of card and allow you to save. 

**Your Arches Information**

*   Version used: dev.6.1.x
*   Operating System and version (desktop or mobile): Windows/Docker WSL2
*   Browser Name and version: Chrome
*   Link to your Arches Install (optional):

**Additional context**

Testing on Arches-HER - example above is on the Heritage Area model

**Ticket Background**
*   Found by: @aj-he

PR #8458, opened 2022-05-22
Ensure url and url_label are always props on url value, re #8451

chiatt:
### Types of changes

### Description of Change
Ensures that both the url and url_label are saved to the url datatype value because if one of the properties is missing on the value, the widget will not work properly.

### Issues Solved

#8451


aj-he (changes requested):
I had a couple of issues.

1. I had a card in my test db that only had the url value. Changing to the branch seemed to cause the config fields to show and the value fields not to populate. I also could not edit the value...
image
image

2. I added a new card and the values worked well. However, when I cleared the "url label" and "url" values to return my card to the original state, I got the following error...
image

chiatt:
Thanks for looking at this @aj-he. I'll have to try a different approach.

jacobtylerwalls:
New effort in #10642

Issue #7316, opened 2021-04-16
v5.1 Manage Permissions - search dropdown not responding to user permission


Jude-Dicken:
Hello - Our Arches v5.1 at https://isleofmanher.im/ is not matching searches matches  in the dropdown (e.g.: your "predictive list") to the user permission level.  Example:

Ballanorris Keeill is set for 'No Access' for anonymous user.  Okay.
Anonymous user searches 'Ballanorris' and the resource 'Ballanorris Keeill' appears in predicative list.  User therefore assumes that the resource is available to click and view.
Anonymous user clicks 'Ballanorris Keeill' from the predictive list and (disappointingly, confusingly for the user) the result says '0'.

Chrome is browser used for testing.
Arches v5.1
Operating System - please query this with our installer and support provider Reuben Osborne (Knowledge Integration): reuben.osborne@k-int.co.uk

Thanks,

Jude Dicken (Miss)
Collections Information Manager
Manx National Heritage (Isle of Man), jude.dicken@mnh.im

chiatt:
Just to add-on to the description of this issue... this is due to terms getting indexed that belong to restricted instances. The terms are not filtered by user instance permissions so they appear as search suggestions in the term search input.

chiatt:
This issue also applies to cases when all of the nodegroups of a resource model are restricted as described here: https://github.com/archesproject/arches/issues/2491
A solution to this issue should apply to nodegroup permissions as well as resource instance permissions.

Issue #2491, opened 2017-10-03
Restricting All Nodes On A Resource Model Does Not Fully Restrict Access To It


mradamcox:
### User Story
Currently Arches provides the ability to implement permissions on a node-by-node basis, but there is no ability to fully restrict a given resource model. This should happen if No Access has been defined on every node in the Resource Model, triggering a "completely restricted" state.

#### Expected Behavior
If a resource model is fully restricted from a user/user group, said user/user group should see no trace of any resources  
1. that resource model would not show up in the filter dropdown in the Search Page (Archaeological Site, in the example below).
![image](https://user-images.githubusercontent.com/10427268/30979936-d9235630-a444-11e7-9a6c-320b8b78a983.png)
2. the url that would result from clicking Archaeological Site in the dropdown shown above must be protected, or return no results (right now it does in fact return all archaeological sites, though as the nodes themselves are restricted, no information shows up in the resource report besides the display name)
3. typing the name of a fully restricted resource into the search bar should not generate suggestions that include that name
4. the initial number of search results should not include these restricted resources. Admittedly, no really sensitive information would be revealed, but I can't actually see any good, universal, argument for why the search results should include a count of restricted resources.

#### Steps to Reproduce (for bugs)
Create a new resource model, with at least one or two nodes, set permissions to No Access for the Guest user group, create one resource with this resource model, log out, go to search, note all the behaviors described above.

#### Your Arches Information
* Version used: 4.0.0

#### Context
This level of restriction is necessary for many organizations whose database will include sensitive records, such as archaeological data.

mradamcox:
I have accomplished some of these tasks for my current project with this commit: https://github.com/legiongis/arches/commit/df0d7723cdc21c928b6ad3d3e79cf92f1b9ac9cb. The `get_doc_type` function is altered to check the user's group assignments against a new setting in settings.py. This is not a 100% comprehensive fix for all the issues above, as described below:

1. (not solved) the resource type still shows up in the dropdown (though that should be real easy to fix with this new settings variable).
2. (solved) though not exactly as described: the url is still available to the user, but 0 results will be returned if they are trying to get a resource type to which they don't have access.
3. (not solved) no effort is made to cull dropdown suggestions
4. (solved) the search results number correctly reflects the accessible resources

see also #2492

bferguso:
We have a similar requirement for our business area as we would like to use the system for both sensitive archeological information as well as public heritage data.

chiatt:
[...]

PR #8466, opened 2022-05-24
Filter instances when all nodegroups are restricted, re #8438 

### LINKED ISSUE #8438: Hidden resources displaying in term search and results
**Describe the bug** - Resources that are hidden from certain user groups are displaying in the term search and search results.

**Expected result** - The term search and Resource Type drop down should only show results for a model only if there are are least 1 card with permissions... otherwise hide it from results and the Resource Type dropdown.

**To Reproduce**

1. As an admin user, go to the graph designer for the Consultation model and set permissions for anonymous users to 'No Access' across all Consultation model cards

1. Log out and go to the Search page as an anonymous user

2. Click on the Resource Type drop down and select Consultation (in this example Consultation and Application Area resources are hidden from anonymous users) - the Consultation results return. When clicked on a 403 page is returned so anonymous users cannot access the record, but they should not be able to see them in the Resource Type drop down or search results.

**Screenshots**

anonuser

**Your Arches Information**
*   Version used: 6.0.1
*   Operating System and version (desktop or mobile): Windows 10
*   Browser Name and version: Chrome 101.0.4951.67
*   Link to your Arches Install (optional):

**Ticket Background**
*   Found by: @StucakeGLHER

chiatt:
### Description of Change
Filters out resource instances from search results of resource models for which all nodegroups are restricted.

### Issues Solved
#8438
This also solves most of #2491 with the exception of concealing displaynames/terms from restricted instances in the search suggestions. However, applying permissions to terms should probably be treated in a different PR because it involves resource instance permissions as well as nodegroup permissions, so we can close #8438 and #2491 when this PR is approved, and address term permissions under #7316

### DIFF
--- a/arches/app/media/js/views/components/search/resource-type-filter.js
+++ b/arches/app/media/js/views/components/search/resource-type-filter.js
@@ -1,15 +1,30 @@
 define([
     'knockout',
+    'arches',
     'views/components/search/base-filter'
-], function(ko, BaseFilter) {
+], function(ko, arches, BaseFilter) {
     var componentName = 'resource-type-filter';
     return ko.components.register(componentName, {
         viewModel: BaseFilter.extend({
-            initialize: function(options) {
+            initialize: async function(options) {
                 options.name = 'Resource Type Filter';
                 this.requiredFilters = ['term-filter'];
                 BaseFilter.prototype.initialize.call(this, options);
+                this.resourceModels = ko.observableArray();
+                const self = this;
+
+                const getData = async function() {
+                    const response = await fetch(arches.urls.api_search_component_data + componentName);
+                    if (response.ok) {
+                        const data = await response.json();
+                        self.resourceModels(data.resources);
+                    } else {
+                        // eslint-disable-next-line no-console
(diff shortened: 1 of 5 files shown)


apeters (changes requested):
This update works when a user has selected a Resource Type to filter by or when that filter is negated.  
This doesn't work when the Resource Type filter isn't applied at all.  
You can observe this by making all the nodegroups of a model inaccessible to the anonymous user.  Then viewing the default search page as that user (with no filters selected), you will notice that the user can still see results of that type.

In the screen shot below, the "Person" resource type is inaccessible to the anonymous user, but it is still visible in the results.

Screen Shot 2022-05-24 at 4 59 11 PM

What this implies is that if any model is made inaccessible, then the Resource Type filter has to be employed (either implicitly or explicitly).  Unfortunately, this won't be a trivial thing to do.

chiatt:
Hey @apeters, the resource type filter should now be applied in all cases with these commits:
1a02ae0
fee47ba

apeters (approved):
this works as expected

Issue #8490, opened 2022-05-28
[ETL manager] Option to select the csv column as an id


njkim:
The single csv importer should be able to handle the resourceid or legacyid.
In order to do that, there should be an option to select the id column along with node aliases.

chiatt:
Resolved in #8491

PR #8491, opened 2022-05-28
Add option to select column as an id for single csv import, #8490 

njkim:
### Types of changes

### Description of Change

Add the option to select column as an id for single csv import, #8490 
If a uuid is selected, id will become resource instance id, if not a uuid, legacy id.

### Issues Solved

#8490

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/etl_modules/import_single_csv.py
```diff
@@ -97,14 +98,19 @@ def write(self, request):
 
         fieldnames = request.POST.get("fieldnames").split(",")
         column_names = [fieldname for fieldname in fieldnames if fieldname != ""]
+        id_label = "Use as an id"
```
chiatt, line 101: I don't think this is true anymore right? We are just using the id_label down on line 164. Maybe the id_label assignment on 164 should be moved up here?

chiatt (changes requested):
This looks good. Just one small change that probably needs to be made.

Issue #8457, opened 2022-05-20
Temp files should save to default_storage rather than disk


chiatt:
Currently temporary files for the branch csv importer are saved to a temporary location on disk. Instead, they should save to default_storage in case an instance is configured to manage files in the cloud rather than on the server.

chiatt:
Resolved in #8511

PR #8508, opened 2022-06-03
Add default_storage to etl_manager, modify datatypes accordingly, re …

ryan86:
…#8457

Add default_storage to etl manager and modify datatypes accordingly.


chiatt (approved):
Looks good to me. Package loader and file widget still work as expected.

Issue #8495, opened 2022-06-01
Prevent unecessary calls to graph endpoint in branch csv loader


chiatt:
The graph endpoint does not need to be called when the component is displaying its status.

chiatt:
Resolved in #8514

PR #8514, opened 2022-06-03
Call graph endpoint only when graph list needed, re #8495

chiatt:
### Types of changes

### Description of Change
Call graph endpoint only when a list of graphs is need to power the template dropdown.

### Issues Solved

#8495


njkim (approved):
works great.

Issue #8528, opened 2022-06-09
Provide functions to disable / enable tiles triggers for bulk data loads


bferguso:
Triggers exist to validate tile cardinality and refresh spatial views when performing a CRUD on the tiles table. When performing bulk loads with large datasets these triggers can cause significant performance issues. It would be good to have a way to disable the triggers when performing bulk loads to increase the load performance.

When the load has been completed, there should be the ability to re-enable the triggers, validate the cardinality of the tiles data and bulk refresh the spatial views.

**User Story**
When inserting data (esp via the schema views) it would be good to be able to disable the triggers and re-enable them after the load. It would also be helpful to be able to identify any tiles with cardinality violations.

Proposed pattern:
```
call __arches_prepare_bulk_load(); -- <- This disables triggers on the tiles table

/*
 * Perform bulk loads
*/

call __arches_complete_bulk_load(); -- <- This re-enables the triggers on the tiles table, refreshes the spatial views, validates cardinality of tiles tables. Raises exception if spatial views can't be refreshed or there are tile cardinality issues.

select * from __arches_get_tile_cardinality_violations(); -- <- This returns a table with resourceinstanceid, nodegroupid, parenttileid, tile_count
```

**Acceptance Criteria**
-   [ ] Functions & procedures completed and added to a migration
-   [ ] Function & procedure names approved
-   [ ] Unit Tests passed
-   [ ] Documentation

**Definition of Done**
Procedures/Functions exist to 
1. disable the tiles triggers
2. re-enable the triggers, validate the cardinality of the tiles data and refresh the spatial views
3. determine which resources & nodegroups violate the cardinality constraints

**Ticket Background**
*   Sponsored by: Self Funded
*   Suggested by: @bferguso
*   Designed by: @bferguso

chiatt:
Resolved in #8529

PR #8529, opened 2022-06-09
8528 - Functions to  disable/enable tiles triggers for bulk load performance

bferguso:
Adds functionality to disable and re-enable triggers on the tiles table to increase bulk load performance.

### Types of changes

### Description of Change
Adds 3 procedures/functions to:
1. Disable triggers on the tiles table
2. Re-enable triggers on the tiles table, refresh spatial views, check for tiles cardinality violations
3. Retrieve a list of tiles cardinality violations

### Issues Solved
#8528 

### Checklist

#### Ticket Background
*   Sponsored by: Self Funded
*   Found by: @bferguso
*   Tested by: @bferguso
*   Designed by: @bferguso

### DIFF
--- a/arches/app/models/migrations/8502_loadevent_indexed_time.py
+++ b/arches/app/models/migrations/8502_loadevent_indexed_time.py
@@ -6,7 +6,7 @@ from django.db import migrations, models
 class Migration(migrations.Migration):
 
     dependencies = [
-        ("models", "8009_etlmodule"),
+        ("models", "8449_filter_warning_etl_module"),
     ]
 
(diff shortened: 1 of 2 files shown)


chiatt:
@njkim We could probably use this to lift the cardinality trigger when moving tiles from load_staging to the tiles table. I think we might not really need the trigger because we validate cardinality in the staging table before load.

bferguso:
Associated issue has pattern and procedure/function names in it

chiatt (approved):
These changes look good to me.

Issue #8513, opened 2022-06-03
Remove `arches_get_resourceid_from_legacyid` postgres trigger


chiatt:
The `get_resourceid_from_legacyid` is helpful if a user is reloading data with the same legacyid with the intent to append data to existing resource instances. However, the following problem can occur:

A user loads a 'Person' resource instance with a legacyid of 'ID1'. Then they load a 'Place' resource instance with the same legacyid. When this happens the function associates the resourceid of the Place with Person already in the system and then saves Place tiles to the Person instance. This is really dangerous, so it makes sense to remove this function. Users can use legacyids for the creation of new resource instances. However, for appending additional tiles to existing resources, a resourceid will need to be used.

chiatt:
Fixed in #8532

PR #8532, opened 2022-06-09
Show legacy id error, re #8513

chiatt:
Informs user that load failed because legacy ids in excel file already exist in the database., re #8513


njkim (approved):
Looks great

Issue #8027, opened 2021-12-16
Improve error/warning logging during data import operations


robgaston:
For the existing data import management commands, Arches could provide much better logging throughout the process which would allow a user to better understand what issues might have occurred during a load operation.  This becomes even more important as we being to implement frontend interfaces that in turn use this same import code on the backend.

We should identify specific errors/warnings that should be emitted during this process and implement those throughout in a consistent way, including suggestions for how a user could resolve/respond to these issues.

> [Improve] Arches current data import/export code to improve performance and error logging

chiatt:
I think one thing that should probably come out of this is to stop collecting error messages in an array. The original intent of the array was to allow the errors to be summarized. However, if there are a lot of messages, that array can get really big, sucking up valuable memory and most likel
y adversely impacting performance.

chiatt:
re #6850

PR #8558, opened 2022-06-16
Convert print statements to django logger statements in csvfile.py, r…

ryan86:
Convert print statements to django logging statements in csvfile.py, re #8027


chiatt (changes requested):
@ryan86 Could you remove the commented out code?

arches/app/utils/data_management/resources/formats/csvfile.py
```diff
@@ -766,24 +747,21 @@ def transform_value(datatype, value, source, nodeid):
...
-                                {
-                                    "type": "ERROR",
-                                    "message": "datatype: {0} value: {1} {2} - {3}".format(
-                                        datatype_instance.datatype_model.classname,
-                                        value,
-                                        source,
-                                        str(e) + " or is not a prefLabel in the given collection.",
-                                    ),
-                                }
+                            logger.warn(
```
chiatt, line 750: I think changing this to a warning is good, but the message itself can be improved. When I get an error parsing a resource instance value, it reads like this:

```
datatype: ResourceInstanceListDataType value: ([{'resourceId': '4d1c5b38-2e05-486e-ab17-61cfcf9e93b1',
'ontologyProperty': 'http://www.cidoc-crm.org/cidoc-crm/P14_carried_out_by', 'resourceXresourceId':
'','inverseOntologyProperty': 'http://www.cidoc-crm.org/cidoc-crm/P14i_performed'}], [{'resourceId': '5c5d91c0-
f51e-47c7-8f43-632d05a30341', 'ontologyProperty': 'http://www.cidoc-crm.org/cidoc-crm/P14_carried_out_by',
 'resourceXresourceId': '', 'inverseOntologyProperty': 'http://www.cidoc-crm.org/cidoc-crm/P14i_performed'}]) on
 line 2 - tuple indices must be integers or slices, not str or is not a prefLabel in the given collection.

```
This is a bit unclear because mention of a collection and prefLabel can give the impression that the value was somehow being read as a concept value rather than a resource-instance value.

The error is caused because Arches cannot parse/read the value in the CSV file. For that reason it seems like the error message should be something like: `The following value could not be interpreted as a [datatype] value: [value]`, and the message`or is not a prefLabel in the given collection` should be moved to the `transform_value_for_tile` method of the concept datatype so that it only appears when the issue is concept related.

chiatt (changes requested):

chiatt (approved):

Issue #8551, opened 2022-06-16
[ETL] The user should be able to change how many and filter -user, date, type, status


njkim:
An admin should be able to decide how many jobs to show per page.
Also the filter box should be able to filter on user, date, job type, job status.

PR #8560, opened 2022-06-16
Update ETL Status Filter, #8551

njkim:
### Types of changes

### Description of Change

Add more term to filter including user, etl-module, status. #8551
Also, add status per page to settings 

### Issues Solved

#8551 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/views/etl_manager.py
+++ b/arches/app/views/etl_manager.py
@@ -54,18 +54,22 @@ class ETLManagerView(View):
                 "failed": "failed",
                 "running": "running",
                 "unloading": "reversing",
-                "unloaded": "unloaded"
+                "unloaded": "unloaded",
             }
-            status_string = status_lookup[filter_string] if filter_string in status_lookup else 'undefined'
+            status_string = status_lookup[filter_string] if filter_string in status_lookup else "undefined"
 
             if filter_string:
-                filtered_events = LoadEvent.objects.filter(
-                    Q(status__icontains=status_string) |
-                    Q(etl_module__name__icontains=filter_string) |
-                    Q(user__username__icontains=filter_string) | 
-                    Q(user__first_name__icontains=filter_string) | 
-                    Q(user__last_name__icontains=filter_string)
-                    ).order_by(("-load_start_time")).prefetch_related("user", "etl_module")
+                filtered_events = (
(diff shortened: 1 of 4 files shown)


arches/app/views/etl_manager.py
```diff
@@ -37,18 +40,41 @@ def get(self, request):
...
                 show = False if "show" in module.config.keys() and module.config["show"] is False else True
                 if self.request.user.has_perm("view_etlmodule", module) and show:
                     response.append(module)
         elif action == "loadEvent":
-            item_per_page = 5
-            all_events = LoadEvent.objects.all().order_by(("-load_start_time")).prefetch_related("user", "etl_module")
-            events = Paginator(all_events, item_per_page).page(page).object_list
-            total = len(all_events)
-            paginator, pages = get_paginator(request, all_events, total, page, item_per_page)
+            status_lookup = {
```
chiatt, line 51: I don't think this will work for a system that is not in English. The keys in here correspond to values that an admin could translate, and if that were the case, the filter would not work. The solution here is probably to use a dropdown in the UI.

arches/app/views/etl_manager.py
```diff
@@ -1,10 +1,13 @@
 import logging
-from django.db import connection
+from datetime import datetime
```
chiatt, line 2: This import appears to be unused.

chiatt (changes requested):
This looks good, but the status filtering approach will need to be changed to support internationalization.

arches/app/views/etl_manager.py
njkim: thanks!

arches/app/views/etl_manager.py
chiatt: Even if the system is in English, there could be other 'status' types that exist in custom etl modules. I think we would need to get all the status types listed in the load_event status column to build the status dropdown.

chiatt (changes requested):

chiatt:
Closing this now really needs to be in 7.5 rather than 6.x, and also because a dropdown solution rather than a string filter is needed to support internationalization in 7.5.

Issue #8550, opened 2022-06-16
[ETL] Add confirmation dialog to undo button


njkim:
Pressing the undo button starts the job immediately. 
We want to ask the user if they’re sure, and this should start an async job (if needed) so that the screen isn’t locked.

chiatt:
Fixed in #8579

PR #8579, opened 2022-06-18
Add confirmation dialog to undo etl manager load, re #8550

ryan86:
Add confirmation dialog to etl manager when undoing a load, re #8550

### DIFF
--- a/arches/app/media/js/views/components/plugins/etl-manager.js
+++ b/arches/app/media/js/views/components/plugins/etl-manager.js
@@ -3,7 +3,9 @@ define([
     'knockout',
     'js-cookie',
     'arches',
-], function($, ko, Cookies, arches) {
+    'viewmodels/alert',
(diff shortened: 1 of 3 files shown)


arches/app/media/js/views/components/plugins/etl-manager.js
```diff
@@ -3,7 +3,9 @@ define([
     'knockout',
     'js-cookie',
     'arches',
-], function($, ko, Cookies, arches) {
+    'viewmodels/alert',
+    'viewmodels/alert-json',
+], function($, ko, Cookies, arches, AlertViewModel, JsonErrorAlertViewModel) {
```
chiatt, line 8: The JsonErrorAlertViewModel appears to be unused. I think it can be removed.

chiatt (changes requested):

chiatt (approved):

Issue #8049, opened 2021-12-27
Allow user to create new instances from search driven resource-instance widget 


chiatt:
When the resource instances widget is configured with a search, the user cannot create a new instance from the dropdown. Users should be allowed to create new instances if graphs are listed in the config for the node.

chiatt:
Resolved in #8588

PR #8588, opened 2022-06-21
Add capability to create related resource in workflows, #8049

njkim:
### Types of changes

### Description of Change

Add the capability to create a resource instance when the renderContext is search, ie. workflow, #8049
Now you can pass the `allowInstanceCreate` param in the components to decide to allow user to create a resource.
Ie. allowInstanceCreation should be independent to renderContext.

Also, add the param that a user can add search string to limit the selection by adding search string url.

### Issues Solved

#8049

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/media/js/viewmodels/resource-instance-select.js
```diff
@@ -386,10 +401,10 @@ define([
                     } else {
                         self.url(arches.urls.search_results);
                         var queryString = new URLSearchParams();
-                        if (!!params.node && ko.unwrap(params.node.config.searchString) !== ""){
-                            var searchUrl = new URL(ko.unwrap(params.node.config.searchString));
+                        if (self.searchString) {
+                            const searchUrl = new URL(self.searchString);
                             queryString = new URLSearchParams(searchUrl.search);
-                            self.allowInstanceCreation = false;
+                            //self.allowInstanceCreation = false;
```
chiatt, line 407: Could you remove commented line?

chiatt (changes requested):
This looks really good. Just one very minor change to make.

arches/app/media/js/viewmodels/resource-instance-select.js
njkim: Thanks for catch it. Fixed.

PR #8604, opened 2022-06-28
dev/6.2.x with django upgrade into v7 

chiatt:
Updates v7 with latest in dev/6.2.x including (primarily) the upgrade to django 3.2.


chrabyrd (approved):
I can create a new project, and I can migrate a 6.1.x project to this branch without conflicts. I gave the application the sniff test in both cases and everything seems to pass 👍

Issue #8614, opened 2022-06-28
Test for celery failure in ETL Manager


ryan86:
Test for celery load failure in ETL manager for both branch_csv and single_csv and report status in load even table as 'failed'

chiatt:
Resolved in #8615

PR #8615, opened 2022-06-28
Add try/except to test for celery failure during single_csv and branc…

ryan86:
…h_csv load, re #8614

### DIFF
--- a/arches/app/tasks.py
+++ b/arches/app/tasks.py
@@ -208,29 +208,38 @@ def on_chord_error(request, exc, traceback):
 def load_branch_csv(userid, files, summary, result, temp_dir, loadid):
     from arches.app.etl_modules import branch_csv_importer
 
-    BranchCsvImporter = branch_csv_importer.BranchCsvImporter(request=None, loadid=loadid, temp_dir=temp_dir)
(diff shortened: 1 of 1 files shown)


arches/app/tasks.py
```diff
@@ -208,29 +208,38 @@ def on_chord_error(request, exc, traceback):
...
-    status = _("Compeleted") if load_event.status == "indexed" else _("Failed")
-    msg = _("Branch Excel Import: {} [{}]").format(summary["name"], status)
-    user = User.objects.get(id=userid)
-    notify_completion(msg, user)
+        load_event = models.LoadEvent.objects.get(loadid=loadid)
+        status = _("Completed") if load_event.status == "indexed" else _("Failed")
+        msg = _("Branch Excel Import: {} [{}]").format(summary["name"], status)
+        user = User.objects.get(id=userid)
+        notify_completion(msg, user)
+    except:
```
chiatt, line 220: Could you modify this (and the other except statements in this pr) so that the error gets logged? 
```
except Exception as e:
    logger.error(e)
```

chiatt (changes requested):
This looks good, but I think it would be helpful to have error messages written to the log.

chiatt (approved):

Issue #8578, opened 2022-06-18
Loading resource instances with 1000's of related resources is really slow


apeters:
The attached json-ld file has 1000's of related resources and takes over 5 min to load.
[https://app.zenhub.com/files/49451058/8f0f5234-bea0-4490-a6ca-25b5a5bbb6d6/download](https://app.zenhub.com/files/49451058/8f0f5234-bea0-4490-a6ca-25b5a5bbb6d6/download
)

chiatt:
Resolved in #8619

PR #8619, opened 2022-06-30
8578 rr load performance

apeters:
re #8578


apeters:
To test this you can check that when a tile that has a related-resource node in it is saved that it updates the tile with the appropriate resourceXResourceId value and that corresponding entries in the related-resources table are made for that tile/node combination.

chiatt:
Thanks @apeters, I'll have a look

arches/app/models/tile.py
```diff
@@ -342,18 +342,24 @@ def get_tile_data(self, user_id=None):
...
+
+    def datatype_post_save_actions(self, request=None):
+        try:
+            userid = str(request.user.id)
+        except:
+            userid = None
+
         tile_data = self.get_tile_data(userid)
-        for nodeid, value in list(tile_data.items()):
+        for nodeid in tile_data.keys():
```
chiatt, line 359: This is a nice change. There are a lot of places in Arches where we should be using .keys() and .values() instead of .items().

arches/app/datatypes/datatypes.py
```diff
@@ -1911,71 +1912,18 @@ def validate(self, value, row_number=None, source="", node=None, nodeid=None, st
...
-            # then those need to be removed from the db
-            resourceXresourceInDb = set(
-                models.ResourceXResource.objects.filter(tileid_id=tile.pk, nodeid_id=nodeid).values_list("pk", flat=True)
-            )
-            to_delete = resourceXresourceInDb - resourceXresourceSaved
-            for rr in models.ResourceXResource.objects.filter(pk__in=to_delete):
-                rr.delete()
+        with connection.cursor() as cursor:
+            cursor.execute(sql)
+            ret = cursor.fetchone()
```
chiatt, line 1925: In order for this to work from the web UI the resource relationship would need to be indexed. Either that or we would need to stop relying on relationship index and get our relationships from the database.

chiatt (changes requested):
This looks really good, but the resource relationship is no longer indexed. I think this would  normally happen in the save method of the ResourceXResource model. Because they are not indexed, relationships created through the UI no longer appear in the related resources visualization.

arches/app/datatypes/datatypes.py
apeters: Good catch @chiatt!  I remembered that we indexed that data when we saved a ResourceXResource instance but of course that was when we were using the ORM.  
Honestly I'm not sure why we're indexing that data at all.  The indexed data has the same form as the table itself.  This is where [pgsync](https://pgsync.com/) might pay off.  
In the mean time I should probably fix that....

chiatt: I think ES was used originally to minimize database queries because our db query count was really, really high at the time. That isn't the case anymore, so I think getting relationship data from the database and removing the resource relationship index altogether would be a better approach than indexing with pgsync. That is unless we have a reason to search within resource relationships in a way that we can't do with postgres. I can't think of what that query would be though.

tests/models/tile_model_tests.py
```diff
@@ -87,336 +91,455 @@ def setUp(self):
     def tearDown(self):
         pass
 
-    def test_load_from_python_dict(self):
+    # def test_load_from_python_dict(self):
```
chiatt, line 94: Did you mean to leave this test commented out?

chiatt (changes requested):
This looks good, but there is a little cruft that could be removed and a disabled test that may need to be reenabled.

chiatt (approved):

Issue #8623, opened 2022-07-01
improve logging of JSON-LD reader


apeters:

apeters:
closed by #8666

PR #8624, opened 2022-07-01
allow users to print detailed logging information to the screen, re #8623

apeters:
re #8623

### DIFF
--- a/arches/app/utils/data_management/resources/formats/rdffile.py
+++ b/arches/app/utils/data_management/resources/formats/rdffile.py
@@ -471,12 +471,15 @@ class JsonLdReader(Reader):
         if indent and indent > 0:
             prefix = prefix + "   " * int(indent)
         self.print_buf.append(prefix + text)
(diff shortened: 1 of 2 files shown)


apeters:
to test this, you'll have to generate a json-ld file for a specific resource instance.
You can do that by viewing the report for any instance and then change the portion of the url from `reports` to `resources`.
That should spit out a json-ld version of the instance.
Then create a directory called `data`, and create a directory under that with the name of the `slug` of the graph the instance belongs to.  Then create a directory under that with the 2 first characters of the instance id.
EG:
I have a `person` resource instance with id "08d5e746-ae91-3060-8ec2-95bd7a47e445"
I would then create a directory structure like so
`data/person/08/08d5e746-ae91-3060-8ec2-95bd7a47e445.json`

Once that's done use the management command like so:
`python manage.py load_jsonld -s ../data -ow overwrite -v 3`

ryan86:
Looks good

Issue #8567, opened 2022-06-17
Make Branch Excel and Single CSV importers compatible with v7


chiatt:
The importers need to allow users to upload data in different languages. The work here is largely related to adapting the importers to support the v7 string datatype.

chiatt:
Resolved in #8635

PR #8635, opened 2022-07-08
Update etl modules for v7, #8567

njkim:
### Types of changes

### Issues Solved

#8567 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/etl_modules/branch_csv_importer.py
```diff
@@ -116,6 +117,11 @@ def create_tile_value(self, cell_values, data_node_lookup, node_lookup, row_deta
                     config["nodeid"] = nodeid
                 except TypeError:
                     config = {}
+                if datatype == "string":
+                    try:
+                        source_value = ast.literal_eval(source_value)
```
chiatt, line 122: This check actually belongs in the string datatype's transform_value_for_tile method. If it's moved into transform_value_for_tile, then this same block can be removed from `import_single_csv` as well.

chiatt (changes requested):
Overall this looks good. There is just some datatype specific logic that should be moved into the string datatype class.

arches/app/etl_modules/branch_csv_importer.py
njkim: Moved, thanks.

chiatt (approved):
Looks good to me!

Issue #8671, opened 2022-07-15
Check for id rather than related ontology in graph model


chiatt:
In the graph model there is an if statement in which we check if the graph has a related ontology. This requires a database query. It would be more efficient just to check if the graph has an ontology_id property.

https://github.com/archesproject/arches/blob/a7aace1f5036e0f60980b5fa3983b65b0d9ed715/arches/app/models/graph.py#L207

chiatt:
Fixed in #8672

PR #8672, opened 2022-07-15
Check for ontology id rather than ontology object, re #8671

chiatt:
### Types of changes

### Description of Change
Checks for ontology_id rather than performing a database query to determine if a graph has an assigned ontology.

### Issues Solved

#8671


robgaston (approved):
looks good, thanks!

Issue #8685, opened 2022-07-21
Admin user unable to delete tiles


chiatt:
In v7 the Admin user gets the error that a provisional user would expect when trying to delete a tile with authoritative data.

![image](https://user-images.githubusercontent.com/1877663/180306199-fd2da988-0fb4-4089-a2a8-ee6a559fbb16.png)

chiatt:
Resolved in #8696

PR #8696, opened 2022-07-25
Prevent error when deleting tiles, re #8685

chiatt:
### Types of changes

### Description of Change
Prevents error when deleting tiles. Also removes some cruft and unnecessary db queries.

### Issues Solved

#8685


aarongundel (approved):
Looks good to me.

Issue #8691, opened 2022-07-21
In v7 related resource are labeled with object rather than primary descriptor


chiatt:
![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/9c8b18a8-e311-4f6b-817e-685a940f7dcd)

chrabyrd:
solved in #8703

Issue #8682, opened 2022-07-21
Resource instance widget improperly updated after new resource instance is created


chiatt:
When a new resource instance is created using the resource instance widget, an entry for the newly created instance should appear below the widget input. Instead, only the word 'null' appears in the input.

In 7.0.x  (incorrect):
![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/8735b0ca-e758-40a0-a24d-8c84149450bf)

In 6.2.x (correct): 
![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/9caa7c88-2a1f-423e-89bb-181fb0a3355f)

chrabyrd:
solved in #8703

PR #8703, opened 2022-07-26
updates displaydescription and displayname returns from methods leveraging the ES index #8682

chrabyrd:
### Types of changes

### Description of Change

This updates the places in code where we're using the `displaydescription` and `displayname` objects retrieved from ElasticSearch. This does NOT update the `ResourceDescriptors` class as it's not being used anywhere that I can tell && having the raw ES return could be beneficial.

In a followup ticket, we should consider refactoring `Resource.get_documents_to_index` and the Resource database model to handle i18n in the application-standard way.

### Issues Solved

#8682 #8691


robgaston (approved):
looks good to me. thanks

Issue #8706, opened 2022-07-26
i18n 'loadmessags' command fails with error


chrabyrd:
```
python manage.py i18n loadmessages --lang=en
System check identified some issues:

WARNINGS:
Traceback (most recent call last):
  File "manage.py", line 29, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/Users/cbyrd/Projects/arches/arches/management/commands/i18n.py", line 48, in handle
    self.load_messages(lang=options["lang"])
  File "/Users/cbyrd/Projects/arches/arches/management/commands/i18n.py", line 60, in load_messages
    ArchesPOLoader(file.file, settings.LANGUAGE_CODE, file.language).load()
AttributeError: 'tuple' object has no attribute 'file'
```
```
> files
[('en', [], False)]
```

chrabyrd:
closing as user error for now

chrabyrd:
resolved in #8711

Issue #8688, opened 2022-07-21
In v7 'null' Appears as subtitle in cards that do not have a subtitle


chiatt:
The string 'null' is automatically saved as the subtitle of a card. This displays under the card title in the resource editor.

![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/44627868-1403-4fcd-9c7e-52e0a38ffc5d)

chrabyrd:
resolved in #8711

PR #8711, opened 2022-07-28
update i18n logic to handle for false positives #8688

chrabyrd:
### Types of changes

### Description of Change

This was tested using the arches-HER package.

- navigate to the graph designer
- navigate to the `Application Area` resource
- navigate to the `References` card
- notice that the card subtitle is not shown to the user

This is a low-level change that likely fixes other i18n false-positives in the application.

### Issues Solved

#8706 
#8688


robgaston:
@chrabyrd these changes look reasonable to me, but the tests are failing specifically around tests that reference handling null values in i18n; could you please take a look at these tests and assess?  I'm not sure if this indicates an issue with our code (ie these tests imply we should actually be allowing for nulls for some reason that I'm missing), or if these tests just need to be modified (ie we actually want to handle nulls in this way and the test assertions just need to be changed).

chrabyrd:
@robgaston  -- apologies for assigning this to you too early! Yeah I thought a test might fail this was just to run the test async from my local system. 

I'd say this is the point where we need @apeters opinion.

robgaston (approved):
looks good - thanks

Issue #8736, opened 2022-08-02
Add resourceid to responseJson in resource-report-abstract.js


ryan86:
Add resourceid to responseJson in resource-report-abstract.js. So that the resourceid can be passed to a report even if responseJson is empty coming back from api.ResourceReport.

chiatt:
No longer appears needed in the latest

PR #8737, opened 2022-08-02
Add resourceid to responseJson in resource-report-abstract.js, re #8736

ryan86:
Add resourceid to resource-report-abstract.js so that a report can receive a resourceid even when reponseJson.report_json is empty coming back from api.ResourceReport, re #8736

### DIFF
--- a/arches/app/media/js/views/components/resource-report-abstract.js
+++ b/arches/app/media/js/views/components/resource-report-abstract.js
@@ -58,8 +58,8 @@ define([
             } 
             else if (self.resourceid) {
                 url = arches.urls.api_resource_report(self.resourceid) + "?v=beta&uncompacted=true";
-
                 self.fetchResourceData(url).then(function(responseJson) {
+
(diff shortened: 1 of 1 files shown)


arches/app/media/js/views/components/resource-report-abstract.js
```diff
@@ -69,6 +69,7 @@ define([
                         self.report({
                             'template': responseJson.template,
                             'report_json': responseJson.report_json,
+                            'resourceid': responseJson.resourceid
```
chiatt, line 72: In 6.2 I don't see a `resourceid` property on the responseJson object. Is it possible there were some backend changes not included in this PR? If not, it does look like there is a `responseJson.report_json.resourceinstanceid`

image

chiatt (commented):
In v7 looks like the resourceid is no longer available on the responseJson object, but is available in the report via the 'report_json' property, so I don't think this change is still needed.

Issue #8742, opened 2022-08-02
Adding new languages causes existing graph publications to break on add


aarongundel:
**Describe the bug**

After adding a new language in Arches, the editor no longer works  (in the language that is added) for graphs that have already been published.  The error is 

```
DoesNotExist at /add-resource/
PublishedGraph matching query does not exist.
```

aarongundel:
More context for this ticket:  The issue happens whenever a new language is added and the graphs are already published.  We either need to make sure that all languages have a publication record or fall back on the default language publication record.  The short-term workaround is to republish the graph after the language is added.

aarongundel:
Note:  Better exception handling (or even better - no exception) would be ideal here.  The user doesn't need to know there's an issue or should know immediately how to fix it.

chiatt:
A command would also be helpful to sync up publications.

PR #8773, opened 2022-08-08
Adds command to synchronize languages

aarongundel:
### Types of changes

### Description of Change

Adds command to sync languages from settings to db.  Removes from context_processors; includes the ability to synchronize publications, thus resolving #8742.  This is now called automatically when a publication cannot be retrieved on the current language, preventing a somewhat obscure error that happens when a sys admin adds a language and a publication does not include that language.

### Issues Solved

#8742 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chiatt (changes requested):
This works like a charm, but it looks like there's an unused import added in resource.py

chiatt (approved):

Issue #8769, opened 2022-08-05
Change default string value in string widget to internationalized object


chiatt:
The default value in the string widget is an empty string. This should be changed to an internationalized object.

chiatt:
Fixed in #8781

PR #8781, opened 2022-08-08
Makes default string widget value an object with empty string, re #8769

chiatt:
### Types of changes

### Description of Change
Makes default string widget value an object with empty string

### Issues Solved

#8769


njkim (approved):
Looks great.

Issue #8723, opened 2022-08-01
Arches Designer `Create Mapping File` interaction fails with JSON error in v7 with some Resource Models


chrabyrd:
**Describe the bug**

Stacktrace: 
```
Environment:


Request Method: GET
Request URL: http://localhost:8000/graph/b9e0701e-5463-11e9-b5f5-000d3ab1e588/export_mapping_file

Django Version: 3.2.14
Python Version: 3.8.6
Installed Applications:
('webpack_loader',
 'django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.gis',
 'arches',
 'arches.app.models',
 'arches.management',
 'guardian',
 'captcha',
 'revproxy',
 'corsheaders',
 'oauth2_provider',
 'django_celery_results',
 'compressor')
Installed Middleware:
['corsheaders.middleware.CorsMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'arches.app.utils.middleware.ModifyAuthorizationHeader',
 'oauth2_provider.middleware.OAuth2TokenMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'arches.app.utils.middleware.SetAnonymousUser']



Traceback (most recent call last):
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
[...]

PR #8785, opened 2022-08-09
Fix issue where mapping files were not able to be exported, re #8723

ryan86:
Fix issue where mapping files could not be exported in some cases, re #8723


chrabyrd:
lgtm!

Issue #8762, opened 2022-08-04
switch-widget template for boolean datatype shows [object Object] in Designer


chrabyrd:
**Describe the bug**

switch-widget template for boolean datatype shows [object Object] in Designer. Also, regardless of template, changing the default value does not update the preview in real time

**To Reproduce**
1. navigate to arches designer
2. create or find a resource model with a boolean datatype
3. navigate to the `card` tab and update the widget template to `switch-widget`
4. notice the UI breaks

**Screenshots**

<img width="1409" alt="Screen Shot 2022-08-04 at 1 04 19 PM" src="https://user-images.githubusercontent.com/9162695/182942507-acbe40d7-6c2b-4814-81ac-ff5c24fa1b0b.png">

chiatt:
This issue is essentially resolved. There are still some refinements to make the patch release. https://github.com/archesproject/arches/issues/8794

PR #8787, opened 2022-08-10
Fix for [object object] error in subtitle for switch widget

aarongundel:
### Types of changes

### Description of Change

Fixes [object object] in the subtitle of the switch widget.  There are still some mysterious issues since the widget is instantiated twice (or more) in the graph designer and the config does not propagate correctly, but this change should make it functional.  

### Issues Solved

#8762 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/views/components/widgets/switch.js
+++ b/arches/app/media/js/views/components/widgets/switch.js
@@ -1,9 +1,10 @@
 define([
+    'arches',
     'knockout', 
     'underscore', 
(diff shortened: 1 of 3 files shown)


chiatt (approved):
This is much better. As we discussed there are still some minor issues editing the subtitle, but it now appears as expected.

Issue #8779, opened 2022-08-08
The application enters an error state if a user with nodegroup restrictions navigates to the resource editor


chrabyrd:
**Describe the bug**

```
Uncaught TypeError: Cannot read properties of undefined (reading 'parentnodegroup_id')
    at card.js:231:1
```

**To Reproduce**
1. create a resource editor user
2. as an administrator, restrict this user's write access to a resource model's nodegroup ( or restrict all access )
3. as the user, attempt to create a new resource of said model

**Screenshots**

<img width="2007" alt="Screen Shot 2022-08-08 at 3 20 38 PM" src="https://user-images.githubusercontent.com/9162695/183524472-081d6ba5-d8b7-44fe-a695-feffa5823b5d.png">

chiatt:
Fixed in #8790

PR #8790, opened 2022-08-10
fix nodegroup restrictions resource editor cause page break

chrabyrd:
### Types of changes

### Description of Change

Updates frontend files to handle for missing nodegroup. The only change of significance is in `arches/app/media/js/views/resource/editor.js`

### Issues Solved

#8779

### DIFF
--- a/arches/app/templates/views/components/cards/default.htm
+++ b/arches/app/templates/views/components/cards/default.htm
@@ -1,7 +1,7 @@
 {% load i18n %}
 <!-- ko foreach: { data: [$data], as: 'self' } -->
 
-<!-- ko if: state === 'editor-tree' && self.card.model.visible() -->
+<!-- ko if: state === 'editor-tree' -->
 {% block editor_tree %}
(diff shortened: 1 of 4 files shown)


arches/app/templates/views/components/cards/default.htm
```diff
@@ -1,7 +1,7 @@
 {% load i18n %}
 <!-- ko foreach: { data: [$data], as: 'self' } -->
 
-<!-- ko if: state === 'editor-tree' && self.card.model.visible() -->
```
chiatt, line 4: This causes cards cards that should not be visible to appear in the tree.

chiatt (changes requested):
This resolves the issue, but a change in default.htm is causing hidden cards to display.

arches/app/templates/views/components/cards/default.htm
chrabyrd: resolved!

chiatt (approved):

Issue #8792, opened 2022-08-10
Copy webpack config files in update project command 


chiatt:
When updating a project from v6 to v7 a user has to manually copy files from their arches installation in site-packages to their project. This could be difficult for some users not familiar with where site-packages, or even their Python installation is located. 

Instead, the update_project command should move the files for the user.

chiatt:
Resolved in #8799

chrabyrd:
resolved in #8799

PR #8799, opened 2022-08-11
Copy webpack config files to project in `updateproject` management command re #8792

chrabyrd:
### Types of changes

### Description of Change

Updates the `updateproject` management command to copy the webpack config files from arches-core to the project

### Issues Solved

#8792

### DIFF
--- a/arches/management/commands/updateproject.py
+++ b/arches/management/commands/updateproject.py
@@ -20,10 +20,13 @@ class Command(BaseCommand):
 
     def update_to_v7(self):
         # copy webpack config files to project
-        project_webpack_path = settings.APP_ROOT + "/webpack/"
+        project_webpack_path = os.path.join(settings.APP_ROOT, "webpack")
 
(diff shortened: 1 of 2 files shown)


arches/management/commands/updateproject.py
```diff
@@ -15,6 +19,13 @@ def handle(self, *args, **options):
             self.update_to_v7()
 
     def update_to_v7(self):
+        # copy webpack config files to project
+        project_webpack_path = settings.APP_ROOT + "/webpack/"
```
chiatt, line 23: I think `os.path.join` should probably be used here and on line 26 to make this command Windows safe.

chiatt (changes requested):
This works great @chrabyrd. Just one minor change needed to ensure windows users don't get an error.

arches/management/commands/updateproject.py
chrabyrd: good eye

chiatt (approved):

PR #8801, opened 2022-08-11
fix for invalid tiles being generated by the jsonld importer

### LINKED ISSUE #8800: json-ld importer sometimes generates invalid tiles
The JSON-LD importer sometimes generates invalid tiles.  This can happen when the tile in question requires a parent tile.   The importer will generate a tile _without_ a parent tile id.  These invalid tiles are not visible in the resource editor, and fortunately, it also creates the _correct_ tile - but if the node is a cardinality 1 node and two tiles are generated without a parent tile - this constitutes a cardinality violation.  This can prevent jsonld imports from moving forward.  

See @aarongundel  for info on testing/qa.

aarongundel:
### Types of changes

### Description of Change

Skips adding new tile if the tile should have a parent tile but doesn't.  Fixes #8800 

### Issues Solved

#8800 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/utils/data_management/resources/formats/rdffile.py
+++ b/arches/app/utils/data_management/resources/formats/rdffile.py
@@ -691,8 +691,7 @@ class JsonLdReader(Reader):
                     # Used to pick the previous tile in loop which MIGHT be the parent (but might not)
                     parenttile_id = result["tile"].tileid if "tile" in result else None
 
- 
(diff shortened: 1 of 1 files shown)


chiatt:
Addressed in recent 7.4.x work.

PR #8804, opened 2022-08-11
fix reverse migration of 6458_language #8784

### LINKED ISSUE #8784: Reversing migration replaces some tile data with `null`
Reversing the `6458_language` migration replaces some tile data with 'null'. This probably happens with tiles that belong to any resource model, but you can confirm it is occurring in the system settings instance. To duplicate, just load HER, check that the following tiles have either data or an empty object in `tiledata`.

```select * from tiles where resourceinstanceid = 'a106c400-260c-11e7-a604-14109fd34195';```

```
c45a7069-1803-4d85-a017-61296a2ae760
4345f530-aa90-48cf-b4b3-92d1185ca439
f49e37b3-3686-4fbd-98c9-ab8241f72a60
```

Next, reverse your migrations back to the migration just before `6458_language`. 
```
python manage.py migrate models 7794_i18n_string_datatype
```
If you check the above tiles again, you'll notice that `tiledata` has been changed to `null`

aarongundel:
### Types of changes

### Description of Change

Fixes #8784 by using the system language for the migration.  The old code was using the first language in the languages table, which is not always a reliable system language - it's not necessarily the same as LANGUAGE_CODE in the django settings.  The important note here is that when a user migrates to a7 - settings.LANGUAGE_CODE is the language they're going to migrate their default strings to, and when they back out of the migration, they're going to lose everything but settings.LANGUAGE_CODE as that is what the string will revert to.  

### Issues Solved

#8784

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/models/migrations/6458_language.py
+++ b/arches/app/models/migrations/6458_language.py
@@ -121,7 +121,9 @@ class Migration(migrations.Migration):
                     return;
                 end;
                 $$
-            """.format(settings.LANGUAGE_CODE)
+            """.format(
+                        settings.LANGUAGE_CODE
+                    )
                 )
             ],
             reverse_sql=[
@@ -155,7 +157,9 @@ class Migration(migrations.Migration):
                     return;
                 end;
                 $$
-            """.format(settings.LANGUAGE_CODE)
+            """.format(
+                        settings.LANGUAGE_CODE
+                    )
                 )
             ],
         ),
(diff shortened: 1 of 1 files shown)


chiatt (approved):
Looks good!

Issue #8813, opened 2022-08-15
resource model views don't create


aarongundel:
**Describe the bug**

The resource model views don't create right now when a slug is not provided.  This happens because the default values expect a string.  

`select __arches_create_resource_model_views(graphid)
from graphs
where isresource = true
    and publicationid is not null
    and name::text not like '%Arches System Settings%';`

**To Reproduce**
1. Try to run the resource model view creation function without providing a name parameter.  This will fail.

PR #8815, opened 2022-08-15
fixes the defaults for resource model creation 

aarongundel:
### Types of changes

### Description of Change

We elected to hardcode "en" as the language we're looking for by default in the __arches_create_resource_model_views - for all other needs, the user can pass in a schema_name manually. 

### Issues Solved

#8813 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/models/migrations/8813_resource_model_view_defaults.py
+++ b/arches/app/models/migrations/8813_resource_model_view_defaults.py
@@ -1,0 +1,295 @@
+from django.db import migrations, models
+
+
+# Add sensible defaults to the creation of resource model views.
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("models", "8400_remove_collector_references"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+        """
+            create or replace function __arches_get_node_value_sql(
+                node public.nodes
+            ) returns text as $$
+            declare
+                node_value_sql text;
+                select_sql text = '(t.tiledata->>%L)';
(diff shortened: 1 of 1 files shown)


arches/app/models/migrations/8813_resource_model_view_defaults.py
```diff
@@ -0,0 +1,295 @@
...
+                            st_collect(
+                                array(
+                                    select st_transform(geom, 4326) from geojson_geometries
+                                    where geojson_geometries.tileid = t.tileid and nodeid = %L
+                                )
+                            )',
+                            node.nodeid
+                        );
+                        datatype = 'geometry';
+                    when 'string' then datatype = 'jsonb';
```
aarongundel, line 35: @robgaston here's the addition you requested.

arches/app/models/migrations/8813_resource_model_view_defaults.py
```diff
@@ -0,0 +1,295 @@
...
+                model_id uuid,
+                schema_name text default null
+            ) returns text as $$
+            declare
+                creation_sql text;
+                node record;
+            begin
+                -- create schema and instance view for model
+                if schema_name is null then
+                    select __arches_slugify(name->>'en') into schema_name
```
aarongundel, line 106: @robgaston here's the main modification in this PR.

robgaston (approved):
looks good - thanks

Issue #8707, opened 2022-07-26
The cancel(undo) button in the editor does not work


njkim:
The cancel button in widget in the editor for many datatypes does not work.

The dirty state may not be working, for the cancel button sometimes appears right away before entering a value.

![Screen Shot 2022-07-26 at 4.13.06 PM.png](https://images.zenhubusercontent.com/5c6c877ade77103362c0ab7e/85f08c3d-077a-4e0e-9a31-153cb25458a8)

robgaston:
fixed in latest

PR #8816, opened 2022-08-16
Prestructure tile data with languages, re #8707

chiatt:
Adds missing languages to tile string nodes before data is delivered to front-end to help address #8707


njkim (approved):
Works great!

Issue #8763, opened 2022-08-04
`node-value` datatype throws PermissionDenied error


chrabyrd:
**Describe the bug**

server logs:
```
[04/Aug/2022 15:11:33] "GET /static/node_modules/ckeditor/plugins/tableselection/styles/tableselection.css HTTP/1.1" 200 1145
Forbidden (Permission denied): /resource/343cc20c-2c5a-11e8-90fa-0242ac120005/tiles
Traceback (most recent call last):
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/Users/cbyrd/Projects/arches/arches/app/utils/decorators.py", line 102, in wrapper
    raise PermissionDenied
django.core.exceptions.PermissionDenied
[04/Aug/2022 15:11:46] "GET /resource/343cc20c-2c5a-11e8-90fa-0242ac120005/tiles?nodeid=546b1630-3ba4-11eb-9030-f875a44e0e11&term=&_=1659643876633 HTTP/1.1" 403 135
```

**To Reproduce**
1. arches-HER webpack branch, core webpack branch
2. navigate the arches designer for `Artefact` resource instance
3. select the `Associated Construction Phase` node
4. navigate to the card tab and see the error in your browser's console

NOTE: this behavior also happens in the Resource Editor

chiatt:
Resolved in #8824

PR #8824, opened 2022-08-16
update node-value-select to handle for edge cases #8763

chrabyrd:
### Types of changes

### Description of Change

The previous logic uses a specific pattern to pull a `resourceId` from the URL, and fire a request that ultimately hits `permission_backend.check_resource_instance_permissions` . This function attempts to fetch the ResourceInstance from the database, and if that object doesn't exist it returns falsey. However the `add-resource` and `graph_designer` views do not yet have a `ResourceInstance` entry in the database, so the function incorrectly returns falsey and that results in an error.

### Issues Solved

#8763


chiatt (approved):
Looks good!

Issue #8724, opened 2022-08-01
`Clone Model` interaction in Arches Designer fails with error


chrabyrd:
**Describe the bug**

Stacktrace: 
```
[01/Aug/2022 12:12:55] "POST /graph/22477f01-1a44-11e9-b0a9-000d3ab1e588/clone HTTP/1.1" 500 132728
Internal Server Error: /graph/42ce82f6-83bf-11ea-b1e8-f875a44e0e11/clone
Traceback (most recent call last):
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/Users/cbyrd/Projects/ENV/lib/python3.8/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/Users/cbyrd/Projects/arches/arches/app/views/graph.py", line 392, in post
    clone_data = graph.copy()
  File "/Users/cbyrd/Projects/arches/arches/app/models/graph.py", line 819, in copy
    widget.node_id = node_map[widget.node_id]
KeyError: '0a5e6b91-860d-11ea-aa4f-f875a44e0e11'
```

**To Reproduce**
1. Ensure you're on core branch `8090-cbyrd-adds-frontend-compiler` + arches-HER branch `update_project_for_webpack_v7`
2. Ensure the application has been successfully webpacked
3. navigate to the graph designer
4. Attempt to clone any Resource Model
5. Notice the interaction fails with above error

chrabyrd:
resolved in #8874

PR #8825, opened 2022-08-16
Fix issue where graphs could not be cloned, re #8724

ryan86:
Fix issue where user was unable to clone a model from the graph designer, re #8724 
### Types of changes

### Issues Solved

#8724 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/models/graph.py
+++ b/arches/app/models/graph.py
@@ -779,7 +779,7 @@ class Graph(models.GraphModel):
             copy_of_self.subtitle = ""
             copy_of_self.description = ""
             copy_of_self.author = ""
-            copy_of_self.serialized_graph = ""
+        copy_of_self.serialized_graph = ""
 
         # returns a list of node ids sorted by nodes that are collector nodes first and then others last
         node_ids = sorted(copy_of_self.nodes, key=lambda node_id: copy_of_self.nodes[node_id].is_collector, reverse=True)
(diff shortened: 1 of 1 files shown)


arches/app/models/graph.py
```diff
@@ -779,6 +779,7 @@ def flatten_tree(tree, node_id_list=[]):
             copy_of_self.subtitle = ""
             copy_of_self.description = ""
             copy_of_self.author = ""
+            copy_of_self.serialized_graph = ""
```
chrabyrd, line 782: I thought about this a little since we last spoke. Since we're cloning the graph reason stands that we're also to clone `published` state. So instead of just nulling out serialized_graph, we should ensure the cloned graph gets saved to the database. After the save we should then re-publish the graph if the original graph was published.

chrabyrd (changes requested):
resolved

arches/app/models/graph.py
chrabyrd: deferred to 3rd developer -- the cloned graph should be unpublished

chrabyrd (changes requested):
I'm still getting the error originally outlined in the ticket. Happy to chat any time if you'd like 👍

chrabyrd (changes requested):
I'm still getting the error originally outlined in the ticket. Happy to chat any time if you'd like

chrabyrd:
fixed in latest

Issue #8722, opened 2022-07-31
Loading internationalized resources results in search


chiatt:
This is probably due to the descriptor not being properly formed for other languages

<img width="510" alt="image" src="https://user-images.githubusercontent.com/1877663/184539729-db61490e-ef95-4c44-86be-27b43a0f76c0.png">

PR #8841, opened 2022-08-18
Moves descriptor conversion from object to text entirely to backend, …

chiatt:
### Types of changes

### Description of Change
Moves descriptor conversion from object to text to the backend so that search results always has a string value as a descriptor. Prioritizes languages in order of active language, default language, first language in list of available languages.

### Issues Solved
#8722


aarongundel (approved):
Looks good and works for me!

Issue #8726, opened 2022-08-01
String values do not render properly in resource instance history


chiatt:
String tile values display as [object Object] in resource instance history.

![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/979ed449-8878-4b63-9d0b-28e8ca56da08)

chrabyrd:
resolved in #8843

PR #8843, opened 2022-08-18
Adds i18n localization tool re #8726

chrabyrd:
### Types of changes

### Description of Change

Adds an i18n utility for localizing complex inputs. I went back and forth with myself on the shape this logic should take and where exactly it should be used, I'm not married to any of it and I'm happy to hear different opinions. Also, I've added a few comments to explain some of the logic behind the changes.

### Issues Solved

#8726


arches/app/utils/i18n.py
```diff
@@ -10,9 +10,44 @@
...
+    """
+    deserialized_json = JSONDeserializer().deserialize(json)
+
+    def recursive_localize(deserialized_json):
+        if isinstance(deserialized_json, list):
+            for obj in deserialized_json:
+                recursive_localize(obj)
+        if isinstance(deserialized_json, dict):
+            for key in deserialized_json.keys():
+                try:
```
chrabyrd, line 39: This calls `get_localized_value` on every `dict`, and if `get_localized_value` throws an exception then the `dict` wasn't an i18n dict. It's possible to add some quick regex here to look at each key in the `dict` and if any of them match the i18n key pattern ('xx", 'xx-XX') only then call `get_localized_value`, but I'm unsure if it's worth it since `get_localized_value` is speedy enough on its own and it already iterates all the keys of the dict.

arches/app/utils/i18n.py
```diff
@@ -58,8 +93,9 @@ def get_localized_value(obj, lang=None, return_lang=False):
 
             if settings.LANGUAGE_CODE in obj:
                 found_lang = settings.LANGUAGE_CODE
-            else:
```
chrabyrd, line 61: The original code basically translated to: "if I can't match any of the keys in the dictionary to a language code, assign the last key in the dict to `found_lang`. If this is called on a non-i18n dictionary it fails silently and in an interesting way. Instead I've removed this logic and replaced it with throwing an error if this `get_localized_value` is called on a non-i18n dict

arches/app/views/resource.py
```diff
@@ -596,7 +596,7 @@ def get(self, request, resourceid=None, view_template="views/resource/edit-log.h
                 resource_type=graph_name,
                 resource_description=resource.displaydescription(),
                 iconclass=resource_instance.graph.iconclass,
-                edits=JSONSerializer().serialize(permitted_edits),
+                edits=localize_serialized_object(JSONSerializer().serialize(permitted_edits)),
```
chiatt, line 599: It looks like `permitted_edits` is getting serialized here, then deserialized at the start of `localize_serialized_object` and then re-serialized at the end of `localize_serialized_object`. Could permitted_edits be localized here first and then serialized? Basically flip the order like this:
`JSONSerializer().serialize(localize_object(permitted_edits)),`

chrabyrd: ya good advice

robgaston (approved):
looks good - thanks!

Issue #8780, opened 2022-08-08
`Manage` link on index page can be accessed by non-permissioned users


chrabyrd:
**Describe the bug**

**To Reproduce**
1. arches-HER webpack branch, arches-core webpack branch
2. as admin, create a guest user
3. login as guest user, notice `Manage` link on index page
4. click manage link and notice `403 Forbidden` error

chrabyrd:
resolved in #8844 8844

PR #8844, opened 2022-08-19
Hide manage link if user is not in resource editor group, re #8780

chiatt:
### Types of changes

### Description of Change
Hides manage link on index page if user is not in resource editor group

### Issues Solved

#8780

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/utils/permission_backend.py
+++ b/arches/app/utils/permission_backend.py
@@ -227,8 +227,10 @@ def get_createable_resource_types(user):
     user -- the user to check
 
     """
-
-    return get_resource_types_by_perm(user, "models.write_nodegroup")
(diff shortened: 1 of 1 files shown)


chrabyrd (changes requested):
This works great and the change makes logical sense. However while testing this I noticed the `Graphs` link is still visible to a Guest user. It wasn't originally scoped in the ticket but I'm hoping the solution is a similar change to different function. Of course if it's a bigger lift than expected it can be solved in another ticket

chrabyrd:
lgtm!

PR #8845, opened 2022-08-19
descriptor changes for i18n

### LINKED ISSUE #8838: Descriptors indexed and saved to resource instance table in active language only
When a document is indexed the displayname is save in the application's active language. If a tile value used to create the displayname has a value in two languages, the array of displaynames in the document will only have one entry. Consequently the displayname will only ever appear in search in the language in which it was last saved, regardless of what the active or default language is. 

This could easily lead to editors unintentionally overwriting the displayname in different languages.

In search the displayname is saved like this even if the tile has the same node in English:
 ```[ { "language": "de","value": "Ein Ding"}],```

It seems like both languages should be saved instead like so:
 ```[ { "language": "de","value": "Ein Ding"}, { "language": "en","value": "A Thing"}],```

This would allow the displayname to appear in whatever language the user has activated in the language dropdown.

To do that, it seems that the format of the displayname would need to change in the resource instance table because it is currently saved without any language information:
```
{
  "name": "Ein Ding",
  "map_popup": "<Description>",
  "description": " <Description>"
}
```

aarongundel:
### Types of changes

### Description of Change

Changes the descriptor name field to be i18njson - to support multiple values for different languages.  Changes how indexing happens - to minimize issues with things that depend on the display_descriptors fields, it just keeps an array of values with a language key.  Changes when descriptor name is saved and allows you to request language specific versions of a string based primary descriptor.  

### Issues Solved

closes #8838 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/datatypes/base.py
+++ b/arches/app/datatypes/base.py
@@ -203,7 +203,7 @@ class BaseDataType(object):
             logger.exception(_("Tile has no authoritative or provisional data"))
 
 
-    def get_display_value(self, tile, node):
+    def get_display_value(self, tile, node, **kwargs):
         """
         Returns a list of concept values for a given node
         """
--- a/arches/app/datatypes/concept_types.py
+++ b/arches/app/datatypes/concept_types.py
@@ -175,7 +175,7 @@ class ConceptDataType(BaseConceptDataType):
     def get_pref_label(self, nodevalue, lang="en-US"):
         return get_preflabel_from_valueid(nodevalue, lang)["value"]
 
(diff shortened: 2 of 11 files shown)


apeters (approved):
This works as described.  One note is that the descriptors field (column) might be better served as an I18n_JSON or I18n_Text field.

robgaston (approved):
looks good to me -- thanks!

Issue #8866, opened 2022-08-24
dev/7.0.x Bug Broken display of resource instance name after alternate language selected


ekansa:
I'm running dev/7.0.x Arches core with the Arches HER project in order to test some internationalization features. 

I have a resource instance with an English, Arabic, and Hebrew name. When I switch the search feature map view to click on the map icon of the resource instance, I first see snippet of JSON text as the name in for the resource instance in the popup. A few moments later this changes to '[object Object]'. It seems some javascript involved in the popup display may have been expecting a simple string rather than a mult-attribute object?

I would expect the Hebew language version of this resource instance name to be displayed in the popup.

Here's the initial view showing a fragment of JSON text:
![arches-partial-json-hebrew-part-1](https://user-images.githubusercontent.com/1286742/186490935-88c0dd36-941c-4fa6-a024-44710e3d54e5.png)

Here's what it looks like a few moments later when '[object Object]' gets displayed:
![arches-partial-json-hebrew-part-2](https://user-images.githubusercontent.com/1286742/186491027-355588f3-11a4-49ce-bc19-230c543ee7f8.png)

robgaston:
may have already been addressed in #8845

chiatt:
Fixed in #8873

PR #8873, opened 2022-08-25
translates descriptors in resource view re: #8866

robgaston:
to fix issues with popups on map and iiif widgets, we should translate resource descriptors in the resource view.  the view now always returns a string for all descriptors, either: the only one available, the user's language translation, the system language translation, or the translated string "Undefined"

fixes: #8866


chiatt (approved):
Looks good!

Issue #8842, opened 2022-08-18
Branches not working in 7.x


aarongundel:
**Describe the bug**

Branches don't seem to be working 7.x - they're all showing as inactive.  There's also a function to publish branches - I'm not sure that makes sense.  In any case, the publication of branches doesn't seem to affect the active/inactive status of the branch so that it can be used in a graph.

**To Reproduce**
1. Go to any new graph
2. Try to add a branch
3. Problems

**Screenshots**

**Expected behavior**

**Your Arches Information**

*   Version used:
*   Operating System and version (desktop or mobile):
*   Browser Name and version:
*   Link to your Arches Install (optional):

**Additional context**

**Ticket Background**
*   Found by: @

aarongundel:
Per discussion with @robgaston and @chiatt remove the active flag from branches; treat all branches as active.  Remove the publish button from branches.

aarongundel:
thanks @ekansa for finding this issue.

ekansa:
@aarongundel  Thanks for following up and investigating so quickly!

chrabyrd:
resolved in #8874

PR #8874, opened 2022-08-25
Updates branch append logic and graph clone logic to handle serialized graphs re #8842

chrabyrd:
### Types of changes

### Description of Change

In plain english the changes are:
- `Node`, `Edge` and `Card` models have have their init methods updated to force their ids to be UUID if they're strings
- `branch-list.htm` has been updated to replace `active` with `published` 
- `models.Graph` 's `append_branch` has been updated to return `self` instead of `branch_copy`, with `self` referring to the graph the branch is added to
- `models.Graph`'s `serialize` method has been updated to force UUIDs in top-level dictionary values. Also the process for stocking `self.widgets` has been reverted to not leverage the serialized graph
- `branch-list.js` and `graph-designer.js` have been updated to refresh the page on branch append, with a TODO in the code to revert

### Issues Solved

#8842 #8724


robgaston (approved):
looks good and functions as expected. thanks

PR #8880, opened 2022-08-26
fixes edit log for nested tiles

### LINKED ISSUE #8879: Edit log fails for resources with tiles nested 2 levels or deeper
Edit log javascript only handles for one degree of child tiles currently, resulting in errors.  Existing logic should be made recursive to handle for any depth.

robgaston:
makes edit log card handling recursive to allow for any depth of tiles

fixes #8879


chiatt (approved):
Looks good!

Issue #8831, opened 2022-08-17
dev/7.0.x Bug Toggling Language back from Arabic to English


ekansa:
I'm running dev/7.0.x Arches core with the Arches HER project in order to test some internationalization features. I haven't installed Hebrew or Arabic .po files, but I'm trying to do some tests that don't depend on those.

In Arches Designer, to edit a branch, I could initially toggle from English to Arabic. That successfully changed the user interface into the desired/expected configuration for a right-to-left written language. However, the language selection options to go back from Arabic to English does not work. The browser reloads (with the spiny GIF), but it reloads back as "Arabic" (again no .po files) rather than the expected English. 

I'm using a the Brave (based on Chrome) browser in Windows 11. Screenshot is attached.

PS. The problem of not being able to toggle out of Arabic and back to English persists when I exit the Arches Designer.
PPS. I also can switch to my third language option, Hebrew, after I selected Arabic. 
PPPS. If I'm in Hebrew and attempt to go back to English (or Arabic) I am similarly stuck in Hebrew.

![arches-ui-bug-internationalization-lang-toggle](https://user-images.githubusercontent.com/1286742/185255071-d4bdf127-6421-405f-9e98-419b2e055a68.png)

njkim:
Could not reproduce. It looks like it is fixed with most recent updates... need to be confirmed.

ekansa:
I just updated to my Arches core (`HEAD is now at 3fd8230f8`). The problem persists. I can switch from English to another language, but once I'm in another language, I can't switch again. The page just reloads and the language state did not change (URL stays the same too).

ekansa:
I'm now at `HEAD is now at 7328756d7 Merge pull request #8875 from archesproject/8861_getting_cardwidget_for_editor`. The problem is persisting for me. I need to manually edit the URL to go back to English after I made a switch to another language.

njkim:
@ekansa It looks like the language switcher does work properly when the language code is in the url (i18n url patterns).

The i18n url pattern can be temporarily disable until the permanent solution is implemented by commenting out the following code block in the urls.py in the project directory.

```
# if settings.SHOW_LANGUAGE_SWITCH is True:
#     urlpatterns = i18n_patterns(*urlpatterns)
```

apeters:
closed by #8904

PR #8904, opened 2022-08-30
Comment out the i18n url patterns, #8831

njkim:
### Types of changes

### Description of Change

Disable i18n url patterns (which includes the language code in the url) temporarily to fix the issue with language switcher. #8831
It is desirable to make both the language switcher and i18n url working in the future.

### Issues Solved

#8831

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


apeters (approved):
looks good to me

apeters (approved):
comments our the logic to add the language code to the url as this seems to only intermittently work.

PR #8931, opened 2022-08-31
Update landing page, re #8923

chiatt:


chrabyrd (approved):
lgtm! -- will test as part of release process

Issue #8935, opened 2022-08-31
iiifannotations API not working


aarongundel:
**Describe the bug**

There's a problem with the iiifannotations API endpoint in 6.2.  

**To Reproduce**
Procedure 1 (with AFS)
1. Using the Arches for Science Project
2. Go through the upload dataset workflow (this does happen in other workflows - but this one is easy to see - also assuming you have run the appropriate prerequisite workflows) 
3. On the Data files and location step, you'll see breakage - see screenshot below.

Procedure 2 (without AFS)
1. Hit the iiifannotations API with the appropriate/necessary parameters.  
2. The same breakage should occur

**Screenshots**

![Screen Shot 2022-08-31 at 4 36 17 PM](https://user-images.githubusercontent.com/124614/187797019-8a70ec79-f31b-47ca-8206-bb68c7d3a6a8.png)

aarongundel:
@chiatt I think there's a problem here.  This originally worked for me, then stopped working.  I think it's because this view gets overwritten by 7131 - so the change made in migration 6837 needs to be made in 7131 as well.

chiatt:
Fixed in #8962

PR #8947, opened 2022-09-02
Cast feature column in annotation view as jsonb to support django 3.2, re #8935

chiatt:
### Types of changes

### Description of Change
Cast feature column in annotation view as jsonb to support django 3.2

### Issues Solved

#8935


aarongundel (approved):
works for me - thanks!

PR #8949, opened 2022-09-02
Set the thumbnail image size for the small images

njkim:
### Types of changes

### Description of Change

Set thumbnail image size to the image size if the image is smaller than the default thumbnail size (300 x 300), archesproject/arches-for-science#943

Fix the issue of the thumbnail image not loading when the image itself is small.
https://iiif.io/api/image/2.0/#size

### Issues Solved

archesproject/arches-for-science#943

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chiatt (approved):
Looks great and works as expected

Issue #8963, opened 2022-09-02
ETL Manager fails when trying to load a csv with "can't start background process" error


dwuthrich:
Tried to load a simple 3 column csv attached ![Screen Shot 2022-09-02 at 2.57.17 PM.png](https://images.zenhubusercontent.com/56f1ad0a6caf56c9d38e1cbd/f11225ef-5591-41a7-9308-fe892ed97d62)into physical thing model and got a background error process as shown in the following screenshot

chiatt:
fixed in #8964 and #8972

PR #8964, opened 2022-09-08
update error message when Celery isn't running, re #8963

apeters:
re #8963


apeters:
this also needs to be cherry picked into dev/6.2.0

chiatt (approved):

Issue #8965, opened 2022-09-09
Warn admin that related resource exists in db only when 'strict' is used


chiatt:
During the resource instance datatype validation, we check if the related resource instance exists in the database. For even a modest load this can generate a very large number of useless database queries and warnings. Load performance would be improved if we did this check only when the 'strict' parameter is passed. 

Additionally there may be unnecessary database queries in the post_tile_save method of the ResourceInstanceDatatype that can be removed.

apeters:
closed by #8971

PR #8971, opened 2022-09-09
Remove related resource warning unless strict validation is used

chiatt:
…sed, re #8965

### Types of changes

### Description of Change
Removes warning in the related resources validation if a resource instance's related resource does not already exist in the database. This validation bogs down data import with postgres and ES queries and can flood the terminal output with warnings that are not useful. In cases where this validation is needed, the 'strict' flag can be passed to the validation function.

### Issues Solved

#8965


apeters (approved):
fixes that warning

Issue #8974, opened 2022-09-13
Management of resource_x_resource instances causes a major slowdown during resource import


apeters:
Using the Physical Thing resource type in Arches for Science, it takes around 32 seconds to import just the first 2 records.
By commenting out the `post_tile_save` method reduces that to just 2 seconds.  
It looks like the Postgres function `__arches_create_resource_x_resource_relationships` is the issue.

PR #8975, opened 2022-09-13
fix sql error in __arches_create_resource_x_resource_relationships, re #8974

apeters:
re #8974


chiatt (approved):
Looks good and significantly improves performance

Issue #8983, opened 2022-09-15
Alias can change upon reloading a graph


chiatt:
If a graph has two nodes with the same name and no custom alias, the alias can change on subsequent graph loads depending on the order in which the nodes are loaded. For example if a graph has two nodes with the name 'foo' then their aliases would be 'foo' and 'foo_n1'. If the graphs are reloaded and 'foo_n1' is imported before 'foo', then the alias will get recalculated to be 'foo' and when the node that previously had the 'foo' alias loads, its alias will be recalculated to be 'foo'.

PR #8984, opened 2022-09-15
Do not overwrite alias on import if it already exists, re #8983

chiatt:
### Types of changes

### Description of Change
Prevent alias from getting recalculated on graph import

### Issues Solved
#8983


njkim (approved):
Looks good

PR #9020, opened 2022-10-11
protects concept.py get_e55_domain from sql injection #9019

chrabyrd:
### Types of changes

### Description of Change

Solution based from: https://docs.djangoproject.com/en/4.1/topics/db/sql/#executing-custom-sql-directly

### Issues Solved

#9019


apeters (approved):
works as expected

Issue #9026, opened 2022-10-12
arches-core proxy for `Time-based blind SQL injection in get_child_edges (GHSL-2022-071) #5`


chrabyrd:

apeters:
closed by #9027

PR #9027, opened 2022-10-12
fixes get_child_edges sql injection #9026

chrabyrd:
### Types of changes

### Description of Change

fixes sql injection in concept model `get_child_edges` method

### Issues Solved

#9026


apeters (changes requested):
this will still fail because of the string formatting done with the "subquery".

@chrabyrd ask my about a url to test this.

Looking at this more closely, it looks like we can incorporate the "subquery" into the full "sql" string variable.  I don't think we need it.  Then we can just pass the "query" concatenated with leading and trailing "%%" as a param to cursor.execute.

chrabyrd:
@apeters yeah we should chat. Re-interpolation failed for me for queries without a subquery

chrabyrd:
I re-ran the results from my branch and status quo through a diff checker: 

the query `http://localhost:8000/concepts/paged_dropdown?conceptid=e480165f-44d2-4ec4-b13a-f246270f06af` results in: https://www.diffchecker.com/bSunlEME

the query `http://localhost:8000/concepts/paged_dropdown?conceptid=e480165f-44d2-4ec4-b13a-f246270f06af&query=art` results in: https://www.diffchecker.com/UJX5fUY4

apeters (approved):
works great!

PR #8967, opened 2022-09-09
Check bnodeid exists in tile data

probonas:
### Types of changes

-   [X] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
Checking that the _bnodeid_ exists in the tile's data before accessing it, allowed the loader to not fail with a _KeyError_ exception but rather raise an exception with an appropriate message that revealed the underlying cause of the error, as seen in the screenshot below.

arches-correct-message

Now fixing the cardinality in the model, allowed me to load the desired file.

arches-successful-load

### Issues Solved

#8966 
#

### Checklist

-   [X] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: Getty Conservation Institute
*   Found by: @probonas
*   Tested by: @probonas
*   Designed by: @probonas

PR #9057, opened 2022-10-27
Check bnodeid exists in tile data

probonas:
In response to the feedback from : https://github.com/archesproject/arches/pull/8967


apeters (approved):
works as expected

Issue #9078, opened 2022-11-01
The single csv importer often fails, when a concept value was included in the csv.


njkim:
with the error
```
...

File "/Users/njk/projects/sf-planning-7/arches/arches/app/datatypes/concept_types.py", line 52, in lookup_labelid_from_label
    result = self.lookup_label(value, collectionid)

UnboundLocalError: local variable 'collectionid' referenced before assignment
```

PR #9079, opened 2022-11-01
Fix concept import in the single csv importer, #9078

njkim:
### Types of changes

### Description of Change

Add the config to properly transform data for concept datatype, #9078 
Also check if the value is empty string before try to transform, if the value is "" turn it into None.
Fix the issue the single csv importer for the concept and resource instance datatype

### Issues Solved

#9078 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/etl_modules/import_single_csv.py
+++ b/arches/app/etl_modules/import_single_csv.py
@@ -280,12 +280,18 @@ class ImportSingleCsv(BaseImportModule):
                                 config = current_node.config
                                 config["path"] = temp_dir
                                 value = (
-                                    datatype_instance.transform_value_for_tile(source_value, **config) if source_value is not None else None
+                                    datatype_instance.transform_value_for_tile(source_value, **config) if source_value is not None and source_value != '' else None
                                 )
                                 errors = datatype_instance.validate(value, nodeid=node, path=temp_dir)
                             else:
-                                value = datatype_instance.transform_value_for_tile(source_value) if source_value is not None else None
-                                errors = datatype_instance.validate(value)
+                                try:
+                                    config = current_node.config
+                                except TypeError:
+                                    config = {}
+                                value = (
+                                    datatype_instance.transform_value_for_tile(source_value, **config) if source_value is not None and source_value != '' else None
+                                )
+                                errors = datatype_instance.validate(value, nodeid=node)
                             valid = True if len(errors) == 0 else False
                             error_message = ""
                             for error in errors:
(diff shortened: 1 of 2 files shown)


arches/app/etl_modules/import_single_csv.py
```diff
@@ -280,12 +280,22 @@ def populate_staging_table(self, loadid, graphid, has_headers, fieldnames, csv_f
                                 config = current_node.config
                                 config["path"] = temp_dir
                                 value = (
-                                    datatype_instance.transform_value_for_tile(source_value, **config) if source_value is not None else None
+                                    datatype_instance.transform_value_for_tile(source_value, **config)
+                                    if source_value is not None and source_value != ""
+                                    else None
```
chrabyrd, line 285: is it possible to change this to:

value = datatype_instance.transform_value_for_tile(source_value, **config) if source_value else None

? Or is there an explicit reason for the `None` and empty string comparison?

arches/app/etl_modules/import_single_csv.py
```diff
@@ -280,12 +280,22 @@ def populate_staging_table(self, loadid, graphid, has_headers, fieldnames, csv_f
...
                                 )
                                 errors = datatype_instance.validate(value, nodeid=node, path=temp_dir)
                             else:
-                                value = datatype_instance.transform_value_for_tile(source_value) if source_value is not None else None
-                                errors = datatype_instance.validate(value)
+                                try:
+                                    config = current_node.config
+                                except TypeError:
+                                    config = {}
```
chrabyrd, line 292: why the try/except here? is it possible to check for the current_node config a different way?

arches/app/etl_modules/import_single_csv.py
```diff
@@ -280,12 +280,22 @@ def populate_staging_table(self, loadid, graphid, has_headers, fieldnames, csv_f
...
+                                value = (
+                                    datatype_instance.transform_value_for_tile(source_value, **config)
+                                    if source_value is not None and source_value != ""
+                                    else None
+                                )
```
chrabyrd, line 297: same note here as the previous value assignment

chrabyrd (changes requested):
looks good, just a few questions about format

arches/app/etl_modules/import_single_csv.py
njkim: good point! thanks.

arches/app/etl_modules/import_single_csv.py
njkim: Yes you are right. There should always be the config. So, remove try and move the line to above combine with file type. Thanks.

chrabyrd (approved):
lgtm!

chrabyrd:
resolved in #9079

PR #9110, opened 2022-11-28
lang delete bug

apeters:
re #9105

### DIFF
--- a/arches/app/models/graph.py
+++ b/arches/app/models/graph.py
@@ -1420,12 +1420,14 @@ class Graph(models.GraphModel):
                 return serialized_graph
             except:
                 self.refresh_from_database()
-                return self.serialize(fields=fields, exclude=exclude, force_recalculation=True, use_raw_i18n_json=use_raw_i18n_json, **kwargs)
+                return self.serialize(
+                    fields=fields, exclude=exclude, force_recalculation=True, use_raw_i18n_json=use_raw_i18n_json, **kwargs
+                )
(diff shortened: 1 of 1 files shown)


arches/app/models/graph.py
```diff
@@ -1409,18 +1409,21 @@ def serialize(self, fields=None, exclude=None, force_recalculation=False, use_ra
...
+                for key in exclude:
+                    if serialized_graph.get(key) is not None:  # explicit None comparison so falsey values will still return
+                        serialized_graph[key] = None
+
+                return serialized_graph
+            except:
+                self.refresh_from_database()
+                return self.serialize(
+                    fields=fields, exclude=exclude, force_recalculation=True, use_raw_i18n_json=use_raw_i18n_json, **kwargs
+                )
```
chrabyrd, line 1425: If `self.refresh_from_database()` fails, this will enter an infinite loop

apeters: Because I set `force_recalculation=True`, it should never enter the `if` part of that method and so, shouldn't ever get into an infinite loop scenario.  I'm not sure what would happen if it doesn't successfully refresh from the db.  Didn't test that.  Probably nothing good....  :(
Let me know if I'm missing something.

chrabyrd: Ya fair actually that makes sense 👍

chrabyrd (approved):
lgtm

PR #9126, opened 2022-12-06
create references to source record

### LINKED ISSUE #9116: [ editable_future_graphs ] Ensure that any models created during the `Graph.copy()` process contain a reference to their source
When working on this issue, please ensure your branch is based on `9053-cbyrd-create-future-publications` https://github.com/archesproject/arches/pull/9112

Part of the logic enabling the editing of a published graph requires creating a copy of the current Graph and having the user apply edits to it. It would be confusing for the end-user if the identifiers like UUIDs changed between the Graph and the copy of the Graph. We therefore need to add a column to attrs to some models that point a copied instance to its source so the frontend can show the correct identifier. Note that this behavior must be parameterized to ensure the original use of the `Graph.copy` method remains unchanged.

Models that should be investigated, but unsure if all models listed below will be affected:
- node
- nodegroup ( probably not nodegroup references node)
- edge
- card
- widget

aarongundel:
Add columns to reference source records for cards, nodes, and edges.   Resolves #9116 

### Types of changes

### DIFF
--- a/arches/app/models/graph.py
+++ b/arches/app/models/graph.py
@@ -776,7 +776,7 @@ class Graph(models.GraphModel):
             function_copy = models.FunctionXGraph(function=function_x_graph.function, config=config_copy, graph=self)
             function_copy.save()
 
-    def copy(self, root=None):
+    def copy(self, root=None, set_source=False):
         """
         returns an unsaved copy of self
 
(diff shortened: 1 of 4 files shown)


chrabyrd:
lgtm!

PR #9128, opened 2022-12-07
don't add child tiles that don't point to parent tiles during jsonld import

apeters:
re #8800 

This is cherry picked from @aarongundel work done here: https://github.com/archesproject/arches/pull/8801


ryan86:
lgtm

Issue #9131, opened 2022-12-08
Report: Related Resources section not getting populated


ryan86:
The related resources section of reports appears to not be getting populated when a resource instances has other resource instances related to it and those graphs are published.

chiatt:
Resolved in #9132

PR #9132, opened 2022-12-08
Remove call to exclude related resources, re #9131

ryan86:
Removes param to exclude related resources on report load, re #9131


apeters (approved):
works as expected

Issue #9147, opened 2022-12-14
__arches_get_tile_cardinality_violations i18n bug


bferguso:
**Arches Version/Branch:** dev/7.2.x

**Describe the bug and how to reproduce it:**
The `__arches_get_tile_cardinality_violations` function lists any tile cardinality violations and breaks them down by graph name. The function wasn't updated as part of the v7 i18n updates so is still assuming the graph name is a string not i18n JSON value. Need to update the graph name reference to work with the i18n changes.

bferguso:
Issue resolved.

PR #9148, opened 2022-12-14
9147 - fix i18n bug in __arches_get_tile_cardinality_violations

bferguso:
### Types of changes

### Description of Change
Fix i18n bug in __arches_get_tile_cardinality_violations - was referring to graphs.name as text, not i18n JSON string.

### Issues Solved

#9147 

### Checklist

#### Ticket Background
*   Sponsored by: Self Funded
*   Found by: @bferguso
*   Tested by: @bferguso
*   Designed by: @bferguso

### Further comments
I've added a diff here - copied the new function to the reverse migration section to show the diff.:
https://github.com/bcgov/arches/compare/9147_tile_cardinality_function_bug...bcgov:arches:9147_diff?expand=1

### DIFF
--- a/arches/app/models/migrations/9147_get_tile_cardinality_violations_i18n.py
+++ b/arches/app/models/migrations/9147_get_tile_cardinality_violations_i18n.py
@@ -1,0 +1,71 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("models", "8974_rr_load_performance_v2"),
+    ]
+
+    sql_string = """
+create or replace function __arches_get_tile_cardinality_violations()
+    returns table(graph_name text, node_names text[], resourceinstanceid uuid, nodegroupid uuid, parenttileid uuid, tilecount bigint) as
+$$
+DECLARE
+BEGIN
+    return query with tile_violations as ( SELECT t.resourceinstanceid,
+                        t.nodegroupid,
+                        t.parenttileid parent_tileid,
+                        count(*)       tilecount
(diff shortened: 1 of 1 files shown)


apeters (changes requested):
Hard coding "en" as the language will break if that language doesn't exist.  Maybe the function should allow users to enter the language code as an argument to the function.  

It doesn't look this function isn't called anywhere, but instead it's left up to users to call it instead.  Is that right?

bferguso:
@apeters - I actually meant to leave a comment about this. I had originally added i18n config to this, but I was a bit confused about the difference in the graph.name i18n JSON structure and the structure everywhere else. Everywhere else the i18n JSON structure is something like 
`{'en': {'value': '[data]', 'direction': '[direction]'}, 'de': {}....}`
What I'm seeing in the graph.name column is just 
`{'en': '[en name]'}`. Wondering why the difference in the structure there, and wasn't sure what the intent was. I was originally pulling the language with `select code from languages where isdefault limit 1;`

This function is called by the `__arches_complete_bulk_load()` function which re-enables the tile validation triggers and checks to ensure there are no tile cardinality issues that were caused when they were disabled for data loading. It is only called by users directly if they use this mechanism for loading data via the views.

bferguso:
@apeters - Are you OK if we just use the default language instead of requiring the user to pass it in as a parameter? I've made a change for that if it's an acceptable option.

apeters (approved):
users will need to remember to make sure the default language in the languages table is in sync with the setting in settings.py

Issue #9165, opened 2023-01-06
Improve Bulk Data Loader error reporting


chiatt:
If data submitted to the bulk data loader has a large number of errors, the error report can be too large for a user to interpret. Instead, the bulk data loader should aggregate and summarize errors to help users understand how to fix their data.

PR #9172, opened 2023-01-12
Improve error reporting in bulk importer modules, #9165

njkim:
### Types of changes

### Description of Change

Improve the error reporting for the bulk data manager, single-csv-importer and branch-csv-importer, #9165 

### Issues Solved

#9165

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/views/etl_manager.py
```diff
@@ -19,15 +19,73 @@ class ETLManagerView(View):
     to get the ETL modules from db
     """
 
+    def dictfetchall(self, cursor):
```
chiatt, line 22: We should probably try to avoid using Python types in function names. Maybe this could be: `fetch_load_errors(self, cursor)`

arches/app/models/models.py
```diff
@@ -1638,6 +1638,21 @@ class Meta:
         db_table = "load_staging"
 
 
+class loadErrors(models.Model):
```
chiatt, line 1641: It's probably good to capitalize the class name here

arches/app/models/migrations/9165_error_reporting.py
```diff
@@ -0,0 +1,39 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("models", "9075_external_oauth"),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name="loadErrors",
```
chiatt, line 17: change class name to 'LoadErrors'

chiatt (changes requested):
This looks good. There are just a few syntax formatting changes that I think would be good to make.

arches/app/models/migrations/9165_error_reporting.py
njkim: The naming issues were fixed. Thanks!

chiatt (approved):

PR #9204, opened 2023-01-24
Upgrade slugify to version that supports Python version >= 3.7

chiatt:
Upgrade slugify to version that supports Python version >= 3.7 
Replaces PR #7455


chiatt:
Resolved in #9209

PR #9211, opened 2023-01-27
Improve resource instance error reporting

chiatt:
Ensure resource instance datatype reports errors for bulk loader, re #9165


njkim (approved):
Looks good!

PR #9212, opened 2023-01-27
9208 external widget observables

### LINKED ISSUE #9208: Text and Rich Text widgets don't respect external observables
The text and rich text widgets don't respect changes to external observables.  When the value of the external observable changes, the value of the widget is not updated.  This makes it possible to initialize the widgets, but not to modify their value after initialization programatically.  They have to be modified by hand.

aarongundel:
Fixes issue with mutating external observables for text and rich text widgets.

### Types of changes

### Description of Change

Fixes #9208

### Issues Solved

#9208

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/views/components/widgets/rich-text.js
+++ b/arches/app/media/js/views/components/widgets/rich-text.js
@@ -69,7 +69,7 @@ define([
             const currentLanguage = self.currentLanguage();
             if(!currentLanguage) { return; }
 
-            if(JSON.stringify(currentValue) != JSON.stringify(ko.toJS(self.value()))){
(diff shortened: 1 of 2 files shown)


chiatt (changes requested):
These changes resolve issues with the use of the string widget in complex workflows. There still seem to be minor issues with cancel/undo in the afs sample taking workflow, but that is something that probably needs to be resolved there, not in these widgets. 

The only change I think that would be good to make here is unwrapping value on line 72  in rich-text.js and line 131 in text.js

chiatt (approved):

Issue #9161, opened 2023-01-06
Allow concepts to define relationship types


chiatt:
Relationship types currently can only be defined in the resource instance widget using ontology properties. In addition to ontology properties users should be able to define relationships using concept values from concept collections. Given the significance and potential long term impact of this change, discussion is needed with core developers to determine how best to implement this both in the resource_x_resource table, as well as in the Resource Instance widget. re #6358, #8169

azaroth42:
Seems like for this to be interoperable with ontologies, and that future RDM will use models for concepts, the ontology would need to be reflected into the arches modeling directly? I can see a lot of benefits in doing this, including being able to manage new core relationships within the graphs from within the UI, rather than reloading an RDF/XML file on the command line. 
But it seems like a lot of work! Would appreciate being kept in the loop on this one, especially given #6358.

chiatt:
@azaroth42 This initial round of work simply allows the use of a concept valueid to describe a relationship. Unfortunately it doesn't yet associate concepts with ontologies, though I don't think it would hinder future work in that direction (e.g.#6358).

PR #9213, opened 2023-01-28
Implement the relationship concept, #9161

njkim:
### Types of changes

### Description of Change

Implement the relationship concept in resource instance data, #9161 
so that the relationship can be defined using concept.

### Issues Solved

#9161

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/media/js/views/components/datatypes/resource-instance.js
```diff
@@ -68,6 +68,9 @@ define([
                 });
                 graph.ontologyProperty = ko.observable(ko.unwrap(graph.ontologyProperty));
                 graph.inverseOntologyProperty = ko.observable(ko.unwrap(graph.inverseOntologyProperty));
+                graph.relationshipConcept = ko.observable(ko.unwrap(graph.relationshipConcept) || 'ac41d9be-79db-4256-b368-2f4559cfbe55');
```
chiatt, line 71: I think it would be nice to declare default relationship uuid up at the top and use the variable here so that it is clear to developers that `ac41d9be-79db-4256-b368-2f4559cfbe55` is the `is related to` concept value.

chiatt (changes requested):
This is looking really good @njkim. I did notice that when I create a collection value with more than one language label and switch between languages in the UI, the language change is not reflected in the relationship. You can see in the images below that the relationship shown in the visualization should be `ist ähnlich` as it is represented in the report, but it is still `is similar to`.

image

image

chiatt (approved):

Issue #9232, opened 2023-02-09
Prevent unnecessary descriptor queries


chiatt:
As of https://github.com/archesproject/arches/commit/55490cec73cfacadd3407170ebce002b527fec72 descriptors are now stored in the resource instance table. This means, in cases when the descriptor is already on a resource, we no longer need to query the functions table to get the primary descriptor function to get the primary descriptor module in order to call `get_primary_descriptor_from_nodes`. Bypassing `get_primary_descriptor_from_nodes` could mean a significant reduction in queries wherever descriptors are used in Arches.

chiatt:
Fixed in #9232

PR #9234, opened 2023-02-09
Use descriptor from resource instance if available, re #9232

chiatt:
### Description of Change
Reduces queries to the function table get primary descriptors.

### Issues Solved

#9232

### DIFF
--- a/arches/app/models/resource.py
+++ b/arches/app/models/resource.py
@@ -74,6 +74,18 @@ class Resource(models.ResourceInstance):
         self.tiles = []
 
     def get_descriptor(self, descriptor, context):
+
+        requested_language = None
+        if context and "language" in context:
+            requested_language = context["language"]
+        language = requested_language or get_language()
+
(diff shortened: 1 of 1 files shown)


arches/app/models/resource.py
```diff
@@ -74,6 +74,18 @@ def __init__(self, *args, **kwargs):
         self.tiles = []
 
     def get_descriptor(self, descriptor, context):
+
+        requested_language = None
+        if context and "language" in context:
+            requested_language = context["language"]
+        language = requested_language or get_language()
+
+        if self.name and self.descriptors:
```
apeters, line 83: I'm not sure if this check is really necessary.  I might just use the try/except block exclusively.  Try to get the value from the model field, if it throws for whatever reason, then recalculate from the function.

chiatt: Ah - good point. Just pushed that change.

aarongundel (approved):
lgtm

PR #9238, opened 2023-02-13
Set descriptors only on index rather than on every descriptor request

chiatt:
### Types of changes

### Issues Solved
re #9232


chiatt:
Sadly it was discovered that this change could cause custom primary descriptor functions to break.

Issue #9251, opened 2023-02-17
Unable to add new users if instance perms applied to one or more instances


chiatt:
If one or more instances is restricted using instance level permissions, new users cannot be added to the system due to the following error:

![image](https://user-images.githubusercontent.com/1877663/219725334-c248b6c4-21dc-4cc5-a21f-7896193754d0.png)

chiatt:
FIxed in #9252

PR #9252, opened 2023-02-17
Fix error preventing new users when resource instance permissions are used

chiatt:
### Types of changes

### Description of Change
Ensure resource has all required properties before modifying permissions and indexing. Fixes error when instances are indexed upon the creation of a new user.

### Issues Solved

#9251


chiatt:
Reviewed and tested by @dwuthrich

PR #9259, opened 2023-02-21
74 performance enhancements

aarongundel:
### Types of changes

### Description of Change

Various performance enhancements

### Issues Solved

#

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/models/resource.py
```diff
@@ -215,9 +232,13 @@ def get_root_ontology(self):
 
         """
         root_ontology_class = None
-        graph_nodes = models.Node.objects.filter(graph_id=self.graph_id).filter(istopnode=True)
-        if len(graph_nodes) > 0:
-            root_ontology_class = graph_nodes[0].ontologyclass
+        try:
+            graph_node = SimpleNamespace(**next((x for x in self.serialized_graph["nodes"] if x["istopnode"] == True), None))
```
chiatt, line 236: Hey @aarongundel, could you change `==` to `is` appease Stickler?

aarongundel: old habits!

arches/app/models/resource.py
```diff
@@ -238,6 +259,46 @@ def get_flattened_tiles(self):
             tiles.extend(tile.get_flattened_tiles())
         return tiles
 
+    @staticmethod
+    def bulk_update(resources, transaction_id=None):
```
chiatt, line 263: @aarongundel, I don't see bulk_update called anywhere

apeters: This method seems an almost duplicate of the "bulk_save" method.  This one, though, isn't calling a "bulk" django method to update tiles and instead calls "update_or_create".  There is a django provided "bulk_update" method that could be used here although in reality it seems this whole method could just be subsumed by our "bulk_save" method with some creative restructuring.

apeters:
@aarongundel What's the best way to test all of this?  Is there some specific testing workflow that you've been using?

arches/app/models/resource.py
```diff
@@ -292,13 +353,13 @@ def bulk_save(resources, transaction_id=None):
         se.bulk_index(documents)
         se.bulk_index(term_list)
 
-    def index(self, context=None):
+    def index(self, context=None, df=None):
```
apeters, line 356: I don't think we need to pass in the DatatTypeFactory.  It is in fact a singleton, so we should just be able to instantiate it w/o penalty.

Also, at the end of this method I see a call to `save` method on the Resource super class.  That doesn't seem right to me.

arches/app/models/resource.py
```diff
@@ -73,6 +75,14 @@ def __init__(self, *args, **kwargs):
         # end from models.ResourceInstance
         self.tiles = []
         self.descriptor_function = None
+        try:
+            self.serialized_graph = (
+                models.PublishedGraph.objects.filter(publication=self.graph.publication.publicationid, language=settings.LANGUAGE_CODE)
+                .first()
+                .serialized_graph
+            )
+        except:
```
chiatt, line 84: I think it would just an AttributeError that would be expected here in the event that no published graph was returned or if self did not have a `serialized_graph` property

arches/app/datatypes/datatypes.py
```diff
@@ -2235,6 +2235,22 @@ def validate(self, value, row_number=None, source="", node=None, nodeid=None, st
...
+            node = models.Node.objects.get(pk=nodeid)
+
+        if node.datatype not in ["resource-instance-list", "resource-instance"]:
+            return
+
+        relationships = tile.data[nodeid]
+        try:
+            for relationship in relationships:
+                relationship["resourceXresourceId"] = str(uuid.uuid4())
+        except:
```
chiatt, line 2251: As we discussed, maybe it would be good to remove this exception and let any unexpected errors get logged.

arches/app/models/resource.py
chiatt: Yeah, I think it would be good to either move the code into the project where it is used, or add the update logic to the bulk_save.

arches/app/models/resource.py
aarongundel: I think that's correct, and I'm fine with making this change - but there's also the chance that the resource. has no graph at initialization and the graph is assigned later - I have seen this a few places in the code.  I think this'll also trigger an AttributeError, but I am not as knowledgeable about the ORM exceptions.

arches/app/models/resource.py
aarongundel: fair enough re datatypefactory - I've removed that.  re: the save method - that is part of a separate commit that was made some time (9+ months) ago, and I'm reluctant to remove it as part of this PR and without more testing about why it was added in the first place.

aarongundel:
> 

The best way to test this is with the Arches for Science create project workflow - the last step where physical things are added to the project should be markedly faster.  Anything that saves tiles should be faster though, so any afs workflow should be faster.

arches/app/models/resource.py
chiatt: Yeah - removing save from the index method will be a bit of work. I've created a ticket for it: https://github.com/archesproject/arches/issues/9278

chiatt (approved):
This looks good to me. There are definitely some good performance gains - particularly in the `Add things to project` step in AfS. Adding a large number of things to a project took 28 seconds. After this change (and the addition of bulk_update in AfS), it takes only 2.6 seconds! Regular tile saves in the editor are 25% faster with 25% fewer db queries. Really nice. 

image

arches/app/datatypes/datatypes.py
```diff
@@ -2235,6 +2235,20 @@ def validate(self, value, row_number=None, source="", node=None, nodeid=None, st
                     errors.append({"type": error_type, "message": message})
         return errors
 
+    def pre_tile_save(self, tile, nodeid, **kwargs):
+        node = kwargs.pop("node", None)
```
apeters, line 2239: I'm pretty sure that we shouldn't need lines 2239-2245 as no other type of node, besides "resource-instance", and "resource-instance-list",  should trigger this method.

If that is the case, I'm not sure if adding the **kwargs param is necessary any more.

arches/app/models/tile.py
```diff
@@ -284,10 +295,18 @@ def check_for_missing_nodes(self):
...
-                node = models.Node.objects.get(nodeid=nodeid)
+                try:
+                    node = SimpleNamespace(**next((x for x in self.serialized_graph["nodes"] if x["nodeid"] == nodeid), None))
+                except:
+                    node = models.Node.objects.get(nodeid=nodeid)
                 datatype = self.datatype_factory.get_instance(node.datatype)
                 datatype.clean(self, nodeid)
                 if self.data[nodeid] is None and node.isrequired is True:
+                    try:
+                        node.cardxnodexwidget_set.all()
```
apeters, line 306: I think we could store off the result from the call to `node.cardxnodexwidget_set.all()` and use it on lines 310 and 311.

arches/app/models/resource.py
```diff
@@ -307,24 +322,29 @@ def index(self, context=None):
...
+            }
             document, terms = self.get_documents_to_index(datatype_factory=datatype_factory, node_datatypes=node_datatypes, context=context)
             document["root_ontology_class"] = self.get_root_ontology()
             doc = JSONSerializer().serializeToPython(document)
             se.index_data(index=RESOURCES_INDEX, body=doc, id=self.pk)
             for term in terms:
                 se.index_data("terms", body=term["_source"], id=term["_id"])
 
-            celery_worker_running = task_management.check_if_celery_available()
+            if len(settings.ELASTICSEARCH_CUSTOM_INDEXES) > 0:
```
chiatt, line 338: Given that the performance cost is significant, maybe it makes sense to also check if any indexes have the "should_update_asynchronously" property?
```
indexes = settings.ELASTICSEARCH_CUSTOM_INDEXES
use_async = any([idx.get('should_update_asynchronously') for idx in indexes])
if indexes and use_async:
    celery_worker_running = task_management.check_if_celery_available()
```

apeters: That would work as well, but honestly, I think we should be caching this value like we do in the DatatypeFactory because checking this every time a resource uses a custom index is still overkill especially when considering bulk upload/indexing. 
That would probably be better to handle in another ticket IMO.

Issue #9264, opened 2023-02-22
Attempting to clone a graph results in IntegrityError


chrabyrd:
Attempting to clone a graph via the `Clone graph` option under the `Manage` dropdown ( in both graph designer index page and graph designer page ) results in the following:

django.db.utils.IntegrityError: insert or update on table "nodes" violates foreign key constraint "nodes_nodegroupid_30e98973_fk_node_groups_nodegroupid"
DETAIL:  Key (nodegroupid)=(68ef6030-b2ed-11ed-99a6-3e22fbbb7736) is not present in table "node_groups".

This is happening in every version from 7.2

chrabyrd:
resolved in #9275

PR #9275, opened 2023-02-23
Fixes clone graph interaction #9264

chrabyrd:
### Types of changes

### Description of Change

- Fixes clone graph interaction by ensuring nodegroups are saved before nodes.
- Removes superfluous `graph.save()` call in `graph.delete()` method that was throwing error when deleting cloned graphs
- updates `GraphSettingsView.get()` method to correctly fail as intended in when originally written

 If cloning a graph from the graph-designer view ( `/graph-designer/` ), it should now correctly open a new window with the cloned graph instead of the original graph. JSONResponse was serializing a published graph and without the `force_recalculation` flag it was using the cached published version. 

### Issues Solved

#9264


arches/app/models/graph.py
```diff
@@ -485,6 +485,9 @@ def save(self, validate=True, nodeid=None):
                 self.update_es_node_mapping(node, datatype_factory, se)
                 self.create_node_alias(node)
                 try:
+                    if node.nodegroup:
+                        node.nodegroup.save()
```
robgaston, line 489: It seems like this call (and the same at ln 506) would result in multiple SQL statements to save the same nodegroup (ie it would save a nodegroup n times where n is the number of nodes in that group).  Can we adjust this logic so that it isn't adding this unnecessary SQL overhead?

robgaston (changes requested):
looks good with the one item for follow up -- seems like we can tighten that up a bit

robgaston (approved):
looks good - thanks!

PR #9286, opened 2023-03-01
Update webpack for Windows support

jkemper-pca:
Fix errors from trying to install Arches v7 in a Windows environment due to system incompatibility with the current webpack configuration.
### Types of changes

### Description of Change

At Parks Canada, we are currently only supporting Windows environments (no Docker either currently) so getting v7 of Arches working natively in a Windows environment is a big focus for us as the new localization support in v7 is mandatory for us to use Arches. These changes are my attempt at solving the issues that exist so far with getting v7 of Arches to run on a Windows environment in the hopes that it will be useful in getting Windows support added soon.
### Issues Solved

- Errors with yarn commands due to Windows not supporting setting variables like `NODE_PATH` and `NODE_OPTIONS`
  - Solved using the [`cross-env` package](https://www.npmjs.com/package/cross-env)
- Errors with path resolution when resolving the template locations on Windows
  - Solved by replacing the backslashes on a Windows environment with forward slashes
- Slower machines (Windows appears slower than Linux at this but that might just be my experience) can timeout with yarn is retrieving all the templates from Arche due to too many connections open at once (might be worse with the 302 redirects that happen with multiple languages enabled). Because this is something that only runs once for a deployment it seemed like the easiest solution would be to simply slow down the requests which has solved the issue in my testing even on the slowest VM available to test with.
#

### Checklist

Unable to run the local tests using [the command in the documentation](https://arches.readthedocs.io/en/stable/creating-a-development-environment/#running-tests) because the test throws an error on python 3.10+. Testing was done by running and using the application.
#### Ticket Background
*   Found by: @ 
*   Tested by: @jkemper-pca  
*   Designed by: @jkemper-pca  

### Further comments

The goal for me when making this changes were to be as non-invasive as possible with the existing functionality because I didn't want to break any existing functions with this change. Both the use of `cross-env` and the `replace()` function used are the solutions found after a wide search of online resources of others dealing with similar issues. 
I hope that these changes are worth using or at least will give some hints as to what is currently broken and ideas for fixing it.

Feel free to reach out to me about anything I did for this PR!

### DIFF
--- a/arches/app/templates/base.htm
+++ b/arches/app/templates/base.htm
@@ -81,7 +81,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
     {% include 'javascript.htm' %}
     
     {% if main_script %}
-        <script src="{% webpack_static '' %}/build/js/{{main_script}}.js"></script>
+        <script src="{% webpack_static '' %}build/js/{{main_script}}.js"></script>
     {% endif %}
 
     {% if app_settings.GOOGLE_ANALYTICS_TRACKING_ID != None %}
--- a/arches/app/templates/base.htm
+++ b/arches/app/templates/base.htm
@@ -81,7 +81,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
     {% include 'javascript.htm' %}
     
     {% if main_script %}
-        <script src="{% webpack_static '' %}build/js/{{main_script}}.js"></script>
+        <script src="{% webpack_static '' %}/build/js/{{main_script}}.js"></script>
     {% endif %}
 
     {% if app_settings.GOOGLE_ANALYTICS_TRACKING_ID != None %}
--- a/arches/install/arches-templates/project_name/webpack/webpack-utils/build-image-filepath-lookup.js
+++ b/arches/install/arches-templates/project_name/webpack/webpack-utils/build-image-filepath-lookup.js
@@ -22,7 +22,7 @@ const _buildImageFilePathLookup = function(publicPath, path, outerAcc, imageDire
 
             return { 
                 ...acc, 
-                [Path.join(publicPath, 'img', filename)]: Path.resolve(__dirname, Path.join(outerPath, subPath))
+                [Path.join(publicPath, 'img', filename).replace(/\\/g, '/')]: Path.resolve(__dirname, Path.join(outerPath, subPath))
             };
         }
     }, outerAcc);
--- a/arches/install/arches-templates/project_name/webpack/webpack-utils/build-javascript-filepath-lookup.js
+++ b/arches/install/arches-templates/project_name/webpack/webpack-utils/build-javascript-filepath-lookup.js
@@ -25,7 +25,7 @@ function buildJavascriptFilepathLookup(path, outerAcc, javascriptDirectoryPath) 
 
             return { 
                 ...acc, 
-                [pathName]: { 'import': Path.join(outerPath, subPath), 'filename': Path.join('js', '[name].js') } 
+                [pathName.replace(/\\/g, '/')]: { 'import': Path.join(outerPath, subPath), 'filename': Path.join('js', '[name].js') }
             };
         }
     }, outerAcc);
--- a/arches/install/arches-templates/project_name/webpack/webpack-utils/build-template-filepath-lookup.js
+++ b/arches/install/arches-templates/project_name/webpack/webpack-utils/build-template-filepath-lookup.js
@@ -21,7 +21,7 @@ const _buildTemplateFilePathLookup = function(path, outerAcc, templateDirectoryP
 
             return { 
                 ...acc, 
-                [Path.join('templates', filename)]: Path.resolve(__dirname, Path.join(outerPath, subPath))
+                [Path.join('templates', filename).replace(/\\/g, '/')]: Path.resolve(__dirname, Path.join(outerPath, subPath))
             };
         }
     }, outerAcc);
--- a/arches/install/arches-templates/project_name/webpack/webpack.common.js
+++ b/arches/install/arches-templates/project_name/webpack/webpack.common.js
@@ -93,8 +93,8 @@ module.exports = () => {
                 plugins: [
                     new CleanWebpackPlugin(),
                     new webpack.DefinePlugin({
-                        ARCHES_CORE_DIRECTORY: `'${ROOT_DIR}'`,
-                        APP_ROOT_DIRECTORY: `'${APP_ROOT}'`
+                        ARCHES_CORE_DIRECTORY: `'${ROOT_DIR.replace(/\\/g ,'/')}'`,
+                        APP_ROOT_DIRECTORY: `'${APP_ROOT.replace(/\\/g ,'/')}'`
                     }),
                     new webpack.ProvidePlugin({
                         jquery:  Path.resolve(__dirname, APP_ROOT, 'media', 'node_modules', 'jquery', 'dist', 'jquery.min'),
@@ -189,6 +189,9 @@ module.exports = () => {
                                                 let serverAddress = ARCHES_NAMESPACE_FOR_DATA_EXPORT;
                                                 if (serverAddress.charAt(serverAddress.length - 1) === '/') {
                                                     serverAddress = serverAddress.slice(0, -1)
+                                                }
+                                                if(process.platform === 'win32'){
+                                                    serverAddress = serverAddress.replace('localhost', '127.0.0.1')
                                                 }
                                                 resp = await fetch(serverAddress + templatePath);
                                             }
--- a/arches/install/arches-templates/project_name/webpack/webpack.common.js
+++ b/arches/install/arches-templates/project_name/webpack/webpack.common.js
@@ -201,7 +201,7 @@ module.exports = () => {
                                                     '\x1b[33m%s\x1b[0m',  // yellow
                                                     `"${templatePath}" has failed to load. Retrying (${failureCount} / 5)...`
                                                 );
-                                                await new Promise(r => setTimeout(r, 2000));
+                                                await new Promise(r => setTimeout(r, 8000));
                                                 return await renderTemplate(failureCount=failureCount);
                                             }
                                         }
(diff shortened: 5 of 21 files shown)


arches/app/templates/base.htm
```diff
@@ -81,7 +81,7 @@
     {% include 'javascript.htm' %}
     
     {% if main_script %}
-        <script src="{% webpack_static '' %}/build/js/{{main_script}}.js"></script>
```
chrabyrd, line 84: Can you remove this change in favor of changing it on a project-by-project basis? There are several installations that would break if this change got commited.

arches/install/arches-templates/project_name/webpack/webpack-utils/build-image-filepath-lookup.js
```diff
@@ -22,7 +22,7 @@ const _buildImageFilePathLookup = function(publicPath, path, outerAcc, imageDire
 
             return { 
                 ...acc, 
-                [Path.join(publicPath, 'img', filename)]: Path.resolve(__dirname, Path.join(outerPath, subPath))
+                [Path.join(publicPath, 'img', filename).replace(/\\/g, '/')]: Path.resolve(__dirname, Path.join(outerPath, subPath))
```
chrabyrd, line 25: wouldn't this change break non-windows installations?

arches/install/arches-templates/project_name/webpack/webpack-utils/build-javascript-filepath-lookup.js
```diff
@@ -25,7 +25,7 @@ function buildJavascriptFilepathLookup(path, outerAcc, javascriptDirectoryPath)
 
             return { 
                 ...acc, 
-                [pathName]: { 'import': Path.join(outerPath, subPath), 'filename': Path.join('js', '[name].js') } 
+                [pathName.replace(/\\/g, '/')]: { 'import': Path.join(outerPath, subPath), 'filename': Path.join('js', '[name].js') }
```
chrabyrd, line 28: wouldn't this change break non-windows installations?

arches/install/arches-templates/project_name/webpack/webpack-utils/build-template-filepath-lookup.js
```diff
@@ -21,7 +21,7 @@ const _buildTemplateFilePathLookup = function(path, outerAcc, templateDirectoryP
 
             return { 
                 ...acc, 
-                [Path.join('templates', filename)]: Path.resolve(__dirname, Path.join(outerPath, subPath))
+                [Path.join('templates', filename).replace(/\\/g, '/')]: Path.resolve(__dirname, Path.join(outerPath, subPath))
```
chrabyrd, line 24: wouldn't this change break non-windows installations?

arches/install/arches-templates/project_name/webpack/webpack.common.js
```diff
@@ -93,8 +93,8 @@ module.exports = () => {
                 plugins: [
                     new CleanWebpackPlugin(),
                     new webpack.DefinePlugin({
-                        ARCHES_CORE_DIRECTORY: `'${ROOT_DIR}'`,
-                        APP_ROOT_DIRECTORY: `'${APP_ROOT}'`
+                        ARCHES_CORE_DIRECTORY: `'${ROOT_DIR.replace(/\\/g ,'/')}'`,
+                        APP_ROOT_DIRECTORY: `'${APP_ROOT.replace(/\\/g ,'/')}'`
```
chrabyrd, line 97: wouldn't this change break non-windows installations?

arches/install/arches-templates/project_name/webpack/webpack.common.js
```diff
@@ -198,6 +201,7 @@ module.exports = () => {
                                                     '\x1b[33m%s\x1b[0m',  // yellow
                                                     `"${templatePath}" has failed to load. Retrying (${failureCount} / 5)...`
                                                 );
+                                                await new Promise(r => setTimeout(r, 8000));
```
chrabyrd, line 204: Can you remove this change in favor of changing it on a project-by-project basis? So far most installations do not have an issue with the speed of these requests ( we give webpack 5 chances to fetch the correct template )

arches/install/arches-templates/project_name/webpack/webpack.common.js
```diff
@@ -190,6 +190,9 @@ module.exports = () => {
                                                 if (serverAddress.charAt(serverAddress.length - 1) === '/') {
                                                     serverAddress = serverAddress.slice(0, -1)
                                                 }
+                                                if(process.platform === 'win32'){
+                                                    serverAddress = serverAddress.replace('localhost', '127.0.0.1')
```
chrabyrd, line 194: Can this be removed? It's hardcoding a server address.

chrabyrd (changes requested):
Overall this looks pretty good! Thanks for submitting it 👍 

I've requested a few changes below.

arches/install/arches-templates/project_name/webpack/webpack-utils/build-image-filepath-lookup.js
jkemper-pca: I'm not a Linux expert but my logic here is that Linux paths do not have backslashes and, since this only replaces existing backslashes with forward slashes, it doesn't affect any paths on anything but Windows environments.

If there are case where this assumption about Linux paths doesn't work then this could be changed to include a environment condition to guaranteed only affecting Windows environments.
Ex.
`if(process.platform === 'win32'){`

I did spin up a Ubuntu VM and was able to run the application without issue.

arches/install/arches-templates/project_name/webpack/webpack-utils/build-javascript-filepath-lookup.js
jkemper-pca: *Copying from other comment because it's the same*

I'm not a Linux expert but my logic here is that Linux paths do not have backslashes and, since this only replaces existing backslashes with forward slashes, it doesn't affect any paths on anything but Windows environments.

If there are case where this assumption about Linux paths doesn't work then this could be changed to include a environment condition to guaranteed only affecting Windows environments.
Ex.
`if(process.platform === 'win32'){`

I did spin up a Ubuntu VM and was able to run the application without issue.

arches/install/arches-templates/project_name/webpack/webpack-utils/build-template-filepath-lookup.js
jkemper-pca: _Copying from other comment because it's the same_

I'm not a Linux expert but my logic here is that Linux paths do not have backslashes and, since this only replaces existing backslashes with forward slashes, it doesn't affect any paths on anything but Windows environments.

If there are case where this assumption about Linux paths doesn't work then this could be changed to include a environment condition to guaranteed only affecting Windows environments.
Ex.
if(process.platform === 'win32'){

I did spin up a Ubuntu VM and was able to run the application without issue.

arches/install/arches-templates/project_name/webpack/webpack.common.js
jkemper-pca: _Copying from other comment because it's the same_

I'm not a Linux expert but my logic here is that Linux paths do not have backslashes and, since this only replaces existing backslashes with forward slashes, it doesn't affect any paths on anything but Windows environments.

If there are case where this assumption about Linux paths doesn't work then this could be changed to include a environment condition to guaranteed only affecting Windows environments.
Ex.
if(process.platform === 'win32'){

I did spin up a Ubuntu VM and was able to run the application without issue.

arches/install/arches-templates/project_name/webpack/webpack.common.js
jkemper-pca: **I have removed this.**

Webpack failing after 5 retries is a consistent issue on both my Windows and Linux tests but both environments are on the lower tier of VM available and so it fails to handle all the requests and repeat requests done in such a short time (Windows seems worse at handling all the requests than Linux but they both will fail all 5 times for a few requests per run). This is mostly an issue with multiple languages enable because of the extra network activity from the 302 redirects for each request.

I will simply make this change locally for my installations.

arches/app/templates/base.htm
jkemper-pca: Unless I'm misunderstanding the purpose of the original this seems like a typo in the original that ends up only affecting Windows. 
The code is appending `"/build/js/{{main_script}}.js"` to the `webpack_static` variable. But the `webpack_static` variable is `"/static/"` therefor the combined string is `"/static//build/js/{{main_script}}.js"`. Note the double "//" between "static" and "build". 
On a default installation on either Windows or Linux you will see this code on pages that use `base.htm` (like /search):
```html
<script src="/static//build/js/views/search.js"></script>
``` 
Linux seems flexible enough to ignore it but Windows throws an error.

If there is a good reason the current functionality exists with the double slashes then I will have to add more code that does a modification only on Windows installs because this change would have to be done on every Windows install or none would be able to load any pages that use `base.htm`.

chrabyrd: ah I  can see your point here. The installations it would break do not have a trailing slash in the static_url. This is ok.

arches/install/arches-templates/project_name/webpack/webpack-utils/build-javascript-filepath-lookup.js
chrabyrd: yep my mistake. When I first looked at this I thought you were replacing slashes the other way.

arches/install/arches-templates/project_name/webpack/webpack-utils/build-template-filepath-lookup.js
chrabyrd: yep my mistake. When I first looked at this I thought you were replacing slashes the other way.

arches/install/arches-templates/project_name/webpack/webpack.common.js
chrabyrd: yep my mistake. When I first looked at this I thought you were replacing slashes the other way.

chrabyrd (approved):

chrabyrd (changes requested):
Overall looks good!

I see you've updated the arches-templates files so that newly created projects will be affected. However it's possible to run Arches without a project and we'd like maintain this functionality as well. 

Would you mind updating the arches-core webpack files and package.json file as well? Once those changes are made this should be good to go 👍

Issue #9294, opened 2023-03-03
Bulk loader modules need to define their own error reporting UI


chiatt:
Different bulk loader modules will need to have error reporting that reflects their schema. Currently the etl manager defines UI that all modules must share. This needs to be modified so that there is a default UI, but individual modules can define their own if needed.

johnatawnclementawn:
I am working on this

PR #9303, opened 2023-03-07
9294 individualized bulk loader error reporting

johnatawnclementawn:
### Types of changes

### Description of Change

These changes provide a default error report for the bulk data loader. In the future, this default can be overridden based on the bulk loader module's schema. 

### Issues Solved

#9294

### DIFF
--- a/arches/app/media/js/views/components/etl_modules/import-single-csv.js
+++ b/arches/app/media/js/views/components/etl_modules/import-single-csv.js
@@ -51,6 +51,8 @@ define([
         this.selectedLoadEvent = params.selectedLoadEvent || ko.observable();
         this.validationErrors = params.validationErrors || ko.observable();
         this.validated = params.validated || ko.observable();
+        this.getErrorReport = params.getErrorReport;
+        this.getNodeError = params.getNodeError;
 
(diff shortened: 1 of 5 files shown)


arches/app/media/js/views/components/etl_modules/import-single-csv.js
```diff
@@ -48,6 +48,10 @@ define([
             return "unique_id_" + self.uniqueId;
         });
 
+        this.selectedLoadEvent = params.selectedLoadEvent || ko.observable();
+        this.validationErrors = params.validationErrors || ko.observable();
+        this.validated = params.validated || ko.observable();
+
```
chiatt, line 54: I believe the `getNodeError` method needs to be passed in here.

chiatt (changes requested):
Hey Johnathan, this looks really good, but when I click on the 'show' button in the error table, I get an error. I believe it's because the getNodeError method is missing and needs to be passed in the params. It's a method rather than an observable, so no need for the `|| ko.observable` logic. 

image

johnatawnclementawn:
Hey @chiatt, I think this should solve the issue.

chiatt (approved):

Issue #9321, opened 2023-03-13
Client side uploads are overwritten by storage


aarongundel:
For files that are uploaded on the client side, the backend datatype factory can/will overwrite them when attempting to create the server side file structure.

PR #9322, opened 2023-03-13
overwritten file fix for #9321

aarongundel:
### Types of changes

### Description of Change

This change is currently only really applicable to projects that implement client side uploads.  Unfortunately this core level fix seems necessary in order to prevent the datatype factory from overwriting the file that gets written.  See Arches for Science for an implementation/usage.

### Issues Solved

#9321 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -1615,7 +1615,7 @@ class FileListDataType(BaseDataType):
                 file_model.tile = tile
                 if models.TileModel.objects.filter(pk=tile.tileid).count() > 0:
                     original_storage = file_model.path.storage
-                    # this temporary storage change _bypasses_ s3 storage temporarily
+                    # Prevents Django's file storage API from overwriting files uploaded directly from client re #9321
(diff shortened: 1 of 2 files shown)


arches/app/datatypes/datatypes.py
```diff
@@ -1614,7 +1614,12 @@ def post_tile_save(self, tile, nodeid, request):
                 file_model.path = file_data
                 file_model.tile = tile
                 if models.TileModel.objects.filter(pk=tile.tileid).count() > 0:
+                    original_storage = file_model.path.storage
+                    # this temporary storage change _bypasses_ s3 storage temporarily
```
chiatt, line 1618: Hey Aaron - just to clarify this a bit more, could you add another comment stating something like: "Prevents Django's file storage api from overwriting files uploaded directly from the client, re #9321"

chiatt (changes requested):
This looks good. Just suggesting another comment to help clarify this change a bit.

chiatt (approved):

Issue #9317, opened 2023-03-13
Add updated translations to Arches


apeters:
Updating translations for the following languages:

- Hebrew (he)
- Arabic (ar)
- Urdu (ur)

chiatt:
Addressed in #9323

PR #9323, opened 2023-03-14
add language updates for Arabic, Hebrew, and Urdu, re #9317

apeters:
re #9317


chiatt (approved):
Everything appears to load as expected in each of the updated languages.

Issue #9370, opened 2023-03-24
Add 'useOntologyRelationship' property to resource instance node configs


chiatt:
If resource-instance nodes are already designed to use ontology relationships, then they will need to have the `useOntologyRelationship` property set in the node config in order for the node config to display properly.

chiatt:
Fixed in #9371

PR #9371, opened 2023-03-25
Add useOntologyRelationship to resource instance node config

chiatt:
### Types of changes

### Description of Change
Adds `useOntologyRelationship` to resource-instance and resource-instance-list node configs so that nodes that use ontology properties will default to that behavior in the node config ui. 

### Issues Solved

#9370


njkim (changes requested):
It works great. 
One question is:
It seems there can be `"ontologyProperty": ""` which will set useOntologyProperty = true.
This is probably ok?

chiatt:
@njkim Yeah - I considered that, and I think that's what we want. Users will just see what they would have seen prior to the availability of concept relationships.

Issue #9344, opened 2023-03-22
Make number widget accessible


apeters:

johnatawnclementawn:
https://github.com/archesproject/arches/pull/9373

johnatawnclementawn:
Resolved in #9373

PR #9373, opened 2023-03-27
9344 Make number widget accessible

johnatawnclementawn:
### Types of changes

### Description of Change

Improve a user's ability to navigate through a number widget and its configuration using a screen reader by including more landmarks and improving semantic HTML structure.

### Issues Solved

#9344 

### Checklist


arches/app/templates/views/components/widgets/number.htm
```diff
@@ -5,7 +5,7 @@
 {% block form %}
 <div class="row widget-wrapper">
     <div class="form-group">
-        <label class="control-label widget-input-label" for="" data-bind="text:label"></label>
+        <label class="control-label widget-input-label" data-bind="text:label"></label>
```
apeters, line 8: I think we should change the label element to a span element.  That way we don't have a label that doesn't point to an input element (which is flagged in the WAVE plugin). 
With that change we should also add this `display: inline-block;` to the `widget-input-label` class.

apeters (changes requested):
looks good to me.

apeters (approved):
looks good

Issue #9369, opened 2023-03-24
Bulk editor: node dropdown should show widget labels instead of node names


apeters:
Because widget labels can be duplicated, we should also prepend the card label to the end of the widget label like so
`Widget Label - (Card Label)`, eg: `Name - (Name of Textual Work)`

apeters:
fixed in #9380

Issue #9366, opened 2023-03-24
Bulk editor: Task status should show all user defined settings used in the operation 


apeters:
For example, when using the bulk editor with a search url, the url used doesn't show up "Bulk Data Edit Details".

PR #9383, opened 2023-03-30
Bug fixes and improvements on bulk editor

njkim:
### Types of changes

### Description of Change

Add additional edit details to edit status summary section, #9366
Update the sql to get example data for preview using search url if available, #9367
Replace the node name / node alias with widget label and card name for the label in the node selector dropdown, #9369

### Issues Solved

#9366, #9367, #9369

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


apeters:
Details section is showing [Object object] value

Screenshot 2023-04-06 at 2 18 17 PM

apeters:
If edit process throws an error, the UI doesn't seem to be able to recover.

Screenshot 2023-04-06 at 2 25 16 PM

Here's the error on the backend
```2023-04-06 15:37:09,928 arches.app.etl_modules.base_data_editor ERROR    invalid input syntax for type json
DETAIL:  Character with value 0x0a must be escaped.
CONTEXT:  JSON data, line 1: "star_record_no: 50445
PL/pgSQL function __arches_edit_staged_data(uuid,uuid,uuid,text,text,text,text) line 43 at assignment

[2023-04-06 15:37:09,928: ERROR/ForkPoolWorker-16] invalid input syntax for type json
DETAIL:  Character with value 0x0a must be escaped.
CONTEXT:  JSON data, line 1: "star_record_no: 50445
PL/pgSQL function __arches_edit_staged_data(uuid,uuid,uuid,text,text,text,text) line 43 at assignment```

apeters:
Using URL based edit, the url doesn't show up on `Bulk Data Edit Details` section.

Also, the buttons didn't show up in the results section (see below).

Screenshot 2023-04-06 at 2 44 09 PM

apeters:
Status still says `running` after "undoing" an edit.

Screenshot 2023-04-06 at 2 53 25 PM

apeters:
We should be using widget labels in the select a node dropdown
Those labels should also be sorted

Screenshot 2023-04-06 at 4 18 25 PM

apeters:
It would be nice to know how many records were affected by the edit operation.

apeters (changes requested):
see my comments above

apeters:
Preview doesn't respect the search url

apeters (commented):
Selecting the "trim" operation along with the "Capitalize" (and presumably other) operations causes both operations to not take effect.

apeters (changes requested):
When "undoing" an edit, we should immediately change the button label to show the status as "running" or "undoing", or something like that.  Ideally it would be nice to show a spinner in the button as well.

Screenshot 2023-04-24 at 1 49 47 PM

apeters:
It would be nice to show the number of resources affected by the edit operation in the "Bulk Data Edit Details" section.

apeters:
I think we should remove the "Cancel Editing" button.  It seems kind of unnecessary.
Screenshot 2023-04-24 at 2 19 15 PM

apeters:
The "Replace" bulk edit operation should include if the replace function was "case sensitive" or not.

apeters:
We should also combine the "Capitalize", "Upper", and "Lower" string operations into a single operation.  Let's call this "Change Case".  The subtitle can specify the possible operations, EG: "Capitalize, Upper Case, Lower Case"

apeters (approved):
everything looks great!

Issue #9362, opened 2023-03-22
Make top level search template accessible


phudson-he:

phudson-he:
PR raised here: #9385

johnatawnclementawn:
Resolved in #9385

PR #9385, opened 2023-03-31
9362 Make top level search template accessible

phudson-he:
### Types of changes

### Description of Change
Make top level search view template have semantic markup, to include article and sections.

### Issues Solved
#9362 

### Checklist

#### Ticket Background
*   Found by: @phudson-he
*   Tested by: @phudson-he
*   Designed by: @phudson-he

### DIFF
--- a/arches/app/templates/views/search.htm
+++ b/arches/app/templates/views/search.htm
@@ -6,9 +6,8 @@
 {% block title %} {{ block.super }} {% trans "Search" %} {% endblock title %}
 
 {% block main_content %}
-    <div class="content-panel flexrow" data-bind="click:function() { menuActive(false); }, visible: true", style="display:none;">
-
-        <div class="flex search-results-panel">
+    <article class="content-panel flexrow" data-bind="click:function() { menuActive(false); }, visible: true", style="display:none;">
+        <section class="flex search-results-panel">
             <div class="search-attribute-widget">
                 <div class="search-dropdown" data-bind="component: {
                     name: 'term-filter',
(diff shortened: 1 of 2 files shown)


arches/app/templates/views/search.htm
```diff
@@ -1,32 +1,13 @@
...
 {% load static %}
 
 {% block title %} {{ block.super }} {% trans "Search" %} {% endblock title %}
 
-
 {% block main_content %}
-    <div class="content-panel flexrow" data-bind="click:function() { menuActive(false); }, visible: true", style="display:none;">
-
-        <div class="flex search-results-panel">
+    <article class="content-panel flexrow" data-bind="click:function() { menuActive(false); }, visible: true", style="display:none;">
```
apeters, line 9: Considering the following from Mozilla should we use section instead of article:

`The <article> [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element represents a self-contained composition in a document, page, application, or site, which is intended to be independently distributable or reusable (e.g., in syndication). Examples include: a forum post, a magazine or newspaper article, or a blog entry, a product card, a user-submitted comment, an interactive widget or gadget, or any other independent item of content.`

Also, should we be using the `aria-label` property on these section and article elements to further describe their meaning?

phudson-he: There are sections under the article. So using the css classes to refer to the containers, "content-panel flexrow" becomes the article and there are 2 sections within, one for "flex search-results-panel" and one for "flex".

We've not added aria-label's to containers previously as there were no audit issues raised, but it does make sense to do so as it further explains the regions on the page, so I will add them accordingly. I'll resolve this when changes pushed etc.

apeters (changes requested):
can you wrap the labels you added in trans tags {% trans "..." %}

apeters (approved):
looks good

Issue #9395, opened 2023-03-31
Update the iiif-manifest model


njkim:
IIIF manifest model needs to be updated to add a new field transactionid
In order to provide a standard way to remove the manifest during workflow if the use decide to quit.
[arches-for-science-prj/#973](https://github.com/archesproject/arches-for-science-prj/issues/973)

Also, it would be nice if the delete logic can be move to the models.py instead of manifest view.

PR #9396, opened 2023-04-03
Add transaction id to iiif manifest model, #9395

njkim:
### Types of changes

### Description of Change

Update iiif_manifest model to add a transcation id so that it can be removed when the workflow (or any transaction) is cancelled, #9395 

### Issues Solved

#9395 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/models/models.py
```diff
@@ -1531,6 +1541,28 @@ class Meta:
         managed = True
         db_table = "iiif_manifests"
 
+    def delete(self, *args, **kwargs):
+        def check_canvas_in_use(canvas_id):
+            canvas_ids_in_use = [annotation.canvas for annotation in VwAnnotation.objects.all()]
```
chiatt, line 1546: It looks like this can just return ids, so it's not necessary to iterate a second time:
```
return [annotation.canvas.canvas_id for annotation in VwAnnotation.objects.all()]
```

chiatt (changes requested):
This looks really good. I've made just a couple minor suggestions that I think would improve things a bit.

arches/app/models/models.py
njkim: I believe this is supposed to return a boolean.

chiatt: Ah - indeed. In that case maybe the `check_canvas_in_use function` isn't needed. If we do something like below, we only have to make the query to the annotation view once:

```
all_canvases = [annotation.canvas for annotation in VwAnnotation.objects.all()]
canvases_in_manifest = self.manifest["sequences"][0]["canvases"]
canvas_ids = [canvas["images"][0]["resource"]["service"]["@id"] for canvas in canvases_in_manifest]
canvases_in_use = []
for canvas_id in canvas_ids:
    if canvas_id in all_canvases:
        canvases_in_use.append(canvas_id)
if len(canvases_in_use) > 0:
    canvas_labels_in_use = [
        item["label"] for item in canvases_in_manifest if item["images"][0]["resource"]["service"]["@id"] in canvases_in_use
    ]
    message = _("This manifest cannot be deleted because the following canvases have resource annotations: {}").format(
        ", ".join(canvas_labels_in_use)
    )
    raise IIIFManifestValidationError(message)

```

chiatt (approved):
This looks good to me!

PR #9402, opened 2023-04-04
9337 Make checkbox widget accessible

johnatawnclementawn:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
Allow the user to substitute and <enter> keypress for a click event on a checkbox.

### Issues Solved

#9337 

### Checklist

-   [ ] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

Issue #9345, opened 2023-03-22
Make radio-boolean widget accessible


apeters:

johnatawnclementawn:
@apeters please see PR #9414

johnatawnclementawn:
Resolved in #9414

PR #9414, opened 2023-04-06
9345 Make radio boolean widget accessible

johnatawnclementawn:
### Types of changes

### Description of Change

Improve a user's ability to navigate through the radio boolean button widget and its configuration using a screen reader/keyboard. Allow the user to substitute and keypress for a click event. Keypress feature also included in #9402.

### Issues Solved

#9345

### Checklist


arches/app/media/js/views/components/widgets/radio-boolean.js
```diff
@@ -27,14 +28,24 @@ define([
          
         WidgetViewModel.apply(this, [params]);
         var self = this;
+        this.ariaCheckedTrueValue = ko.observable();
```
apeters, line 31: we should be able to remove these variables and all the associated code, as we can just compare the value against `true` or `false` in the UI

apeters (changes requested):

apeters (approved):
looks good

Issue #8702, opened 2022-07-26
Check a project's arches version for compatibility 


chiatt:
In addition to checking graphs for compatibility with arches, it is necessary to also check a project's compatibility. This is because a project may depend on a particular arches feature independent of graph requirements. As an example, a project's graphs may be compatible with Arches 6.0.0 and 6.0.1, but the project itself may only work as expected with 6.0.1. 

We should add a function that checks a project's arches dependency on load of a page and alerts the user if the wrong version of arches is being used.

PR #9420, opened 2023-04-11
Enforcing arches version compatibility in a project, re #8702

chiatt:
### Types of changes

### Description of Change

- Allows an admin to define a project version and enforce min and max arches version compatibility 
- Displays project version with Arches version under the DEBUG flag in main navigation (lower left)
- Ensures new projects automatically define as the min and max allowable version of arches, the version of arches used to create the project.

### Issues Solved
#8702


arches/app/utils/compatibility.py
```diff
@@ -0,0 +1,40 @@
...
+
+    """
+
+    if min_arches is None or max_arches is None:
+        logger.error(
+            _("Either a minimum or maximum compatible Arches version is not specified. Unable to check {0} compatibility".format(target))
+        )
+        return
+
+    if len(min_arches) <= 3 and len(max_arches) <= 3:
```
apeters, line 26: I'm not sure that we should **require** that a project creator specify both a min and max compatible version, or either for that matter.  A required min version might make sense, but a max version is harder to understand.  
At the moment packages aren't pip installed, so upgrading to keep packages and arches versions in sync is more difficult.  

Was any thought put to using a library like [this](https://github.com/python-semver/python-semver) to manage version numbers and comparisons?

arches/app/templates/base-manager.htm
```diff
@@ -272,7 +272,10 @@ <h4 style="flex: 1" class="ep-form-alert-title" data-bind="text: alert().title">
                 {% if app_settings.DEBUG %}
                 <div class="debug-notice">
                     <div>{% trans 'DEBUG' %}</div>
+                    <div>{% trans 'Arches' %}</div>
```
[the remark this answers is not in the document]
chiatt: Good point - I've fixed it.

arches/app/utils/compatibility.py
apeters: @chiatt Yeah, that all makes sense to me.

chiatt:
To QA:

run `pip install -r requirements.txt`

Next add this setting to your project:
APP_VERSION = semantic_version.Version(major=1, minor=0, patch=0, prerelease=('a', '0'))
Confirm it displays like this at the bottom of the nav bar:
image

Next add these settings to your project. Note that the beta version is lower than 
MIN_ARCHES_VERSION = "7.4.0-b0"
MAX_ARCHES_VERSION = "7.4.0"

Not that the app still runs as expected (on the dev/7.4.x branch)
Change the MAX_ARCHES_VERSION to 7.3.0 and Arches should stop running. You should get an error message like this: ```arches.app.utils.compatibility.CompatibilityError: 'This project is incompatible with Arches 7.4.0b0.'```

aarongundel (approved):
Works as described for me.  My only complaint is that it feels a little heavy handed to have the application fail to start.  Ideally it would give something more visual for feedback rather than failing without anything being served up.  That being said, this is an advanced feature, at least it's logging it, and I don't expect many people will know about it - shouldn't affect more than a handful of projects once it gets out into the wild.  Therefore, 💯

chiatt:
Thanks @aarongundel - I totally agree about having Arches fail to start. I agree a more visual approach would be better. Maybe we can add that refinement in a patch release.

Issue #9435, opened 2023-04-18
Search resource type filter shows inactive models 


SDScandrettKint:
Issue found on 6.2.2

When a model is set to inactive it still appears on the Resource Type filter dropdown.

PR #9440, opened 2023-04-19
Hide inactive models in Resource Type filter

SDScandrettKint:
Added a filter in the resource-type-filter.js to check for if model is active before adding it to array. Now only active models show in resource type filter.

### Description of Change

In version 6.0.0, only active models would show in the resource type filter, however, this feature broke upon the release of 6.1.0. Adding a filter to check if the resource model is active now shows only active models in resource type dropdown.

### Issues Solved

#9435

### Checklist

#### Ticket Background


chiatt (approved):
Looks good to me!

PR #9451, opened 2023-04-25
8162 Map Permissions Available for Map Layers

csmith-he:
### Types of changes

### Description of Change
This allows access to Basemaps (except default) and Overlays to be restricted based on user group permissions (set in the django interface via the new Object Permissions interface)

### Issues Solved
#8162

#### Ticket Background
Sponsored by: https://github.com/HistoricEngland
Found by: @
Tested by: @csmith-he 
Designed by: @RobOatesHistoricEngland and @csmith-he 

### Further comments

Includes migration 9437_include_map_source_and_map_layer_permissions.  Please remember to apply migrations for this feature.


chiatt (changes requested):
Hi @csmith-he This is a great start. However it would be a big improvement if, in addition to negating permission with 'no_access_to_maplayer', you could add positive permissions as well (user can read, update, delete, create). This would provide much better control over how we allow users to manage layers in the map layer manager. The nodegroup model is a good example showing the positive permissions: https://github.com/archesproject/arches/blob/30f0c6de0349c7c1b775c4304c84fb3f73f4a0b3/arches/app/models/models.py#L538-L544

Also, the logic for managing nodegroup and instance permissions is in `utils/permission_backend.py`.  It would be nice if the permission check for maplayers was done there as well. If you add the positive permissions, then the pattern for checking the permissions on a layer would be similar to the check for nodegroup permissions: https://github.com/archesproject/arches/blob/30f0c6de0349c7c1b775c4304c84fb3f73f4a0b3/arches/app/utils/permission_backend.py#L24-L46

I don't think you'd need to worry about the `CachedObjectPermissionChecker`, but rather just use the `ObjectPermissionChecker`

arches/app/models/migrations/9437_include_map_source_and_map_layer_permissions.py
```diff
@@ -0,0 +1,15 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('models', '9436_relational_view_null_geom_support'),
```
chiatt, line 7: It looks like somebody added a new migration, so this dependency needs to be updated to: `9055_add_branch_publication_to_node`

csmith-he: dependency updated

csmith-he:
@chiatt I've made changes to the code based on your very helpful feedback.  Could you take another look and see if this fulfils the requirements for bringing into dev/7.4.x, please?  Thanks.

aj-he:
@csmith-he looks like the migration order has changed as the checks are failing.

csmith-he:
@chiatt - sorted things out having seen your message on Slack.  The build has completed successfully

chiatt (approved):

Issue #9453, opened 2023-04-25
Indexing fails if Arches tries to index instances belonging to an inactive graph


chiatt:
If a user creates instances and then makes their graph inactive, they will still have indexed instances of the inactive graph showing up in search. Reindexing should remove those instances from the index, however, reindexing fails if a user tries to index instances of inactive graphs. This could present a big problem for admins. There may be a better solution, such as modifying search to filter out instances from inactive graphs. For the time being, however, we need to prevent indexing of instances of inactive graphs.

This issue currently only applies to v6, as v7.0 - v7.3 versions of Arches don't use the active flag.

PR #9454, opened 2023-04-25
Only index resources of active graphs

chiatt:
### Types of changes

### Description of Change
Prevents index from failing if resources exist that belong to inactive resource models

### Issues Solved
#9453


apeters (approved):
looks good to me

Issue #9472, opened 2023-04-28
Update `Graph.append_branch` to handle appending branch with multiple nodes that share a name with the graph


chrabyrd:

chrabyrd:
resolved in #9473

PR #9473, opened 2023-04-28
Updates Graph.append_branch to handle branch with multiple nodes that share a name with a node in the graph  #9472

chrabyrd:
### Types of changes

### Description of Change

When a branch has multiple nodes with the same name, and the resource model where you're appending the branch _also_ includes a node with the same name, there will be an alias conflict. This occurs because the `make_name_unique` function will not consider that there could be a node with the updated alias that already exists in the branch.

TO VERIFY:

The easiest way to verify the change is load the sfplanning project. Then navigate to the `Community Input` Resource Model and add the `Activity branch (ARM WG)` branch to the model. This action will fail on `dev/7.4.x` but pass in this branch.

Alternatively, create a graph. Add a node to the graph with a specific name/alias. Then create a branch. Add a node with the same name as the node in the graph you created. Then, add another node with the same name. It should automatically update the alias to `${name}_n1`. Then attempt to append the branch to the graph. On `dev/7.4.x` it will fail because the branch node with the alias `${name} will be updated to `${name}_n1` without considering that `${name}_n1` already exists in the branch.

### Issues Solved

#9472


robgaston (approved):
looks good - thanks!

PR #9478, opened 2023-05-03
9477 edtf and table alias fix for spatial views

### LINKED ISSUE #9477: Fix for spatial view DB functions for EDTF value error and table alias error
When creating and updating rows in a spatial view attribute table, there is a chance that the nodegroupid UUID value used to create a table alias causes the dynamic SQL to fail. The table alias needs to be wrapped in double speech marks to stop this.

EDTF display values are not generated correctly and results in "null" being displayed.

RobOatesHistoricEngland:
Fixes #9477 

- Fixes EDTF datatype values coming through as null.
- Fixes an issue where the UUID that forms part of a table alias can sometimes cause a compile error with PGSQL. This was achieved by simply wrapping the aliases in quotes.

### Types of changes

#### Ticket Background
*   Designed by: @aj-he @RobOatesHistoricEngland

### DIFF
--- a/arches/app/models/migrations/9477_fix_for_spatial_view_dbf_function_table_alias_error.py
+++ b/arches/app/models/migrations/9477_fix_for_spatial_view_dbf_function_table_alias_error.py
@@ -114,19 +114,19 @@ class Migration(migrations.Migration):
 
                         ',
                         att_table_name,
-                        node_create, 
-                        geometry_node_id, 
+                        node_create,
+                        geometry_node_id,
                         tile_create,
                         att_comments,
                         att_table_name);
-                    
+
                     execute att_view_tbl;
-                    
+
                     return att_table_name;
                 end;
-                            
+
             $BODY$;
-                
+
             -- FUNCTION: public.__arches_trg_fnc_update_spatial_attributes()
 
             -- DROP FUNCTION IF EXISTS public.__arches_trg_fnc_update_spatial_attributes();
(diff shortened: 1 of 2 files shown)


aj-he:
@robgaston I've assigned to you but reassign as your team requires.

aj-he:
@chiatt I've added you to this as I know @robgaston is often on other things.

arches/app/models/migrations/9477_fix_for_spatial_view_dbf_function_table_alias_error.py
```diff
@@ -0,0 +1,258 @@
...
+
+                    return trigger_tile;
+                end;
+
+            $BODY$;
+        """
+
+    reverse_sql_string = """
+            DROP FUNCTION IF EXISTS public.__arches_create_spatial_view_attribute_table;
+            DROP FUNCTION IF EXISTS public.__arches_trg_fnc_update_spatial_attributes;
```
chiatt, line 253: Could you also return these functions to their previous state rather than dropping them? It looks like they are also defined in the 8042_3_spatialview_db_functions migration.

chiatt (changes requested):
Hi @RobOatesHistoricEngland, thanks for submitting these fixes. I added a couple of requests inline, but in addition to those could you remove any commented code in the SQL statements?

arches/app/models/migrations/9477_fix_for_spatial_view_dbf_function_table_alias_error.py
RobOatesHistoricEngland: Function now returning to previous state, rather than it getting dropped.

chiatt (approved):

PR #9479, opened 2023-05-03
Hide Search Export button by user group list #8010

khodgkinson-he:
### Types of changes

### Description of Change
Adds code to restrict search exports to specified user groups if the ENABLE_EXPORT_FOR_SPECIFIC_GROUPS array is populated.

### Issues Solved

#8010

### DIFF
--- a/arches/app/models/migrations/8010_export_permissions.py
+++ b/arches/app/models/migrations/8010_export_permissions.py
@@ -6,7 +6,7 @@ class Migration(migrations.Migration):
 
     dependencies = [
         migrations.swappable_dependency(settings.AUTH_USER_MODEL),
-        ('models', '9055_add_branch_publication_to_node'),
+        ('models', '9558_add_plugin_help'),
     ]
 
     def add_permissions(apps, schema_editor, with_create_permissions=True):
--- a/arches/app/models/migrations/8010_export_permissions.py
+++ b/arches/app/models/migrations/8010_export_permissions.py
@@ -1,0 +1,42 @@
+from django.conf import settings
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('models', '9055_add_branch_publication_to_node'),
+    ]
+
+    def add_permissions(apps, schema_editor, with_create_permissions=True):
+        db_alias = schema_editor.connection.alias
+        Group = apps.get_model("auth", "Group")
+        User = apps.get_model("auth", "User")
+        resource_exporter_group = Group.objects.using(db_alias).create(name='Resource Exporter')
+
+        try:
+            users = User.objects.using(db_alias)
+            users.groups.add(resource_exporter_group)
+            print('added users group')
+        except Exception as e:
+            print(e)
+
+    def remove_permissions(apps, schema_editor, with_create_permissions=True):
+        db_alias = schema_editor.connection.alias
+        Group = apps.get_model("auth", "Group")
+        resource_exporter_group = Group.objects.using(db_alias).get(name='Resource Exporter')
+        User = apps.get_model("auth", "User")
+
+        try:
+            users = User.objects.using(db_alias)
+            users.groups.remove(resource_exporter_group)
+            print('removed users group')
+        except:
+            pass
+
+        resource_exporter_group.delete()
+
+    operations = [
+        migrations.RunPython(add_permissions, reverse_code=remove_permissions),
+    ]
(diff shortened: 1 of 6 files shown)


chiatt (commented):
I think this is a good first approach, but it seems like a custom permission or even a custom user group would be a little better suited to addressing this problem. This could be a custom permission 'can export data' that would apply to search export and possibly other export options as they become available in the bulk data manager. 

Another option would be to add a permission to the SearchComponent model which would control whether or not a search component is available at all. This would then be useful if an admin wanted to hide search export and perhaps other components such as advanced search from some groups. However, I'm not sure if this would cause problems in the UI if a default SearchComponent was not available to user. @apeters has been thinking about changes to search, so I'm wondering what he thinks about this direction.

apeters:
@khodgkinson-he
I do agree with @chiatt that this might be better served via a new user group called something like `Data Exporter`.  We could then add users directly to that group or use a custom permission as @chiatt suggests.  We can then check to see if the user is part of that group and (dis)allow data export similar to what you've done in `search.py` view.  I think @chiatt may have an example of how you might achieve that.

khodgkinson-he:
@apeters @chiatt
As ever with these things, we are forced to weigh costs vs benefits.  

Whilst the improvements you suggest may be desirable, they would involve significant further refactoring.

(This fix for issue #8010 has already undergone major refactoring since its' first submission in PR #8011.)

Sadly our resources for this ticket are now very limited, what is MVP for submission?

chiatt:
Hi @khodgkinson-he, 

Sorry to ask you to change course after this work is already on it's second round. The change in my thinking is mainly because I know there are going to be other avenues for export from the UI, and we want to avoid having different patterns for managing access to them. I don't think this was a concern when the first round of work was submitted.

I started writing up a detailed explanation of an MVP, and then decided that it might be faster for me just to draft the changes in a PR. This is what I had in mind:

https://github.com/archesproject/arches/pull/9516

I generally would never respond to a PR with another PR, but this is quite a different approach and without reflecting on the work that you and @csmith-he have done, I wouldn't have come up with this suggestion.

By adding the Resource Exporter group, we can also prevent users from accessing export features of the Bulk Data Manager (when they are added). We could also add more nuanced custom permissions to the Resource Exporter group. The drawback to this approach is that users would have to be added to the 'Resource Exporter' group, but the migration could be updated to handle that for existing users.

What do you think? Would this approach meet your needs?

khodgkinson-he:
Good Afternoon @chiatt 

  I think that looks outstanding, sir, thank you! ^_^

  You appear with #9516  to be providing all the functionality we require, and it certainly looks like the way to be going. (In fact, we shall raise a ticket internally for the next sprint to incorporate this change.)

  Whilst discussing the migration code, we wondered whether it might be a better default to grant **all** users membership of Resource Exporter from the start?  This would minimise any 'surprise' changes to users on upgrading version, and ensure the revocation of Search Exporter rights would have to be a deliberate choice where applied..  ..we'd be interested in your thoughts?

  Many thanks once again!

khodgkinson-he:
Good Morning @chiatt  -  I've amended the code to come into line with #9516, and modified the migration to grant all users membership of Resource Exporter, as discussed.  Does this fit better?

Many thanks once again for all your help with this! ^_^

khodgkinson-he:
..finding an issue with the method_decorator... will dig further on the morrow..

chiatt:
Thanks @khodgkinson-he, at a glance this looks great. I haven't actually tested it yet, so I haven't noticed and issue with the decorator. Hopefully it doesn't give you too much trouble. Let me know if I can help.

arches/app/views/search.py
```diff
@@ -211,6 +216,7 @@ def get_resource_model_label(result):
         return ""
 
 
+@method_decorator(group_required("Resource Exporter"), name="dispatch")
```
chiatt, line 219: I see the issue with the decorator. It should actually just be `@group_required("Resource Exporter")` because it is directly on the view method rather than a view class.

arches/app/models/migrations/8010_export_permissions.py
```diff
@@ -0,0 +1,36 @@
+from django.conf import settings
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('models', '9055_add_branch_publication_to_node'),
```
chiatt, line 9: Could you update to:
`('models', '9558_add_plugin_help'),`

chiatt (changes requested):
Hi @khodgkinson-he  Just a couple really minor changes and we can merge this in. Hopefully tomorrow?

I figured out the issue with the decorator. It looks like I might have suggested the wrong syntax here: #9516. There's an inline comment with the proper syntax. Also, if you could update the migration dependency that would be great.

chiatt (approved):

Issue #9481, opened 2023-05-03
Replace psycopg2-binary with psycopg2


chiatt:
According to the psycopg2 docs, psycopg2-binary should really only be used [in development](https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary). This change should be made in both v6.2.x and v7.4.x.

PR #9482, opened 2023-05-03
Replace psycopg2-binary with psycopg2, re #9481

chiatt:
### Description of Change
Replaces psycopg2-binary with psycopg2 and upgrades to 2.9.6

### Issues Solved

# 9481


johnatawnclementawn:
Works as expected!

Issue #9441, opened 2023-04-19
String can be force entered into resource layer service styling causing ValueError: invalid literal for int() with base 10


SDScandrettKint:
Issue found on 6.2.2.

In the resource layer service styling user input boxes, found at /map_layer_manager, a user can force input a string which causes an entire Arches instance to crash with the following error:
```
Internal Server Error: /map_layer_manager
Traceback (most recent call last):
  File "/mnt/c/dev/env/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/mnt/c/dev/env/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/mnt/c/dev/env/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/mnt/c/dev/env/lib/python3.8/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/mnt/c/dev/env/lib/python3.8/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/mnt/c/dev/env/lib/python3.8/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/mnt/c/dev/arches/arches/app/views/map.py", line 50, in get
    context = self.get_context_data(
  File "/mnt/c/dev/arches/arches/app/views/base.py", line 109, in get_context_data
    map_layer = datatype.get_map_layer(node)
  File "/mnt/c/dev/arches/arches/app/datatypes/datatypes.py", line 1207, in get_map_layer
    "expanded_weight": int(node.config["weight"]) * 2,
ValueError: invalid literal for int() with base 10: ''
```

Normally, whenever anything but a number is inputted the input is replaced in two scenarios:
- A letter is inputted after the default number, causing the input to revert to 1
- If default number is deleted it is automatically replaced with 1
However, this can be bypassed by force by highlighting the input and typing any string.

Steps to reproduce:
- See the Line Style "Weight" OR "Halo Weight" input boxes as possibilities: 
![image](https://user-images.githubusercontent.com/119508494/233033578-17b08044-9bf2-401d-8afd-fdb08330b85e.png)
[...]

PR #9485, opened 2023-05-04
9441 Resource overlay styling and clustering input fixes

SDScandrettKint:
### Types of changes

### Description of Change
Previously, a string could be force inputted all of the service styling and clustering config inputs. This change checks in the front end whether the number is valid i.e not an empty string or a combination of integers and strings. Additionally, the value is saved to the node config when the save button is used, rather than when the value is inputted in order to fix the double save button issue that existed previously.

### Issues Solved

#9441

### Checklist

#### Ticket Background

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -779,45 +779,6 @@ class GeojsonFeatureCollectionDataType(BaseDataType):
             except ValueError:
                 layer_def = "[]"
         else:
-            # default_values = {"radius":2,
-            #                   "haloRadius":4,
-            #                   "weight":2,
-            #                   "haloWeight":4,
-            #                   "outlineWeight":2}
-            # try:
-            #     for inp, val in default_values.items():
-            #         # check if input is string
-            #         if isinstance(node.config[inp], str):
-            #             print(inp, node.config[inp])
-            #             print(type(node.config[inp]))
-
-            #             node.config[inp] = int(val)
-            #             print(inp, node.config[inp])
-            #             print(type(node.config[inp]))
-
-            #             node.config[inp].save()
-                    
-
-            # except (AttributeError, KeyError):
-            #     pass
-            # for inp, val in default_values.items():
-            #     #try:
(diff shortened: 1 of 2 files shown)


chiatt (changes requested):
Hi @SDScandrettKint, thanks for submitting these changes. Could you make a minor refactor?  For each of these variables:

```
haloWeightValue
weightValue
outlineWeightValue
haloRadiusValue
radiusValue
clusterDistanceValue
clusterMaxZoomValue
clusterMinPointsValue
simplificationValue
```

Could you change this pattern:
```
weightValue = self.config.weight();
if(weightValue === ""){weightValue = 2;} 
else (weightValue = Number(weightValue));
```

To this pattern?
```
const weightValue = self.config.weight; 
weightValue() === "" ? weightValue(2) : weightValue(Number(weightValue()));
```

You would also need to move the line `const weightValue = self.config.weight` out of the case block and above just above `switch (layer.id) {` on line 140. Like so:

```
var updateMapStyle = function() {
    _.each(overlays, function(layer) {
        const weightValue = self.config.weight;
        const haloWeightValue = self.config.haloWeight; 
        switch (layer.id) {
        case "resources-fill-" + params.nodeid:
            layer.paint["fill-color"] = self.config.fillColor();
            break;
        case "resources-line-halo-" + params.nodeid:
            haloWeightValue() === "" ? haloWeightValue(4) : haloWeightValue(Number(haloWeightValue()));
            layer.paint["line-width"] = haloWeightValue;
            layer.paint["line-color"] = self.config.lineHaloColor();
            break;
        case "resources-line-" + params.nodeid:
            weightValue() === "" ? weightValue(2) : weightValue(Number(weightValue()));
```

If you make that change:
1. Declaring `weightValue` with `const` will scope the variable to the method it is in (which should clear up some of the Stickler errors in the PR).
 2. If you assign self.config.widget to weightValue rather than just its value, then you don't have to update self.config.weight. It will already be updated.
 3. The ternary tightens up the code a bit.

chiatt (approved):

PR #9508, opened 2023-05-08
removes python-docx from arches core requirements

aarongundel:
Removes python-docx from core requirements - it doesn't appear to be used.


chiatt (approved):
Looks good!

PR #9536, opened 2023-05-16
Fix issue where files were not rendering in last step of data upload …

ryan86:
…workflow, re: 
archesproject/arches-for-science-prj#1156, 
archesproject/arches-for-science-prj#1157,
archesproject/arches-for-science-prj#1158


chiatt (changes requested):
This looks good - just one print statement that should be removed.

chiatt (approved):

Issue #9546, opened 2023-05-18
String based widgets no longer show the asterisk to indicate they are required


apeters:
The asterisk fails to display on either the text widget or rich-text widget when the node is defined as required.  

re #9281

Issue #9281, opened 2023-02-25
String nodes no longer enforcing required value


chiatt:
To reproduce - create a card with a string node and the string widget. Make the node required. Notice in the widget there is not `*` indicating that the node is required. Save a the card with no value and the node will save despite the required status.

re #9546

chiatt:
The asterisk now appears because #9546 is now resolved. However the node requirement is not enforced if you try to save a string without a value. This is because we no longer assign `None` to a string node. Instead we save the object with and empty string for each language. It's not ideal, but it allowed us to overcome dirty state issues. 

As a result this evaluation returns False and the missing value is never identified: https://github.com/archesproject/arches/blob/b32545022825912f1651fd46f32de7da26459260/arches/app/models/tile.py#L304

One solution might be to add a datatype method that reports whether a value is empty based on the nature of the data. For the string datatype, that might mean that all values in all languages are an empty string. I think this would also make sense for other datatypes that store JSON. For example, the geojson datatype might return empty if there is a geojson object that has no features.

chiatt:
I think to start on this we should create a new method on the base datatype - maybe something like `value_is_empty` which in the base datatype (`arches/app/datatypes/base.py`) would just return `True` if the node value is `None`. On the string datatype, however, it would check each of the value properties for each of the languages in the string object and return `True` if none of the values had actual strings.

For the time being, I think we should just do this for the string datatype, but we may consider doing something similar for the geojson and annotation datatypes.

PR #9547, opened 2023-05-18
fix asterisk not showing up for required string nodes, re #9546

apeters:
re #9546
re #9281

### DIFF
--- a/arches/app/templates/views/components/widgets/rich-text.htm
+++ b/arches/app/templates/views/components/widgets/rich-text.htm
@@ -8,7 +8,12 @@
         <div class="widget-inline-tools-collapser" data-bind=" click: function() {
             showi18nOptions(!showi18nOptions());
         }">
-            <label class="control-label widget-input-label" for="" data-bind="text:label"></label>
+            <span">
+                <label class="control-label widget-input-label data-bind="text:label"></label>
+                <!-- ko if: node -->
+                <i data-bind="css: {'ion-asterisk widget-label-required': node.isrequired}"></i>
+                <!-- /ko -->
+            </span>
             <span>
                 <i class="fa" data-bind="{
                     css: {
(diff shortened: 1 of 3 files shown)


arches/app/templates/views/components/widgets/rich-text.htm
```diff
@@ -8,7 +8,12 @@
         <div class="widget-inline-tools-collapser" data-bind=" click: function() {
             showi18nOptions(!showi18nOptions());
         }">
-            <label class="control-label widget-input-label" for="" data-bind="text:label"></label>
+            <span">
```
chiatt, line 11: Looks like there is an accidental quote that ended up in this tag.

arches/app/templates/views/components/widgets/rich-text.htm
```diff
@@ -8,7 +8,12 @@
...
+                <label class="control-label widget-input-label data-bind="text:label"></label>
```
chiatt, line 12: A double quote is needed after `widget-input-label`

chiatt (changes requested):
It looks like there are a couple of typos preventing webpack from building. Also, this fixes the UI, but the node requirement is actually not enforced if you try to save a string without a value. This is because we no longer assign `None` to a string node. Instead we save the object with an empty string for each language. It's not ideal, but it allowed us to overcome dirty state issues. 

As a result this evaluation returns False and the missing value is never identified: https://github.com/archesproject/arches/blob/b32545022825912f1651fd46f32de7da26459260/arches/app/models/tile.py#L304

One solution might be to add a datatype method that reports whether a value is empty based on the nature of the data. For the string datatype, that might mean that all values in all languages are an empty string. I think this would also make sense for other datatypes that store JSON. For example, the geojson datatype might return empty if there is a geojson object that has no features.

chiatt (approved):

apeters:
@bferguso (@chiatt ), you're right it looks like the text widget is the only one where we "lost" the asterisk (the rich-text widget seems to be ok) .  Looks like it happened somewhere between dev/7.3 and dev/7.4  This may have happened as a result of the accessibility work.
This should be fixed.  I'll make a new ticket

PR #9556, opened 2023-05-24
9302 not null search fix

apeters:
re #9302


apeters:
This is hard to test.  You need a very large dataset.  You can test on the pir-etl.fargeo machine.  It has this fix applied.

chiatt (approved):
Looks good to me!

Issue #9563, opened 2023-05-25
Add table for Temporary Files


aarongundel:
Sometimes there are files that are needed for temporary purposes that don't persist and aren't tied to the arches graph/tiles.  One example is screenshot files created for AfS reporting - they don't persist past the creation of the report.  I've run into this in other contexts as well.  The files table doesn't seem to be an appropriate venue for this as it requires a tile id.  A table that doesn't require a tile id seems more appropriate.

PR #9565, opened 2023-05-25
Adds temporary files table 

aarongundel:
### Types of changes

### Description of Change

Add a temporary files table so that we're not necessarily requiring a tile id in order to create a file.

### Issues Solved

#9563


arches/app/models/migrations/9563_temporary_files.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 3.2.15 on 2023-05-25 16:05
+
+import arches.app.models.fields.i18n
+import datetime
```
chiatt, line 4: It looks like these imports are not used:
```
import arches.app.models.fields.i18n
import datetime
```

chiatt (changes requested):

chiatt (approved):

Issue #9570, opened 2023-05-26
Add the ability to search for special reserved characters.


apeters:
Right now users can't search for reserved characters like "?" or "*" because those are used as wildcards.
Some users use those characters in their data as flags that indicate that that data needs to be reviewed.  Without this functionality, searching for those characters is impossible.

PR #9574, opened 2023-05-30
reserved character search

apeters:
re #9570


apeters:
To test this you'll need to first confirm that you can't search for a string that contains a wild card character (either ? or *).
Once you pull this change you'll need to reindex your data.
Then confirm that you can now use a wild card search and that  you can escape the wild card with a back slash to be able to search for that character as well (eg "Test\?" to search for the string "Test" followed by a question mark symbol)

chiatt (approved):
Looks good to me!

Issue #9585, opened 2023-05-31
Add ability for map layer to be hidden by default


bferguso:
Currently map layers are visible by default, and are hidden when the `no_access` map layer permission is assigned to a user or group. The consequence of this is that sensitive layers that should only be visible to a small group will be visible to new layers and groups by default.

Adding the ability to have a layer hidden by default and only visible to users/groups it is granted to would be help maintain the security of sensitive layers. The layer default visibility should be configurable in the mapbox_spec_json file, the Map Layer Manager and the Django Admin panel.

PR #9586, opened 2023-06-01
9585 map layer public flag

bferguso:
### Types of changes

### Description of Change

Adds functionality to allow map layers to have restricted permissions by default. A new boolean `ispublic` column has been added to the `map_layers` table and associated Model class. It affects the visibility of the map layer as follows:
1. If value is `true`, the map layer is visible by Users/Groups without explicit permissions. The map layer is not visible to Users/Groups assigned the `no_access_to_maplayer` permission.
2. If value is `false`, the map layer is not visible to Users/Groups without explicit permissions. The map layer is visible to Users/Groups assigned the `read_maplayer` permission.

The `ispublic` value can be controlled in both the Map Layer Manager and Django Administration tool.

It can also be controlled by the `ispublic` attribute in the mapbox JSON spec `arches-metadata` section:

```
  "arches-metadata": {
    "ispublic": false,
    "addtomap": false,
     ...
  },
```

### Issues Solved

#9585 

### Checklist

#### Ticket Background
*   Sponsored by: Self Funded
*   Found by: @bferguso
*   Tested by: @bferguso
*   Designed by: @bferguso


chiatt (approved):
Looks good to me!

Issue #9590, opened 2023-06-01
Graph Designer - Adding node with duplicate name causes exception


bferguso:
In the Graph Designer, when adding a node with a duplicate name a 500 error occurs on the server:
```
[01/Jun/2023 18:14:40] "POST /int/arches-fossils/node/df3ee1ae-9c1c-11ec-964d-5254008afee6 HTTP/1.1" 500 163
```

and the following javascript exception occurs:
```
Uncaught TypeError: can't access property "find", self.graph.nodegroups() is undefined
    nodegroup webpack://arches_fossils/../../../../arches-core/arches/app/media/js/views/graph/graph-designer/node-form.js?:35
```

The processing spinner stays active due to the JS error above.

Branch: dev/7.4.x @ 54a426456ee0f948978ff389405b8e721b16e1dd

PR #9601, opened 2023-06-05
cbyrd better error messaging for duplicate node names #9590

chrabyrd:
### Types of changes

### Issues Solved

#9590

### DIFF
--- a/arches/app/models/graph.py
+++ b/arches/app/models/graph.py
@@ -965,6 +965,11 @@ class Graph(models.GraphModel):
         """
         node["nodeid"] = uuid.UUID(str(node.get("nodeid")))
         old_node = self.nodes.pop(node["nodeid"])
+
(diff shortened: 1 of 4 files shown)


arches/app/models/graph.py
```diff
@@ -965,6 +965,11 @@ def update_node(self, node):
         """
         node["nodeid"] = uuid.UUID(str(node.get("nodeid")))
         old_node = self.nodes.pop(node["nodeid"])
+
+        for sibling_node in self.get_sibling_nodes(old_node):
```
chiatt, line 969: It seems like this same logic is used in validate_node_name which gets called in graph.validate.

chiatt: I'm wondering if we are somehow skipping the validate method or if errors raised in validate are not getting passed back to the frontend properly.

Issue #9591, opened 2023-06-01
Graph Designer - Allow multiple values switch does not operate


bferguso:
When on a node that has the `Place node(s) in a separate card` switch visible, the `Allow multiple values.` switch is inoperable.

When toggling the `Place node(s) in a separate card` switch it automatically toggles the `Allow multiple values.` switch to the opposite `Place node(s) in a separate card` value. (eg `Place node(s) in a separate card` -> `True`, then `Allow multiple values.` -> `False`)

No javascript exceptions are thrown.

Branch: dev/7.4.x @ 54a426456ee0f948978ff389405b8e721b16e1dd
Browser: Firefox Developer

chrabyrd:
This feature appears to be functioning as normal. The `allow multiple values` switch should only be available once the user has chosen to place nodes in a separate card. The user will need to hit the `save edits` button after placing nodes in a separate card for the `allow multiple values` switch to be enabled. The switch should become disabled when the `place nodes in separate card` switch is disabled. A fix has been made to not show 2 sets of save/discard edit buttons if the values of both switches have been updated.

bferguso:
I'll see if I can reproduce, but the behaviour I was experiencing was that the `allow multiple values` switch wasn't available regardless of the position of the `place cards in separate card`. The strange thing was that if I turned the `place nodes in separate card` off the `allow multiple values` automatically turned on.

<img width="625" alt="image" src="https://github.com/archesproject/arches/assets/6750642/8e66c839-8062-4c4e-a2f3-f5dc434e6452">

I think part of the issue was having to save to make the `allow multiple values switch work, but there still seems to be a UI behavioural issue as shown above.

chrabyrd:
:wave: @bferguso ! Yes, if you have a model and a set of steps that would reproduce this behavior, I'd be interested to see them. Thank you 👍

PR #9612, opened 2023-06-07
resolves issue with 2 sets of save/discard buttons appearing in some edge cases #9591

chrabyrd:
### Types of changes

### Issues Solved

#9591 

TO REPRODUCE:
1. In the graph designer create a node, then a child node, save the edits.
2. click 'place nodes in separate card' switch to be truthy, save the edit.
3. Interact with the `allow multiple values swtich` and the `place nodes in separate card` switch, trying various combinations
4. Notice that only one set of save/edit buttons appears


chiatt (approved):
Looks good!

PR #9613, opened 2023-06-07
NOT FOR INDIVIDUAL REVIEW! Editable future graphs feature branch

chrabyrd:
NOT FOR INDIVIDUAL REVIEW! This branch is for the holistic review of several individual branches making up the editable_future_graphs feature


arches/app/views/graph.py
```diff
@@ -538,6 +596,74 @@ def get(self, request, template="text"):
...
+        )
+        context["nav"]["title"] = self.graph.name
+        context["nav"]["help"] = {"title": _("Managing Published Graphs"), "template": "graph-publications-help"}
+
+        return render(request, "views/graph/model-history.htm", context)
+
+    def post(self, request, graphid):
+        publication_id = JSONDeserializer().deserialize(request.body)
+
+        user_language = translation.get_language()
```
apeters, line 637: translation is not defined

apeters (changes requested):

apeters (commented):
this [line](https://github.com/archesproject/arches/pull/9613/files#diff-f435fb6934ed65137d7c2d7094c0b3e78e61ecc6aae7c6b57884d3d573f3e58eR463) is throwing an error: 
```
NameError: name 'PublishedModelError' is not defined
```

arches/app/media/js/views/graph/graph-publication-data.js
```diff
@@ -0,0 +1,12 @@
...
+    function removeTrailingCommaFromObject(string) {
+        return string.replace(/,\s*}*$/, "}");
+    }
+
+    const graphPublicationDataHTML = document.querySelector('#graphPublicationData');
+    const graphPublicationData = graphPublicationDataHTML.getAttribute('graphPublicationData');
+
+    const parsedGraphPublicationData = JSON.parse(removeTrailingCommaFromObject(graphPublicationData));
+
+    return parsedGraphPublicationData;
```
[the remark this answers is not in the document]
chrabyrd: agreed

arches/app/templates/views/components/datatypes/resource-instance.htm
```diff
@@ -24,7 +24,7 @@
     <div class="rr-table">
         <div data-bind="foreach: config.graphs" style="display: flex; flex-direction: column;">
             <div class='rr-table-row'>
-                <div class="rr-table-row-initial" data-bind="style:{color: $parent.isEditable === true || '#aaa'}">
```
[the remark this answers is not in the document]
chrabyrd: out of date -- should be fixed in latest

arches/app/models/migrations/9053_add_future_graphs.py
```diff
@@ -0,0 +1,59 @@
+# Generated by Django 3.2.15 on 2022-11-30 16:31
```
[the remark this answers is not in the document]
chrabyrd: It depends on if we're going to merge the work in branch-by-branch, or the plan is to merge in the feature branch only.

chrabyrd:
> this [line](https://github.com/archesproject/arches/pull/9613/files#diff-f435fb6934ed65137d7c2d7094c0b3e78e61ecc6aae7c6b57884d3d573f3e58eR463) is throwing an error:
> 
> ```
> NameError: name 'PublishedModelError' is not defined
> ```

yep, good 🎣  and fixed in latest push

apeters (changes requested):
loading a package throws and error
`no such file or directory .......templates/views/components/widgets`

also missing `...media/js/views/components/widgets`

chrabyrd:
> loading a package throws and error `no such file or directory .......templates/views/components/widgets`
> 
> also missing `...media/js/views/components/widgets`

Which package are you attempting to load? Could not reproduce loading the sfplanning package

apeters:
I copied the `pkg` folder out of the accessibility project

apeters (changes requested):
I you clone a graph and then delete that clone, the original graph become unusable and inaccessible from the UI. 
It throws a `'NoneType' object has no attribute 'get_relatable_resources'`

apeters (changes requested):
Resource type filter in search has duplicate and invalid entries.
- Screenshot 2023-10-20 at 11 01 18 AM

`Create new Publication` and `Revert Changes` buttons don't reflow properly on smaller screens.
- Screenshot 2023-10-20 at 12 51 25 PM

Creating a new model creates 2 entries in the model hist page.  That doesn't seem right.
The model below has only been published once, but there are 2 entries.
- Screenshot 2023-10-20 at 1 04 09 PM

Reports appear to be broken.  They don't show any data.

apeters (changes requested):
if you make an edit to the current graph that's active, save it, and then revert the change, the graph changes to the `inactive` state.  See [this](https://recordit.co/A0dxssiGXC).

apeters (changes requested):
Adding a new language and then updating the a widget label on the current graph (not publishing a new graph) throws an error when viewing the report or editor of an instance of that graph.

apeters (commented):
I don't know if this is a bug or a feature, although it feels like a bug.  I noticed that when editing the current graph once you've saved change but not applied the change that the endpoint to edit a resource instance redirects (403) to the report.  From the end users perspective it feels like something is broken.

chrabyrd:
> I you clone a graph and then delete that clone, the original graph become unusable and inaccessible from the UI. It throws a `'NoneType' object has no attribute 'get_relatable_resources'`

good catch, ended up being an insidious little bug -- the publication is not erased on graph copy

chrabyrd:
> Resource type filter in search has duplicate and invalid entries.
> 
> * Screenshot 2023-10-20 at 11 01 18 AM

fixed in latest

> `Create new Publication` and `Revert Changes` buttons don't reflow properly on smaller screens.
> 
> * Screenshot 2023-10-20 at 12 51 25 PM

Yeah this fell out of a conversation w/Rob Dennis and I. The UX will definitely need further tweaking but I say punt until operational && we can have another UX discussion.

> Creating a new model creates 2 entries in the model hist page. That doesn't seem right. The model below has only been published once, but there are 2 entries.
> 
> * Screenshot 2023-10-20 at 1 04 09 PM

Ah this is by design. As of this branch a graph _cannot_ exist without a publication. So when a graph is first created saved, it gets a publication.

> Reports appear to be broken. They don't show any data.

fixed in latest

chrabyrd:
> this

We should have a chat about this, there's some complexity here.

chrabyrd:
> Adding a new language and then updating the a widget label on the current graph (not publishing a new graph) throws an error when viewing the report or editor of an instance of that graph in the new language.

Can you expand on this, maybe include a video?

chrabyrd:
> I don't know if this is a bug or a feature, although it feels like a bug. I noticed that when editing the current graph once you've saved change but not applied the change that the endpoint to edit a resource instance redirects (403) to the report. From the end users perspective it feels like something is broken.

This is a feature, happy to chat about i.

apeters:
Exporting a card as a branch throws an error. `'dict' object has no attribute 'is_immutable'`

apeters:
Branches that are "Active" still show up as "inactive" in the append branch window.Screenshot 2023-12-08 at 3 52 07 PM

apeters:
It would be nice to see the counts of resources that are associated with each published graph.  It's hard to know which graph has data associated with it.
Screenshot 2023-12-08 at 4 20 01 PM

apeters:
a couple of graph properties don't get persisted when re-publishing an existing graph:
`resource models that may be related` and `slug`
Screenshot 2023-12-08 at 5 55 49 PM
Screenshot 2023-12-08 at 5 55 43 PM

chrabyrd:
> Exporting a card as a branch throws an error. `'dict' object has no attribute 'is_immutable'`

fixed in latest

chrabyrd:
> Branches that are "Active" still show up as "inactive" in the append branch window.Screenshot 2023-12-08 at 3 52 07 PM

fixed in latest

chrabyrd:
> It would be nice to see the counts of resources that are associated with each published graph. It's hard to know which graph has data associated with it. Screenshot 2023-12-08 at 4 20 01 PM

added in latest

apeters (commented):
I found an issue that, unfortunately, I can't reproduce but has happened twice on separate occasions.   I believe I was editing the current graph and only focusing on the graph name itself, testing with different languages, saving edits, undoing edits, etc..., just trying to find any issues when this issue came up.

Here is the detail of the error and additional context:
```
AttributeError
'NoneType' object has no attribute 'get_relatable_resources'
/Users/alexei/Work/Projects/arches/env_8x/arches/arches/app/models/graph.py, line 1596, in serialize
arches.app.views.graph.GraphDesignerView
/Users/alexei/Work/Projects/arches/env_8x/bin/python3
```

Screenshot 2024-02-08 at 4 40 31 PM

arches/app/views/graph.py
```diff
@@ -210,23 +225,15 @@ def get(self, request, graphid):
         if self.graph.ontology is not None:
```
jacobtylerwalls, line 225: Clicking on an existing graph in AFS throws on this line because self.graph is None (in turn, because self.editable_future_graph is None, earlier). I ran the `manage.py graph publish --update` command.

arches/app/views/graph.py
```diff
@@ -449,19 +468,13 @@ def post(self, request, graphid=None):
...
             return JSONErrorResponse(e.title, e.message, {"status": "Failed"})
-        except PublishedModelError as e:
-            return JSONErrorResponse(e.title, e.message)
 
     @method_decorator(group_required("Graph Editor"), name="dispatch")
     def delete(self, request, graphid):
         if self.action == "delete_node":
             data = JSONDeserializer().deserialize(request.body)
             try:
                 graph = Graph.objects.get(graphid=graphid)
```
jacobtylerwalls, line 477: Clicking the delete node button just after publishing a graph (before dismissing the success banner) raises a 500. This seems unlikely, but I could see it happening if a user made a mistake and didn't bother dismissing the banner.


https://github.com/archesproject/arches/assets/38668450/84226bee-dd78-4379-af4b-3f7601f4b1cc

jacobtylerwalls (commented):
Hi! Just pulled this down to test drive it. I'm not planning on combing the diff here, but I did glance at models.py just to get a flavor of how my database was going to change on migrating. In doing that, I noticed one thing that I've tangled with before, and wanted to share a solution.

jacobtylerwalls:
In this workflow, nodes are created without an alias:

1. Graph Designer > New Resource Model. Edit name of top node.
2. Save. At this point the "create publication" and "revert changes" buttons appear.
3. Add a child node. Alias placeholder text appears.
4. Save Edits. Node alias is generated.
5. Add a grandchild node. Select it in sidebar. Alias placeholder text appears.
6. Discard Edits.
7. Create new publication.

Result: grandchild node has null node alias

chrabyrd:
> I found an issue that, unfortunately, I can't reproduce but has happened twice on separate occasions. I believe I was editing the current graph and only focusing on the graph name itself, testing with different languages, saving edits, undoing edits, etc..., just trying to find any issues when this issue came up.
> 
> Here is the detail of the error and additional context:
> 
> ```
> AttributeError
> 'NoneType' object has no attribute 'get_relatable_resources'
> /Users/alexei/Work/Projects/arches/env_8x/arches/arches/app/models/graph.py, line 1596, in serialize
> arches.app.views.graph.GraphDesignerView
> /Users/alexei/Work/Projects/arches/env_8x/bin/python3
> ```
> 
> Screenshot 2024-02-08 at 4 40 31 PM

I can shed a _little_ more light into this as it came up when I was chasing other issues. That happens when something has caused the graph to not properly update/serialize previously, but making it far enough in the process to reassign the `root` node. I've made some adjustments to `update_from_editable_future_graph` that have allowed the `transaction.atomic()` block to be re-added. This should hopefully correct any instances of this behavior

arches/app/etl_modules/import_single_csv.py
```diff
@@ -36,7 +36,7 @@ def get_graphs(self, request):
             GraphModel.objects.all()
             .exclude(pk=settings.SYSTEM_SETTINGS_RESOURCE_MODEL_ID)
             .exclude(isresource=False)
-            .exclude(publication_id__isnull=True)
```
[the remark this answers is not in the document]
chrabyrd: Good catch. With this feature a Graph is _always_ published unless it's an editable future graph. I've updated all filter calls throughout the app to instead key on the `source_identifier__isnull` bool

arches/app/models/models.py
```diff
@@ -106,6 +102,11 @@ def __init__(self, *args, **kwargs):
         if isinstance(self.cardid, str):
             self.cardid = uuid.UUID(self.cardid)
 
+    def save(self, *args, **kwargs):
+        if self.pk == self.source_identifier_id:
+            self.source_identifier_id = None
```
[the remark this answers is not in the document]
chrabyrd: ++

arches/app/views/graph.py
chrabyrd: This is a good catch. `graph publish --update` is really only useful in germs of adding or updating languages. There wasn't a great method to create editable_future_graphs in bulk for a project upgrade. So I added the management command `create_editable_future_graphs` to accommodate

chrabyrd:
> In this workflow, nodes are created without an alias:
> 
> 1. Graph Designer > New Resource Model. Edit name of top node.
> 2. Save. At this point the "create publication" and "revert changes" buttons appear.
> 3. Add a child node. Alias placeholder text appears.
> 4. Save Edits. Node alias is generated.
> 5. Add a grandchild node. Select it in sidebar. Alias placeholder text appears.
> 6. Discard Edits.
> 7. Create new publication.
> 
> Result: grandchild node has null node alias

This appears to be an issue outside of the editable_future_graph work. It looks like a node alias is not generated until the `Save Edits` button is pushed. I'll spin up a separate ticket but IMO this is out of scope

arches/app/models/models.py
```diff
@@ -1138,7 +1184,7 @@ class TileModel(models.Model):  # Tile
     resourceinstance = models.ForeignKey(ResourceInstance, db_column="resourceinstanceid", on_delete=models.CASCADE)
     parenttile = models.ForeignKey("self", db_column="parenttileid", blank=True, null=True, on_delete=models.CASCADE)
     data = JSONField(blank=True, null=True, db_column="tiledata")
-    nodegroup = models.ForeignKey(NodeGroup, db_column="nodegroupid", on_delete=models.CASCADE)
+    nodegroup_id = models.UUIDField(db_column="nodegroupid", null=True)
```
jacobtylerwalls, line 1187: Heads up that the first failing test on the last push looks like it was due to this change, so let's add a breaking changes notice. There could be applications that need to check their views & tests.

tests/base_test.py
```diff
@@ -118,7 +119,11 @@ def setUpClass(cls):
             oauth_client_secret=OAUTH_CLIENT_SECRET,
             jwt_algorithm=test_settings.JWT_ALGORITHM,
         )
-        cursor.execute(sql)
+
+        try:
+            cursor.execute(sql)
+        except psycopg2.errors.UniqueViolation:  # entry already in database
+            pass
```
jacobtylerwalls, line 126: Hopefully this won't be a problem after #10711 but I can recheck it

jacobtylerwalls: I looked into this a little further. This is still a problem after #10711. It's related to the TODO statements I added showing where we have some package loads in test case setup that _depend_ on running _outside_ of a transaction, which isn't ideal:

https://github.com/archesproject/arches/blob/350e4a7cd2cca3b17176cf31bbcbf5b5282ff695/tests/views/api_tests.py#L41-L44

I'll open an issue and ref them in the TODO statements. If we fix that, I expect we wouldn't need this workaround.

However, I don't think this workaround is doing what it appears. When I revert the bugfix that causes the api test to fail, and I run `python3 manage.py test tests.views.api_tests tests.views.auth_tests --settings=tests.test_settings`, this except isn't reached because Django reraises `UniqueViolation` as `IntegrityError`. However, if I change this to catch IntegrityError, then it causes far more failures because the transaction is aborted and doesn't clean up after itself and all subsequent tests fail.

I'm feeling like we should just remove the workaround and wait for the follow-up issue to be fixed. WDYT?

chrabyrd: Hmm let's keep this fix for now but you should request changes on this PR so it can't get merged in. That way the fix is still there in case the followup doesn't work or @apeters decides to merge it in knowing that the followup is not far behind.

jacobtylerwalls (changes requested):
Marking changes requested to ensure we discuss setUpClass workaround before merge

arches/management/commands/graph.py
```diff
@@ -73,34 +74,57 @@ def add_arguments(self, parser):
...
+
+
+    def create_editable_future_graphs(self):
+        print("\nBEGIN Create editable_future_graphs...")
+
+        for graph in self.graphs:
+            print("\nCreating editable_future_graph for %s" % graph.name)
+            graph.create_editable_future_graph()
+
+            print("\n%s has been updated. Creating a new publication for %s" % graph.name, graph.name)
```
jacobtylerwalls, line 101: Ran this command and got:
```py
  File "/web_root/arches/arches/management/commands/graph.py", line 101, in create_editable_future_graphs
    print("\n%s has been updated. Creating a new publication for %s" % graph.name, graph.name)
TypeError: not enough arguments for format string
```

jacobtylerwalls:
Ran the AfS view tests against this branch and found some `TileModel.nodegroup` references to fix. That gave me the idea to grep for things in core arches possibly not exercised by a test.

This looks relevant. It's fine if nodegroup returns something, but if it's None, then `.pk` will obviously raise.
https://github.com/archesproject/arches/blob/a1c2b429f5aaa54af95a4111862eb87d56709107/arches/app/utils/data_management/resources/formats/rdffile.py#L828

Also, we should update this to use nodegroup_id so we don't reintroduce the db queries that #10546 optimized out. It's not a pure replace; I figure we'll have to extract the pk's from the set of model objects.

https://github.com/archesproject/arches/blob/d4d960801140e3f62a40869c55910de6c7316ab2/arches/app/models/resource.py#L260

Resource.copy() is currently broken (#10712), but let's fix the breaking change here while we're fresh:

https://github.com/archesproject/arches/blob/a1c2b429f5aaa54af95a4111862eb87d56709107/arches/app/models/resource.py#L736

This is a little paranoid, but we could add some safety against null nodegroups here. Speaking of, are you allowed to create future versions of the system settings graph?

https://github.com/archesproject/arches/blob/a1c2b429f5aaa54af95a4111862eb87d56709107/arches/app/models/system_settings.py#L119-L120
https://github.com/archesproject/arches/blob/a1c2b429f5aaa54af95a4111862eb87d56709107/arches/app/models/system_settings.py#L130-L133

Another null guard:
https://github.com/archesproject/arches/blob/a1c2b429f5aaa54af95a4111862eb87d56709107/arches/app/utils/data_management/resources/formats/csvfile.py#L1101

Another null guard:
https://github.com/archesproject/arches/blob/a1c2b429f5aaa54af95a4111862eb87d56709107/arches/app/utils/data_management/resources/formats/rdffile.py#L206-L230

tests/models/graph_tests.py
```diff
@@ -1071,24 +1072,130 @@ def test_appending_a_branch_with_an_invalid_ontology_class(self):
...
-        self.assertEqual(cm.exception.code, 1012)
-
-    def test_appending_published_branch_to_unpublished_graph(self):
-        graph = Graph.objects.get(node=self.rootNode)
-        admin = User.objects.get(username="admin")
-        branch = Graph.objects.get(graphid=self.NODE_NODETYPE_GRAPHID)
-        branch.publish(user=admin)
-        self.addCleanup(branch.unpublish)
-
-        graph.append_branch("http://www.nasa.gov/", graphid=self.NODE_NODETYPE_GRAPHID)
```
jacobtylerwalls, line 1094: Are these tests being deleted because the sense of what they are doing is captured in the tests that are being added?

jacobtylerwalls:
[stream of consciousness review, sorry...]

Re: the part of the comment above about `.nodegroup` adding queries now that's it's a property -- it occurs to me that now we won't be able to `select_related("nodegroup")`'able to avoid N+1 queries. Looping over tiles and accessing `.nodegroup` will hit the DB every time. (That is, without some ORM gymnastics like adding a custom model manager or doing a highly custom prefetch.)

If all we need is the nullability, have we considered just using making the existing foreign key nullable? Then the following would still work without breaking changes, other than needing to check for None:
 - `select_related("nodegroup")`
 - `.nodegroup = ...`
 - `.filter(nodegroup= ... )`

tests/models/graph_tests.py
chrabyrd: Nah they're being deleted for relevance. It's not longer possible to append a node to a publish graph, and there's no such thing as an unpublished graph after this feature

releases/8.0.0.md
```diff
@@ -0,0 +1,37 @@
...
+### Breaking changes
+- The `TileModel` entity has been updated. Instead of `nodegroup` acting as a foreign key to the `Nodegroup` table, it has been replaced with `nodegroup_id`, a simple UUID field. There is now a `nodegroup` class method that will retrieve the Nodegroup from the `nodegroup_id` field, but any direct database references to `nodegroup` should be updated.
+
+### Upgrading Arches
+
+1. You must be upgraded to at least version   before proceeding. If you are on an earlier version, please refer to the upgrade process in the []()
+
+### Upgrading an Arches project
+
+1. Create editable_future_graphs for your Resource Models using the command `python manage.py graph create_editable_future_graphs`. This will publish new versions of each Graph.
```
jacobtylerwalls, line 37: For a graph with a slug, after running this command, I cannot reverse migrate:

```py
django.db.utils.IntegrityError: could not create unique index "graphs_slug_32b94acf_uniq"
DETAIL:  Key (slug)=(my_slug) is duplicated.
```

arches/app/templates/views/graph/model-history.htm
```diff
@@ -0,0 +1,195 @@
...
+                        border-right: 1px solid #ddd; 
+                        cursor: pointer;
+                    "
+                >
+                    <div style="width: 60px; flex: none;">
+                        <strong>{% trans "Notes:" %}</strong>
+                    </div>
+                    <div 
+                        style="
+                            overflow-y: scroll;
```
jacobtylerwalls, line 164:
```suggestion
                            overflow-y: auto;
```
Seems to prevent this situation on MacOS:
always-scrollbar

jacobtylerwalls:
1. Create new graph
2. Click "Make changes without publishing"
3. make edit; click "Save edits"
4. Quit Designer
5. Graph Manager > Delete the original graph from "Manage..." dropdown
6. Manager refreshes to show only the graph marked "(Has unpublished changes)" > Delete that one from "Manage..." dropdown

Raises `Graph.DoesNotExist`

arches/app/views/graph.py
```diff
@@ -210,23 +225,15 @@ def get(self, request, graphid):
         if self.graph.ontology is not None:
             branch_graphs = branch_graphs.filter(ontology=self.graph.ontology)
```
jacobtylerwalls, line 226: 1. Create a branch from scratch (in graph designer)
2. Save Edits
3. Quit Designer
4. [You're on Branches tab of Graph Designer:] Click on that branch


Raises

```py
  File "/Users/jwalls/prj/arches/arches/app/views/graph.py", line 225, in get
    if self.graph.ontology is not None:
       ^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'ontology'
```

chrabyrd:
> [stream of consciousness review, sorry...]
> 
> Re: the part of the comment above about `.nodegroup` adding queries now that's it's a property -- it occurs to me that now we won't be able to `select_related("nodegroup")` to avoid N+1 queries. Looping over tiles and accessing `.nodegroup` will hit the DB every time. (That is, without some ORM gymnastics like adding a custom model manager or doing a highly custom prefetch.)
> 
> If all we need is the nullability, have we considered just making the existing foreign key nullable? Then the following would still work without breaking changes, other than needing to check for None:
> 
> * `select_related("nodegroup")`
> * `.nodegroup = ...`
> * `.filter(nodegroup= ... )`

No worries! Always glad to have to re-think/defend decisions so the best way forward is implemented. Ended up fiddling around with the code a bit to remind myself why the choice was made.

The changes made here are to support the switching of Graphs between versions. eg:

1. Create a Graph with a single string node, named 'AAA'. It's a de-facto nodegroup.
2. Create a Resource from the Graph, it has a Tile referencing 'AAA". 
3. Edit the Graph and delete the 'AAA' node. Then add a 'BBB' number node. 
4. Attempt to publish the new Graph.

An error will occur if the link between the Nodegroup/Tile still exists because in Postgres you can't have a ForiegnKey reference a non-existent entity. 

So let's say we make `Tile.nodegroup` nullable with the `on_delete=models.SET_NULL` arg. Going through the above workflow will probably work, but the only linkage between NodeGroup and Tile is  on the tile. Once the new graph is published, the 'AAA' nodegroup will be deleted and the Tile will effectively be orphaned. Attempting to view the resource report or edit the resource in the editor will end in error. 

A different solution I imagined was having both nodegroup and nodegroup_id fields on the TileModel. We could use Django signals or something so that whenever a NodeGroup is created the Tile table could be iterated and that link re-established. However it's easy for the Tiles table to grow large and I think that would be a noticeable performance hit.

So I landed on the current solution. Probably not ideal but the best path I can see right now. As far as the nullguarding goes I'm inclined to just built an Exception into the `Tile.nodegroup` property instead of playing whack-a-mole with the callsites. WDYT?

releases/8.0.0.md
chrabyrd: Could you include the steps to repro please?

arches/app/templates/views/graph/model-history.htm
chrabyrd: can't seem to repro but setting it to auto doesn't affect the UX negatively for me, so happy to include this

releases/8.0.0.md
jacobtylerwalls: 1. Graph Manager > Add a slug and save (field is called "URI Slug for API Access")
2. Create new publication
3. Check db shell: you should have 2 graph rows with the slug you created
4. Reverse migrate: `python manage.py migrate models 9511`

Result: django attempts to reenforce the unique constraint on slug, but PG complains that multiple rows exist

I found this in a more roundabout way just loading AfS on 7.6 and upgrading to 8.0 and running `manage.py graph create_editable_future_graphs` which gave me the duplicate slugs.

chrabyrd:
> 1. Create new graph
> 2. Click "Make changes without publishing"
> 3. make edit; click "Save edits"
> 4. Quit Designer
> 5. Graph Manager > Delete the original graph from "Manage..." dropdown
> 6. Manager refreshes to show only the graph marked "(Has unpublished changes)" > Delete that one from "Manage..." dropdown
> 
> Raises `Graph.DoesNotExist`

I cannot seem to repro. Are there any specific changes you're making?

1hSUFL2hFc

arches/app/views/graph.py
chrabyrd: I also cannot seem to repro this either. Happy to chat if you wanna show off your workflow && compare

jacobtylerwalls:
> Are there any specific changes you're making?

Now I can't reproduce either. No worries.

releases/8.0.0.md
chrabyrd: This was a good catch && took some thinking on how to work around it. Should be fixed in latest

arches/app/views/api.py
```diff
@@ -1095,8 +1146,28 @@ def get(self, request, resourceid):
         perm = "read_nodegroup"
 
         resource = Resource.objects.get(pk=resourceid)
-        graph = Graph.objects.get(graphid=resource.graph_id)
+        published_graph = models.PublishedGraph.objects.get(publication=resource.graph_publication, language=translation.get_language())
```
jacobtylerwalls, line 1149: Found that an import went missing here (assuming after tons of merges). Reproduced by clicking "View Report" button from Edit Log activity:

```py
2024-04-02 16:52:28   File "/web_root/arches/arches/app/views/api.py", line 1149, in get
2024-04-02 16:52:28     published_graph = models.PublishedGraph.objects.get(publication=resource.graph_publication, language=translation.get_language())
2024-04-02 16:52:28 NameError: name 'translation' is not defined
```

jacobtylerwalls:
Thanks for the walkthrough on the missing nodegroup flow, I get it now.

> As far as the nullguarding goes I'm inclined to just built an Exception into the Tile.nodegroup property instead of playing whack-a-mole with the callsites. WDYT?

I think I was wrongly assuming that some of these other workflows should still succeed if there are unmigrated historical tiles pointing to nonexistent nodes. But now that I see we want them to fail, then, yeah, let's not play whackamole. As you suggested something really simple like this would do:

```diff
diff --git a/arches/app/models/models.py b/arches/app/models/models.py
index aa5a0d6e3..909e9ce5a 100644
--- a/arches/app/models/models.py
+++ b/arches/app/models/models.py
@@ -1199,8 +1199,7 @@ class TileModel(models.Model):  # Tile
 
     @property
     def nodegroup(self):
-        nodegroup_query = NodeGroup.objects.filter(pk=self.nodegroup_id)
-        return nodegroup_query.first()
+        return NodeGroup.objects.get(pk=self.nodegroup_id)
 
     def is_fully_provisional(self):
         return bool(self.provisionaledits and not any(self.data.values()))
```

jacobtylerwalls:
@robgaston raised the idea of making a small effort to handle other locations (e.g. search export) that naively assume a tile's nodegroup always exists -- and wrap with something more appropriate than a bare error like the below. ("Error: ad-hoc graph migration; contact your administrator" or whatever). But I can also see that being follow-up work.

I'm not totally certain what the UX is like when Debug=False for 500s like this.

```py
  File "/web_root/arches/arches/app/search/search_export.py", line 264, in get_node
    return self.node_lookup[nodeid]

During handling of the above exception ('52aa2007-c450-11e9-b5d4-a4d18cec433a'), another exception occurred:
  File "/usr/local/lib/python3.10/dist-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.10/dist-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.10/dist-packages/django/contrib/auth/decorators.py", line 23, in _wrapper_view
    return view_func(request, *args, **kwargs)
  File "/web_root/arches/arches/app/views/search.py", line 278, in export_results
    export_files, export_info = exporter.export(format, report_link)
  File "/web_root/arches/arches/app/search/search_export.py", line 160, in export
    resource_obj = self.flatten_tiles(
  File "/web_root/arches/arches/app/search/search_export.py", line 313, in flatten_tiles
    node = self.get_node(nodeid)
  File "/web_root/arches/arches/app/search/search_export.py", line 266, in get_node
    self.node_lookup[nodeid] = models.Node.objects.get(pk=nodeid)
  File "/usr/local/lib/python3.10/dist-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/django/db/models/query.py", line 637, in get

Exception Type: DoesNotExist at /search/export_results
Exception Value: Node matching query does not exist.
```

chrabyrd:
> @robgaston raised the idea of making a small effort to handle other locations (e.g. search export) that naively assume a tile's nodegroup always exists -- and wrap with something more appropriate than a bare error like the below. ("Error: ad-hoc graph migration; contact your administrator" or whatever). But I can also see that being follow-up work.
> 
> I'm not totally certain what the UX is like when Debug=False for 500s like this.
> 
> ```python
> Traceback (most recent call last):
>   File "/web_root/arches/arches/app/search/search_export.py", line 264, in get_node
>     return self.node_lookup[nodeid]
> 
> During handling of the above exception ('52aa2007-c450-11e9-b5d4-a4d18cec433a'), another exception occurred:
>   File "/usr/local/lib/python3.10/dist-packages/django/core/handlers/exception.py", line 55, in inner
>     response = get_response(request)
>   File "/usr/local/lib/python3.10/dist-packages/django/core/handlers/base.py", line 197, in _get_response
>     response = wrapped_callback(request, *callback_args, **callback_kwargs)
>   File "/usr/local/lib/python3.10/dist-packages/django/contrib/auth/decorators.py", line 23, in _wrapper_view
>     return view_func(request, *args, **kwargs)
>   File "/web_root/arches/arches/app/views/search.py", line 278, in export_results
>     export_files, export_info = exporter.export(format, report_link)
>   File "/web_root/arches/arches/app/search/search_export.py", line 160, in export
>     resource_obj = self.flatten_tiles(
>   File "/web_root/arches/arches/app/search/search_export.py", line 313, in flatten_tiles
>     node = self.get_node(nodeid)
>   File "/web_root/arches/arches/app/search/search_export.py", line 266, in get_node
>     self.node_lookup[nodeid] = models.Node.objects.get(pk=nodeid)
>   File "/usr/local/lib/python3.10/dist-packages/django/db/models/manager.py", line 87, in manager_method
>     return getattr(self.get_queryset(), name)(*args, **kwargs)
>   File "/usr/local/lib/python3.10/dist-packages/django/db/models/query.py", line 637, in get
>     raise self.model.DoesNotExist(
> 
> Exception Type: DoesNotExist at /search/export_results
> Exception Value: Node matching query does not exist.
> ```

I was able to adjust the search_export logic to not fail when a nodegroup isn't present. The output could be seen as confusing though; for example if the user exports a search result that contains 1 resource instance, and that resource instance is out-of-date, the export will succeed but the csv will be blank. 

I'll file a followup ticket to fix this. And I suppose this could spark a larger discussion of keeping cards/cards_x_nodes_x_widgets/edges/nodes/nodegroups between graph versions instead of deleting them 🤔  . It would result in more DB clutter but could be useful for cases such as this.

arches/app/models/models.py
```diff
@@ -1152,6 +1197,10 @@ def __init__(self, *args, **kwargs):
         if not self.tileid:
             self.tileid = uuid.uuid4()
 
+    @property
+    def nodegroup(self):
```
jacobtylerwalls, line 1201: Wondering if you considered making this a cached property (given how unlikely noedgroup-deletions-in-production are)?

> Be careful with your own custom properties - it is up to you to implement caching when required, for example using the [cached_property](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.functional.cached_property) decorator.

https://docs.djangoproject.com/en/4.2/topics/db/optimization/#understand-cached-attributes

chrabyrd:
> Thanks for the walkthrough on the missing nodegroup flow, I get it now.
> 
> > As far as the nullguarding goes I'm inclined to just built an Exception into the Tile.nodegroup property instead of playing whack-a-mole with the callsites. WDYT?
> 
> I think I was wrongly assuming that some of these other workflows should still succeed if there are unmigrated historical tiles pointing to nonexistent nodes. But now that I see we want them to fail, then, yeah, let's not play whackamole. As you suggested something really simple like this would do:
> 
> ```diff
> diff --git a/arches/app/models/models.py b/arches/app/models/models.py
> index aa5a0d6e3..909e9ce5a 100644
> --- a/arches/app/models/models.py
> +++ b/arches/app/models/models.py
> @@ -1199,8 +1199,7 @@ class TileModel(models.Model):  # Tile
>  
>      @property
>      def nodegroup(self):
> -        nodegroup_query = NodeGroup.objects.filter(pk=self.nodegroup_id)
> -        return nodegroup_query.first()
> +        return NodeGroup.objects.get(pk=self.nodegroup_id)
>  
>      def is_fully_provisional(self):
>          return bool(self.provisionaledits and not any(self.data.values()))
> ```

Ah yeah after re-introducing the loud fail I remember why I took it out in the first place. It breaks serialization of Tiles and I'd like to not touch the long-untouched serializer if possible. I added it back in there with the most recent set of changes that un-breaks report for out of date Graphs.

arches/app/models/models.py
chrabyrd: My gut says to shy away from caching in this instance because of the potential for unexpected behavior during graph-switching or updating

jacobtylerwalls: sounds wise 👍

apeters (changes requested):
`discard edits` never goes away after editing a card and then discarding the edits

Screenshot 2024-04-12 at 9 55 53 AM

apeters:
We should add a cancel button to this popup.
Screenshot 2024-04-12 at 12 17 14 PM

chrabyrd:
> `discard edits` never goes away after editing a card and then discarding the edits
> 
> Screenshot 2024-04-12 at 9 55 53 AM

This issue appears is dev/7.6.x -- and looks to be related to i18n. Agreed that it should be resolved but I believe it's out-of-scope for this PR. I ticketed the issue in #10873

tests/fixtures/data/json/resource_export_business_data_truth.json
```diff
@@ -41,6 +42,12 @@
             "data": {
               "677f236e-09cc-11e7-8ff7-6c4008b05c4c": null
             },
+            "nodegroup": {
+              "cardinality": "1",
+              "legacygroupid": null,
+              "nodegroupid": "677f236e-09cc-11e7-8ff7-6c4008b05c4c",
+              "parentnodegroup_id": null
+            },
```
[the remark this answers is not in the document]
chrabyrd: @apeters 

This was fun, it took a while to track down. 

It's here: https://github.com/archesproject/arches/blob/dev/7.6.x/arches/app/utils/betterJSONSerializer.py#L209

Essentially because `nodegroup` is now a property on `TileModel` it gets serialized since `TileModel` doesn't override the default `serialize` method.

To fix this, I added `TileModel.serialize()` with a default exclude of `['nodegroup']`

jacobtylerwalls (approved):
Thanks for the updates, all my comments are addressed 👍

apeters (approved):
This looks great @chrabyrd !  thanks for sticking it out ❤️

Issue #9632, opened 2023-06-12
Add lookup resource instance by ES query


ryan86:
Add method in resource utils to get resource instance(s) by ES query

PR #9633, opened 2023-06-12
Add lookup resource instance by ES query, re #9632

ryan86:
Add lookup resource instance by query method re #9632 .

### DIFF
--- a/arches/app/media/js/utils/resource.js
+++ b/arches/app/media/js/utils/resource.js
@@ -25,6 +25,28 @@ define(['arches'], function(arches) {
         },
 
         /**
+         * lookupResourceInstanceDataByQuery - gets mlutiple resource instance data from Elastic Search
+         *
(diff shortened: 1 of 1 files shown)


arches/app/media/js/utils/resource.js
```diff
@@ -24,6 +24,28 @@ define(['arches'], function(arches) {
             }
         },
 
+        /**
+         * lookupResourceInstanceDataByQuery - gets mlutiple resource instance data from Elastic Search
```
chiatt, line 28: Can you change to something like this?
``` * lookupResourceInstanceDataByQuery - gets resource instance(s) from Elasticsearch using custom query```

chiatt (changes requested):
Looks great - Just one very minor doc string change.

chiatt (approved):

Issue #9634, opened 2023-06-12
Add listener to manifest manager formData


ryan86:
Add listener to manifest manager formData to enable workflows to access file upload data.

PR #9635, opened 2023-06-12
Add listener to manifest manager to allow workflow to access file upl…

ryan86:
Add listener to manifest manager to allow workflows to access file upload data, re #9634


chiatt (approved):
Looks good!

Issue #9579, opened 2023-05-31
Add flag to es management command to recalculate primary descriptors


bferguso:
If the resources being reindexed already have calculated primary descriptors stored in the `resource_instances` table those values are used when reindexing the resources. If the primary descriptors function has been modified new values are not used when indexing resources using the es management command.

It would be helpful if a `recalculate_descriptors` argument was added to the `es` management command so that a recalculation of the resource primary descriptors can be done.

bferguso:
Completed with PR #9636

PR #9636, opened 2023-06-12
Ensure descriptor recalc is performed after bulk edit, re #9579

chiatt:
### Description of Change
Adds argument to bulk load logic to recalculate descriptors on bulk load.

### Issues Solved

#9579


bferguso:
Looks good!

Issue #9651, opened 2023-06-14
Map layer model ispublic=True by default


bferguso:
To be consistent, if the MapLayer `ispublic` attribute isn't set it should default to `True`. Currently it defaults to `False` so layers are hidden unless specifically granted permission.

PR #9655, opened 2023-06-14
9651 map layer default public

bferguso:
### Types of changes

### Description of Change

Change the default value of MapLayer.ispublic in the model from False to True

### Issues Solved

#9651 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @bferguso
*   Tested by: @bferguso
*   Designed by: @bferguso

### Further comments
Set dev/7.4.x as target as we probably want to either add this to the 7.4.0 if possible or include it in the next patch release. Should be applied to dev/7.5.x too.


chiatt (approved):
Looks good!

PR #9675, opened 2023-06-19
Gives dev user admin page perms

chiatt:
Extends permissions of the dev user so that they can perform essential actions in the admin page.


arches/management/commands/add_test_users.py
```diff
@@ -46,17 +47,28 @@ def add_users(self):
...
                     "RDM Administrator",
                     "Resource Reviewer",
                     "System Administrator",
+                    "Developer",
                 ],
             },
         )
 
+        try:
+            dev_group = Group.objects.create(name='Developer')
```
chrabyrd, line 64: Should the Developer group be created here, or should it be treated like other groups?

chiatt: That's a good question, but I'm reluctant to add another group that is not required for most Arches deployments. I think it is good to keep the number of groups that appear in the admin page as small as possible. If we were going to make changes to the default list of groups, we should actually consider removing `crowdsource editor` and `mobile project administrator`.

chrabyrd:
lgtm!

PR #9680, opened 2023-06-20
fix for download_ready_email_notification template

aarongundel:
The download ready email template will not always have a zip file available - sometimes the zip is not generated if the files are too big to be zipped.


aarongundel:
@chiatt this seemed minor enough for 7.4.1 - but feel free to rebase if you feel otherwise.

chiatt (approved):
Looks good!

PR #9685, opened 2023-06-20
fixes null check errors in the resource instance select widget

aarongundel:
fixes null check errors in the resource instance select widget.


chiatt (changes requested):
The fix for the relationship status works great, but `.iconclass` also needs to be null checked on line 273 in the makeObject method.

chiatt (approved):

PR #9696, opened 2023-06-27
adds template loading to package load process

aarongundel:
Adds template loading to package load command.  For this to work, the arches templating application must be installed, there must be properly named templates in the "templates" directory inside of the package, and PR https://github.com/archesproject/arches-templating/pull/10 must be merged


chiatt (approved):
Looks good!

PR #9699, opened 2023-06-28
fix for transaction reversal 403 

### LINKED ISSUE #9698: Transaction reversal fails with 403 unexpectedly
This applies mostly for workflows - workflows have a delete button.  This delete button allows you to roll back the workflow and move on.  This delete button calls the transaction reversal method.  That method fails with a 403 if the user executing the call is not a superuser and no transactions exist - effectively causing the "delete" button to fail.

aarongundel:
fixes #9698 - where transaction reversal fails if no records exist in the transaction table for non-superusers.  The fix modifies the user_created_transaction method to return "true" if there are no transactions to be rolled back.  Since there are no transactions, this will effectively make the reverse method a no-op that returns 200, rather than causing it to fail with 403.

### DIFF
--- a/arches/app/utils/permission_backend.py
+++ b/arches/app/utils/permission_backend.py
@@ -522,7 +522,10 @@ def user_created_transaction(user, transactionid):
     if user.is_authenticated:
         if user.is_superuser:
             return True
-        if EditLog.objects.filter(transactionid=transactionid, userid=user.id).count() > 0:
+        if EditLog.objects.filter(transactionid=transactionid).count() > 0:
+            if EditLog.objects.filter(transactionid=transactionid, userid=user.id).count() > 0:
+                return True
+        else:
             return True
     return False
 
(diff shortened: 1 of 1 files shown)


arches/app/utils/permission_backend.py
```diff
@@ -522,7 +522,10 @@ def user_created_transaction(user, transactionid):
     if user.is_authenticated:
         if user.is_superuser:
             return True
-        if EditLog.objects.filter(transactionid=transactionid, userid=user.id).count() > 0:
+        if EditLog.objects.filter(transactionid=transactionid).count() > 0:
```
chiatt, line 525: Using `exists` might be more performant here than `count`. 
https://dev.to/codereviewdoctor/why-queryset-exists-is-more-efficient-than-queryset-count-2f3h

chiatt (changes requested):
It seems like we should use `exists` here rather than `count`. That probably makes sense for line 526 as well.

chiatt (approved):
Looks good!

Issue #9054, opened 2022-10-25
Design and implement a pattern for database migrations in a project to migrate tiles


robgaston:
To facilitate evolving project models, we need to build a way for Arches projects to define migrations.

Currently, there is not an obvious way to define migrations that are specific to an Arches project.  We need to establish a pattern for this, and design/build tools for updating business data in response to model changes.

This initial work should be to establish the pattern and define the basic requirements for migrating tiles based on model changes.  Based on these requirements, we should open additional issues for developing features to support them.

(RDM Phase II, 7.2.x)

PR #9701, opened 2023-06-29
9054 cbyrd resource instance migration

chrabyrd:
### Types of changes

### Issues Solved

#9054


chiatt:
This feature could be used - not only for tile data, but other objects that must be migrated after graphs and reference data are loaded (e.g. report configs and spatial views).

chiatt (changes requested):
Could you add an entry in the changelog?

chiatt:
Moving this effort over to: https://github.com/archesproject/arches/pull/12840

PR #9703, opened 2023-06-29
purges count method in favor of exists

aarongundel:
purges count in favor of exists (where logical) to improve performance.


chiatt (approved):
Looks good!

Issue #9648, opened 2023-06-13
Provisional edits for grouping cards break the resource & report pages


johnatawnclementawn:
Requirements to reproduce:
1. Resource with n>0 grouping cards
2. Access to both provisional user & resource reviewer accounts

Steps to reproduce:
1. With the provisional user account, create a resource & add data to the grouping card tiles
2. With the resource reviewer account, visit [http://localhost:8000/report/](http://localhost:8000/report/${resourceinstance_id}) or [http://localhost:8000/resource/](http://localhost:8000/resource/${resourceinstance_id})
4. Observe broken page

Screenshots:
1. Provisional Editor: 'boolean' is a grouping card encompassing number & string
![ProvisionalEditor](https://github.com/archesproject/arches/assets/22161487/b92a2aa3-85e6-4e43-925f-e64d7d767f5e)

2. Resource Reviewer:
![ResourceReviewer](https://github.com/archesproject/arches/assets/22161487/27750f86-f569-4e64-ac4a-50ed51f92da2)

PR #9704, opened 2023-06-29
Fixes errors when creating provisional edits

chrabyrd:
### Types of changes

### Issues Solved

#9648

### DIFF
--- a/arches/app/models/migrations/9648_add_empty_key_value_pairs_to_tiles.py
+++ b/arches/app/models/migrations/9648_add_empty_key_value_pairs_to_tiles.py
@@ -1,0 +1,20 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("models", "9477_fix_for_spatial_view_dbf_function_edtf_displaying_null"),
+    ]
+
+    def forwards_func(apps, schema_editor):
+        TileModel = apps.get_model("models", "TileModel")
+
+        for tile in TileModel.objects.all():
+            tile.save()
+
+    def reverse_func(apps, schema_editor):
+        pass
+
+    operations = [
+        migrations.RunPython(forwards_func, reverse_func),
+    ]
--- a/arches/app/models/migrations/9648_add_empty_key_value_pairs_to_tiles.py
+++ b/arches/app/models/migrations/9648_add_empty_key_value_pairs_to_tiles.py
@@ -10,7 +10,7 @@ class Migration(migrations.Migration):
         TileModel = apps.get_model("models", "TileModel")
 
         for tile in TileModel.objects.all():
-            tile.save()
+            tile.save()  # should trigger new code in `TileModel.save` method that adds empty key/value pairs to tile data
 
     def reverse_func(apps, schema_editor):
         pass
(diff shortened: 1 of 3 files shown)


arches/app/models/migrations/9648_add_empty_key_value_pairs_to_tiles.py
```diff
@@ -0,0 +1,20 @@
...
+
+        for tile in TileModel.objects.all():
+            tile.save()  # should trigger new code in `TileModel.save` method that adds empty key/value pairs to tile data
+
+    def reverse_func(apps, schema_editor):
+        pass
+
+    operations = [
+        migrations.RunPython(forwards_func, reverse_func),
+    ]
```
chiatt, line 20: It's a bit surprising but the updated save method isn't used here, so that code needs to be repeated here. These seem to work forward and backward.

```
    def forwards_func(apps, schema_editor):
        TileModel = apps.get_model("models", "TileModel")
        Node = apps.get_model("models", "Node")

        for tile in TileModel.objects.filter(data={}, provisionaledits__isnull=False):
            for node in Node.objects.filter(nodegroup_id=tile.nodegroup_id):
                if not str(node.pk) in tile.data:
                    tile.data[str(node.pk)] = None
            tile.save() 

    def reverse_func(apps, schema_editor):
        TileModel = apps.get_model("models", "TileModel")

        for tile in TileModel.objects.filter(provisionaledits__isnull=False):
            if bool(tile.provisionaledits and not any(tile.data.values())):
                tile.data = {}
                tile.save() 
```

arches/app/datatypes/datatypes.py
```diff
@@ -1535,7 +1535,7 @@ def append_to_document(self, document, nodevalue, nodeid, tile, provisional=Fals
             for f in tile.data[str(nodeid)]:
                 val = {"string": f["name"], "nodegroup_id": tile.nodegroup_id, "provisional": provisional}
                 document["strings"].append(val)
-        except KeyError as e:
+        except (KeyError, TypeError) as e:
```
chiatt, line 1538: Maybe we don't need the KeyError any longer?

chiatt (changes requested):
This is looking really good, but the migration didn't seem to work for me. I added a suggestion inline that seems to work ok.

arches/app/models/migrations/9648_add_empty_key_value_pairs_to_tiles.py
chrabyrd: Ah -- that _is_ strange and I wonder why it worked on my local but not yours. I'm happy to instead of relying on the `Tile.save()` method to duplicate the logic here.  But I want to make sure we're thinking the same thing about the logic. 

There's a reason I didn't write a `reverse_func`. The `forwards_func` is straightforward but without collecting some metadata there's no way to reliably reverse the changes made to the database -- I assume there can be cases of blank k/v pairs in tile data before we apply these changes. It could be that it's safe to delete those k/v pairs on reverse but I wanted to err on the side of caution.

The functions listed above work well, but don't match the logic in the `TileModel` -- that is that the logic added to `TileModel` creates blank k/v pairs regardless of provisional edits.

arches/app/datatypes/datatypes.py
chrabyrd: Resolving this as-is for now, please let me know if that needs to be changed.

chiatt (approved):
Looks good to me!

Issue #9706, opened 2023-06-30
Importing concept-list node business data with the --create_concepts flag does not work


atapscott:
Subject: RDM

When importing business data with csv format, using the *--create_concepts create* flag, cells mapped to a node with a concept-list data type do not get properly processed.
This is due to how it is checked if the text string, used as preferred label, is already present in the collection.
Specifically, this happens in [csvfile.py](https://github.com/archesproject/arches/blob/master/arches/app/utils/data_management/resources/formats/csvfile.py) line 625, in the import_business_data() method:
```
elif row[node["file_field_name"]] not in list(concepts_to_create[node["arches_nodeid"]].values()):
    for concept_value in concept:
        concepts_to_create[node["arches_nodeid"]][str(uuid.uuid4())] = concept_value
```
*row[node["file_field_name"]]* contains a list of concept values separated by commas, and this joined string is used to compare against every single value individually (stored in *concepts_to_create[node["arches_nodeid"]]*). The result is that all values inside a concept-list are considered new no matter if they exist or not inside the creation (unless the concept-list only contains one concept, in which case it works correctly).
This results in a duplicate key problem when saving the data, because the program attempts to generate the same concepts multiple times.
To solve this, the existence check could be performed in the following line, with each of the concepts instead of the whole joined string.

PR #9707, opened 2023-06-30
9706 concept list create import

atapscott:
This fix is specifically to enable the dynamic concept creation (using the --create_concepts create flag) when importing data from a csv document.
Currently, concept-list nodes can only import single value, not comma-separated values (as mentioned in the docs).

### Types of changes

### Description of Change
It is a simple, small, change to move the comparison from the whole joined concept string to individual values.

### Issues Solved
Solves issue #9706 I created.
#

### Checklist

#### Ticket Background
*   Sponsored by: Arcadia Foundation, MAPHSA project.


chiatt (approved):
Looks good to me!

chiatt:
Hi @atapscott, Thanks very much for this fix. I didn't catch that the target branch was stable/7.4 rather than dev/7.4.x before merging. We only develop against the dev branches as that is what ends up in future releases. I'm moving this commit into dev/7.4.x and I will be removing it from stable/7.4.x.

PR #9718, opened 2023-07-05
yaml templating for template configuration

aarongundel:
changes load templates call to use yaml configuration files


chiatt (approved):
Looks good to me!

PR #9720, opened 2023-07-05
Upgrade django to 3.2.20

chiatt:
Upgrade django to 3.2.20

re https://docs.djangoproject.com/en/4.2/releases/3.2.20/


aarongundel (approved):
LGTM

Issue #9746, opened 2023-07-07
Saving a card with related resource nodes either times out or takes much longer than cards w/o those node data types.


apeters:
If a card is saved that has a related resource node and that node is populated, then the save operation can be orders of magnitude slower.  On very large datasets (100M tiles, 13M resource instances) this operation can time out.

If you have a card w/o a related resource node, or the card does have a related resource node, but you don't save a value to that node then the save operation is similar to saves of other cards with other datatypes.

apeters:
link to original and optimized analyzed execution plans:
https://explain.depesz.com/s/Zsneq

PR #9747, opened 2023-07-08
fix type cast issue with __arches_create_resource_x_resource_relationship function

apeters:
re #9746


apeters:
link to original and optimized analyzed execution plans:
https://explain.depesz.com/s/Zsneq

chiatt (approved):
Looks good to me!

PR #9753, opened 2023-07-10
uses serialized graph in tile model

aarongundel:
Uses serialized graph in tile model.  This was already used in quite a few places here, there were only a few changes to make.

### DIFF
--- a/arches/app/models/tile.py
+++ b/arches/app/models/tile.py
@@ -570,7 +570,7 @@ class Tile(models.TileModel):
                     
             nodes = [(node for node in self.serialized_graph["nodes"] if node["nodegroup_id"] == self.nodegroup_id)]
             nodes = [SimpleNamespace(**next(node, None)) for node in nodes]
-        except:
+        except TypeError: # handle if serialized_graph is None
             nodegroup = models.NodeGroup.objects.get(pk=self.nodegroup_id)
             nodes = nodegroup.node_set.all()
 
--- a/arches/app/models/tile.py
+++ b/arches/app/models/tile.py
@@ -335,7 +335,7 @@ class Tile(models.TileModel):
         for nodeid, value in self.data.items():
             try:
                 node = SimpleNamespace(**next((x for x in self.serialized_graph["nodes"] if x["nodeid"] == nodeid), None))
-            except:
+            except TypeError: # will catch if serialized_graph is None
                 node = models.Node.objects.get(nodeid=nodeid)
             datatype = self.datatype_factory.get_instance(node.datatype)
             error = datatype.validate(value, node=node, strict=strict, request=request)
--- a/arches/app/models/tile.py
+++ b/arches/app/models/tile.py
@@ -106,10 +106,13 @@ class Tile(models.TileModel):
                         tile = Tile(tile_obj)
                         tile.parenttile = self
                         self.tiles.append(tile)
+        self.load_serialized_graph()
+
+    def load_serialized_graph(self):
         try:
             published_graph = self.resourceinstance.graph.get_published_graph()
             self.serialized_graph = published_graph.serialized_graph
-        except:
+        except Exception as e:
             self.serialized_graph = None
 
     def save_edit(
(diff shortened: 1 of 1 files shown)


arches/app/models/tile.py
```diff
@@ -525,7 +530,10 @@ def delete(self, *args, **kwargs):
             try:
                 super(Tile, self).delete(*args, **kwargs)
                 for nodeid in self.data.keys():
-                    node = models.Node.objects.get(nodeid=nodeid)
+                    try:
+                        node = SimpleNamespace(**next((x for x in self.serialized_graph["nodes"] if x["nodeid"] == nodeid), None))
+                    except:
```
chiatt, line 535: It looks like this is probably a TypeError we are excepting here if self.serialized_graph is None. If that's the case, I think we should probably be explicit about that even if the odds of other errors are very low.

chiatt (changes requested):

arches/app/models/tile.py
```diff
@@ -598,8 +614,14 @@ def get_blank_tile_from_nodegroup_id(nodegroup_id, resourceid=None, parenttile=N
         tile.resourceinstance_id = resourceid
         tile.parenttile = parenttile
         tile.data = {}
+        tile.load_serialized_graph()
```
jacobtylerwalls, line 617: Because `load_serialized_graph()` does a db query, we're probably giving back all the intended performance savings. I noticed the two implementations perform roughly the same (this is with a nodegroup from AfS):


```py
python3 manage.py shell
>>> from timeit import timeit
>>> from arches.app.models.tile import Tile
>>> timeit("Tile.get_blank_tile_from_nodegroup_id('0434ccce-1d9f-11eb-a29f-024e0d439fdb')", globals=globals(), number=5000)  # pr
3.2668450839992147
>>> timeit("Tile.get_blank_tile_from_nodegroup_id('0434ccce-1d9f-11eb-a29f-024e0d439fdb')", globals=globals(), number=5000)  # dev branch
3.308565291983541
```

I can take a closer look at this tomorrow, but what do you think about that point in general, in other words only trying to do this (potential) optimization only if we already have the published graph in hand?

arches/app/models/tile.py
```diff
@@ -557,8 +565,16 @@ def flatten_tiles(obj):
...
-        nodegroup = models.NodeGroup.objects.get(pk=self.nodegroup_id)
-        for node in nodegroup.node_set.all():
+        try:
+            nodegroup = SimpleNamespace(**next((x for x in self.serialized_graph["nodegroups"] if x["nodegroupid"] == self.nodegroup_id), None))
+                    
+            nodes = [(node for node in self.serialized_graph["nodes"] if node["nodegroup_id"] == self.nodegroup_id)]
+            nodes = [SimpleNamespace(**next(node, None)) for node in nodes]
+        except:
+            nodegroup = models.NodeGroup.objects.get(pk=self.nodegroup_id)
+            nodes = nodegroup.node_set.all()
```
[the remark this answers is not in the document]
aarongundel: This one should stay; this is meant to catch if serialized graph is None - in which case the nodegroup wouldn't be set.  But the catch should be more explicit so that the typeerror is caught.

arches/app/models/tile.py
aarongundel: The idea here is that it's likely the caller of this method will not just grab the empty tile and be finished, but that they'll set some nodes in the tile and then save it back to the database - at which point this becomes a bit more valuable because then the serialized graph doesn't have to be loaded on save.  So this is a prefetch of the the serialized graph for that tile object.  However, it is true that this is not strictly needed here, and it's probably better left for when it is actually needed - the optimization is a bit premature.

arches/app/models/tile.py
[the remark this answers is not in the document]
jacobtylerwalls: Yep I'm with you on why the except clause is here, I'm just suggesting rewriting the content of the except clause.

> in which case the nodegroup wouldn't be set

I didn't follow this part -- Django will lazily fetch `nodegroup`, so it's better to rely on that than always doing the extra query AFAIU.

jacobtylerwalls (approved):
Thanks, looks great!

Issue #9758, opened 2023-07-11
Move call to save a resource instance out of resource.index method


chiatt:
Ideally we would not make changes to data in postgres when we are indexing a resource in ElasticSearch

chiatt:
The reason why we are calling save within index is because we need to calculate and set the descriptors on the database instance record. We can't put it in resource.save because we don't call resource.save when saving a tile. Instead, we just save the tile and then reindex the resource. Calling save on the resource with in save ensure that save is always called after all tiles are saved.

chrabyrd:
resolved in #9967

PR #9763, opened 2023-07-11
removes the save call from resource.index re #9758

aarongundel:
Research suggests this call is likely no longer needed, so this PR purges the save call from index to improve performance.


chiatt:
It looks like the purpose of this line was to save the descriptors to the resource instance in postgres. Removing it seems harmless, but we lose the performance benefit of having the descriptors available on the instance. There is a call to resource.save in `index_database` doing the same thing. 

I think these save calls should probably be replaced by a change to `resource.calculate_descriptors`. Calling resource.save just to set the descriptors is inefficient because we end up resaving tiles. To avoid this, we could pass an argument to calculate_descriptors indicating that the resource instance should be updated with the descriptors. Then we could use raw sql to update the instance in postgres efficiently.

chiatt:
Closing in favor of #9967

Issue #9561, opened 2023-05-25
JSON-LD Assertion Error where parent concept-node value unknown, when child concept-node has value. 


khodgkinson-he:
**Steps to Recreate :** 

1. Add Semantic Node to Resource Model. "New Node"
2. Add Concept Node below "New Node". Set Concept Collection. "New Node Defaultless"
3. Add Concept Node below "New Node Defaultless". Set Concept Collection. Set Default Value on Card tab. "New Node WithDefault"

![image](https://github.com/archesproject/arches/assets/53860784/fc456a68-cc8e-4cf3-83cf-4eee5b1da30b)

5. In resource instance, save card with empty "New Node Defaultless"

![image](https://github.com/archesproject/arches/assets/53860784/b7ddc6f9-fc79-4147-8cbd-638b95c95b71)

6. Get JSON-LD for resource.

![image](https://github.com/archesproject/arches/assets/53860784/399db3fe-4659-43a8-be54-1180eebf93ba)

The Unknown value of the parent "New Node Defaultless" is not contained within it's concept collection and so (correctly) fails validation.

**Parent Concept-Node value not in concept collection.**

![image](https://github.com/archesproject/arches/assets/53860784/71de29d5-924a-459f-a18f-4990a1d4f8ac)

@aj-he @chiatt @apeters @robgaston I started digging into this one with regards to https://github.com/archesproject/arches-her/issues/870.  

We implemented a quick workaround (https://github.com/HistoricEngland/arches/pull/253) on the HistoricEngland fork to treat the parent node as Undefined rather than Unknown.  Though this was suitable in our instance, it could lead to data loss in other situations.

..as to a more general solution, we've been discussing options such as:

- Blocking default values for children of no-default parents..
- Enforcing default values for parents of default value children..
- Disabling some of the validation on the Graph checking..

..though there may, of course, be some we've missed..

It'd be great if we could get your thoughts on the matter, and any preferred solutions?

Many thanks!  ^_^

PR #9783, opened 2023-07-17
9561 Json-Ld fix : Omit Edge from Null Domain

khodgkinson-he:
Issue : JSON-LD concept model handling does not handle null **domain nodes** whilst there is data in the **range node** on rdffile.py edge relationship generation.

ArchesProject GitHub Giveback Issue : #9561 

Fix : Do not parse the null. Domain node as Undefined (omitted from JSON) substitution used for Unknown (NULL value).

@chiatt @aj-he : Happy to rebase to dev/7.5.x if that'd be of help!  ^_^

### DIFF
--- a/arches/app/utils/data_management/resources/formats/rdffile.py
+++ b/arches/app/utils/data_management/resources/formats/rdffile.py
@@ -149,6 +149,11 @@ class RdfWriter(Writer):
             # Nothing to do here
             if pkg["r_uri"] is None and pkg["range_tile_data"] is None:
                 return
+            
+            # JSON-LD fails assert if domain node empty while range node has data. 
+            # Unknown!=Undefined, but reasonable substitution to omit edge from null domain.
+            if pkg["d_uri"] is None:
+                return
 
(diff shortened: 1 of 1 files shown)


aj-he:
@chiatt assigned to you for review. Delegate as you see fit.

chiatt:
Thanks @khodgkinson-he, I think dev/7.4.x is the right target for this change.

chiatt (changes requested):
I think that we should probably inform the admin why there is data missing from the response. 
Maybe something like this?
```
if pkg["d_uri"] is None and pkg["range_tile_data"]:
    self.logger.warn(_("Unable to return range value because domain is None, re https://github.com/archesproject/arches/pull/9783/files"))
    return
```

khodgkinson-he:
@chiatt : Excellent suggestion to inform admin logs!  I've tweaked that code accordingly, sir! ^_^

chiatt:
Thanks @khodgkinson-he!

chiatt (approved):

PR #9793, opened 2023-07-18
Integrates Vue

chrabyrd:
### Types of changes

### Issues Solved

#9789


arches/webpack/webpack-utils/build-vue-filepath-lookup.js
```diff
@@ -0,0 +1,39 @@
...
+        
+        if (fs.lstatSync(Path.join(path, name)).isDirectory() ) {
+            return buildVueFilePathLookup(
+                Path.join(path, name), 
+                acc, 
+                outerPath
+            );
+        }
+        else {
+            let subPath = (Path.join(path, name)).split(/frontend(.*)/s)[1];  // splits only on first occurance
```
chiatt, line 20: As you mentioned @chrabyrd, we might want to change this directory as the location for vue components. Maybe something like `../media/vue_components`? In any case we can let the team weigh in and change it to something else before the 7.5 release if we need to.

chiatt (approved):
Looks good to me!

Issue #9623, opened 2023-06-09
Quotation marks in resource model abstract causes Arches crash


SDScandrettKint:
Issue found in Arches 7.3.0 and present when testing in dev/7.4.x 

Original post regarding issue - https://community.archesproject.org/t/application-feedback-input-that-could-kill-the-application/1997

Steps for my reproducing of the error:

Adding a description/abstract in a resource model which comprises only text inside quotations. 
Then saving and publishing the graph will result in an Unpublished Model Error:
![image](https://github.com/archesproject/arches/assets/119508494/7cf88349-434d-4401-b079-395a08da1e83)
Refreshing will show the following django error:
![image](https://github.com/archesproject/arches/assets/119508494/dc342889-f7ed-4148-b5ab-c3428735ac72)

The database entry looks as follows:
![image](https://github.com/archesproject/arches/assets/119508494/4fab54a8-80ff-4c02-8302-45f2f9143f49)

Interestingly, when I try to save a model description with just plain text (no quotations) and publish the graph it successfully saves, but the database description value stays as null and refreshing the page shows the text didn't actually save. It should save with the database entry {"en": WITH THE TEXT HERE}.

However, if I were to input and save the following description {"en": "This text should appear"}, that would save correctly and Arches would show the text. This is now able to be changed and saved successfully until I attempt to add a description that starts and ends with quotation marks, then the "en" part is deleted.    

Writing a description that either either starts but doesn't end or ends but doesn't start with a quotation mark e.g. HELLO "WORLD" successfully saves to the database.
![image](https://github.com/archesproject/arches/assets/119508494/92d882bf-782b-43ad-a790-23d8437aab33)
![image](https://github.com/archesproject/arches/assets/119508494/ba279e5d-0d65-4f65-aa1e-3f12fa12d3f6)

FURTHER UPDATES:

This doesn't just occur for description but also subtitle and presumably any text input that has internationalisation features.

PR #9794, opened 2023-07-19
Fix for allowing double quotation marks in i18n inputs (dev/7.4.x)

SDScandrettKint:
This fix allows users to have strings that start and end with double quotation marks (" ") in i18n text inputs.
### Types of changes

### Description of Change
Inputting a string that starts and ends with double quotation marks (e.g. "Hello, world") into any Resource Model text input box that used internationalisation would save incorrectly - e.g. as "Hello, world" rather than {"en":""\Hello, world"\"}. Strings which included double quotations not at the start and end (e.g. "Hello", world) would save correctly (e.g. {"en": ""\Hello"\, world"}. 
This change converts inputs with double quotations at the start and end to the raw literal for the character - "\ .

### Issues Solved

#9623

### Checklist

#### Ticket Background

### DIFF
--- a/arches/app/models/fields/i18n.py
+++ b/arches/app/models/fields/i18n.py
@@ -22,6 +22,10 @@ class I18n_String(object):
         ret = {}
 
         if isinstance(value, str) and value != "null":
+            # fix for issue #9623 - using double quotation marks in i18n input
+            if value.startswith('"') and value.endswith('"'):
+                value = value.replace('"', r'\"')
+
             try:
                 ret = json.loads(value)
             except:
--- a/arches/app/models/fields/i18n.py
+++ b/arches/app/models/fields/i18n.py
@@ -22,12 +22,20 @@ class I18n_String(object):
         ret = {}
 
         if isinstance(value, str) and value != "null":
-            # fix for issue #9623 - using double quotation marks in i18n input
-            if value.startswith('"') and value.endswith('"'):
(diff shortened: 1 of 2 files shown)


arches/app/models/fields/i18n.py
```diff
@@ -22,6 +22,10 @@ def _parse(self, value, lang, use_nulls):
         ret = {}
 
         if isinstance(value, str) and value != "null":
+            # fix for issue #9623 - using double quotation marks in i18n input
```
chiatt, line 25: You can remove this line. We can always trace back to the ticket through a git blame.

chiatt (changes requested):
This does prevent the error from occurring, but it may result in some unintended behavior. If a string has already been saved with quotes escaped, the current change will escape them a second time. When that happens you can end up with something like this: 

image

I think the solution might be to check if there are quotes around the string and if those quotes are already escaped. This regex pattern might work: 
```
import re
unescaped_quotes = r'^(?<!\\)".*"(?<!\\)$'
if re.match(unescaped_quotes, value):
        value.replace('"', r'\"')
``` 
@apeters pointed out that because this is such an important part of the Arches code, we'd really need a unit test to ensure that strings are getting parsed properly.

apeters:
I just pushed a small fix that I think handles for the issue that @chiatt described.  I also added a test.

chiatt (approved):

Issue #9777, opened 2023-07-14
Graphs with "/" in name cause error in create_package


SDScandrettKint:
Found/tested in Arches 6 & 7

If a graph has a slash in its name, the `create_package` process will cause the following error because it is recognising the slash as a directory:
```
Traceback (most recent call last):
  File "manage.py", line 30, in <module>
    execute_from_command_line(sys.argv)
  File "/mnt/c/arches_dev/v7_development/env_v7dev/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/mnt/c/arches_dev/v7_development/env_v7dev/lib/python3.8/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/mnt/c/arches_dev/v7_development/env_v7dev/lib/python3.8/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/mnt/c/arches_dev/v7_development/env_v7dev/lib/python3.8/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/mnt/c/arches_dev/v7_development/arches/arches/management/commands/packages.py", line 359, in handle
    self.create_package(options["dest_dir"])
  File "/mnt/c/arches_dev/v7_development/arches/arches/management/commands/packages.py", line 520, in create_package
    self.export_resource_graphs(os.path.join(dest_dir, "graphs", "resource_models"), "true")
  File "/mnt/c/arches_dev/v7_development/arches/arches/management/commands/packages.py", line 407, in export_resource_graphs
    with open(output_file, "w") as f:
FileNotFoundError: [Errno 2] No such file or directory: 'testingpackage3/graphs/resource_models/An example name / or another.json'
```

PR #9795, opened 2023-07-19
Fix for graphs with forward slash ("/") in name when using create_package (dev/7.4.x)

SDScandrettKint:
Graphs with forward slash in cause an error when running create_package.

### Types of changes

### Description of Change
If a graph name has a forward slash ("/") in the name, replace it with a dash before writing the file. The graph name will still include the forward slash but the file written will have replaced it with a dash so linux doesn't recognise it as a directory.

### Issues Solved

#9777 

### Checklist

#### Ticket Background
*   Sponsored by: 

### Further comments

Unsure if back slash also needs changing for windows paths.


chiatt (approved):
Looks good to me!

PR #9810, opened 2023-07-21
 9726 Give screen reader context for how many results are shown

### LINKED ISSUE #9726: Search Results and Pagination - 4.1.3 Status Message - Screen reader does not read out number of results
As a user reliant on a screen reader, I should be able to determine the number of results (be that many or none) displayed when I enter a search term.

**Linked Work Item:**
#9659 

**Expected Result:**
When a user enters a valid or non-valid search term, selects a result from the drop-down and hits Enter, the screen reader should then read out the number of results returned by the user's search.

**Actual Result:**
The number of search results is not read out by the screen reader. Instead, the screen reader reads, "Main landmark, Search results region, edit blank" (or something to that effect) every time, no matter how many results are displayed.

**Evidence:**

(Use NVDA to replicate)

johnatawnclementawn:
### Types of changes

### Description of Change

Screen readers need context for how many results are show, not just total number of results. 

### Issues Solved

Resolves #9726

### DIFF
--- a/arches/app/templates/views/search.htm
+++ b/arches/app/templates/views/search.htm
@@ -33,7 +33,7 @@
                     <div class="col-xs-6">
                         <!-- Result Count and ordering Block -->
                         <div class="search-count-container">
-                            <p class="search-title" aria-live="polite" data-bind="text: $root.translations.shownTotalSearchResults(sharedStateObject?.searchResults?.results?.hits?.hits?.length, total())"></p>
+                            <p class="search-title" aria-live="polite" data-bind="text: $root.translations.shownTotalSearchResults(hits(), total())"></p>
                         </div>
                     </div>
                     <div class="col-xs-6">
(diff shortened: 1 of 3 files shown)


arches/app/templates/views/search.htm
```diff
@@ -33,7 +33,7 @@
                     <div class="col-xs-6">
                         <!-- Result Count and ordering Block -->
                         <div class="search-count-container">
-                            <p class="search-title" aria-live="polite" data-bind="text: {% quoted_trans 'Results: ' %} + total()"></p>
+                            <p class="search-title" aria-live="polite" data-bind="text: $root.translations.shownTotalSearchResults(sharedStateObject?.searchResults?.results?.hits?.hits?.length, total())"></p>
```
apeters, line 36: @johnatawnclementawn Instead of referencing this nested object value here for the returned hits, I'd create an observable in the .js file and use that instead here.

apeters (changes requested):

apeters (approved):
that's a nice improvement

PR #9849, opened 2023-07-28
Fix incorrect language labels on concept select widgets #9848

### LINKED ISSUE #9848: Concept label language is sometimes French when it should be English
In the first step of the New Project workflow, the concept label for projects is sometimes spelled with a 'c' and sometimes not. Presumably the label presented is the French label rather than the English one. 

Image

jacobtylerwalls:
### Types of changes

### Description of Change
Before, the `prefLabel` on a concept select widget control used the last label returned from the db, which may or may not correspond to the desired language.

Now, we prefer the user's selected language or the system default language when there are labels in multiple languages. (This requires sorting the best language last given how the recursive search overwrites previous values.)

### Issues Solved

closes #9848

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @chiatt
*   Designed by: @ 

### Further comments
Alternatives considered:

- doing the sort in the SQL query
  - it works, but it requires ugly CASE WHEN statements repeated in both parts of the recursive and non-recursive query, as well as additional interpolated query values
- short-circuiting inside the part of `_findNarrower` that does the overwriting, with a call to a helper function `get_preflabel_from_conceptid`
  - it possibly works, but the helper function adds overhead/queries, and the short-circuiting makes a dense part of `_findNarrower` denser. Or, it doesn't work, if there are nested concept labels that wouldn't be visited because of the short-circuiting.

Eager to hear suggestions about more idiomatic/reliable ways to do this kind of lookup/sorting.

### DIFF
--- a/arches/app/models/concept.py
+++ b/arches/app/models/concept.py
@@ -1189,7 +1189,7 @@ class Concept(object):
         cursor.execute(
             """
             WITH RECURSIVE children AS (
-                SELECT d.conceptidfrom, d.conceptidto, c2.value, c2.valueid as valueid, c.value as valueto, c.valueid as valueidto, c.valuetype as vtype, 1 AS depth, array[d.conceptidto] AS conceptpath, array[c.valueid] AS idpath        ---|NonRecursive Part
+                SELECT d.conceptidfrom, d.conceptidto, c2.value, c2.valueid as valueid, c.value as valueto, c.valueid as valueidto, c.valuetype as vtype, c.languageid, 1 AS depth, array[d.conceptidto] AS conceptpath, array[c.valueid] AS idpath        ---|NonRecursive Part
                     FROM relations d
                     JOIN values c ON(c.conceptid = d.conceptidto)
                     JOIN values c2 ON(c2.conceptid = d.conceptidfrom)
@@ -1198,7 +1198,7 @@ class Concept(object):
                     and c.valuetype in ('prefLabel', 'sortorder', 'collector')
                     and (d.relationtype = 'member' or d.relationtype = 'hasTopConcept')
                     UNION
-                    SELECT d.conceptidfrom, d.conceptidto, v2.value, v2.valueid as valueid, v.value as valueto, v.valueid as valueidto, v.valuetype as vtype, depth+1, (conceptpath || d.conceptidto), (idpath || v.valueid)   ---|RecursivePart
+                    SELECT d.conceptidfrom, d.conceptidto, v2.value, v2.valueid as valueid, v.value as valueto, v.valueid as valueidto, v.valuetype as vtype, v.languageid, depth+1, (conceptpath || d.conceptidto), (idpath || v.valueid)   ---|RecursivePart
                     FROM relations  d
                     JOIN children b ON(b.conceptidto = d.conceptidfrom)
                     JOIN values v ON(v.conceptid = d.conceptidto)
@@ -1206,7 +1206,7 @@ class Concept(object):
                     WHERE  v2.valuetype = 'prefLabel'
                     and v.valuetype in ('prefLabel','sortorder', 'collector')
                     and (d.relationtype = 'member' or d.relationtype = 'hasTopConcept')
-                ) SELECT conceptidfrom::text, conceptidto::text, value, valueid::text, valueto, valueidto::text, depth, idpath::text, conceptpath::text, vtype FROM children ORDER BY depth, conceptpath;
+                ) SELECT conceptidfrom::text, conceptidto::text, value, valueid::text, valueto, valueidto::text, languageid, depth, idpath::text, conceptpath::text, vtype FROM children ORDER BY depth, conceptpath;
             """,
             [conceptid],
         )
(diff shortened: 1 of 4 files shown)


arches/app/models/concept.py
```diff
@@ -1264,8 +1265,18 @@ def _findNarrower(val, path, rec):
                             _findNarrower(child, path, rec)
                 val.children.sort(key=lambda x: (x.sortorder, x.text))
 
-        for row in rows:
-            rec = dict(list(zip(column_names, row)))
+        def best_language_last(rec):
```
apeters, line 1268: We now have several places in the code that try to score/rank the match of a concept label for a given language.  

- https://github.com/archesproject/arches/blob/master/arches/app/models/concept.py#L797
- https://github.com/archesproject/arches/blob/master/arches/app/models/concept.py#L1403
- https://github.com/archesproject/arches/blob/master/arches/app/utils/i18n.py#L52

It would be nice to centralize this logic into a shared method if possible.

apeters (changes requested):

arches/app/views/search.py
```diff
@@ -172,7 +172,10 @@ def search_terms(request):
                 if len(result["top_concept"]["buckets"]) > 0:
                     for top_concept in result["top_concept"]["buckets"]:
                         top_concept_id = top_concept["key"]
-                        top_concept_label = get_preflabel_from_conceptid(top_concept["key"], lang)["value"]
+                        top_concept_label = get_preflabel_from_conceptid(
+                            top_concept["key"],
+                            lang=lang if lang != "*" else None
```
jacobtylerwalls, line 177: (`*` is only useful for elastic searches and will never match on a language code in the values table.)

arches/app/models/concept.py
jacobtylerwalls: Smart. After our huddle I added a `rank_label()` util in [72e3ccd](https://github.com/archesproject/arches/pull/9849/commits/72e3ccd8ad83f8dfd08d5ad1d9103386ceb81a7d) and was able to use it in the first two links you posted. (The third felt like it was okay as is.)

arches/app/utils/i18n.py
```diff
@@ -101,6 +101,38 @@ def get_localized_value(obj, lang=None, return_lang=False):
         return {found_lang: obj[found_lang]} if return_lang else obj[found_lang]
 
 
+def rank_label(value, *, kind='prefLabel', label_lang, sought_lang=None):
```
apeters, line 104: this pattern of forcing named parameters (using the *) has never been used in Arches.  It took me a while to even find out what this meant.  I would suggest that we remove this and then give `label_lang` a default value of `None`

apeters (changes requested):
It would be nice to add a test for this.  Something that would rank values of all the different combinations and then assert that the ranking between them is correct.  Maybe this could be a new file under `tests/utils/i18n.py`

arches/app/utils/i18n.py
jacobtylerwalls: No problem!

Just to add context for what motivated it, what I like about keyword-only params is that they help guard against confusing very similar arguments for each other. ("What's a label_lang -- the language _this_ label is in or the language this label _should_ be in?") Forcing someone to type out the param name hopefully helps readers/writers of the code reinforce which argument value belongs to which argument, but yeah, just a matter of style, and if we don't use `*` anywhere yet, this isn't the place to do it :-)

jacobtylerwalls: I could probably find a better param name than `label_lang`!

jacobtylerwalls: My understanding of `label_lang` is that we'll always have it, and defaulting to None and guessing English would be less ideal. That is, for "projet" we'll never be in doubt that it's `'fr`'--but let me know if I'm wrong about that.

jacobtylerwalls: Please forgive my waffling--it's just a rank function, so we can be permissive and then just rank it low if the language of the label is unknown 👍

jacobtylerwalls: Updated in [dd84ea7](https://github.com/archesproject/arches/pull/9849/commits/dd84ea794f518db63154519f29c909765cc35e87)

jacobtylerwalls:
Added tests in f1e415b. Let me know how this looks!

tests/utils/i18n.py
```diff
@@ -0,0 +1,57 @@
...
+class I18nTests(SimpleTestCase):
+    def test_rank_label(self):
+        translation.activate("es-ES")
+        self.addCleanup(translation.activate, "en")
+
+        old_system_lang = settings.LANGUAGE_CODE
+        settings.LANGUAGE_CODE = "en-US"
+        self.addCleanup(setattr, settings, "LANGUAGE_CODE", old_system_lang)
+
+        cases = [
```
jacobtylerwalls, line 17: I could also sort these by score, if preferable.

jacobtylerwalls:
Thanks, I like that idea. I adjusted the tests and am happy to add some or adjust further, too.

apeters (approved):
This looks great!

Issue #9850, opened 2023-07-31
Add support for connecting to Elasticsearch via Elastic Cloud


RobOatesHistoricEngland:
Whilst looking in to #9700, HE have a requirement to connect to Elasticsearch 8.x via Elastic Cloud. Specifying the Cloud ID and API Key such as...

```python
ELASTICSEARCH_CONNECTION_OPTIONS = {
    "cloud_id": "<CLOUD_ID>",
    "api_key": "<API_KEY>",
}
```
...is not sufficient. An obscure error message results:

```
[31/Jul/2023 06:53:01] "GET /search_component_data/advanced-search HTTP/1.1" 200 44595
2023-07-31 06:53:01,801 arches.app.search.search ERROR    2023-07-31 06:53:01.801778: WARNING: search failed for query: {'bool': {'should': [], 'must': [], 'must_not': [], 'filter': [{'terms': {'graph_id': ['5e8dcba3-3ac8-11eb-b1c6-c0b8832fa043', '1fe4bf71-3a3c-11eb-8023-c0b8832fa043', 'ff623370-fa12-11e6-b98b-6c4008b05c4c', '86cd5919-45c3-4ef7-9557-d39ceb85210f']}}]}} 
Exception detail: memoryview: a bytes-like object is required, not 'str'
Traceback (most recent call last):
  File "/web_root/arches/arches/app/search/search.py", line 151, in search
    ret = self.es.search(**kwargs).body
  File "/root/.local/lib/python3.9/site-packages/elasticsearch/_sync/client/utils.py", line 414, in wrapped
    return api(*args, **kwargs)
  File "/root/.local/lib/python3.9/site-packages/elasticsearch/_sync/client/__init__.py", line 3863, in search
    return self.perform_request(  # type: ignore[return-value]
  File "/root/.local/lib/python3.9/site-packages/elasticsearch/_sync/client/_base.py", line 285, in perform_request
    meta, resp_body = self.transport.perform_request(
  File "/root/.local/lib/python3.9/site-packages/elastic_transport/_transport.py", line 329, in perform_request
    meta, raw_data = node.perform_request(
  File "/root/.local/lib/python3.9/site-packages/elastic_transport/_node/_http_urllib3.py", line 157, in perform_request
    body_to_send = gzip.compress(body)
  File "/usr/local/lib/python3.9/gzip.py", line 548, in compress
    f.write(data)
  File "/usr/local/lib/python3.9/gzip.py", line 284, in write
[...]

PR #9859, opened 2023-08-02
Add support for Elasticsearch API key authentication #9700 AB#57245

### LINKED ISSUE #9700: Add support for using Elasticsearch API Key authentication
There is a requirement to be able to use API Keys instead of username/password basic auth to connect to and use and Elasticsearch instance. 

It is proposed that we be able to specify api_key details in the settings instead of username and password.

The Elasticsearch python client docs provide the following example
```py
# Set 'api_key' on the constructor
client = Elasticsearch(
    "http://localhost:9200",
    api_key=("id", "api_key")
)
client.search(...)
```

RobOatesHistoricEngland:
### Types of changes

### Description of Change
Update settings.py and settings.py-tpl to show how to implement Elasticsearch API key credentials. This will only work with a non-Elastic Cloud instance of Elasticsearch. #9850 is required if you need to connect to Elastic Cloud using an API key. 

### Issues Solved

Fixes #9700

### Checklist

#### Ticket Background

### DIFF
--- a/arches/settings.py
+++ b/arches/settings.py
@@ -75,12 +75,8 @@ ELASTICSEARCH_CONNECTION_OPTIONS = {"timeout": 30}
 # Set the ELASTIC_PASSWORD environment variable in either the docker-compose.yml or .env file to the password you set for the elastic user,
 # otherwise a random password will be generated.
 
-# An API key can be generated using a POST to http://localhost:9200/_security/api_key with a minimum body of { "name": "<API_NAME>" }
-# ensuring you specify the basic authorization username and password credentials e.g. "elastic", "E1asticSearchforArche5". <API_NAME>
-# can be whatever you like e.g. "Arches API Key".
-
-# The response will return "api_key" and "encoded" - use "encoded" as <API_KEY>, not "api_key" as it has already been base64 encoded.
-# Alternatively, an API key can be generated in Kibaba - go to Management, Stack Management, Security, API keys and create a new API key.
+# API keys can be generated via the Elasticsearch API: https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html
+# Or Kibana: https://www.elastic.co/guide/en/kibana/current/api-keys.html
 
 # a prefix to append to all elasticsearch indexes, note: must be lower case
 ELASTICSEARCH_PREFIX = "arches"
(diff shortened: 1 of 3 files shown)


aj-he:
@apeters I've assigned you initially - please reassign as required.

arches/settings.py
```diff
@@ -67,6 +67,17 @@
 # Uncomment this line for a development setup after running the ubuntu_setup.sh script (do not use in production)
 # ELASTICSEARCH_CONNECTION_OPTIONS = {"timeout": 30, "verify_certs": False, "basic_auth": ("elastic", "E1asticSearchforArche5")}
 
+# If you need to connect to Elasticsearch via an API key instead of username/password, use the syntax below:
+# ELASTICSEARCH_CONNECTION_OPTIONS = {"timeout": 30, "verify_certs": False, "api_key": "<API_KEY>"}
```
apeters, line 71: In the docs it looks like the "api_key" is a tuple not a string.  Can you verify this?
See: https://elasticsearch-py.readthedocs.io/en/v8.9.0/api.html#module-elasticsearch

Example from the docs
```
# Set 'api_key' on the constructor
client = Elasticsearch(
    "http://localhost:9200",
    api_key=("id", "api_key")
)
```

RobOatesHistoricEngland: Hi @apeters,

When you generate an API key, the following information is returned from Elasticsearch's `_security/api_key` API request:

``` json
{
    "id": "nuAyjYkBlSnQtat8wfiM",
    "name": "my-api-key",
    "api_key": "a9KzcSMmQampvQ6zj2JHFw",
    "encoded": "bnVBeWpZa0JsU25RdGF0OHdmaU06YTlLemNTTW1RYW1wdlE2emoySkhGdw=="
}
```

Instead of using the `id` and `api_key`, you can simply use the base64 `encoded` value on it's own. My reasoning is why store two secrets, when one will suffice. The `encoded` value is "{id}:{api_key}" base64 encoded, similarly to what happends you you are using basic authentication with username and password.

I can verify this approach works. Would you like the comments expanded to show both methods?

apeters: @RobOatesHistoricEngland after taking a closer look at the docs, it looks like you're right.
From the docs `api_key: Union[str, Tuple[str, str], None] = None`.  It still might be nice to document both methods especially since they give the example of using both the `id` and the `key`.

RobOatesHistoricEngland: Have tested the tuple implementation and is working, as expected.

apeters (approved):
looks good to me

PR #9860, opened 2023-08-02
Add support for Elastic Cloud #9850 AB#57245

RobOatesHistoricEngland:
### Types of changes

### Description of Change
The search engine factory was changed to supply either the `cloud_id` or `hosts` to the constructor. These settings are mutually exclusive and if both are supplied, an error results. If the `cloud_id` is present - that will be used, otherwise it will fallback to use `hosts`.

Elastic Cloud requires the serialized query object to be UTF-8 encoded. To implement this, a `utf_encode` variable is added to the JSONSerializer instance and a check is made within the serialize() method and encodes when present.

### Issues Solved

#9850 

### Checklist

#### Ticket Background

### DIFF
--- a/arches/app/utils/betterJSONSerializer.py
+++ b/arches/app/utils/betterJSONSerializer.py
@@ -21,7 +21,7 @@ from arches.app.models.fields.i18n import I18n_JSON, I18n_String
 
 
 class UnableToSerializeError(Exception):
-    """ Error for not implemented classes """
+    """Error for not implemented classes"""
 
     def __init__(self, value):
         self.value = value
@@ -32,7 +32,7 @@ class UnableToSerializeError(Exception):
 
 
 class UnableToSerializeMethodTypesError(Exception):
-    """ Error for not implemented classes """
+    """Error for not implemented classes"""
 
     def __init__(self, value):
         self.value = value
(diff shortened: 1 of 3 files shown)


aj-he:
@apeters initially assigned to you for review. reassign as required

arches/app/utils/betterJSONSerializer.py
```diff
@@ -79,7 +79,9 @@ def serialize(self, obj, **options):
         options.pop("fields", None)
         options.pop("exclude", None)
         options.pop("force_recalculation", False)
-        return json.dumps(obj, cls=DjangoJSONEncoder, sort_keys=sort_keys, **options.copy())
+        result = json.dumps(obj, cls=DjangoJSONEncoder, sort_keys=sort_keys, **options.copy())
+
+        return result.encode("utf-8") if hasattr(self, "utf_encode") else result
```
apeters, line 84: Right now if someone sets `self.utf_encode = False` the result will still be utf encoded.  
I think what we should do is add the attribute to the __init__ method and set it to False

```
def __init__(self, **options):
    self._options = options
    self.utf_encode = False
```

Then here we can check for the value of the attribute.

apeters (changes requested):

apeters (changes requested):
See comment above

arches/app/utils/betterJSONSerializer.py
RobOatesHistoricEngland: Thanks @apeters - you are right. I was just checking for the existence of `utf_encode`, but as I was setting the value to `True` it would make sense that someone would set the value to `False` and expect the result to not be utf8 encoded. I will make the change, as suggested.

apeters (approved):
thanks @RobOatesHistoricEngland!  This looks good to me.

Issue #9749, opened 2023-07-09
Enable Branch Excel format to append/overwrite/delete existing tile data


chiatt:

njkim:
- The importer should be able to use the tileid if the row has the tileid.
- The importer will check if the tileid is already available in the system and the cardinality.
    - the tileid is given and available
        replace the existing tile (overwrite)
    - the tileid is not given or not in the system
        - cardinality == 'n'
            create a new tile (append)
        - cardinality == '1'
            - if there is already a tile for the nodegroup
                ignore
            - if there is not any tile 
                create a new tile (append)
    - the tileid is available but no data at all in the excel sheet
        delete the existing tile (delete) and _do we delete all the children?_

chiatt:
Delete will be handled in archesproject/arches#9973

PR #9874, opened 2023-08-04
Move import details to module, re #9749

chiatt:
### Types of changes

### Description of Change
Moves import details from the base etl component to module so that they can be tailored to the module, re #9749


njkim (approved):
Looks good! Works as expected.

Issue #9886, opened 2023-08-08
Add migration to create record for tile excel importer


chiatt:
The tile excel importer has configs to register it using the command line, however, we should make it part of core by default.

PR #9894, opened 2023-08-10
Fix typos and remove config from python file, re #9886

ryan86:
Fix typos and remove config from python file, re #9886


chiatt (approved):
Looks good!

PR #9895, opened 2023-08-10
Clean empty strings to None (to enforce field validation)

### LINKED ISSUE #9281: String nodes no longer enforcing required value

jacobtylerwalls:
### Types of changes

### Description of Change
Before, a required string field could escape validation because an object shaped like the following didn't clean to None:
```py
{
    "en": {"value": "", "direction": "ltr"},
    ...
}
```

Now, they clean to None and the field validation works.

### Issues Solved
Closes #9281

### Checklist

#### Ticket Background
*   Sponsored by:
*   Found by: @chiatt 
*   Designed by: @ 

### Further comments
In #9281 @chiatt suggested implementing a new `is_empty()` method, but `clean()` seemed to already be designed for this, and there's already an implementation for the GeoJSON use case mentioned in the ticket:

https://github.com/archesproject/arches/blob/6346ab34e10e5cf2331108947e097d8285c183cc/arches/app/datatypes/datatypes.py#L892-L895

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -6,6 +6,7 @@ from arches.app.utils.file_validator import FileValidator
 import filetype
 import base64
 import re
+import itertools
 import logging
 import os
 from pathlib import Path
@@ -161,6 +162,13 @@ class StringDataType(BaseDataType):
     def clean(self, tile, nodeid):
         if tile.data[nodeid] in ["", "''"]:
             tile.data[nodeid] = None
+        elif isinstance(tile.data[nodeid], dict):
+            for language_dict in tile.data[nodeid].values():
+                if language_dict["value"]:
+                    break
+            else:
+                # No non-empty value was found.
+                tile.data[nodeid] = None
 
     def append_to_document(self, document, nodevalue, nodeid, tile, provisional=False):
         if nodevalue is not None:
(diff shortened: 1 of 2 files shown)


arches/app/datatypes/datatypes.py
```diff
@@ -161,6 +162,13 @@ def validate_from_rdf(self, value):
     def clean(self, tile, nodeid):
         if tile.data[nodeid] in ["", "''"]:
             tile.data[nodeid] = None
+        elif isinstance(tile.data[nodeid], dict):
+            for language_dict in tile.data[nodeid].values():
+                if language_dict["value"]:
```
jacobtylerwalls, line 167: Assuming this key is present seemed legitimate.

arches/app/datatypes/datatypes.py
```diff
@@ -6,6 +6,7 @@
 import filetype
 import base64
 import re
+import itertools
```
jacobtylerwalls, line 9:
```suggestion
```

arches/app/datatypes/datatypes.py
chiatt: Yeah - that seems safe to me

chiatt (approved):
Great solution!

Issue #9897, opened 2023-08-10
Rename TileExcel -> TileExcelImporter, BranchCSV -> BranchExcel


chiatt:
Branch CSV does not actually process csv files, but rather excel files and so should be renamed in the system

The python file, class, and component names of TileExcelImporter should all match 

tile_excel_importer.py
TileExcelImporter
tile-excel-importer.js

PR #9899, opened 2023-08-10
Update module names to be more accurate and consistent #9897

chiatt:
### Description of Change
Update module names to be more accurate and consistent. BranchCSV is now BranchExcel and the TileExcelImporter uses the same name across files.

### Issues Solved

#9897


njkim (approved):
It looks good

Issue #9908, opened 2023-08-11
Reversing migration on 9744_add_tile_excel_importer causes error


chiatt:
This error occurs when a user has records in the load events table:
```
django.db.utils.IntegrityError: update or delete on table "load_event" violates foreign key constraint "load_errors_loadid_fc491ee4_fk_load_event_loadid" on table "load_errors"
DETAIL:  Key (loadid)=(cfc6f3a9-1247-49af-ade9-2b351e30d2aa) is still referenced from table "load_errors".
```

jacobtylerwalls:
Fixed in #9914

PR #9914, opened 2023-08-11
Remove load_errors on reverse migration of tile excel importer, re #9908

ryan86:
Remove load errors in reverse migration of tile excel importer, re #9908


jacobtylerwalls (approved):
+1 tested.

(After some trial and error I found that to get data into my `load_errors` table it was easiest to just reupload the same file.)

PR #9918, opened 2023-08-14
Add missing migration from #9451

jacobtylerwalls:
### Types of changes

-   [x] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
Refs #9917

Adds migration missing from #9451
### Issues Solved

#

### Checklist

-   [ ] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: 
*   Found by: @jacobtylerwalls
*   Tested by: @jacobtylerwalls
*   Designed by: @

PR #9921, opened 2023-08-14
Add workflow step to check for missing migrations #9917

### LINKED ISSUE #9917: Add CI step to fail PRs without necessary migrations
Currently, checking out dev/7.5.x and running `python manage.py makemigrations` produces this migration:

```sh
Migrations for 'models':
  arches/app/models/migrations/9747_alter_maplayer_options.py
    - Change Meta options on maplayer
```

Presumably because #9451 branched off before #9291 and only altered the migration order instead of remaking migrations.

***
I've worked on several projects that have a CI step that runs `python manage.py makemigrations --check` to fail PRs without any necessary migrations made. That would prevent this situation going forward.

jacobtylerwalls:
### Types of changes

### Description of Change
Add a step to fail PRs that having missing migrations.

### Issues Solved

Closes #9917

### Checklist

#### Ticket Background
*   Sponsored by: Farallon Geographics
*   Designed by: @ 

### Further comments
Until Django 4.2, you would have needed to also add the `--dry-run` flag to avoid actually making the migration, but we're about to upgrade to 4.2 anyway, and the action filesystem is ephemeral, so not bothering to add a flag that will become redundant as soon as we're on 4.2.

To test locally, you can use `echo $?` to get the output code of the prior bash command (it should be 1 if there are missing migrations.)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,6 +47,10 @@ jobs:
         run: |
           yarn && yarn build_test
 
+      - name: Check for missing migrations
+        run: |
+          python manage.py makemigrations --check
+
       - name: Run Arches unit tests
         run: |
           coverage run manage.py test tests --pattern="*.py" --settings="tests.test_settings"


jacobtylerwalls:
We can either leave this in a failed state until merging #9918 or we can cherry-pick that over here, either way.

chiatt:
Thanks @jacobtylerwalls, I'll just merge #9918 first and then re-run the build here

chiatt:
Hey @jacobtylerwalls, could you update your branch with dev/7.5.x (now with your CI fix) so that this PR clears all checks?

jacobtylerwalls:
done!

jacobtylerwalls:
The missing migration in the guardian app is fixed in a later version, so the Django upgrade PR will take care of it.

PR #9938, opened 2023-08-17
Move save into calculate_descriptors, re #9758

chiatt:
Uses django model .save rather than the proxy model's save method to save resource descriptors to the database.
re #9758


chiatt:
Closing in favor of #9967

PR #9959, opened 2023-08-23
9923 save workflow history to database

### LINKED ISSUE #9923: Move workflow history from local storage to the database
Currently if a user is going through a workflow (workflow A), they might find half-way that they needed to complete a different workflow (workflow B). If they were to open another tab, and complete workflow B, they would find that they might no longer be able to finish workflow A. If they do not delete workflow A, then the data they had already collected with it would persist and might be difficult to identify and remove. 

This is because workflow step data is saved to local storage and overwritten with each workflow to prevent users from hitting local storage limit. To work around this problem and allow for developers to inspect workflow history, we should start saving workflow step information to a workflow history table. 

The workflow history model should include json fields that store the data we currently save in local storage. Additionally there should be fields for:

- the date the workflow was started
- the user that started the workflow
- the workflow status (complete/incomplete)

johnatawnclementawn:
### Types of changes

### Description of Change

Saves workflow history to the database, rather than local storage

### Issues Solved

Resolves #9923


chiatt:
Moving @johnatawnclementawn's commits to branch `9926_workflow_history`

chiatt (commented):
-

Issue #9944, opened 2023-08-19
Bulk load does not populate resourcegraphid to and from in resource_x_resource


chiatt:
When loading data using the bulk loader, the resourceinstanceto_graphid and resourceinstancefrom_graphid are often not populated. I haven't confirmed whether it is an issue with single csv, but it seems to occur with branch and tile excel.

PR #9965, opened 2023-08-24
Add graph-ids to resource_x_resource records when using bulk importer, #9944 

njkim:
### Types of changes

### Description of Change

Add the graphids of resourceinstance-from and -to to resource_x_resource records when using bulk_importer

### Issues Solved

#9944

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/models/migrations/9944_update_refresh_relationship_function.py
```diff
@@ -0,0 +1,150 @@
...
+        AND x.resourceinstancefrom_graphid is null;
+
+        UPDATE resource_x_resource x
+        SET resourceinstanceto_graphid = r.graphid
+        FROM resource_instances r
+        WHERE r.resourceinstanceid = x.resourceinstanceidto
+        AND x.resourceinstanceto_graphid is null;
+    """
+
+    revert_previous_relationship = """
```
chiatt, line 137: Because this is an empty string instead of SQL, it causes an error when a user tries to reverse the migration. To do nothing, on reversal, it could be something like `SELECT 1=1`.

chiatt (changes requested):
This added the to and from graphids as expected, but not all of the relationships were created. If you load this package, you'll see that there are 4 relationships. If you delete all of the architect and building instances and then reload with the bulk loader, there will be only two relationships. [architect_pkg.zip](https://github.com/archesproject/arches/files/12464670/architect_pkg.zip)

chiatt (approved):
Looks good!

bferguso:
@chiatt, @njkim - This is causing significant performance issues in the SQL views. Loading 125k resources was taking about 3 minutes and is now taking multiple hours.

I did the test above, inserting the all the architect data and then all the building data only using SQL inserts and I can see the issue where only half the relationships are being added. However if the instances are created prior to adding the relationships (eg add in the following order `architect.intances`, `architect.name`, `building.instance`, `building.name`, `architect.designed`, `building.occupied by`) we get the correct number of relationships.

I'm guessing this is the same issue with the bulk import - creating a relationship before the `to` resource exists. Does the bulk importer use the `__arches_prepare_bulk_load();` and `__arches_complete_bulk_load();` calls? 

I'm wondering if we should do the following rather than doing it for each insert:
1. Turn the `join` into a `left join` to create all the relationships, but with a null graph ID if the `to` relationship doesn't exist and
2. add something to the `__arches_complete_bulk_load` that cleans up any missing relationship graph IDs

chiatt:
Coincidently @apeters was noticing this morning that a migration was taking a very, very long time and he traced the issue back to this change. Your suggestion would definitely improve performance, but discussing it with @apeters and @njkim, we are concerned that `__arches_refresh_tile_resource_relationships` could be called outside of an import, and there wouldn't be a guarantee that a user would run `__arches_comple_bulk_load` afterward. This would leave them with incomplete resource relationships. @njkim is working on a solution that should avoid an expensive join.

bferguso:
To be clear - for me it's the two updates at the end of the function causing the problem (lines 56-66 in this PR). As a workaround I added a left join to the resources table to pickup the graph IDs if the `to` resource exists and it was as fast as before. It seems like a bit of overkill to me to do two full table updates on a row-level insert. It's a bit of a strange case to be adding a relationship to a resource that doesn't exist yet - I'm sure there's a use case outside of an import for that, but not sure what it would be.

PR #9967, opened 2023-08-26
Reduce db queries when indexing, re #9758

chiatt:
### Description of Change
Uses django model save method rather than the proxy model's save method when saving descriptors to the database. Saves resources in index_database cmd only when recalculating descriptors, and prevents redundant indexing when tiles are saved with child tiles.

This doesn't solve #9758, but helps clarify why resource.save is being called.


arches/app/models/tile.py
```diff
@@ -475,13 +475,13 @@ def save(self, *args, **kwargs):
                     transaction_id=transaction_id,
                 )
 
-            if index:
-                self.index()
-
             for tile in self.tiles:
                 tile.resourceinstance = self.resourceinstance
                 tile.parenttile = self
                 tile.save(*args, request=request, index=index, **kwargs)
```
aj-he, line 481: @chiatt i would pass index=false here as you are going to handle indexing afterwards by the looks of things

chiatt: Ah, good catch @aj-he. Fixed it. Thanks!

apeters (changes requested):
I have a few minor suggestions and a larger more general comment/suggestion.
Minor suggestions:

- Declare `self.name` and `self.descriptors` in the `Resource.__init__` method
- Remove lines 115-119 from resource.py
- Remove check for `self.descriptors` on line 143 of resource.py

The larger issue I have is that now a call to `get_descriptor` actually ends up saving the resource via the call to `save_descriptors`.  This seems unexpected to me.
This was one of the issues we had previously except that a call to `index` triggered a resource save.

I think we need to acknowledge that we need to call `save_descriptors` every time we save a tile.  Right now that call is stuck within an `if block` in the `tile.save` method (again loosely tying it to indexing).  If we do that then we can remove the call to `save_descriptors` from `get_descriptors` and then what these methods do becomes clearer.

Happy to chat with you @chiatt about this.

chrabyrd:
@chiatt @apeters this should be ready for re-review now 👍

chrabyrd:
@chiatt @apeters this should be ready for review 👍

chrabyrd:
Can confirm both issues and one other: refreshing the page in the resource editor removes its descriptors

chrabyrd:
@apeters @chiatt This should be ready to review again. I had to change the logic in `primary_descriptors.py` and I'm happy to chat about it anytime

arches/app/models/resource.py
```diff
@@ -146,36 +148,36 @@ def get_descriptor(self, descriptor, context):
...
-
-        for descriptor in descriptors:
-            if len(self.descriptor_function) == 1:
-                module = self.descriptor_function[0].function.get_class_module()()
-                self.descriptors[language][descriptor] = module.get_primary_descriptor_from_nodes(
-                    self, self.descriptor_function[0].config["descriptor_types"][descriptor], context
-                )
-                if descriptor == "name" and self.descriptors[language][descriptor] is not None:
-                    self.name[language] = self.descriptors[language][descriptor]
+        for lang in settings.LANGUAGES:
```
apeters, line 158: I think we should only attempt to save the descriptor if the `context["tile"].sortorder` > 0

apeters (changes requested):
This fails if you reorder a tile in the UI to be the top tile and then edit that tile.
https://recordit.co/WcAIMMR39q

This is because I think the sortorder on the client isn't being managed well.

arches/app/models/tile.py
```diff
@@ -537,11 +537,14 @@ def delete(self, *args, **kwargs):
                         node = SimpleNamespace(**next((x for x in self.serialized_graph["nodes"] if x["nodeid"] == nodeid), None))
                     except TypeError: #will catch if serialized_graph is None
                         node = models.Node.objects.get(nodeid=nodeid)
+
                     datatype = self.datatype_factory.get_instance(node.datatype)
                     datatype.post_tile_delete(self, nodeid, index=index)
-                if index:
```
chiatt, line 539: This may be necessary, but I'm concerned about creating a resource proxy model instance for every tile save whether or not we need to index. That would decrease the performance of loading data using the cli and probably in other cases as well.

apeters: This is on the tile.delete method

chiatt: Thanks @apeters updated my comment - I meant that it applies to both save and delete.

apeters: @chiatt even though `save_descriptors` gets called on every save or delete it won't mean that the descriptors will be recalculated every time.  I think they only time they get recalculated is when you save or delete a tile with `sortorder = 0`

chrabyrd: not in the code's current state @apeters. We need to handle the case where a cardinality `n` nodegroup  has multiple tiles and the user deletes the first tile.

apeters: @chrabyrd I just checked and confirmed that delete of the first tile updates the primary name with the second tile. I thought that was the case because of lines 45-48 in the primary_descriptors function.

arches/app/functions/primary_descriptors.py
```diff
@@ -37,16 +37,19 @@ def get_primary_descriptor_from_nodes(self, resource, config, context=None):
...
-                if len(tiles) == 0:
-                    tiles = models.TileModel.objects.filter(nodegroup_id=uuid.UUID(config["nodegroup_id"])).filter(
+                tile = context.get('tile')
+
+                if not tile or tile.sortorder:
+                    tile = models.TileModel.objects.filter(nodegroup_id=uuid.UUID(config["nodegroup_id"])).filter(
                         resourceinstance_id=resource.resourceinstanceid
-                    )
-                for tile in tiles:
+                    ).order_by('sortorder').first()
```
apeters, line 48: this handles for the delete of the first tile

arches/app/models/tile.py
chrabyrd: @apeters  Yes, this is because we are recalculating the descriptors on every tile save, regardless of `sortorder`

arches/app/models/resource.py
```diff
@@ -146,36 +148,36 @@ def get_descriptor(self, descriptor, context):
...
+            language = self.get_descriptor_language({"language":lang[0]})
+            if context:
+                context["language"] = language
             else:
-                self.descriptors[language][descriptor] = None
+                context = {"language":language}
+            if not self.descriptor_function:
+                self.descriptor_function = models.FunctionXGraph.objects.filter(
+                    graph_id=self.graph_id, function__functiontype="primarydescriptors"
+                ).select_related("function")
```
jacobtylerwalls, line 167: This should be deindented so we don't multiply the number of queries by the number of languages.

Spotted when applying the test case from #10316 to this branch.

jacobtylerwalls (changes requested):

arches/app/models/resource.py
chrabyrd: ah -- this isn't terrible because it's only making that query if there's not already a descriptor function, and the logic sets the descriptor function.

But it is an improvement to readability to happy to move it out of the block

chrabyrd: ah -- this isn't terrible because it's only making that query if there's not already a descriptor function, and the logic sets the descriptor function.

But it is an improvement to readability to happy to move it out of the block

jacobtylerwalls: Interesting, I was definitely getting 10 queries instead of 2:
```py
(venv311) jwalls@Jacobs-MacBook-Pro arches % python3 manage.py test tests.models.resource_test.ResourceTests.test_recalculate_descriptors_one_query_for_descriptor_function --settings="tests.test_settings"
Found 1 test(s).
F
FAIL: test_recalculate_descriptors_one_query_for_descriptor_function (tests.models.resource_test.ResourceTests.test_recalculate_descriptors_one_query_for_descriptor_function)
  File "/Users/jwalls/prj/arches/tests/models/resource_test.py", line 254, in test_recalculate_descriptors_one_query_for_descriptor_function
    self.assertEqual(len(function_x_graph_selects), 1)
AssertionError: 10 != 1

Ran 1 test in 2.091s

FAILED (failures=1)
```

jacobtylerwalls: I can take a look to see if the test setup is not realistic 👍

jacobtylerwalls: Ah, what I think is is this: the check against `self.descriptor_function` was testing truth value, not is None, so even though the logic "sets" the descriptor_function, it didn't help in the case of a resource without a descriptor function. But we'll be fixing this soon anyway in #10316 👍

jacobtylerwalls (approved):
🚢 it!

Issue #9968, opened 2023-08-28
Resources are incorrectly deleted when PUTing a JSON-LD file without supplying either a graph id or graph slug.


apeters:
If a user doesn't supply either a graph id or graph slug, then the system should respond back with a meaningful error message and should not delete the resource from the system.

PR #9969, opened 2023-08-28
fixes bug where not supplying a graphid or slug to the json-ld api would delete the resource but not save the new one, re #9968

apeters:
fixes bug where not supplying a graphid or slug to the json-ld api would delete the resource but not save the new one, re #9968

see this thread for more context:
https://community.archesproject.org/t/api-for-update-resource/2094/1


chiatt (approved):
Looks good!

Issue #9982, opened 2023-08-31
Node-value report fails in search context


johnatawnclementawn:
The node-value widget relies on logic that extracts the resourceid from the current url. However, when report is rendered in a search context (e.g. in the search details), the [updateTiles](https://github.com/archesproject/arches/blob/6346ab34e10e5cf2331108947e097d8285c183cc/arches/app/media/js/viewmodels/node-value-select.js#L39-L58) method fails because the [value that is extracted from the url](https://github.com/archesproject/arches/blob/6346ab34e10e5cf2331108947e097d8285c183cc/arches/app/media/js/viewmodels/node-value-select.js#L21) which should be a resourceid, is actually `search`.

PR #9983, opened 2023-08-31
9883 hide empty node value link

### LINKED ISSUE #9883: Resource Report -  2.4.4 Link Purpose (In Context) - 1 link on the page is both empty and broken

johnatawnclementawn:
### Types of changes

### Description of Change

In a report context, when a node with datatype node-value is empty or the node value lookup fails, a broken link is created. These changes prevent an empty link from being created and causing accessibility problems.

### Issues Solved

Resolves #9883 

### Further comments
This solves the immediate accessibility issues. However, there is a deeper bug that should be addressed as detailed in: #9982


apeters (approved):
looks good!

Issue #9988, opened 2023-09-01
Ensure bulk_disambiguated_resource_instance endpoint respects resource_id ordering


chrabyrd:

chrabyrd:
resolved in #9995

PR #9995, opened 2023-09-05
enforces ordering in bulk disambiguated endpoint #9988

chrabyrd:
### Types of changes

### Issues Solved

#9988


njkim:
The original order isn't retained for me.

It looks like the json is reordered by this line, even though the ordered dictionary is passed,
https://github.com/archesproject/arches/blob/dev/7.5.x/arches/app/utils/betterJSONSerializer.py#L79
through
https://github.com/archesproject/arches/blob/dev/7.5.x/arches/app/utils/response.py#L18

So if resorting by the key isn't desirable, it seems sort_keys needs to be set to False.

njkim (changes requested):
The original order isn't retained for me.

It looks like the json is reordered by this line, even though the ordered dictionary is passed,
https://github.com/archesproject/arches/blob/dev/7.5.x/arches/app/utils/betterJSONSerializer.py#L79
through
https://github.com/archesproject/arches/blob/dev/7.5.x/arches/app/utils/response.py#L18

So if resorting by the key isn't desirable, it seems sort_keys needs to be set to False.

chrabyrd:
> The original order isn't retained for me.
> 
> It looks like the json is reordered by this line, even though the ordered dictionary is passed, https://github.com/archesproject/arches/blob/dev/7.5.x/arches/app/utils/betterJSONSerializer.py#L79 through https://github.com/archesproject/arches/blob/dev/7.5.x/arches/app/utils/response.py#L18
> 
> So it seems sort_keys=False needs to be passed, if resorting by the key isn't desirable.

good catch!

njkim (approved):
It looks good and works as expected.

PR #10003, opened 2023-09-05
fixes #10002

### LINKED ISSUE #10002: select2 disable doesn't respect observable value
A select2 drop down that is initially disabled works as expected.  You can modify that observable value to a false value and select2 respects that as well, enabling the dropdown. However, when it is changed back to disabled again the select2 dropdown does not change back and stays enabled.   The issue appears to be within the select2 binding.

aarongundel:
Fixes issue with select2 not disabling itself to respect observable value.


chiatt (approved):
Looks good to me!

PR #10004, opened 2023-09-05
changes manifest manager manifest selection verbiage

### LINKED ISSUE #9980: Change Edit a Service to Select a Service in the Image Service Manager
AFS users sometimes need to find and select a service from the manifest manager. When the manifest manager opens, there is no option to simply select a service - only create and edit. This is also a little inaccurate because some services are remotely hosted and cannot be edited. To help users select an image service, the 'Edit a Service' tab should be titled 'Select a Service'.  Within the content of the `Select a Service` tab, `Edit an existing Service` should be `Select an Existing Service`. 

image

aarongundel:
resolves #9980 - Changes 'Edit a Service' in manifest manager to 'Select a Service'


chiatt (approved):
LGTM!

Issue #10006, opened 2023-09-06
escape characters in the string data cause the bulk editor to fail


apeters:
The bulk editor fails when performing operations on string data that contains escape (backslash) characters.

To reproduce, save a string with a value that contains a backslash: eg "testing \something"

Run any of the bulk editor operations (trim, capitalize, etc...) on that field.
Notice that the bulk operation fails.

apeters:
closed by #10008

PR #10008, opened 2023-09-06
fix bulk edit operations when data contains escape chars, re #10006

apeters:
re #10006

To reproduce, save a string with a value that contains a backslash: eg "testing \something"

Run any of the bulk editor operations (trim, capitalize, etc...) on that field.
Notice that the bulk operation fails.


njkim (approved):
It works as expected. Looks good.

PR #10010, opened 2023-09-07
10009 node script deprecation

aarongundel:
Changes how node is installed in the ubuntu setup script.  Also adds note that removes support for Ubuntu 18.04 - which is now [out of standard support](https://ubuntu.com/about/release-cycle).  

Tested and verified working with Ubuntu 20.04.

### DIFF
--- a/arches/install/ubuntu_setup.sh
+++ b/arches/install/ubuntu_setup.sh
@@ -40,9 +40,14 @@ function install_postgres {
 }
 
 function install_yarn {
-  wget --quiet -O - https://deb.nodesource.com/setup_16.x | sudo -E bash -
   sudo apt-get update
-  sudo apt-get install -y nodejs
+  sudo apt-get install -y ca-certificates curl gnupg
+  sudo mkdir -p /etc/apt/keyrings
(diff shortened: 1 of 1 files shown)


arches/install/ubuntu_setup.sh
```diff
@@ -40,9 +41,14 @@ function install_postgres {
 }
 
 function install_yarn {
-  wget --quiet -O - https://deb.nodesource.com/setup_16.x | sudo -E bash -
   sudo apt-get update
-  sudo apt-get install -y nodejs
+  sudo apt-get install -y ca-certificates curl gnupg
+  sudo mkdir -p /etc/apt/keyrings
+  curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | sudo gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
```
chiatt, line 47: We are calling curl here, but it is installed on line 52. Maybe line 52 should be moved before this line - just to be safe, or just remove line 52 altogether.

chiatt (changes requested):
I've tried this a couple of times on ubuntu 20.04 and node v10.19.0 is installed which is too low for our version of webpack.

chiatt (approved):

Issue #10022, opened 2023-09-11
Ctrl+S doesn't work for saving nodes and card edits.


SDScandrettKint:
The common keyboard shortcut "CTRL+S" used to save work has recently been merged into Arches with regards to resource edits but the functionality is not implemented for saving nodes or card edits in the graph designer.

PR #10023, opened 2023-09-11
Ctrl+S functionality for adding nodes and saving card edits.

SDScandrettKint:
### Types of changes

### Description of Change
Adds functionality for use of keyboard shortcut CTRL+S in saving node and card edits in the resource model designer.

### Issues Solved

#10022 

### Checklist

#### Ticket Background


apeters (dismissed):
fixed missing or unnecessary semi-colons, simplified code a bit.

SDScandrettKint:
[This issue ](https://github.com/archesproject/arches/issues/10027)is causing some issues with the Ctrl+S in functionality in resources. When a string card with cardinality of 1 (only one value able to be added) has data added, moving away to a different node and back to the string to keep repeating the addition of data and navigation away, then using Ctrl+S, will result in multple tiles of the same data. 
I believe that if data persisted this may not happen.

apeters:
Good catch @SDScandrettKint!  This is because and event listener is registered every time the widget is rendered.  The problem is that the event listener isn't **unregistered** when the widget is hidden/destroyed.  Then listeners keep getting added and when a user finally saves using crtl+s all the extra save methods are called and multiple tiles are saved.

apeters:
I think for this to work we need to remove the event listener when we dispose of the widget.
For example, this works in the `card-component.js` file:

```
        var keyListener = function(e){
            if (e.ctrlKey && e.key === "s") {
                e.preventDefault();
                if (self?.tile?.dirty() == true && 
                    self?.tile?.parent?.isWritable === true) {
                        self.saveTile();
                }
            }
        };
        document.addEventListener("keydown", keyListener);

        this.dispose = function(){
            document.removeEventListener("keydown", keyListener);
        };
```

apeters:
@SDScandrettKint can you make those changes suggested above to the files?

SDScandrettKint:
I've added the suggested changes to dispose of the eventlisteners once the cards/nodes are navigated away. This is effective in the resource editor view, preventing the issue above. 
A nice feature in the future could be to store all card edits so that a user can use Ctrl+S to save everything in at once in bulk?

Though added, disposing the eventlistener is probably not necessary for saving graph nodes, as it's not possible to navigate away from a dirty node, but it could be useful if this somehow gets bypassed.

It doesn't seem to work for card edits in the graph view, as multiple edits from separate cards can be made and saved with ctrl+s.  This doesn't seem to be causing any issues though, e.g.
saving_all_card_edits
Perhaps the navigation to different cards in the graph functions differently to navigating to different cards in the resource editor, so disposing of the eventlistener doesn't have an effect.

apeters:
When saving with the button, you have to explicitly save both edits.  By using CRTL+S both edits are saved.  This actually seems like an improvement.

PR #10025, opened 2023-09-11
[6.2.x] Fix version number of arches-django-revproxy in release note

jacobtylerwalls:
Fix the version number of `arches-django-revproxy` in release notes. Discovered installing 6.2 a couple weeks ago that the given version number doesn't exist.


chiatt (approved):
Nice catch

Issue #10011, opened 2023-09-07
Allow specific node widgets to be targeted by CSS rules


bferguso:
Currently there is no easy way for a specific widget associated with a node to be targeted by a CSS rule. nth selectors could theoretically be used, however that would be relatively brittle and would break with any kind of page reorganization (reordering of widgets, etc).

If the node alias was assigned to the widget HTML containers this would allow formatting of widgets on a node-by-node basis using only CSS rules.

PR #10034, opened 2023-09-12
10011 add node alias and widget name as class to HTML elements

bferguso:
### Types of changes

### Description of Change

Adds the associated node alias and widget name as classes to HTML container elements to allow specific node elements to be targeted by CSS rules.

_Implementation Notes_
- Currently just added at a top level `div` for each widget/report/display_value section to not litter many HTML elements with the same classes. 
- `report` elements do not appear to have a top level `div` so applied to the `dt`, and `dd` elements.
- widget name is only available when in the `form` elements are visible. 
- `node_alias` is available for `form`, `report` and `display_value` blocks.
- left `config_form` section out of scope as this is primarily intended for user presentation.

### Issues Solved

#10011 

### Checklist

#### Ticket Background
*   Sponsored by: Self Funded
*   Found by: @bferguso
*   Tested by: @bferguso
*   Designed by: @bferguso

### DIFF
--- a/arches/app/media/js/viewmodels/widget.js
+++ b/arches/app/media/js/viewmodels/widget.js
@@ -62,11 +62,11 @@ define([
             this.config = ko.observable(this.config);
         }
 
-        this.widgetCssClasses = function() {
+        this.nodeCssClasses = ko.pureComputed(function() {
             return [self.node?.alias(),
                 self.widget?.widgetLookup[ko.unwrap(self.widget?.widget_id)].name
                 ].join(" ").trim();
-        }
+        });
 
         this.disposables = [];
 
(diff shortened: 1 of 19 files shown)


chiatt (commented):
This looks really good. I'm wondering if it would make sense to include the graph slug in the nodeCssClasses. That way if you had a node alias shared by two or more resource models, you could style them differently.

bferguso:
> This looks really good. I'm wondering if it would make sense to include the graph slug in the nodeCssClasses. That way if you had a node alias shared by two or more resource models, you could style them differently.

@chiatt - good idea. As far as I can see the graph slug doesn't seem to be available in the widget.js component right now. There is a graph object available through the form (only when the edit widgets are available) but it doesn't include the slug. We'll need to figure out how to make it available so we can include it as a class. If you know off the top how we can do that I'd be interested in any suggestions. Otherwise I'll dig into that when I have a minute.

chiatt (approved):

robgaston (approved):

chiatt:
I've added an issue for the graph slug implementation so that we can handle it in a different PR: #10119

Issue #9975, opened 2023-08-30
Single CSV importer should report the number of resources/tiles created


chiatt:
The Single CSV importer should report the number of resources and tiles that were successfully imported.

PR #10037, opened 2023-09-13
Move resource count into base-import, re #9975

chiatt:
Moves resource count into base-import, re #9975


njkim (approved):
This looks good!

Issue #9987, opened 2023-09-01
Excel export should write concept ids, not labels, by default


chiatt:
This could be a checkbox in the exporter UI allowing a user to decide how they want their data exported

chrabyrd:
resolved in #10047

PR #10047, opened 2023-09-15
adds option for users to export concepts as UUIDS when using bulk tile exporter #9987

chrabyrd:
### Types of changes

### Issues Solved

#9987

### DIFF
--- a/arches/app/etl_modules/branch_excel_exporter.py
+++ b/arches/app/etl_modules/branch_excel_exporter.py
@@ -45,7 +45,7 @@ class BranchExcelExporter(BaseExcelExporter):
         self.loadid = loadid if loadid else None
 
 
-    def run_export_task(self, load_id, graph_id, graph_name, resource_ids):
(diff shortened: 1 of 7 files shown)


arches/app/etl_modules/branch_excel_exporter.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, request=None, loadid=None):
         self.loadid = loadid if loadid else None
 
 
-    def run_export_task(self, load_id, graph_id, graph_name, resource_ids):
+    def run_export_task(self, load_id, graph_id, graph_name, resource_ids, *args, **kwargs):
```
jacobtylerwalls, line 48: I take it the reason this is dynamic is so that the BranchExcelExporter doesn't have to implement the new argument. If that's the reason, we could also just let that implementation accept a default argument value of None (or UUIDs, since that's what it seems to do) and then raise NotImplementedError if it got labels. Not important, but there's something nice about the explicitness IMO. Up to you!

jacobtylerwalls (approved):
Tested, works well, feel free to merge!

Issue #9974, opened 2023-08-30
Users should be able to name exported files


chiatt:
In the tile excel and branch excel bulk export modules, users should be able to assign a name for their exported file.

chrabyrd:
resolved in #10049

PR #10049, opened 2023-09-18
allows naming of exported files from bulk_data_manager #9974

chrabyrd:
### Types of changes

### Issues Solved

#9974


jacobtylerwalls (approved):
Tested, works well.

Is the scope of this to just allow naming the .zip and not the extracted underlying files? If so, good to merge.

Issue #10029, opened 2023-09-11
Manifest canvases fail to delete due to missing method


chiatt:
The manifest manager view used to have a method, `check_canvas_in_use` but it was removed. That method is still called within `delete_canvases` resulting in an error if a user tries to delete a canvas.

Here is the former method: https://github.com/archesproject/arches/commit/b3ac5284569db52596af3966b04841a2441dc678

It was removed here:
https://github.com/archesproject/arches/commit/456ab3e9b44625385778f76979f4b7cafa05260a

jacobtylerwalls:
Fixed in #10054

PR #10054, opened 2023-09-20
Fix canvas delete function, #10029

njkim:
### Types of changes

### Description of Change

Fix the issue of not being able to delete a canvas, #10029

### Issues Solved

#10029

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/views/manifest_manager.py
+++ b/arches/app/views/manifest_manager.py
@@ -113,7 +113,7 @@ class ManifestManagerView(View):
             manifest.manifest["sequences"][0]["canvases"] += canvases
 
         def check_canvas_in_use(canvas_id):
-            return models.VwAnnotation.objects.filter(canvas__contains=canvas_id).exists()
+            return models.VwAnnotation.objects.filter(canvas=canvas_id).exists()
 
         def delete_canvases(manifest, canvases_to_remove):
             canvas_ids_remove = [canvas["images"][0]["resource"]["service"]["@id"] for canvas in canvases_to_remove]
--- a/arches/app/views/manifest_manager.py
+++ b/arches/app/views/manifest_manager.py
@@ -112,11 +112,15 @@ class ManifestManagerView(View):
         def add_canvases(manifest, canvases):
             manifest.manifest["sequences"][0]["canvases"] += canvases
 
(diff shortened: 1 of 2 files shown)


arches/app/views/manifest_manager.py
```diff
@@ -112,11 +112,15 @@ def create_canvas(image_json, file_url, file_name, image_id):
         def add_canvases(manifest, canvases):
             manifest.manifest["sequences"][0]["canvases"] += canvases
 
+        def check_canvas_in_use(canvas_id):
+            canvas_ids_in_use = [annotation.canvas for annotation in models.VwAnnotation.objects.all()]
+            return canvas_id in canvas_ids_in_use
```
jacobtylerwalls, line 117: We don't need all the rows from the table to check if a single row exists containing `canvas_id`, so we can:
```suggestion
            return models.VwAnnotation.objects.filter(canvas__contains=canvas_id).exists()
```

jacobtylerwalls (approved):
Fix works great 👍

(Noticed we can tighten up the db query, but we can future that if you prefer.)

arches/app/views/manifest_manager.py
njkim: > We don't need all the rows from the table to check if a single row exists containing `canvas_id`, so we can:

Absolutely! Thanks.

arches/app/views/manifest_manager.py
```diff
@@ -112,11 +112,14 @@ def create_canvas(image_json, file_url, file_name, image_id):
         def add_canvases(manifest, canvases):
             manifest.manifest["sequences"][0]["canvases"] += canvases
 
+        def check_canvas_in_use(canvas_id):
+            return models.VwAnnotation.objects.filter(canvas__contains=canvas_id).exists()
```
jacobtylerwalls, line 116: I think this is what we need, to avoid causing #10055 (sorry!)
```suggestion
            return models.VwAnnotation.objects.filter(canvas=canvas_id).exists()
```

jacobtylerwalls (approved):

PR #10071, opened 2023-09-25
Add a button to copy node UUIDs to clipboard in graph designer tree

SDScandrettKint:
Adds a button to copy node UUIDs to clipboard in the graph designer tree.
### Types of changes

### Description of Change
Adds a button that shows up in the graph designer node tree once 'show IDs is clicked' that allows the current node UUID to be copied to the clipboard. This provides a quicker way of obtaining a node UUID than the previous method of using inspect element.

When show IDs is off:
image
When show IDs is on:
image

### Issues Solved

#

### Checklist

#### Ticket Background

### Further comments

At the moment the tooltip is not showing up even though the text has been set and the structure is the same as all other buttons on the bar- any suggestions/help would be greatly appreciated! :)

### DIFF
--- a/arches/app/media/js/views/graph/graph-designer/graph-tree.js
+++ b/arches/app/media/js/views/graph/graph-designer/graph-tree.js
@@ -248,6 +248,11 @@ define([
         },
         toggleGrid: function(){
             this.showGrid(!this.showGrid());
+        },
+
+        copyNodeId: function(node, e) {
+            var nodeId = node.nodeid
+            navigator.clipboard.writeText(nodeId);
         }
     });
     return GraphTree;
--- a/arches/app/media/js/views/graph/graph-designer/graph-tree.js
+++ b/arches/app/media/js/views/graph/graph-designer/graph-tree.js
@@ -251,6 +251,17 @@ define([
         },
 
         copyNodeId: function(node, e) {
+            const copyUUIDelement = $(copy_node_id)
+
+            function restoreTitle(e) {
+                copyUUIDelement.tooltip('hide') 
+                copyUUIDelement.attr('data-original-title', 'Copy UUID to Clipboard')
+                copyUUIDelement.off('mouseleave', restoreTitle)
+            };
+            copyUUIDelement.attr('data-original-title', 'Copied to Clipboard!')
+            copyUUIDelement.tooltip('show')
+            copyUUIDelement.on('mouseleave', restoreTitle)
+
             var nodeId = node.nodeid
             navigator.clipboard.writeText(nodeId);
         }
--- a/arches/app/media/js/views/graph/graph-designer/graph-tree.js
+++ b/arches/app/media/js/views/graph/graph-designer/graph-tree.js
@@ -251,16 +251,14 @@ define([
         },
 
         copyNodeId: function(node, e) {
-            const copyUUIDelement = $(copy_node_id)
-
             function restoreTitle(e) {
-                copyUUIDelement.tooltip('hide') 
-                copyUUIDelement.attr('data-original-title', 'Copy UUID to Clipboard')
-                copyUUIDelement.off('mouseleave', restoreTitle)
+                $(e.target).tooltip('hide') 
+                $(e.target).attr('data-original-title', 'Copy UUID to Clipboard')
+                $(e.target).off('mouseleave', restoreTitle)
             };
-            copyUUIDelement.attr('data-original-title', 'Copied to Clipboard!')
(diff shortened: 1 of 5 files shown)


arches/app/templates/views/graph/graph-designer/graph-tree.htm
```diff
@@ -53,6 +53,9 @@
             <span style="font-weight:bold;" data-bind="text: ': ' + node.nodeid"></span>
             <!-- /ko -->
             <span class="spacer"></span>
+            <!-- ko if: graphTree.showIds-->
+            <i class="jstree-node-action-icon ion-ios-copy-outline" role="presentation" data-bind="click: graphTree.copyNodeId.bind(graphTree)" data-toggle="tooltip" data-placement="auto" data-container="body" data-original-title="{% trans "Copy UUID to Clipboard" %}"></i>
+            <!-- /ko -->
```
aarongundel, line 58: The reason the tooltip isn't working is because the button is not actually rendered if showids is false - when the tooltips are all initialized.  Instead of encasing this in a ko block, it might be more effective to use the the visible binding - then the dom element will be available to the tooltip generator when it needs to be - just not visible to the user.  
```suggestion
           <i class="jstree-node-action-icon ion-ios-copy-outline" role="presentation" data-bind="click: graphTree.copyNodeId.bind(graphTree), visible: graphTree.showIds()" data-toggle="tooltip" data-placement="auto" data-container="body" data-original-title="{% trans "Copy UUID to Clipboard" %}"></i>
```

aarongundel (changes requested):
This is a cool feature.  Just minor feedback re: the tooltips.

arches/app/templates/views/graph/graph-designer/graph-tree.htm
SDScandrettKint: Thanks! This has been committed

SDScandrettKint:
I thought it would be a good feature to update the tooltip text with "Copied to Clipboard!" once the button has been clicked. This works great and as intended for the graph top node, but for some reason doesn't work as intended for the inner nodes. For some reason the updated tooltip shows in the top left and doesn't follow the mouseleave events and I can't figure out why:

tooltips_title_change_issue

The latest commit has the code for this function, but in short an id is added to the htm element and the following JS updates the tooltip:
```
            function restoreTitle(e) {
                copyUUIDelement.tooltip('hide') 
                copyUUIDelement.attr('data-original-title', 'Copy UUID to Clipboard')
                copyUUIDelement.off('mouseleave', restoreTitle)
            };
            copyUUIDelement.attr('data-original-title', 'Copied to Clipboard!')
            copyUUIDelement.tooltip('show')
            copyUUIDelement.on('mouseleave', restoreTitle)
```

Any help/feedback would once again be great :)

aarongundel:
The tooltips are a bit of a pain - what you're suggesting is going to be somewhat difficult because the tooltips are static - generated once on page load based on all the DOM nodes currently loaded and then never changed.  This is why the tooltip didn't work originally - tooltips in arches aren't dynamic.  Look at page-view.js line 144.  page-view.js executes once - on page load.  jquery's already present in graph-tree.js.  You might be able to use it with a hover event from knockout to update it and then change it back with a setTimeout - I've had to do something like this before.  @SDScandrettKint

SDScandrettKint:
The latest commit seems to fix the tooltip updates.  Removing the reference to the HTML line via it's ID and instead using jquery to get the element through `$(e.target)` fixes the issue.  An example can be seen below.

working_copy_nodeid_to_clipboard_tooltip_updates

arches/app/media/js/views/graph/graph-designer/graph-tree.js
```diff
@@ -248,6 +248,20 @@ define([
         },
         toggleGrid: function(){
             this.showGrid(!this.showGrid());
+        },
+
+        copyNodeId: function(node, e) {
+            function restoreTitle(e) {
+                $(e.target).tooltip('hide') 
+                $(e.target).attr('data-original-title', 'Copy UUID to Clipboard')
```
chiatt, line 256: 'Copy UUID to Clipboard' and 'Copied to Clipboard' need to be added to arches.translations. Also, 'Copy UUID to Clipboard' should probably be changed to 'Copy Node ID to Clipboard' just to be explicit.

chiatt (changes requested):

chiatt (approved):

chiatt (approved):
Looks good to me!

PR #10075, opened 2023-09-26
10072 file upload

aarongundel:
Fixes broken file uploads in 7.4.x after tile performance fix.


jacobtylerwalls (approved):
Tested, works well.

I suppose we'll need to be on guard against this in the future, that the `node` argument to `validate` might be an ersatz-node.

Issue #10074, opened 2023-09-26
Bulk Editor: UI updates


apeters:
A couple of UI updates:
- When starting a bulk edit operation there is a noticeable lag when the new status shows up in the Task Status tab.
- Where the preview returns no values alert the user instead of showing an empty table
- Don't allow users to start a bulk edit operation without first previewing.

PR #10076, opened 2023-09-26
bulk editor UI updates

apeters:
re #10074


njkim (approved):
Works as described, Looks good.

PR #10078, opened 2023-09-26
Size up small text in accessibility mode #9964

### LINKED ISSUE #9964: All Components - Some text is too small
As a user with low vision, I expect all text in the page to be of an appropriate size to ensure easy reading. 

**Linked work item:** 
#9391 #9387  

**Expected Result:**
All text should be an appropriate size to ensure users with low vision can easily read text.

**Actual Result:**
Text at the bottom of the left nav and 'Details'/'Related Resources' text on each search result is too small. Text which is very small is difficult to read, particularly for those with low vision. 

**Evidence:**

Image

johnatawnclementawn:
### Types of changes

### Description of Change

Bring small text into compliance with accessibility standards 

### Issues Solved

Resolves #9964

### DIFF
--- a/arches/app/media/css/accessibility.scss
+++ b/arches/app/media/css/accessibility.scss
@@ -93,6 +93,7 @@ a, button {
 
 .debug-notice {
     color: $debug;
+    font-size: 1.2rem;
 }
 
 .sidenav-sm {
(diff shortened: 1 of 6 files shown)


johnatawnclementawn:
Self approving small change per Alexei

Issue #9145, opened 2022-12-13
Edit coordinates: Polygon with many coordinates fills up screen


SDScandrettKint:
**Arches Version/Branch:** dev/6.2.x

**Describe the bug:**
When drawing a polygon that has a large number of coordinates, the edit coordinates screen doesn't function properly, e.g.

![image](https://user-images.githubusercontent.com/119508494/207304023-a21f9738-755b-4ddb-910a-8ad3c5f11d40.png)

Additionally, the Edit Cordinates banner is linked to the overall y scroll of the page so on a smaller screen it moves from its intended position of the top of the page to the location of the scroll.

**To reproduce:**
1. Draw a polygon with at least 6 points (6 appears to be the cut-off point before the last set of coordinates becomes buried).
2. Click on the polygon so the 'Edit Coordinates' button can be interacted with
3. Click on 'Edit Coordinates'

PR #10080, opened 2023-09-27
9145 Fix for the visual edit coordinates bug with multiple coordinates in a polygon (dev/7.4.x)

SDScandrettKint:
CSS changes to fix the edit coordinates bug.
### Types of changes

### Description of Change
CSS fixes for the bug that caused the Edit Coordinates panel to be crammed with coordinates for a polygon with many coordinates.

### Issues Solved

#9145 

### Checklist

#### Ticket Background

### Further comments

`map-coordinate-editor-crs-selector` change from fixed to absolute prevents "Coordinate preference" panel from moving down the screen when the overall scroll is used.
image

`map-coordinate-editor-header` change from fixed to absolute prevents "Edit Coordinates" panel from moving down the screen when the overall scroll is used.
image

`map-coordinate-editor-list` change fixes the ability to scroll through all coordinates and keeps the add new coordinate at the bottom when scrolling.

### DIFF
--- a/arches/app/media/css/arches.scss
+++ b/arches/app/media/css/arches.scss
@@ -12859,7 +12859,7 @@ ul.select2-choices:after {
 .map-coordinate-editor-list {
     padding-top: 70px;
     padding-left: 5px;
-    height: 350px;
+    height: 250px;
     overflow-y: auto;
 }
 
--- a/arches/app/media/css/arches.scss
+++ b/arches/app/media/css/arches.scss
@@ -12859,7 +12859,7 @@ ul.select2-choices:after {
 .map-coordinate-editor-list {
     padding-top: 70px;
     padding-left: 5px;
-    height: 250px;
+    height: 350px;
     overflow-y: auto;
 }
 
--- a/arches/app/media/css/arches.scss
+++ b/arches/app/media/css/arches.scss
@@ -12836,7 +12836,7 @@ ul.select2-choices:after {
 }
 
 .map-coordinate-editor-crs-selector {
-    position: fixed;
+    position: absolute;
     margin-top: 25px;
     background: #fff;
     padding-bottom: 8px;
@@ -12857,8 +12857,10 @@ ul.select2-choices:after {
 }
 
 .map-coordinate-editor-list {
-    padding: 0 5px;
-    margin-top: 70px;
+    padding-top: 70px;
+    padding-left: 5px;
+    height: 250px;
+    overflow-y: auto;
 }
 
 .map-coordinate-editor-list a:focus i {
@@ -12902,7 +12904,7 @@ ul.select2-choices:after {
 }
 
 .map-coordinate-editor-header {
-    position: fixed;
+    position: absolute;
     background: #fff;
     padding: 6px 0;
     font-size: 1.2em;


arches/app/media/css/arches.scss
```diff
@@ -12857,8 +12857,10 @@ ul.select2-choices:after {
 }
 
 .map-coordinate-editor-list {
-    padding: 0 5px;
-    margin-top: 70px;
+    padding-top: 70px;
+    padding-left: 5px;
+    height: 250px;
```
chiatt, line 12862: Could you increase the height to around 350px?

chiatt (changes requested):
This is a nice improvement and fixes the bug. Eventually it would be nice to update the layout so that there isn't a need to hardcode the height. For the time being if you could increase the height of the `map-coordinate-editor-list`, then I think we can merge this in.

arches/app/media/css/arches.scss
SDScandrettKint: Resolved in the latest commmit [1d812e4](https://github.com/archesproject/arches/pull/10080/commits/1d812e43a6668f4f152484f3c7ed874ed638fc30)

SDScandrettKint:
After changing the height in the latest commit, I've noticed that a height of 350px makes the "done" button in-line with the bottom of the container, and the final coordinate is cut in half, and the add new coordinate is missing:
image

When set to 250px, the UI looks as such:
image

I do agree that not hard coding the height would be the best for the future.

chiatt:
I see - thanks for trying that @SDScandrettKint. Please change it back to 250px for the time being.

SDScandrettKint:
After some further digging into this problem and fix, I've noticed that it negatively affects the "Map card" template which isn't affected by this bug and works well without the fix.  The CSS changes to the `map-coordinate-editor-list` which change the height are the reason.

On the "Map card" template, it is the `workbench-card-sidepanel-body` that has the scroll bar.  This allows it to scroll through existing coords, the add new X & Y option and the done button (which are all displayed on the `new-provisional-edit-card-container`).

The "Default card" template has a respective `workbench-card-sidepanel` but this has the overflow-y set to hidden.  I believe when I submitted my first PR for this (a year ago) I tried to add overflow scroll to this div but it is likely to be used in many other places and the overflow will have been set to hidden for a reason.  Adding overflow to `map-coordinate-editor-list` needs a height setting before it can be used to scroll, and of course it is used by the "Map card" template so changing height (and even the scroll) affects that template negatively.

A solution would be to add `style="overflow-y: scroll;"` to [line 33 of map.htm ](https://github.com/SDScandrettKint/arches-development/blob/ea5140503a0681b7b70996492268ec533a2e3e9a/arches/app/templates/views/components/map.htm#L33) so overflow is added for only the `map workbench-card-sidepanel-body`.  
With keeping the CSS changes to `map-coordinate-editor-crs-selector` and `map-coordinate-editor-header` in this PR, the final outputs would look like the following:

Map card template:
map_card_template_coords_scroll


Default card template:
default_card_template_coords_scroll


Some notable differences/issues:
- The coordinate reference doesn't stay on top for the default card like it does for the map card
- (Very minor) when selecting edit coordinates, the map card shows the top of the coordinates list first, but the default card shows the bottom of the list first (could be its own separate issue).

chiatt (approved):

Issue #10084, opened 2023-09-27
Search export notifications fail


chiatt:
The failure results in the following error:

```
Task arches.app.tasks.update_user_task_record[f1c3100d-19b5-442e-b1bc-2d30d006997a] raised unexpected: TypeError('Object of type UUID is not JSON serializable')

  File "/arches/app/tasks.py", line 173, in update_user_task_record
    notify_completion(msg, task_obj.user, notiftype_name, context)
  File "/arches/app/tasks.py", line 346, in notify_completion
    notif = models.Notification.objects.create(notiftype=notif_type, message=msg, context=context)
```

This appears to be due to the 'link' in the context being a UUID: ```{'greeting': 'Hello,\nYour request to download a set of search results is now ready.', 'link': UUID('32ad6223-806a-4b56-96b2-fa19ea946785'), 'button_text': 'Download Now', 'closing': 'Thank you', 'email': None, 'name': None, 'email_link': 'http://localhost:8000/files/export_deliverables/Arches_2023-09-27T163324.308743.zip'}```

That link is assigned here: https://github.com/archesproject/arches/blob/6346ab34e10e5cf2331108947e097d8285c183cc/arches/app/tasks.py#L89

PR #10085, opened 2023-09-27
Ensure link property is a string not a uuid, re #10084

chiatt:
Ensure link property is a string not a uuid, re #10084


aarongundel (approved):
LGTM

PR #10092, opened 2023-09-29
Update the fix for the missing graph ids in resource x resource, #9944

njkim:
### Types of changes

### Description of Change

Improve the fix for the missing graphids in the resource_x_resource records, #9944 
Address the performance issue in some cases

### Issues Solved

#9944

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/models/migrations/9944_improve_refresh_relationship_function.py
```diff
@@ -0,0 +1,170 @@
...
+
+    update_check_excess_tiles_trigger = """
+        create or replace procedure __arches_complete_bulk_load() AS
+        $$
+            DECLARE
+                cardinality_violations bigint;
+            BEGIN
+                alter table tiles enable trigger __arches_check_excess_tiles_trigger;
+                alter table tiles enable trigger __arches_trg_update_spatial_attributes;
+            END
```
bferguso, line 137: @njkim - Can the 2 updates added to `etl_modules/save.py` be added here too? That way any missing graph IDs will be filled in when bulk load is completed.

njkim: @bferguso Sorry for the delay. It turned out this query can be still slow for the the big dataset with some missing resources and is unnecessary for the bulk editing(which is using the same save_to_tiles() method).
So we moved the query to db function as you suggested but also added a flag whether the user (the editing module) wants to run the update query or not.

arches/app/etl_modules/save.py
```diff
@@ -66,7 +66,7 @@ def save_to_tiles(userid, loadid, finalize_import=True, multiprocessing=True):
             }
         finally:
             try:
-                cursor.execute("""CALL __arches_complete_bulk_load();""")
+                cursor.execute("""CALL __arches_complete_bulk_load(%s);""",[finalize_import])
```
apeters, line 69: I think we're trying to do too much in these functions (database as well as python functions). 
I would get rid of both `__arches_prepare_bulk_load` and `__arches_complete_bulk_load` altogether.  I would move the enabling and disabling of the triggers to the `save_to_tiles` function and call them directly. 

I would then create another function that encapsulates the sql below into another sql function.
```UPDATE resource_x_resource x
SET resourceinstancefrom_graphid = r.graphid
FROM resource_instances r
WHERE r.resourceinstanceid = x.resourceinstanceidfrom
AND x.resourceinstancefrom_graphid is null;
UPDATE resource_x_resource x
SET resourceinstanceto_graphid = r.graphid
FROM resource_instances r
WHERE r.resourceinstanceid = x.resourceinstanceidto
AND x.resourceinstanceto_graphid is null;```

I would then call that function after the call to `save_to_tiles` where needed.  Same goes for `__arches_refresh_spatial_views`.  That should be pulled out of the `save_to_tiles` method and used upstream as needed.

**Review by apeters — CHANGES_REQUESTED** [review]

File: arches/app/etl_modules/save.py (hunk 2 above)
Comment (chiatt) [reply arches/app/etl_modules/save.py:69]: It would be more explicit to call specific functions from `save_to_tiles`, but `__arches_prepare_bulk_load` and `__arches_complete_bulk_load` have been available since v6.2, and were introduced to support custom sql loads using spatial views. It seems like removing them completely should wait until v8 to maintain backward compatibility.

**Review by apeters — CHANGES_REQUESTED** [review]

File: arches/app/etl_modules/save.py [hunk 3]
```diff
@@ -14,7 +14,10 @@
 def save_to_tiles(userid, loadid, finalize_import=True, multiprocessing=True):
```
apeters, line 14: at the moment, `multiprocessing` doesn't appear to do anything

apeters (changes requested):

apeters:
I guess I'm suggesting something like this:
```
def save_to_tiles(userid, loadid, multiprocessing=True):
    with connection.cursor() as cursor:
        try:
            cursor.execute("""SELECT * FROM __arches_staging_to_tile(%s)""", [loadid])
            saved = cursor.fetchone()[0]
            if not saved:
                cursor.execute(
                    """UPDATE load_event SET status = %s, load_end_time = %s WHERE loadid = %s""",
                    ("failed", datetime.now(), loadid),
                )
                return {"success": False, "data": "failed"}
            else:
                cursor.execute(
                    """SELECT g.name graph, COUNT(DISTINCT l.resourceid)
                        FROM load_staging l, resource_instances r, graphs g
                        WHERE l.loadid = %s
                        AND r.resourceinstanceid = l.resourceid
                        AND g.graphid = r.graphid
                        GROUP BY g.name
                    """, [loadid]
                )
                resources = cursor.fetchall()
                number_of_resources = {}
                for resource in resources:
                    graph = json.loads(resource[0])[settings.LANGUAGE_CODE]
                    number_of_resources.update({ graph: { "total": resource[1] } })
                cursor.execute(
                    """SELECT g.name graph, n.name, COUNT(*)
                        FROM load_staging l, nodes n, graphs g
                        WHERE l.loadid = %s
                        AND n.nodeid = l.nodegroupid
                        AND n.graphid = g.graphid
                        GROUP BY n.name, g.name;
                    """, [loadid]
                )
                tiles = cursor.fetchall()
                for tile in tiles:
                    graph = json.loads(tile[0])[settings.LANGUAGE_CODE]
                    number_of_resources[graph].setdefault('tiles', []).append({'tile': tile[1], 'count': tile[2] })

                number_of_import = json.dumps({ "number_of_import": [{ "name": k, "total": v["total"], "tiles": v["tiles"] } for k, v in number_of_resources.items()]})
                cursor.execute(
                    """UPDATE load_event SET (status, load_end_time, load_details) = (%s, %s, load_details || %s::JSONB) WHERE loadid = %s""",
                    ("completed", datetime.now(), number_of_import, loadid),
                )

        except (IntegrityError, ProgrammingError) as e:
            logger.error(e)
            cursor.execute(
                """UPDATE load_event SET status = %s, load_end_time = %s WHERE loadid = %s""",
                ("failed", datetime.now(), loadid),
            )
            return {
                "status": 400,
                "success": False,
                "title": _("Failed to complete load"),
                "message": _("Unable to insert record into staging table"),
            }

        try:
            index_resources_by_transaction(loadid, quiet=True, use_multiprocessing=multiprocessing, recalculate_descriptors=True)
            user = User.objects.get(id=userid)
            user_email = getattr(user, "email", "")
            user_firstname = getattr(user, "first_name", "")
            user_lastname = getattr(user, "last_name", "")
            user_username = getattr(user, "username", "")
            cursor.execute(
                """
                    UPDATE edit_log e
                    SET (resourcedisplayname, userid, user_firstname, user_lastname, user_email, user_username) = (r.name ->> %s, %s, %s, %s, %s, %s)
                    FROM resource_instances r
                    WHERE e.resourceinstanceid::uuid = r.resourceinstanceid
                    AND transactionid = %s
                """,
                (settings.LANGUAGE_CODE, userid, user_firstname, user_lastname, user_email, user_username, loadid),
            )
            cursor.execute(
                """UPDATE load_event SET (status, indexed_time, complete, successful) = (%s, %s, %s, %s) WHERE loadid = %s""",
                ("indexed", datetime.now(), True, True, loadid),
            )
            return {"success": True, "data": "indexed"}
        except Exception as e:
            logger.exception(e)
            cursor.execute(
                """UPDATE load_event SET (status, load_end_time) = (%s, %s) WHERE loadid = %s""",
                ("unindexed", datetime.now(), loadid),
            )
            return {"success": False, "data": "saved"}

```

apeters (approved):
works well!  thanks @njkim for all your patience

PR #9926, opened 2023-08-15
Fix race condition when saving cards in graph designer #9858

jacobtylerwalls:
### Types of changes

-   [x] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
The frontend saves cards twice, once to update the card tree, and once to update the permissions tree. If these two requests are pending, and both requests "think" a new CardXNodeXWidget must be created (on a fast architecture like an M1 Mac), then the second will fail with an IntegrityError.

https://github.com/archesproject/arches/blob/9959ce34a4485fc636bdf4e258de5f84daab96ca/arches/app/media/js/views/graph-designer.js#L294-L295

Alleviated this race condition with `get_or_create()`, so that we can be sure we don't have multiple requests each trying to insert a new CardsXNodesXWidgets.

### Issues Solved
Closes #9858

### Checklist

-   [ ] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: 
*   Found by: @jacobtylerwalls 
*   Tested by: @jacobtylerwalls 
*   Designed by: @

PR #10096, opened 2023-09-29
[WIP] Move workflow history from local storage to database #9926

jacobtylerwalls:
### Types of changes

### Description of Change
TODO:
 - [ ] add transactionid to the model so we can send it with the requests and issue deletes when deleting the workflow
 - [ ] send completed: true after completing workflow
 - [ ] reject edits after completed: true
 - [ ] smoke test the workflows
 - [ ] add PR comments describing motivations for changes

### Issues Solved

#

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/media/js/viewmodels/workflow-step.js
```diff
@@ -219,6 +222,31 @@ define([
             );
         };
 
+        this.setToWorkflowHistory = function(key, value) {
+            const workflowid = self.workflow.id();
+            const workflowHistory = {
+                workflowid,
+                completed: false,
+                stepdata: {
+                    // Django view will patch in this key, keeping existing keys
```
jacobtylerwalls, line 231: Letting the Django view just combine incoming incomplete workflow data avoids requiring each workflow step to their own GET requests to get the current state of the object. I noticed that when doing the GET before a POST back, the data might already be updated on the server by some other POST before the current request can post it back, causing lost data. There were also performance implications and lots of issues awaiting in all the right places while the UI initialized.

(Or, the whole workflow infra could have been refactored to manipulate some shared state object on the frontend, but I figured that wasn't in scope // this was cleaner.)

arches/app/media/js/viewmodels/workflow.js
```diff
@@ -288,15 +275,18 @@ define([
             if (furthestValidStepIndex !== self.furthestValidStepIndex()) {
                 self.furthestValidStepIndex(furthestValidStepIndex);
             }
-        });
+    });
 
-        this.updateStepPath = function() {
+        this.updateStepPath = async function() {
+            // Fetch step data in one query, rather than per step
```
jacobtylerwalls, line 281: Before settling on the current approach, I think I noticed like 45 or so GET requests going out when starting the new project workflow in afs. Now it's down to 1.

There are still n POST requests that go out (1 per step), but we don't await them on the frontend, so it's not a big deal.

arches/app/media/js/viewmodels/workflow.js
```diff
@@ -423,12 +416,27 @@ define([
...
-                window.location.assign(self.quitUrl);
-            }
+
+            const workflowid = self.id();
+            const workflowHistory = {
+                workflowid,
+                completed: true,
+            };
+
+            fetch(arches.urls.workflow_history + workflowid, {
```
jacobtylerwalls, line 430: We're currently not awaiting these.

arches/app/views/workflow_history.py
```diff
@@ -0,0 +1,75 @@
...
+
+class WorkflowHistoryView(View):
+
+    def get(self, request, workflowid):
+        if not user_is_resource_reviewer(request.user):
+            return JSONErrorResponse(_("Request Failed"), _("Permission Denied"), status=403)
+        try:
+            workflow_history = models.WorkflowHistory.objects.get(workflowid=workflowid, user=request.user)
+        except models.WorkflowHistory.DoesNotExist:
+            workflow_history = {}
```
jacobtylerwalls, line 23: Or we could return a 404 and then check for that on the frontend when initializing.

arches/app/views/workflow_history.py
```diff
@@ -0,0 +1,75 @@
...
+    def post(self, request, workflowid):
+        if not user_is_resource_reviewer(request.user):
+            return JSONErrorResponse(_("Request Failed"), _("Permission Denied"), status=403)
+
+        data = json.loads(request.body)
+        stepdata = data.get("stepdata", {})
+        componentdata = data.get("componentdata", {})
+
+        # TODO(Django 5.0) rewrite as a simpler update_or_create()
+        # call using different `defaults` vs. `create_defaults`
```
jacobtylerwalls, line 35: `update_or_create()` almost does what we need, but you can't use F expressions to reference the current state of a field inside a create statement (makes sense!).

I need the current state of the workflow fields to combine the keys. In Django 5.0 you can provide different behavior for create and update, so we could slim this down a bit. Right now this is just a riff on how `update_or_create()` works.

See [Django forum](https://forum.djangoproject.com/t/feature-idea-update-or-create-to-allow-different-defaults-for-create-and-update-operations/18300)

arches/app/models/models.py
```diff
@@ -1570,6 +1570,19 @@ class Meta:
         db_table = "plugins"
 
 
+class WorkflowHistory(models.Model):
+    workflowid = models.UUIDField(primary_key=True)
+    stepdata = JSONField(null=False, default=dict)
+    componentdata = JSONField(null=False, default=dict)
```
jacobtylerwalls, line 1576: I decided to split the workflow data into two JSON fields since I saw that the requests were being made separately by separate frontend viewmodels. So this way those viewmodels can patch in partial data without needing to know or care about the rest of the data.

arches/app/views/workflow_history.py
```diff
@@ -0,0 +1,75 @@
...
+                    defaults = {
+                        "stepdata": stepdata,
+                        "componentdata": componentdata,
+                        "user": request.user,
+                        "completed": data.get("completed", False),
+                    },
+                )
+            except IntegrityError:
+                # Wrong user for given workflowid.
+                return JSONErrorResponse(_("Request Failed"), _("Permission Denied"), status=403)
```
jacobtylerwalls, line 51: Would be nice to add a test case for this, but noticed the test suite doesn't run `add_test_users` yet, so we don't have more than 1 users who is a resource reviewer. (I added add_test_users() to the afs tests, so it wouldn't be a big lift.)

jacobtylerwalls: The UI should prevent you from getting into this scenario, given that we currently limit you to retrieving your own workflows.

arches/app/utils/response.py
```diff
@@ -49,7 +49,9 @@ def __init__(self, content=b"", *args, **kwargs):
 
 
 class JSONErrorResponse(JSONResponse):
-    def __init__(self, title=None, message=None, content={}, status=500, *args, **kwargs):
+    def __init__(self, title=None, message=None, content=None, status=500, *args, **kwargs):
+        if not content:
+            content = {}
```
jacobtylerwalls, line 54: New workflow tests passed in isolation but failed when combined with the `views.resource_tests.CommandLineTests`, traced it to this mutable default argument.

There's a possibility that JSONErrorRepsonses are being generated always with the first delete message rather than the correct message (like the failing test).

Not sure if this is important enough to backport. (I guess it would affect what surfaces in the UI.) If so, I can open a PR.

jacobtylerwalls:
Need to push up one more change to move step locking logic to the database also.

arches/app/views/workflow_history.py
```diff
@@ -0,0 +1,75 @@
...
+from arches.app.utils.permission_backend import user_is_resource_reviewer
+from arches.app.utils.response import JSONErrorResponse, JSONResponse
+from arches.app.models import models
+
+
+class WorkflowHistoryView(View):
+
+    def get(self, request, workflowid):
+        if not user_is_resource_reviewer(request.user):
+            return JSONErrorResponse(_("Request Failed"), _("Permission Denied"), status=403)
```
chiatt, line 19: For AFS being a resource reviewer is required, but it's possible for other applications to have workflows that only require a user to be a resource editor. They would have to adjust their permissions, but it's possible. I think instead of this check, maybe we just use the group required decorator? ```@method_decorator(group_required("Resource Editor"), name="dispatch")``` above the get and post methods?

arches/app/views/workflow_history.py
chiatt: I think what you have here is good.

arches/app/media/js/viewmodels/workflow-step.js
chiatt: I think this is a nice solution, and glad you added the comment.

chiatt (changes requested):
This looks great. The only change I think we should make is to lower the permission requirement to from Resource Reviewers to Resource Editors leaving the possibility open for developers to create workflows that collect provisional data.

arches/app/views/workflow_history.py
jacobtylerwalls: Cool, I'll relax this to Resource Editor.

Thanks for showing me the decorators. They seem to return 302 (redirected to nowhere) instead of 403:
```py
FAIL: test_get_workflow_history (tests.views.workflow_tests.WorkflowHistoryTests)
  File "/Users/jwalls/prj/arches/tests/views/workflow_tests.py", line 71, in test_get_workflow_history
    self.assertEqual(response.status_code, 403)
AssertionError: 302 != 403

FAIL: test_no_edits_after_completed (tests.views.workflow_tests.WorkflowHistoryTests)
  File "/Users/jwalls/prj/arches/tests/views/workflow_tests.py", line 159, in test_no_edits_after_completed
    self.assertEqual(response.status_code, 403)
AssertionError: 302 != 403

FAIL: test_post_workflow_history (tests.views.workflow_tests.WorkflowHistoryTests)
Partial updates of componentdata and stepdata are allowed.
  File "/Users/jwalls/prj/arches/tests/views/workflow_tests.py", line 85, in test_post_workflow_history
    self.assertEqual(response.status_code, 403)
AssertionError: 302 != 403

Ran 5 tests in 0.103s
```

That also means no banner in the UI like this if we keep using a JSONErrorResponse:
Screenshot 2023-10-16 at 6 03 03 PM


Which approach do you think is better?

chiatt: Ah - interesting. I think the 403 is better.

jacobtylerwalls: 👍 
relaxed to Resource Editor in [23c4e11](https://github.com/archesproject/arches/pull/10096/commits/23c4e114578ec730b4594aa04ab24ef1c3769301)

arches/app/models/models.py
```diff
@@ -1570,6 +1570,19 @@ class Meta:
...
+    created = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey(db_column="userid", null=True, on_delete=models.SET_NULL, to=settings.AUTH_USER_MODEL)
```
jacobtylerwalls, line 1578: Might be nice to let this cascade delete.

chiatt (approved):

Issue #10090, opened 2023-09-29
Search term-filter should include exact matches for resource displaynames


whatisgalen:
This could be classified as either a bug or a feature enhancement.

Use Case:
A user is typing a search term in Search and receives a list of dropdown terms from which to select. Should a user select a term displaying a node value alongside the name of the node/nodegroup, this term-filter is applied. However, when the term-filter is applied, it still gets treated more or less as a search term. 
![Screen Shot 2023-09-28 at 5 23 15 PM](https://github.com/archesproject/arches/assets/8366985/c8c4c4c9-630d-480c-aa48-ae666daff702)

The expected behavior is that a resourceinstance-specific value is presented in the dropdown and thus only that resource instance should show up in the search query results.
![Screen Shot 2023-09-28 at 5 04 02 PM](https://github.com/archesproject/arches/assets/8366985/91d7bb03-3cdf-43ca-b605-9563b0739f20)

The search-term objects that get sent to the frontend via the search method ``search_terms`` should be able to get the ``resourceinstanceid`` associated with the node value since it is present in the ``terms`` index.

apeters:
To clarify, the underlying issue that @whatisgalen is trying to fix is that users might expect to get an exact match to a single resource instance when selecting an item from the "Term Search" dropdown.  Right now those terms in the dropdown are really only used as suggestions.

<img width="388" alt="Screenshot 2023-09-29 at 1 07 51 PM" src="https://github.com/archesproject/arches/assets/823773/4329a1fe-5de5-4d37-aced-31713d835632">


After a conversation with both @whatisgalen and @chiatt about this ticket and [associated PR](https://github.com/archesproject/arches/pull/10094) 
what might be nice to have would several things:
1. A separate tab (bucket) in the term search dropdown that explicitly highlights "exact resource instance" matches
2. Renaming "Term Matches" to "Suggested Searches"
3. Removing the context section (text in parenthesis) of the term suggestions 
4. Only showing the "Concepts" tab to logged in, and presumably more advanced, users

whatisgalen:
Following conversation about the intial PR's associated with this issue, it was agreed that something ideal would be a new category of search term dropdown, i.e. "Exact Matches". 

(Incidentally, it was also mentioned that the label "Term Matches" should be something like "Term Suggestions" so as not to mislead the user in thinking that some logic specific to that node or its value would be applied.)

Therefore I have removed the "bug" label and will rename the issue to fit the description of a more ideal solution in PR #10102

[...]

PR #10102, opened 2023-10-01
10090 exact match dev7.5.x

whatisgalen:
### Types of changes

### TESTING THIS PR (PLEASE READ)
Because this PR changes the es mapping, you will need to reindex (delete the old and re-create with new mapping) the resources index only. it also has language support so testing with a different script is advised.

### Description of Change

This PR creates a new category of search term matches in the Search term-filter UX. See images below:
Screen Shot 2023-10-01 at 4 45 18 PM
Screen Shot 2023-10-01 at 4 45 30 PM

A category of search terms called "Exact Matches" which shows any resources whose displayname is an exact or near-exact match with the entered term. On the backend, there is a significant refactor:

- the `mappings.py` file changes the property type from `"keyword"` to `"text"` and adds an analyzer so that this field can be searchable in a more optimal way akin to how values in the `terms` index are searchable:
```
"displayname": {
      "type": "nested",
      "properties": {
          "value": {
              "type": "text",
              "analyzer": "folding",
              "fields": {
                  "keyword": {
                      "type": "keyword"
                  }
              }
          },
          "language": {
              "type": "keyword"
          }
      }
  },
```

- the method `search_terms` in `views/search.py` now handles a third and different case for querying the `resources` index with the search term for a match among `displayname`. The query logic is different and the logic handling the results is different, however a context object is passed to the frontend in the same way as in the cases of `terms` and `concepts` indexes.

- the `term_filter.py` search component class now has a case for the `term.type` value of `"exactmatch"`:
```
if term["type"] == "exactmatch":
    ids_query = Ids(ids=[term["resourceinstanceid"]])
    search_query.must(ids_query)
```

### Issues Solved

#10090 

### Checklist

#### Ticket Background
*   Sponsored by: City of Los Angeles Office of Historic Record
*   Found by: @whatisgalen 
*   Tested by: @ 
*   Designed by: @whatisgalen

### DIFF
--- a/arches/app/media/css/arches.scss
+++ b/arches/app/media/css/arches.scss
@@ -10797,11 +10797,11 @@ table.table.dataTable {
 .resource_search_widget_dropdown ul .select2-disabled .group {
     padding: 0px;
     border-top: 1px solid #C1D4F3;
-    width: 383px;
+    width: 450px;
 }
 
 .resource_search_widget_dropdown ul .select2-disabled div span span button {
-    width: 195px;
+    width: 150px;
     height: 35px;
 }
 
--- a/arches/app/media/js/bindings/term-search.js
+++ b/arches/app/media/js/bindings/term-search.js
@@ -96,10 +96,21 @@ define([
                 },
                 formatResult: function(result, container, query, escapeMarkup) {
                     var markup = [];
-                    var indent = result.type === 'concept' || result.type === 'term' ? 'term-search-item indent' : (result.type === 'string' ? 'term-search-item' : 'term-search-group');
+                    var indent = result.type === 'concept' || result.type === 'exactmatch' || result.type === 'term' ? 'term-search-item indent' : (result.type === 'string' ? 'term-search-item' : 'term-search-group');
                     if (result.type === 'group') {
                         _.each(result.text, function(searchType, i) {
-                            var label = searchType === 'concepts' ? arches.translations.termSearchConcept : arches.translations.termSearchTerm;
+                            var label;
+                            switch (searchType) {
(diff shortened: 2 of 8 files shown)


apeters (changes requested):
@whatisgalen When you first proposed this to me and @chiatt and demoed the issue and potential solution to us I was in agreement that there was an issue.

After having taking a look at the actual implementation and worked with the solution a bit I think we need to step back and reexamine the actual problem that we're trying to fix.

That problem as I understand it is two fold.

1. Terms that show up in the dropdown have a contextual bit in parenthesis that can add more confusion then help about what is actually being searched, or even what might be returned from a search.  ie, since the model name is shown one might assume that only instances of that type will be returned, which is not the case
2. Maybe more importantly, search results aren't scored, and therefore ordered, in way that that the most relevant results are at the top.  For example if the name of a resource happens to match a simple string in another resource, then those two resources will get the same score.  One would (correctly) assume that the resource that matched on the display name would receive a higher score and therefore show up nearer (or at) the top of the search results.

What I recommend is that instead of adding an `Exact Match` tab in the already crowded term search filter, we instead improve the `term` and `string` search by including the display name in the search DSL.  I would also weight this result, so that its relevance score reflects a "better" match than other types of results.  This way if a user selects a term that happens to almost exactly match a resource display name that resource will show up nearer the top of the results, which will give a more natural search experience.

If you have any questions about this @whatisgalen, I'm happy to chat with you about the details.

apeters:
Here's an example of boosting a query
```

whatisgalen:
closing in favor of #10871

PR #10104, opened 2023-10-02
Improve responsiveness for search toolbar items #9984

### LINKED ISSUE #9984: Map Filter - ANDROID 1.3.4 Orientation - Three buttons are overlapping the map on Android (Details, Time Filter, Saved & Search)
As a mobile user, I expect all components on the page to reflow into a usable format on mobile, with no components cut off or obscured.

**Linked Work Item:**
#9493 

**Expected Result:**
When using the site on mobile, all components should be visible and should not overlap one another, should not be cut off at the edge of the page and should not require the user to scroll in two dimensions in order to see them. 

**Actual Result:**
Four buttons are overlapping the map on Android and not all of the buttons are visible or accessible. This will impact user's ability to access the map and to the buttons themselves.

**Evidence:**

Image

johnatawnclementawn:
### Types of changes

### Description of Change

Makes search toolbar buttons more responsive at smaller screen sizes.

### Issues Solved

Resolves #9984

### DIFF
--- a/arches/app/media/css/abstracts/_breakpoint-settings.scss
+++ b/arches/app/media/css/abstracts/_breakpoint-settings.scss
@@ -2,8 +2,11 @@
 // across team members. It will improve communication between
 // stakeholders, designers, developers, and testers.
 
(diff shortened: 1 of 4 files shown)


apeters (approved):
looks good

PR #10110, opened 2023-10-03
removes the switching of file storage

### LINKED ISSUE #10109: Remove faking of file storage for zero byte files
Remove faking of file storage for zero byte files - it is not needed anymore and was never really a good way to avoid the zero byte files in other storages (s3)

aarongundel:
Fixes #10109 which would temporarily switch the file storage to avoid writing zero byte files to the cloud to prevent duplication - this is no longer necessary.  Also outputs the file's path - which can sometimes be changed by the file storage backend.


chiatt (approved):
Looks good to me!

Issue #9625, opened 2023-06-09
We should create a command line function to update serialized graphs when a language is enabled


chrabyrd:
We should add a command line function that will add serialized graphs for each language in the system without creating a new graph_x_published_graph value.

To reproduce:

1. Get a project running with only the default language enabled, ensure it has at least 1 Resource model
2. Enable support for multiple languages by uncommenting or adding a value in the `LANGUAGES` setting in `settings.py`
3. Manually rebuild webpack to update the language switcher
4. Select the non-default language from the switcher
5. Attempt to access the graph designer, resource editor, or resource report pages
6. Notice the application breaks because there's no serialized graph for that language

chiatt:
We have a `graphs publish` command. Maybe we could add an argument to that command? `graphs publish --update`?

apeters:
I have run into this problem as well.  It would be nice to get this fixed soon.

chrabyrd:
resolved in #10116

PR #10116, opened 2023-10-11
adds update flag to graph publish command #9625

chrabyrd:
### Types of changes

### Description of Change

Adds `--update` flag to graph publish management command

### Issues Solved

#9625

### DIFF
--- a/arches/app/models/graph.py
+++ b/arches/app/models/graph.py
@@ -1765,6 +1765,37 @@ class Graph(models.GraphModel):
             if graphs_with_matching_slug.exists() and graphs_with_matching_slug[0].graphid != self.graphid:
                 raise GraphValidationError(_("Another resource model already uses the slug '{self.slug}'").format(**locals()), 1007)
 
+    def update_published_graphs(self, user=None, notes=None):
+        """
+        Changes information in in GraphPublication models without creating
+        a new entry in graphs_x_published_graphs table
+        """
+        published_graphs = models.PublishedGraph.objects.filter(publication_id=self.publication_id)
+
+        for language_tuple in settings.LANGUAGES:
+            translation.activate(language=language_tuple[0])
+
+            serialized_graph = JSONDeserializer().deserialize(
+                JSONSerializer().serialize(self, force_recalculation=True)
+            )
+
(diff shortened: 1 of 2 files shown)


arches/app/models/graph.py
```diff
@@ -1765,6 +1765,37 @@ def validate_fieldname(fieldname, fieldnames):
             if graphs_with_matching_slug.exists() and graphs_with_matching_slug[0].graphid != self.graphid:
                 raise GraphValidationError(_("Another resource model already uses the slug '{self.slug}'").format(**locals()), 1007)
 
+    def update_published_graphs(self, user=None, notes=None):
```
apeters, line 1768: neither "user" or "notes" are used.  We should just remove those.

arches/management/commands/graph.py
```diff
@@ -87,7 +94,12 @@ def publish(self, username):
         graphids = []
         for graph in self.graphs:
             print(graph.name)
-            graph.publish(user)
+            
+            if self.update:
+                graph.update_published_graphs(user)
```
apeters, line 99: the passed in user isn't used by the function.  we can just remove this here.

apeters (changes requested):

arches/app/models/graph.py
```diff
@@ -1765,6 +1765,37 @@ def validate_fieldname(fieldname, fieldnames):
...
+            serialized_graph = JSONDeserializer().deserialize(
+                JSONSerializer().serialize(self, force_recalculation=True)
+            )
+
+            published_graph_query = published_graphs.filter(language=language_tuple[0])
+            if not len(published_graph_query):
+                published_graph = models.PublishedGraph.objects.create(
+                    publication=self.publication,
+                    serialized_graph=serialized_graph,
+                    language=models.Language.objects.get(code=language_tuple[0]),
```
apeters, line 1787: This fails if a language is added to settings.py that doesn't exist in the Languages table.
This points out an issue.  The languages defined in settings.py are strictly for the UI.  Languages defined in the table are for business data (strings, concepts, etc...).  I'm guessing that the published graph should be using the langugages from `settings.py`

apeters (changes requested):

apeters (approved):
looks good

PR #10123, opened 2023-10-12
9792 organize landingpage css for accessibility

### LINKED ISSUE #9792: _landingpage.scss should only be applied to the splash page
Styles in that file are being incorrectly applied to the main site (see the map layer manager, settings tab for an example)

johnatawnclementawn:
### Types of changes

### Description of Change

Styles in `_landingpage.scss` were improperly being applied elsewhere in Arches, mainly due to selectors that were too broad. Additionally, most of these styles were created to make the index page more accessible, so they have been moved to the accessibility specific stylesheet. Finally, I've removed some styles that were duplicated in `index.css`

### Issues Solved

Resolves #9792


arches/app/media/css/pages/_landingpage.scss
```diff
@@ -1,75 +1,5 @@
 .navbar-toggle {
```
apeters, line 1: I don't see this class used anywhere.  Can we just get rid of it?

apeters (changes requested):

arches/app/media/css/pages/_landingpage.scss
johnatawnclementawn: @apeters for screen sizes < desktop, the navbar is pushed into a dropdown:
Screenshot 2023-10-16 at 7 16 45 PM

apeters (approved):
looks good

PR #10132, opened 2023-10-16
Fix validation of required list nodes #10077

### LINKED ISSUE #10077: Marking image or resource instance list nodes as required does not work
Resource instance list and file list nodes escape validation for missing values because an empty list is not None:

https://github.com/archesproject/arches/blob/6346ab34e10e5cf2331108947e097d8285c183cc/arches/app/models/tile.py#L298

Version: 7.4
Found while testing: #10075

jacobtylerwalls:
### Types of changes

### Description of Change
Previously, required nodes having list datatypes such as file-list escaped validation for missing values.

### Issues Solved

Closes #10077

### Checklist

#### Ticket Background
*   Sponsored by: GCI
*   Designed by: @

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -1289,6 +1289,11 @@ class FileListDataType(BaseDataType):
             errors.append({"type": "ERROR", "message": message})
         return errors
 
+    def clean(self, tile, nodeid):
+        super().clean(tile, nodeid)
+        if tile.data[nodeid] == []:
+            tile.data[nodeid] = None
+
     def append_to_document(self, document, nodevalue, nodeid, tile, provisional=False):
         try:
             for f in tile.data[str(nodeid)]:
@@ -1796,6 +1801,11 @@ class DomainListDataType(BaseDomainDataType):
 
         return terms
 
+    def clean(self, tile, nodeid):
+        super().clean(tile, nodeid)
+        if tile.data[nodeid] == []:
+            tile.data[nodeid] = None
+
     def append_to_document(self, document, nodevalue, nodeid, tile, provisional=False):
         domain_text_values = set([])
         for tile in document["tiles"]:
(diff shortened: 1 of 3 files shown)


arches/app/models/tile.py
```diff
@@ -287,7 +287,7 @@ def check_for_missing_nodes(self):
                 node = models.Node.objects.get(nodeid=nodeid)
                 datatype = self.datatype_factory.get_instance(node.datatype)
                 datatype.clean(self, nodeid)
-                if self.data[nodeid] is None and node.isrequired is True:
+                if not self.data[nodeid] and node.isrequired is True:
```
chiatt, line 290: I found while testing this that if the boolean datatype is required and a user tries to set the value to `False`, the value will be reported as missing and fail to save. I looked into it a bit and I think what we need is a 'clean' method on the datatype similar to what is used on the string datatype: https://github.com/archesproject/arches/blob/e62ade0409c7e888dd31443d3411c6b3a70f7a79/arches/app/datatypes/datatypes.py#L101

chiatt (changes requested):

arches/app/models/tile.py
jacobtylerwalls: Thanks, that makes more sense.

arches/app/datatypes/datatypes.py
```diff
@@ -2114,6 +2124,11 @@ def to_json(self, tile, node):
     def collects_multiple_values(self):
         return True
 
+    def clean(self, tile, nodeid):
```
chiatt, line 2127: Nice catch adding clean to the other list datatypes. The ResourceInstanceDataType is actually a list as well (just always limited to one entry). If this clean method was moved to ResourceInstanceDataType, then it would inherit in ResourceInstanceListDataType covering both.

chiatt (changes requested):

chiatt (approved):

Issue #10127, opened 2023-10-13
Adding files to tile causes exponential growth in File objects


jacobtylerwalls:
Start with 0 files in the files table.
Save a file to a tile. Files  = 1
Add a second file to that same tile, save. Files = 3
Add a third file to that same tile, save. Files = 7
Add a fourth file to that same tile, save. Files = 14
etc.

Looks like unnecessary files are being re-created here; instead I figure we could use the tile data to determine the IDs of the existing file objects:
https://github.com/archesproject/arches/blob/df4d0916fd54dcc53253de8c698db8bfa38cabc7/arches/app/datatypes/datatypes.py#L1598-L1603

jacobtylerwalls:
Fixed in #10133

Issue #10058, opened 2023-09-21
UI shows "unsaved" upon saving second file


jacobtylerwalls:
### To reproduce ###

1. Add a file to file-list node.
2. Add a second one, save edit.

Observe UI says "unsaved" even though the "Save Edit" button is gone, indicating all the data is saved to the tile.

### Version ###
7.5

![unsaved](https://github.com/archesproject/arches/assets/38668450/223e7862-5b3d-4a89-91c6-60756cae4e4b)

jacobtylerwalls:
Fixed in #10133

PR #10133, opened 2023-10-17
fixes exponential files upload bug #10127

chrabyrd:
### Types of changes

### Description of Change

I originally sought a backend solution for this similar to what @jacobtylerwalls was thinking in the ticket. There appears to be no great way to differentiate `File` objects with posted file data. Turns out the `formData` was not getting cleared after a tile save and was therefore submitting polluted data on multiple uploads.

This also includes a small performance improvement since I was already touching the backend.

### Issues Solved

#10127 
#10058


jacobtylerwalls (approved):
looks good! 👍

PR #10158, opened 2023-10-19
Fix saving unsaved data in finish workflow action #10157

### LINKED ISSUE #10157: [workflows] this.finishWorkflow is not a function
Advance to a workflow step that allows you to have unsaved data and a Save and Complete Workflow action, such as: AfS X/Y Upload Workflow > File Interpretation step.

Click Save and Complete Workflow.

Result:
```js
Uncaught (in promise) TypeError: this.finishWorkflow is not a function
```

Regression in 7cccf96d2d755fda5f3fc9172c5586bee5351442 (unreleased)

jacobtylerwalls:
### Types of changes

### Description of Change
https://github.com/archesproject/arches/commit/7cccf96d2d755fda5f3fc9172c5586bee5351442 relied on a recursive call to `self.finishWorkflow()` to handle marking the workflow as complete, but besides the `TypeError` from using `this` instead of `self`, it was also executing a bunch of unnecessary requests instead of only sending one POST for the unsaved data and one POST for marking complete. Extracted a new function to mitigate that.

### Issues Solved
Closes #10157


arches/app/media/js/viewmodels/workflow.js
```diff
@@ -440,6 +434,16 @@ define([
             });
 
             window.location.assign(self.quitUrl);
+        }
+
+        this.finishWorkflow = function() {
+            if (self.activeStep().hasUnsavedData()) {
+                self.activeStep().save().then(function() {
+                    self.markWorkflowComplete();
+                });
```
chrabyrd, line 443: nit: this can be reduced to `self.activeStep().save().then(self.markWorkflowComplete)`

arches/app/media/js/viewmodels/workflow.js
```diff
@@ -440,6 +434,16 @@ define([
...
+                return;
+            }
+            self.markWorkflowComplete();
```
chrabyrd, line 446: nit ( since we're refactoring ): replacing the `return` statement with `if/else` logic is slightly more readable.

chrabyrd (changes requested):
overall looks good, just a few nits about stuff that was already there since we're refactoring.

chrabyrd (approved):
lgtm!

PR #10163, opened 2023-10-19
Fix race condition in concept select widget #10162

### LINKED ISSUE #10162: Race condition in concept-select.js
There's a bug in project statements.  It appears to be related to concepts.  

1.) Create a new project with the new project workflow.  Add a statement in step 2.
2) View this project resource in the UI.  Click on "Statement About Project". You'll see 
```
Cannot read properties of null (reading 'find')
TypeError: Cannot read properties of null (reading 'find')
```

jacobtylerwalls:
### Types of changes

### Description of Change
`select2` can't cope with async requests each trying to initialize the widget.

### Issues Solved
Closes #10162

### Further comments
You might see another TypeError from clicking on Project Selection, but it traces from CKEDITOR, and there's discussion [here](https://github.com/ckeditor/ckeditor4/issues/5460).

### DIFF
--- a/arches/app/media/js/viewmodels/concept-select.js
+++ b/arches/app/media/js/viewmodels/concept-select.js
@@ -175,7 +175,9 @@ define([
                             setSelectionData(value);
                         } else {
                             $.ajax(arches.urls.concept_value + '?valueid=' + ko.unwrap(value), {
-                                dataType: "json"
+                                dataType: "json",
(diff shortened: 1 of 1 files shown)


jacobtylerwalls:
Thanks @chrabyrd and @aarongundel for pointing to the multiple requests as the source of the issue!

chrabyrd (approved):
lgtm!

PR #10164, opened 2023-10-19
Place JSON configs for card components etc in correct place #10160

### LINKED ISSUE #10160: `load_package` operation doesn't copy json files into directories as expected
`load_package` (with either `-a` or `-s`) is overwriting each .json file for a group of card components, plugins, or reports as it registers them, and then placing it outside the expected `card_components/` directory and without the `*.json`.

[See disco example](https://github.com/archesproject/disco/blob/main/disco/card_components)

jacobtylerwalls:
### Types of changes

### Description of Change
Before, on running `manage.py packages -o load_package`, JSON files for further card components, reports, or plugins beyond the first would overwrite the previous one in place.

Noticed this in DISCO, where [here](https://github.com/archesproject/disco/blob/main/disco/card_components) I would have expected a folder containing two card components (instead, one card component only, called "card_components".

Will open a separate cleanup PR to DISCO.

### Issues Solved
Closes #10160

### DIFF
--- a/arches/management/commands/packages.py
+++ b/arches/management/commands/packages.py
@@ -802,10 +802,12 @@ class Command(BaseCommand):
                 modules.extend(glob.glob(os.path.join(extension, "*.py")))
 
                 if len(modules) > 0:
+                    if os.path.exists(module_dir) is False:
+                        os.mkdir(module_dir)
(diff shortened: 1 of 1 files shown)


chiatt (approved):
Looks good to me!

Issue #10130, opened 2023-10-13
Resource results show as provisional when resource does not exist


chiatt:
If a resource is indexed, but there is no resource in the database, the search result will appear in search as provisional. Ideally there wouldn't be any indexed instances without a corresponding instance in postgres, but when it happens we can hopefully indicate that in the search results rather than leaving users confused about the presence of provisional instances.
![image](https://github.com/archesproject/arches/assets/1877663/318f3ce4-e1c3-4676-bc7d-4235d50c0daa)

chiatt:
@chrabyrd Looked into this issue and found that this state doesn't occur if a resource is manually deleted. It only seems to occur as a side effect of the failure of an afs workflow. It's not clear which one or what kind of failure is needed to reproduce.

chiatt:
This issue is caused by changing an empty tile from {} to a tile with all nodeids as keys with a value of 'None' [done here](https://github.com/archesproject/arches/pull/9704). This changed how we assign the provisional property to a resource instance document.

jacobtylerwalls:
Fixed in #10179

PR #10179, opened 2023-10-20
Properly assign provisional flag during index, re #10130

chiatt:
### Types of changes

### Description of Change
Prevents the provisional flag from getting assigned to resources that have no tiles. Also, fixes logic to assign the provisional flag to resources that have tiles, but none with authoritative data.

### Issues Solved

#10130

To QA:
1. Using a provisional editor, create a resource with one provisional tile. Confirm in search that it has the provisional flag.
2. Using an authoritative editor, create a second, authoritative tile. Confirm it still has a provisional flag.
3. Delete all tiles from the resource instance. Confirm in search that it has no provisional flag.

### DIFF
--- a/arches/app/models/resource.py
+++ b/arches/app/models/resource.py
@@ -428,7 +428,7 @@ class Resource(models.ResourceInstance):
         document["numbers"] = []
         document["date_ranges"] = []
         document["ids"] = []
-        tiles_have_authoritative_data = any([any(val for val in t.data.values()) for t in tiles])
+        tiles_have_authoritative_data = any(any(val is not None for val in t.data.values()) for t in tiles)
         document["provisional_resource"] = "true" if tiles and not tiles_have_authoritative_data else "false"
 
         terms = []
--- a/arches/app/models/resource.py
+++ b/arches/app/models/resource.py
@@ -428,7 +428,8 @@ class Resource(models.ResourceInstance):
         document["numbers"] = []
         document["date_ranges"] = []
         document["ids"] = []
-        document["provisional_resource"] = "true" if sum([len(t.data) for t in tiles]) == 0 else "false"
+        tiles_have_authoritative_data = any([any(val for val in t.data.values()) for t in tiles])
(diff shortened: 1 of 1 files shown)


arches/app/models/resource.py
```diff
@@ -428,7 +428,8 @@ def get_documents_to_index(self, fetchTiles=True, datatype_factory=None, node_da
         document["numbers"] = []
         document["date_ranges"] = []
         document["ids"] = []
-        document["provisional_resource"] = "true" if sum([len(t.data) for t in tiles]) == 0 else "false"
+        tiles_have_authoritative_data = any([any(val for val in t.data.values()) for t in tiles])
```
jacobtylerwalls, line 431: Adding authoritative tiles with values testing false like `False` or `[]` won't meet this check.
```suggestion
        tiles_have_authoritative_data = any(any(val is not None for val in t.data.values()) for t in tiles)
```

jacobtylerwalls (changes requested):
Thanks for the testing instructions.

> Using a provisional editor, create a resource with one provisional tile. Confirm in search that it has the provisional flag.
Using an authoritative editor, create a second, authoritative tile. Confirm in search that it has no provisional flag.

I'm still seeing the provisional flag in this workflow. Do you see the same?

arches/app/models/resource.py
chiatt: That's a good consideration, but I think we want nodes with False to not meet this check because they can be legit boolean values and empty lists should get converted to `None` with the datatype's 'clean' method.

jacobtylerwalls: I stored a boolean value of False (authoritative tile), and then set a breakpoint at this line, and the value of `tiles_have_authoritative_data` was False. My thought was that that would be unexpected?

chiatt: Ah, sorry. I should have more carefully looked at your suggested change. That's right, we should be checking against `None`

jacobtylerwalls (approved):
Updates look good!

PR #10195, opened 2023-10-25
Add new configuration to disable automatic map extent change behaviour in map filter

### LINKED ISSUE #9564: Add new setting to disable the automatic map extent change of the map filter as the results list changes
We have had a lot of feed back from users about the behavior of the map filter constantly changing its extent as the results change.

Often users use the map as a reference/aid for doing quick or advanced searches. By involuntarily moving the map away from their focus area as they try different search criteria, it causes a lot of frustration.

Applying a map-filter to constrain the results is not always what the user want to do as it can lead to the wrong results. 

Therefore, it should be possible for the map-filter's automatic extent change behavior to be disabled.

We have done this be adding a setting to settings.py that allows this behavior to be turned off as part of the configuration.

We could also look at adding the setting to the Arches System Settings model so that the toggle can be in the UI there?

Alternatively, should the user have the option of overriding the setting in the map-filter and how can that be handled in the UI? New toggle in the filter panel?

aj-he:
### Types of changes

### Description of Change

Allow the user to disable the map filter's ability to automatically change the extent of the map in reaction to some events.

As discussed in #9564, a configuration panel for the map filter has been added, which includes a switch to enable and disable the behaviour. The state is saved to a cookie so that this does not need to be reset each time the map is opened.

image

A setting has been added to settings.py to allow the system to be given a default behaviour if necessary, but it is anticipated that this will be left to the user to decide.

### Issues Solved

fixes #9564 

### Checklist

#### Ticket Background
*   Found by: @aj-he 
*   Tested by: @aj-he @csmith-he 
*   Designed by: @aj-he


chrabyrd (approved):
Overall this works well 👍

chiatt (changes requested):

chrabyrd (changes requested):
:wave: @aj-he 

Overall this ticket works well! I appreciate the time and effort put into it.

This has kicked off an internal discussion about where the configuration UX belongs. Having a configuration tab on the search map is a bit limited, if there were other user-defined search settings in the future they may not be related to the map.

Can you instead move this to be an item in the top nav, functioning similarly to the `Help` tab? That way if there are more user-defined search settings in the future they could go in there, even if not related specifically to the map.

aj-he:
Hi @chrabyrd 

I think that creating a configuration UX that sits at the base-manager it quite a big change in scope for this ticket.

It would suggest an application wide user configuration mechanism, which would need more thought. It might even be something that would need to sit within the User Profile and saved to the database.

I think I should drop the UI configuration for this ticket, and just use the system setting and basic mechanism - this will at least provide us with the basic functionality we require.

A new user config pattern can then be developed in another ticket to support system wide user configs.

Let em know what you think?

aj-he:
closed in favour of #10207 which removes the configuration UX, which should be a larger UX enhancement

PR #8305, opened 2022-03-25
add graph icon to resource instance widget

aj-he:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

Adds resource graph icon to the resource instance widgets in to better inform the user on what they are selecting/relating

### Issues Solved

#8303 

### Checklist

-   [ ] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: Historic England
*   Found by: @aj-he 
*   Tested by: @aj-he 
*   Designed by: @aj-he

PR #10211, opened 2023-10-30
Fix race condition in resource-instance-select.js #10210

### LINKED ISSUE #10210: Race condition in resource-instance-select.js
Same issue as #10162 but for the resource-instance-select widget.

No functional impact (that we have found as of yet), but `select2` manages its own references to the DOM and can't handle parallel init'ing, so we should avoid firing multiple simultaneous requests each trying to initialize a select2 widget.

Traceback from DISCO: Sample Taking Workflow:
```js
Cannot read properties of null (reading 'find')
TypeError: Cannot read properties of null (reading 'find')
    at constructor.updateSelection (webpack://disco/./media/node_modules/select2/select2.js?:2453:42)
    at eval (webpack://disco/./media/node_modules/select2/select2.js?:2518:26)
    at eval (webpack://disco/../../arches/arches/app/media/js/viewmodels/resource-instance-select.js?:524:15)
```

jacobtylerwalls:
### Types of changes

### Description of Change
If a document has >1 resource instance select widgets, they may both try to initialize select2 widgets, which manage references to the DOM and cannot handle parallel init'ing.

### Issues Solved
Fixes #10210

Sourced to a36a9465c94a29831c499a7be89610658a225519.

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @ 

### Testing instructions

Ensure that icons still appear in the resource instance select widget per #8305.

### DIFF
--- a/arches/app/media/js/viewmodels/resource-instance-select.js
+++ b/arches/app/media/js/viewmodels/resource-instance-select.js
@@ -93,7 +93,7 @@ define([
                     .then(function(json){
                         self.graphLookup[graphid] = json.graph;
                         self.graphLookupKeys(Object.keys(self.graphLookup));
-                        self.value.valueHasMutated();
+                        self.graphIds.push(json.graph.graphid);
                         return json.graph;
                     });
             }


aj-he (commented):
👋 @jacobtylerwalls 

I can't see any issues with this change, the behaviour seems to be fine.

I have noticed however that the graph icons are no longer showing against the resources that come up in the list. I only get the `fa fa-question`, no matter what I set.

I have tracked this to the following line, where the Graph objects are turned into a dict - https://github.com/archesproject/arches/blob/9aea9d1fbd02d63afef388d53ce3719b8a49ca58/arches/app/views/api.py#L359

After this line, the iconclass property no longer has a value, so this needs to be fixed. I'll pop some screen shots below.

aj-he:
Resources no longer have their model icons...
only-question-icons

The Graph objects retrieved from the DB by graph api has the value...
graph-object-has-iconclass

... but this is then stripped out by the JSONSerializer...
iconclass-stripped-out

jacobtylerwalls:
Thanks for flagging that, I'll take a look.

jacobtylerwalls:
Hey @aj-he I can't reproduce that myself, I still have icons in the UI and I when I break at the line in your screenshot, I still have a graphid:
Screenshot 2023-11-01 at 12 37 56 PM


This is the value of `exclusions`:

```py
['cards', 'domain_connections', 'edges', 'nodegroups', 'nodes', 'widgets']
```

Is yours different?

For me, findings were the same on this branch and on dev/7.5.

jacobtylerwalls:
Whoops, you called out `iconclass`, not `graphid`. Taking another look :eyes:

jacobtylerwalls:
I still can't reproduce, but it looks like it could be a symptom of the graph publication lacking the iconclass. I know your graph object has an iconclass, but could you verify whether its related graph publication record has an iconclass? Or just a quick retest after a republishing.

aj-he:
My mistake @jacobtylerwalls. I had been playing with graphs in the database and had forgotten to repubilsh them! Sorry for wasting your time.

aj-he (approved):
LGTM

Issue #10199, opened 2023-10-26
Search Export Fails card ordering for deeply embedded nodes.


khodgkinson-he:
If the user creates a data item on a deeply embedded node..

e.g. ![image](https://github.com/archesproject/arches/assets/53860784/f3e9153a-fa56-47b3-ad6e-2ee25800c782)

and then seeks to **Export Search Results** for it, whilst EXPORT_DATA_FIELDS_IN_CARD_ORDER = True

![image](https://github.com/archesproject/arches/assets/53860784/608ee32f-cd52-4508-a849-c334dfabceff)

an error occurs..

![image](https://github.com/archesproject/arches/assets/53860784/ca471d9e-68e0-414a-84e2-8f558c76a513)

Note : Bug Fix required for 6.2 and above.

PR #10214, opened 2023-10-31
Handle multiple Subcard levels in Search Export.  #10199

khodgkinson-he:
### Types of changes

### Description of Change

In the example below, I am attempting to nest:

**Discovery
└─> Location Data
____└─>Geometry**

However as **Geometry Card** is processed before it's parent **Location Data Card** is added, it can never be appended..

image

To address this I have introduced a loop whilst sub-cards have been added to the main card list, to check if they have sub-cards in turn.

### Issues Solved

#10199 

### Further comments

This is likely an historical remnant of when Arches only allowed sub-cards nested two levels deep.

### DIFF
--- a/arches/app/search/search_export.py
+++ b/arches/app/search/search_export.py
@@ -68,10 +68,10 @@ class SearchResultsExporter(object):
                 if main_card.nodegroup_id == sub_card.nodegroup.parentnodegroup_id:
                     if sub_card not in main_card_list:
                         sub_cards_to_add.append(sub_card)
-                        subcards_added[0] = True
+                        subcards_added = True
             index_number = main_card_list.index(main_card) + 1
             main_card_list[index_number:index_number] = sub_cards_to_add
-            
+        return subcards_added    
         
 
     def return_ordered_header(self, graphid, export_type):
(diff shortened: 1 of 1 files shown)


aj-he:
@chiatt reassign as required

arches/app/search/search_export.py
```diff
@@ -93,11 +98,15 @@ def return_ordered_header(self, graphid, export_type):
...
+        subcards_added = [True]
+        while subcards_added[0]:
+            subcards_added = [False]
+            self.insert_subcard_below_parent_card(sorted_card_list, card_list_no_sort, subcards_added)
 
-        # Cards in subcard_list_with_sort are added after cards with no sort
+            # Cards in subcard_list_with_sort are added after cards with no sort
 
-        self.insert_subcard_below_parent_card(sorted_card_list, subcard_list_with_sort)
+            self.insert_subcard_below_parent_card(sorted_card_list, subcard_list_with_sort, subcards_added)
```
chiatt, line 109: @khodgkinson-he This works like a charm, but just be consistent with other parts of Arches, could you change this to use recursion rather than a while loop? Something this should work: 
```
def order_cards(subcards_added=True):
        if subcards_added == True:
            subcards_added = False
            x = self.insert_subcard_below_parent_card(sorted_card_list, card_list_no_sort, subcards_added)
            y = self.insert_subcard_below_parent_card(sorted_card_list, subcard_list_with_sort, subcards_added)
            order_cards(y)

    order_cards()
```

With that change you would need to change line 71 to `subcards_added = True` and then add a return statement to line 74, `return subcards_added`

chiatt (changes requested):

arches/app/search/search_export.py
khodgkinson-he: @chiatt Glad you like.. ^_^

I've refactored that as an inline function as recommended, any issues please let me know!

chiatt (approved):

PR #10225, opened 2023-10-31
Removes full-screen overlay for frontend errors when running webpack development server. #10224

chrabyrd:
### Types of changes

### Description of Change

Removes full-screen overlay for frontend errors when running webpack development server.

### Issues Solved

#10224

### DIFF
--- a/arches/install/arches-templates/project_name/webpack/webpack.config.dev.js
+++ b/arches/install/arches-templates/project_name/webpack/webpack.config.dev.js
@@ -20,16 +20,7 @@ module.exports = () => {
                 devServer: {
                     historyApiFallback: true,
                     client: {
-                        overlay: {
-                            errors: true,
-                            warnings: false,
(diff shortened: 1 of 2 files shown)


jacobtylerwalls:
Just wondering about the rationale, is it because we demo with the development server?

chrabyrd:
> Just wondering about the rationale, is it because we demo with the development server?

Yes, that and the overlays can be a hinderance for development and don't offer much gain otherwise. It's not surfacing any legitimate errors that don't already appear in the dev console.

jacobtylerwalls (approved):
Noticed upgrading an Arches implementation to 7.5: without this change (which is okay by me!) we would have needed to increase the lower bound on `webpack-dev-server` to 4.13.0, which is where `runtimeErrors` was introduced.

I almost PR'd that until I realized we're probably just going to merge this instead. LGTM

chiatt (approved):
Looks good!

Issue #9180, opened 2023-01-17
Tile Excel Export Not Working for Windows Deployment


csmith-he:
**Branch:** keystone/main on HE fork created from arches dev/6.1.x

**Description of Issue:** When Arches is deployed to a Windows server environment, the tile excel export does not produce a file but rather returns a server 500 error.

**Work around**

After some investigation, the work around we have employed is to specify the dir and delete parameters in the function in line 270 of arches/arches/app/views/search.py:

`with NamedTemporaryFile() as tmp:`

tmp will only be created if delete is set to false.  The directory is specified so that we can get the temporary file generated located in a directory that will be cleared down at various points

In order to allow this to work on both local environments (where we haven't had a problem as these run on docker containers using linux) and server environments is to add the following settings to settings.py:

`TILE_EXCEL_EXPORT_TEMP_DIRECTORY = None`
`TILE_EXCEL_EXPORT_TEMP_FILE_DELETE = True`

In settings_local.py for the project, these settings are also added with the TILE_EXCEL_EXPORT_TEMP_DIRECTORY set as a folder on the server and TILE_EXCEL_EXPORT_TEMP_FILE_DELETE set to false.

**Expected behaviour:** The behaviour we would hope to see is that delete could be set to True and the tile excel export file still end up being generated.

jacobtylerwalls:
Superseded by #10322

jacobtylerwalls:
Fixed in #10322

PR #10237, opened 2023-11-03
9180 Handles Tile Excel Export Generation for Windows

csmith-he:
### Types of changes

### Description of Change

Addition of an If/Else statement to handle how NamedTemporaryFile is generated based on whether the operating system is Windows or not.  Also the addition, in settings.py, of default settings relating to the arguments required for NamedTemporaryFile in a Windows deployment.

### Issues Solved

#9180

### Checklist

Tested on local deployments, Linux and Windows, to ensure correct behaviour in both instances.

#### Ticket Background
*   Found by: @csmith-he 
*   Tested by: @csmith-he 
*   Designed by: @csmith-he

### DIFF
--- a/arches/app/views/search.py
+++ b/arches/app/views/search.py
@@ -21,6 +21,7 @@ from datetime import datetime
 import logging
 import os
 import json
+import sys
 from django.contrib.auth import authenticate
 from django.contrib.gis.geos import GEOSGeometry
 from django.core.cache import cache
@@ -260,7 +261,13 @@ def export_results(request):
         exporter = SearchResultsExporter(search_request=request)
         export_files, export_info = exporter.export(format, report_link)
         wb = export_files[0]["outputfile"]
-        with NamedTemporaryFile() as tmp:
+        
+        if sys.platform != "win32":
+            tmp = NamedTemporaryFile()
+        else:
+            tmp = NamedTemporaryFile(dir=settings.TILE_EXCEL_EXPORT_TEMP_DIRECTORY, delete=settings.TILE_EXCEL_EXPORT_TEMP_FILE_DELETE)
+                    
+        with tmp:
             wb.save(tmp.name)
             tmp.seek(0)
             stream = tmp.read()
--- a/arches/app/views/search.py
+++ b/arches/app/views/search.py
@@ -260,10 +260,10 @@ def export_results(request):
     elif format == "tilexl":
         exporter = SearchResultsExporter(search_request=request)
         export_files, export_info = exporter.export(format, report_link)
-        wb = export_files[0]["outputfile"]
-        
+        wb = export_files[0]["outputfile"]  
+
         if sys.platform == "win32":
-            tmp = NamedTemporaryFile(dir=settings.TILE_EXCEL_EXPORT_TEMP_DIRECTORY, delete=settings.TILE_EXCEL_EXPORT_TEMP_FILE_DELETE)
+            tmp = NamedTemporaryFile(dir=settings.TEMP_DIRECTORY, delete=False)
         else:
             tmp = NamedTemporaryFile()
                     
@@ -273,6 +273,7 @@ def export_results(request):
             stream = tmp.read()
             export_files[0]["outputfile"] = tmp
             return zip_utils.zip_response(export_files, zip_file_name=f"{settings.APP_NAME}_export.zip")
+
     else:
         exporter = SearchResultsExporter(search_request=request)
         export_files, export_info = exporter.export(format, report_link)
--- a/arches/app/views/search.py
+++ b/arches/app/views/search.py
@@ -262,10 +262,10 @@ def export_results(request):
         export_files, export_info = exporter.export(format, report_link)
         wb = export_files[0]["outputfile"]
         
-        if sys.platform != "win32":
-            tmp = NamedTemporaryFile()
-        else:
+        if sys.platform == "win32":
             tmp = NamedTemporaryFile(dir=settings.TILE_EXCEL_EXPORT_TEMP_DIRECTORY, delete=settings.TILE_EXCEL_EXPORT_TEMP_FILE_DELETE)
+        else:
+            tmp = NamedTemporaryFile()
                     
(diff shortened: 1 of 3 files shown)


arches/app/views/search.py
```diff
@@ -260,7 +261,13 @@ def export_results(request):
...
         export_files, export_info = exporter.export(format, report_link)
         wb = export_files[0]["outputfile"]
-        with NamedTemporaryFile() as tmp:
+        
+        if sys.platform == "win32":
+            tmp = NamedTemporaryFile(dir=settings.TILE_EXCEL_EXPORT_TEMP_DIRECTORY, delete=settings.TILE_EXCEL_EXPORT_TEMP_FILE_DELETE)
+        else:
+            tmp = NamedTemporaryFile()
+                    
+        with tmp:
```
jacobtylerwalls, line 270: What do you think about just making this platform-agnostic? This should work on all platforms, combined with an `os.unlink(tmp.name)` to get rid of the tempfile after the context manager exits.

```suggestion
        with NamedTemporaryFile(delete=False) as tmp:
```

arches/settings.py
```diff
@@ -519,6 +519,9 @@
 
 OAUTH2_PROVIDER = {"ACCESS_TOKEN_EXPIRE_SECONDS": 604800}  # one week
 
+TILE_EXCEL_EXPORT_TEMP_DIRECTORY = None
+TILE_EXCEL_EXPORT_TEMP_FILE_DELETE = True
+
```
jacobtylerwalls, line 524: I'm surprised the temp files weren't cleaned up--is that's a Python bug that's been fixed in later versions? If so, then we can avoid increasing the surface area of the settings here a bit. Depending on the outcome of the prior comment, by adding `os.unlink(tmp.name)` we may not need to specify a directory at all, but let me know if I'm not grokking your use case! 🙏 

If we do keep the directory setting, it might be nice to use it on all platforms.

arches/app/views/search.py
```diff
@@ -21,6 +21,7 @@
 import logging
 import os
 import json
+import sys
```
jacobtylerwalls, line 24: Should we apply the identical patch to `app/tasks.py`?

jacobtylerwalls (commented):
Hi @csmith-he! I was interested in this one as I hazily remember looking into cross-platform temp files on a prior project.

arches/app/views/search.py
csmith-he: @jacobtylerwalls - I am thinking back a while here from when I first did the work on an internal system but I think I did try this and found that Windows (or maybe it was just our Keystone setup on Windows server) that required a specified directory for the temp file.  I'll give your suggestion a whirl on my local Windows setup to see if things will work as you've suggested

chiatt (commented):
It seems like it would be better to do away with NamedTemporaryFile altogether because I don't think it will work in cases where files need to be saved to cloud storage. Instead we could follow the pattern [here](https://github.com/archesproject/arches/blob/d084723fd602d81cc88f357162878fc27e5a069c/arches/app/etl_modules/base_excel_exporter.py#L50C11-L50C11) and use `save_virtual_workbook`. If we need to save a file we really should be using the TempFile model rather than writing to the application server.

chiatt:
Maybe it's not necessary to switch to using the TempFile model if @jacobtylerwalls's suggestion does the trick.

csmith-he:
@chiatt and @jacobtylerwalls I'm getting permission errors when writing the NamedTemporaryFile on a Windows server for a task passed of to celery; I've tried various ways of trying to get it to work but with the same result so unfortunately we're not going to be able to use Jacob's solution.

I'll give Cyrus's solution a whirl to see if that will work on a Windows server setup.

jacobtylerwalls:
Sounds good. Do you have the whole traceback you could share? I have a hunch it's coming from the excel workbook library, during the `wb.save()` call. That might be a symptom that we need to just close the temp file before wb.save() tries to reopen it. That would be consistent with a PermissionError on windows.

chiatt:
> @chiatt and @jacobtylerwalls
> 
> I've spent quite a few days trying both methods and I'm getting permissions errors with both of them. The issue seems to be when building the excel file; it doesn't matter if the account being used has write permissions to the folder the file is being written into (I have also changed the location of the TEMP folder in environment variables to ensure it's looking at a folder I know things can be written into but the same issue occurs).
> 
> image
> 
> This image provides the traceback for Jacob's solution but a similar one is returned for Cyrus's solution. Happy to have anything I've missed pointed out to me.
> 
> With the feature deadline for dev/7.5.x coming up imminently, is there some mileage in accepting the original methodology so that the functionality can work on Windows machines as well as Linux machines and then revisiting this for dev/7.6.x (which I gather is going to be the long term support version)?
> 
> Thanks, both!

Thanks for giving both of those methods a try @csmith-he. If we are going to add a setting for a temp file directory, I think it should be more generic. My concern is that you may run into the same issue with the bulk exporter (if you haven't tested it yet) and it has two formats - not just tile excel. If it were called just 'TEMP_DIRECTORY', it could be used for other purposes as they come up.

As for `TILE_EXCEL_EXPORT_TEMP_FILE_DELETE`, it seems like we could just make the value `True` and always delete in this case. 
```
tmp = NamedTemporaryFile(dir=settings.TILE_EXCEL_EXPORT_TEMP_DIRECTORY, delete=True)
```

csmith-he:
@chiatt I've set the ```delete``` in ```NamedTemporaryFile```  always to False for windows systems now because the file seems to be deleted before it's exported otherwise and then you get an error when it tries to return the zip file.  It does mean the temp folder will need to be cleared out periodically but I couldn't get this to work in Windows environments with ```delete=True```.

As you have suggested, I have changed ```TILE_EXCEL_EXPORT_TEMP_DIRECTORY``` to the more generic ```TEMP_DIRECTORY``` so it can be used elsewhere if necessary.

Modified code is now also in tasks.py - not sure why I hadn't put it in there in the first place but it does need to be in there.

Hopefully that's all ok - if not then please give me a shout.

PR #10239, opened 2023-11-06
Add missing `{{ block.super }}` to javascript.htm instructions

jacobtylerwalls:
### Types of changes

### Description of Change
`{{ block.super }}` is necessary to avoid losing all of the strings defined in core arches.

### Issues Solved
Noticed upgrading an arches project to 7.5 and including a new string.


chiatt (approved):
Looks good!

Issue #10204, opened 2023-10-27
Manifest canvases without a name, should not have the name 'undefined'


chiatt:
If a manifest has a canvas without a name, the default name rendered in the canvas gallery is 'undefined'. Instead, if there is no name, the name should simply be blank.

PR #10245, opened 2023-11-07
default canvas names should be blank - not 'undefined' #10204

aarongundel:
sets the default canvas name to 'blank' - corrects issue where it can occasionally be 'undefined'.


chiatt (approved):
Looks good to me

PR #10247, opened 2023-11-07
Update to roadmap

chiatt:
Updates roadmap


apeters (changes requested):
I think we should probably mention the thumbnail work that I think is going to make it into 7.5

apeters (approved):
Looks good!

PR #10062, opened 2023-09-22
9868 Zoom 400% reflow issue for related resources

phudson-he:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
Fix issues on smaller devices when zoomed in 400% for related resources

### Issues Solved
#9868 

### Checklist

-   [x] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: Historic England
*   Found by: @phudson-he
*   Tested by: @phudson-he
*   Designed by: @phudson-he

PR #10273, opened 2023-11-15
9868 jmc related resource & map reflow improvements in search

### LINKED ISSUE #9868: Related Resources - 1.4.10 Reflow - At 400%, most of the content of the panel is cut off
As a user reliant on the highest zoom setting to perceive content, all text and components should be visible and accessible at 400% zoom and should reflow into a single column. No content should be obscured by other content or cut off by the edge of the page. Users should not be forced to scroll in two dimensions to view content.

**Linked work item:**
#9388  

**Expected Result:**
At 400%, all content should reflow into a single column without any content being cut off, obscured or needing two dimensional scrolling to view.

**Actual Result:**
When the user zooms in to 400% to view the Related Resources panel, most of the left hand side and the bottom of the panel is cut off, including the entire map. 

**Evidence:**

Image

johnatawnclementawn:
### Types of changes

### Description of Change

Builds on [Paul's](@phudson-he) work in #10062, fixing reflow of Related Resources and Map in Search. Map was fixed along with RR because the map tabs were subject to the same issues presented in the RR viewer and the overall containers that render the content are the same.

### Issues Solved

Resolves #9868


apeters (approved):
looks good

PR #10275, opened 2023-11-16
fixes resource.load_tiles method to only filter if given perm #10274

chrabyrd:
### Types of changes

### Issues Solved

#10274


arches/app/models/resource.py
```diff
@@ -250,7 +250,7 @@ def load_tiles(self, user=None, perm=None):
         """
 
         self.tiles = list(models.TileModel.objects.filter(resourceinstance=self))
-        if user:
+        if user and perm:
```
chiatt, line 253: I can't think of a case in which we would load instance tiles for a user without checking if that user should have access to those tiles. I think a more secure way to address this is to change the default value for `perm` from None to `'read_nodegroup'`.

chiatt (changes requested):

arches/app/models/resource.py
chrabyrd: done 👍

chiatt (approved):

Issue #10281, opened 2023-11-18
Unneeded plugins directory in new arches applications


chiatt:
A new arches application contains a plugins directory. It appears to be unused and should probably be removed from the app template.

Issue #10280, opened 2023-11-18
makemigrations fails to recognize new models in arches applications


chiatt:
When a user creates a new Arches App, the app contains a models directory rather than a models.py file. If an app developer creates models within the models directory and then runs `makemigrations`, django will not find the models and no migration will be created. If a models.py file is added outside of the models directory, and the user runs `makemigrations`, django will still fail to find the new models. Using a models.py file and removing the models directory fixes this issue.

New apps should probably not add a models directory, but just a models.py file. This would also be consistent with the default structure of a new Django app.

jacobtylerwalls:
> New apps should probably not add a models directory, but just a models.py file. This would also be consistent with the default structure of a new Django app.

👍 

Apps that anticipate having a large number of models can use the folder structure if they [remember to import the models](https://docs.djangoproject.com/en/4.2/topics/db/models/#organizing-models-in-a-package) in `__init__.py` or add a wildcard import there.

chiatt:
Good point. That should probably be added to the docs.

PR #10290, opened 2023-11-22
updates models pattern && removes plugins directory #10280

chrabyrd:
### Description of Change

Adds `models.py` to arches-app-templates
Removes `models` and `plugins` directories from arches-app-templates

### Issues Solved

#10280 
#10281


chiatt (approved):
Looks good!

PR #10300, opened 2023-11-28
Make primary descriptor fn aware of which description it's calculating #10299

### LINKED ISSUE #10299: Make `get_primary_descriptor_from_nodes()` aware of which descriptor (name, map_popup, description)
Arches for Science has a use case for overriding `AbstractPrimaryDescriptorsFunction.get_primary_descriptor_from_nodes()` to read a calculated value from the db instead of running python logic to calculate a descriptor from scratch. To do that, the function needs knowledge of *which* descriptor's configuration was passed in, e.g. "map_popup", "name", or "description".

See the use case here:

> @chiatt We should consider patching this in Arches 7.5. As it stands, this function has no knowledge of what descriptor was passed in (which we need if we want to be able to read the JSON field from the db and look up the given descriptor by key).

_Originally posted by @jacobtylerwalls in https://github.com/archesproject/arches-for-science/pull/1452#discussion_r1401247567_

jacobtylerwalls:
### Types of changes

### Description of Change
See use case in #10299: when overriding this function, it may be useful to know _which_ kind of descriptor is being resolved.

### Issues Solved
Closes #10299

### DIFF
--- a/arches/app/functions/primary_descriptors.py
+++ b/arches/app/functions/primary_descriptors.py
@@ -10,7 +10,7 @@ logger = logging.getLogger(__name__)
 
 
 class AbstractPrimaryDescriptorsFunction(BaseFunction):
-    def get_primary_descriptor_from_nodes(self, resource, config, context=None):
+    def get_primary_descriptor_from_nodes(self, resource, config, context=None, descriptor=None):
         """
         Arguments:
         resource -- the resource instance to which the primary decriptor will be assigned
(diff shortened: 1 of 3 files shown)


bferguso:
@jacobtylerwalls, @chiatt  - Unless I'm missing something, I think this should already be available in the config parameter. This is how I'm determining this in my custom descriptors:
```
            if config["type"] == "name":
                .... logic ...
```

jacobtylerwalls:
Thanks for linking out to your implementation @bferguso. That approach would require altering the function config.

What do you think about the current approach on this PR? Is it a benefit that the function configs wouldn't have to be altered?

bferguso:
Hi @jacobtylerwalls  - I think that makes perfect sense. I'd forgotten that I'd modified my local function configs to make it work so I think what you've proposed is great.

chiatt (approved):
Looks good!

Issue #10307, opened 2023-11-29
Advanced search allows for searching against a node that doesn't exist in that nodes collection


chiatt:
As described in [this forum post](https://community.archesproject.org/t/searching-for-concept-brings-up-results-outside-of-collection/2189) users can select a concept value in advanced search that does not exist in the collection of the queried node.

re: #10306

johnatawnclementawn:
Resolved in #10308

Issue #10306, opened 2023-11-29
Concepts outside of collection selectable in concept widget


chiatt:
As described [in this forum post](https://community.archesproject.org/t/searching-for-concept-brings-up-results-outside-of-collection/2189) concepts can be saved to a node that do not exist in the collection assigned to that node. 

This could result in users saving incorrect data to a tile. 

re: #10307

johnatawnclementawn:
Resolved in #10308

PR #10308, opened 2023-11-29
Prevent concept widget from allowing selection of concepts outside collection

chiatt:
### Types of changes

### Description of Change
Previously a second db query was performed to identify exact matches and place them at the top of a concept widget result set. This allowed concepts to appear that existed outside of a nodes collection. This fix identifies exact matches limited within a nodes collection. 

### Issues Solved

#10307, #10306


johnatawnclementawn (approved):
Looks good to me!

PR #10317, opened 2023-11-30
show large thumbnail preview when hovering on search result thumbnail #10310

chrabyrd:
### Types of changes

### Issues Solved

#10310


chiatt (changes requested):
This looks really good, but I noticed that for landscape images it would be nice to perhaps have a light gray background or maybe just a border so that the title doesn't appear pushed to the right by whitespace. 
image

chrabyrd:
> This looks really good, but I noticed that for landscape images it would be nice to perhaps have a light gray background or maybe just a border so that the title doesn't appear pushed to the right by whitespace. image

@chiatt resolved in latest

arches/app/templates/views/components/search/search-results.htm
```diff
@@ -8,6 +8,23 @@
...
+                        event: { 
+                            mouseover: mouseoverThumbnail,
+                            mouseout: mouseoutThumbnail,
+                        }
+                    "
+                >
+                    <img style="max-width: 70px;" data-bind="attr:{src: thumbnail}">
+                </div>
+
+                <img style="position: absolute; left: 380px; display: none;" data-bind="attr:{src: thumbnail}">
```
chiatt, line 25: Absolute positioning with left/right is a problem for RTL. Is there another way to do this?

chiatt (changes requested):

arches/app/templates/views/components/search/search-results.htm
chrabyrd: fixed in latest

chiatt (approved):

PR #10320, opened 2023-11-30
add code to support thumbnail generation (currently noop)

apeters:


chiatt (approved):
This looks good. When it comes time to writing make_thumbnail, we should be mindful of potential Windows issues. https://github.com/archesproject/arches/pull/10322/files

PR #10322, opened 2023-12-01
Add Windows compatibility to tile excel export #9180

jacobtylerwalls:
Hey @csmith-he :wave: Thanks for the additional details on #9180--very helpful!

***
I opened this draft just to demo my idea at https://github.com/archesproject/arches/pull/10237#discussion_r1384194720.

Really, it's a lot like your solution in #10237, but it just programmatically deletes the file (and thus nothing left behind to clean up.)

From the traceback you posted, the failure on `wb.save()` suggests to me that the file was opened with `delete=True`, which I would [expect to fail on Windows](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) with PermissionError if reopened (as wb.save() tries to do).

***
I'd be grateful if you'd be willing to see if this diff solves your use case. Or if it doesn't, if it perhaps produces a different traceback.

### DIFF
--- a/arches/app/tasks.py
+++ b/arches/app/tasks.py
@@ -65,12 +65,13 @@ def export_search_results(self, userid, request_values, format, report_link):
         exporter = SearchResultsExporter(search_request=new_request)
         export_files, export_info = exporter.export(format, report_link)
         wb = export_files[0]["outputfile"]
-        with NamedTemporaryFile() as tmp:
+        with NamedTemporaryFile(delete=False) as tmp:
             wb.save(tmp.name)
             tmp.seek(0)
             stream = tmp.read()
             export_files[0]["outputfile"] = tmp
             exportid = exporter.write_export_zipfile(export_files, export_info, export_name)
+        os.unlink(tmp.name)
     else:
         exporter = SearchResultsExporter(search_request=new_request)
         files, export_info = exporter.export(format, report_link)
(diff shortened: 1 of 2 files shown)


jacobtylerwalls:
Or, if this is still a no-go on Windows, we can add the specific directory back in and at least still have a cross/platform implementation.

csmith-he:
@jacobtylerwalls This is working on local and server instances of the application on Windows - thank you for taking the time to put this together.

How do you want to handle getting this code into 7.5.x?

It might also be an idea to look at handling if tmp isn't created and if result in search.py isn't created so the code will fail gracefully.

jacobtylerwalls:
Terrific! Appreciate your giving this a look.

> How do you want to handle getting this code into 7.5.x?

I'll close the other PR, open this one for review, and I've just added you as a co-author. When you're satisfied, just send me back an approval!

> It might also be an idea to look at handling if tmp isn't created and if result in search.py isn't created so the code will fail gracefully.

Yeah, happy to think this through. What do you figure would be most helpful, should we catch PermissionError and reraise it as something else? (I guess I was thinking PermissionError would be fine, but I'm open to ideas!)

csmith-he:
> > It might also be an idea to look at handling if tmp isn't created and if result in search.py isn't created so the code will fail gracefully.
> 
> Yeah, happy to think this through. What do you figure would be most helpful, should we catch PermissionError and reraise it as something else? (I guess I was thinking PermissionError would be fine, but I'm open to ideas!)

I wonder if maybe getting it into a logger.error or something might be helpful - maybe with a try/catch around it so it's easy to see in the code exactly where the issue is?

jacobtylerwalls:
Added a little log:
```
2023-12-04 20:26:09,612 arches.app.views.search ERROR    Temp file could not be created.
```

csmith-he (approved):
I'm happy with this. Approved

PR #10325, opened 2023-12-04
Remove `mock` from dev dependencies

jacobtylerwalls:
`mock` was just a backport of the unittest.mock module, which has been available in the stdlib since Python 3.3.

Noticed this when trying to run tests after forgetting to install the dev dependencies.

Other dev dependencies are already being removed in 7.5, so it seemed a good time to propose one more cleanup.


chiatt (approved):
Looks good!

PR #10265, opened 2023-11-13
add accessibility fixes for the public facing side of Arches

apeters:
to test, make sure to do a yarn.install first

PR #9814, opened 2023-07-21
fixes broken card tree #9805

chiatt:
### Types of changes

-   [x] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
Reverts knockout because 3.5.1 breaks card tree in designer

### Issues Solved

#9805

PR #10329, opened 2023-12-04
Remake yarn lockfile

jacobtylerwalls:
Refs #9814
Refs #10265

### Further comments
Needed to do this to get going on updating arches for science work after the selectWoo changes. A knockout downgrade from #9814 also involved here.


chrabyrd (approved):
lgtm!

PR #10323, opened 2023-12-01
Rate-limit login views

jacobtylerwalls:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
Add rate limiting to auth views. Expose `RATE_LIMIT` as a setting. See [accepted possibilities](https://django-ratelimit.readthedocs.io/en/stable/rates.html#rates-chapter), e.g. `12/5m`, or a python callable.

PR #10339, opened 2023-12-05
Add system check for cache backends capable of rate-limiting

jacobtylerwalls:
Adds a system check (warning) if your configuration is not using a cache backend that supports rate-limiting the auth views.

Refs #10323

```
arches % python3 manage.py check
System check identified some issues:

WARNINGS:
?: (arches.W001) Cache backend does not support rate-limiting
        HINT: Your cache: django.core.cache.backends.locmem.LocMemCache
        Supported caches: ('django.core.cache.backends.memcached.PyLibMCCache', 'django.core.cache.backends.memcached.PyMemcacheCache', 'django.core.cache.backends.redis.RedisCache')
```

Other manage.py commands also surface checks, e.g. `runserver` and `migrate`. The check can be silenced with `SILENCED_SYSTEM_CHECKS`.

@chiatt I'll defer to you on what you think the release note should be, if any.

### DIFF
--- a/arches/apps.py
+++ b/arches/apps.py
@@ -1,0 +1,21 @@
+from django.conf import settings
+from django.core.checks import register, Tags, Warning
+
+@register(Tags.security)
+def check_cache_backend(app_configs, **kwargs):
+    errors = []
+    supported_by_django_ratelimit = (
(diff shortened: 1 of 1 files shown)


aarongundel (approved):
Looks good and works for me.

PR #10340, opened 2023-12-05
Add system check for using dummy cache in production

jacobtylerwalls:
The dummy cache shouldn't be used in production. Users won't get:
- rate-limiting on auth views
- map tile caching

Add a system check that only fires if DEBUG is False.

### DIFF
--- a/arches/apps.py
+++ b/arches/apps.py
@@ -1,0 +1,16 @@
+from django.conf import settings
+from django.core.checks import register, Tags, Error
+
+
+@register(Tags.security)
+def check_cache_backend_for_production(app_configs, **kwargs):
+    errors = []
+    your_cache = settings.CACHES["default"]["BACKEND"]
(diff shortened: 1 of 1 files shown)


tests/base_test.py
```diff
@@ -55,6 +55,14 @@
 
 
 class ArchesTestRunner(DiscoverRunner):
+    def __init__(self, *args, **kwargs) -> None:
+        kwargs["debug_mode"] = True
+        # Unless the user has something other than the Django default, give them
+        # what they probably want.
+        if kwargs["pattern"] == "test*.py":
+            kwargs["pattern"] = "*.py"
```
jacobtylerwalls, line 63: Since I was already patched in here to turn on `--debug-mode`, figured I'd set a better default for `--pattern` so that `manage.py test` works more like it used to before 7.5.

chiatt (approved):
Looks good!

PR #10347, opened 2023-12-05
updates rate limit to multiply by 2 for new auth frontend

aarongundel:
rate limit for 7.x arches - doubling for the log in view.

### DIFF
--- a/arches/app/views/auth.py
+++ b/arches/app/views/auth.py
@@ -85,7 +85,11 @@ class LoginView(View):
     @method_decorator(
         ratelimit(
             key="post:username",
-            rate=("{}/{}".format(int(settings.RATE_LIMIT.split("/")[0]) * 2, settings.RATE_LIMIT.split("/")[1])),
+            rate=(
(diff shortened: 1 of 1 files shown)


arches/app/views/auth.py
```diff
@@ -81,7 +82,13 @@ def get(self, request):
                 },
             )
 
-    @method_decorator(ratelimit(key="post:username", rate=settings.RATE_LIMIT, block=False))
+    @method_decorator(
+        ratelimit(
+            key="post:username",
+            rate=("{}/{}".format(int(settings.RATE_LIMIT.split("/")[0]) * 2, settings.RATE_LIMIT.split("/")[1])),
+            block=False,
+        )
```
jacobtylerwalls, line 90: django-ratelimit accepts `None` or python callables, so let's make this a little more robust and check it's a string before splitting it

jacobtylerwalls (changes requested):

jacobtylerwalls (approved):
Tested, works well!

Issue #10353, opened 2023-12-06
Provide hook to allow custom map configuration


bferguso:
Currently there is no way to provide application-specific map configuration without overriding the entire `viewmodels/map.js` file. This would be helpful to perform configurations such as adding a map scale, or making images available for map styling.

PR #10354, opened 2023-12-06
10353 add map config hook

bferguso:
### Types of changes

### Description of Change
Adds a hook to perform configuration of the map at the beginning and end of the map.on('load') event.

### Issues Solved

#10353 

### Checklist

#### Ticket Background
*   Sponsored by: Self Funded
*   Found by: @bferguso
*   Tested by: @bferguso
*   Designed by: @bferguso

### DIFF
--- a/arches/app/media/js/utils/map-configurator.js
+++ b/arches/app/media/js/utils/map-configurator.js
@@ -1,0 +1,13 @@
+define([], function(){
+
+    let mapConfigurator = {
+        preConfig: function(map) {
+            // This can be used to configure the map as the beginning of the map.on('load') event
+        },
+        postConfig: function(map) {
+            // This can be used to configure the map as the end of the map.on('load') event
+        },
+    };
+
+    return mapConfigurator;
+});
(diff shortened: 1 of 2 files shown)


bferguso:
@jacobtylerwalls - I was just overriding the noop js/utils/map-configurator.js by implementing that file in the project, much like overriding any other base javascript file. I guess that has the downside of the base javascript file essentially being dead code or treated as an abstract class, but has the advantage of not having to affect/extend/override any other files.

bferguso:
It's effectively the same pattern that was used with the map-popup-provider.js. Yes, it would mean a single customization file, but that object could delegate to any number of other classes. I've been trying to minimize the number of templates that need to be overridden as it makes the upgrade path more brittle (something I'm dealing with right now with the 7.4.x->7.5.x upgrade).

jacobtylerwalls (approved):
Got it, thanks for pointing me to map-popup-provider.js.

Yeah, this seems like a nice way to support both avoiding new templates plus having a way to do per-template config via delegation if you have new templates for other reasons, like plugin development.

Thanks for walking me through it!

PR #10356, opened 2023-12-06
Update `update_fields` in custom save() methods #10352

### LINKED ISSUE #10352: [Django 4.2] Harden `save()` methods against update_fields coming from `update_or_create()`
From Django 4.2, when custom save methods set model model fields, they [need to update](https://docs.djangoproject.com/en/4.2/releases/4.2/#setting-update-fields-in-model-save-may-now-be-required) `update_fields`, otherwise they won't take effect.

In this example for a fully provisional tile, the TileModel.save() method is supposed to increment the sortorder by 1 (3 -> 4). After an `update_or_create()`, it doesn't actually take effect, see refreshed value from db at the end.
```py
>>> from arches.app.models import models
>>> tiles = models.TileModel.objects.filter(resourceinstance="852a4792-672f-46b4-9b49-f8e470261d23")
>>> [t.sortorder for t in tiles]
[3]
>>> first = tiles[0]
>>> obj, created = models.TileModel.objects.update_or_create(pk=first.pk, nodegroup=first.nodegroup)
>>> obj.sortorder
4
>>> obj.refresh_from_db()
>>> obj.sortorder
3
```

AfS, at least, does make a TileModel.objects.update_or_create() call. We should just quickly audit the custom save methods and make sure they update `update_fields`.

jacobtylerwalls:
### Types of changes

### Description of Change
`update_or_create()` will break on Django 4.2 for models with custom save methods that set model fields unless we account for this by updating the `update_fields` argument. See [release notes](https://docs.djangoproject.com/en/4.2/releases/4.2/#setting-update-fields-in-model-save-may-now-be-required).

Added two tests (for the more important models of the three affected) that fail on the target branch.

I'm fairly certain we're okay with the proxy model save methods; just need to look at the ones that actually accept `**kwargs`.

### Issues Solved
Closes #10352

### Checklist

#### Ticket Background
*   Sponsored by: Farallon
*   Designed by: @

### DIFF
--- a/arches/app/models/models.py
+++ b/arches/app/models/models.py
@@ -937,9 +937,13 @@ class ResourceXResource(models.Model):
 
         if not self.created:
             self.created = datetime.datetime.now()
+            if (update_fields := kwargs.get("update_fields")) is not None:
+                update_fields.add("created")
         self.modified = datetime.datetime.now()
+        if (update_fields := kwargs.get("update_fields")) is not None:
+            update_fields.add("modified")
 
-        super(ResourceXResource, self).save()
+        super(ResourceXResource, self).save(*args, **kwargs)
 
     def __init__(self, *args, **kwargs):
         super(ResourceXResource, self).__init__(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


chiatt (approved):
This looks great. Glad you caught this issue!

PR #10363, opened 2023-12-07
Harden patching of `update_fields` against more iterables

jacobtylerwalls:
Follow-up to 60f59e3 (#10352)

Django sends the `update_fields` argument a set, which is what I patched in #10356, but arches itself sends lists, which is supported also.

Handle more iterables.


chiatt (approved):
Very nice refactor

Issue #10331, opened 2023-12-04
Items in related resources list have vertical scrollbar


chiatt:
In the resource-instance widget, added items have vertical scroll bar

![image](https://github.com/archesproject/arches/assets/1877663/38c39a54-4a4f-4083-89fb-6c920c2f2ba0)

PR #10372, opened 2023-12-07
fix css nits, re #10331

apeters:
re #10331


chiatt (approved):
This looks much better

PR #10377, opened 2023-12-08
fixes issue with django_storage path not implemented

aarongundel:
A fix for the pathing issue that comes about when using non local file system storages.  

Also some black auto-formatting.


njkim (approved):
lgtm

Issue #10351, opened 2023-12-06
Resource reports don't allow users to scroll if content overflows the view


apeters:

whatisgalen:
following

jacobtylerwalls:
Closed in #10379

PR #10379, opened 2023-12-08
10351 fix report no scroll

apeters:
re #10351

### DIFF
--- a/arches/app/media/css/arches.scss
+++ b/arches/app/media/css/arches.scss
@@ -1137,11 +1137,6 @@ div.wf-multi-tile-card-info div {
     color: #ccc;
 }
 
-.resource-component-abstract {
-    overflow-y: scroll;
-    height: 100%;
-}
-
 .workflow-plugin {
     flex-grow: inherit;
     background-color: #fff;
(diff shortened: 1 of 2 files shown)


jacobtylerwalls:
If I'm looking at the right place, then I too don't see a horizontal scroll:

report-small-width

apeters:
@jacobtylerwalls this was supposed to fix a vertical scroll issue, I didn't look into horizontal scroll

jacobtylerwalls:
@apeters Ah, sorry. Do you recall how to reproduce the original problem without this change, then?

apeters:
@jacobtylerwalls well, any longish report (that flows beyond the bottom of the page) shouldn't be able to be scrolled.  sometime if you don't hide the null values in a report, then it becomes long enough to see the issue.

jacobtylerwalls:
Will take a quick peek at arches-for-science, since there's an [extra `!important` over there](https://github.com/archesproject/arches-for-science/blob/a98815b8f248ca477562c910a06786830e31d1b0/arches_for_science/media/css/report_templates.css#L1-L5)

jacobtylerwalls (approved):
Looks great. I'll follow-up with the same change in AfS.

PR #10385, opened 2023-12-08
Demote no-ontologies UI error to a logged warning #9649

### LINKED ISSUE #9649: False error message when importing graph
When importing a graph without an ontology, the following error message appears. One would think that the graph failed to load, but despite the error it loads anyway.

image

jacobtylerwalls:
### Types of changes

### Description of Change
There's nothing wrong with importing graphs into an environment without ontologies in their db. Demote this UI error to a logged warning.

### Issues Solved

Closes #9649

### DIFF
--- a/arches/app/utils/data_management/resource_graphs/importer.py
+++ b/arches/app/utils/data_management/resource_graphs/importer.py
@@ -92,7 +92,7 @@ def import_graph(graphs, overwrite_graphs=True, user=None):
                 ]:
                     errors.append("The ontologyid of the graph you're trying to load does not exist in Arches.")
             else:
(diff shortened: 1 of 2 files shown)


chiatt:
I'm not sure if we should reduce this to a warning or actually prevent the graph from loading. If the graph loads, it will break until the ontologies are loaded.

jacobtylerwalls:
But only for graphs with ontologies, right? I think that's what's checked in the `if` branch directly above. What if you never use ontologies in arches?

chiatt:
> But only for graphs with ontologies, right? I think that's what's checked in the `if` branch directly above. What if you never use ontologies in arches?

Ah - indeed. I just needed to expand the code above your change to see that 'if'.

chiatt (approved):
Looks good!

PR #10395, opened 2023-12-11
Apply `update_fields` patching to new thumbnail field 

jacobtylerwalls:
### Types of changes

### Description of Change
Apply the fixes from #10352 to the new thumbnail code.

### Further comments
I saw some advice on the Django forum that what we probably want to be doing is to use `pre_save()` for these kinds of assignments to self so that `update_fields` is tracked for us. For another day, perhaps (haven't tested that out).


chiatt (approved):
Looks good!

Issue #9333, opened 2023-03-17
Add Image Datatype


chiatt:
Currently images are managed in arches using the file-list datatype. However, there is commonly data associated with images such as caption, credit, and location that currently must be recorded in separate nodes in a card. With an image datatype this information could be saved alongside the url, perhaps accompanied by other image information such as dimensions and pixel depth. This would reduce the number of nodes needed to track metadata about images. It would also simplify design around components such as the 'PhotoGalleryCard' or that are not intended for use with other file types.

azaroth42:
+1 to this issue.
Also, the URL to the image should be able to be outside of Arches, e.g. point to a IIIF instance, or to any image on the web.

apeters:
closed by #9334

PR #10408, opened 2023-12-12
Avoid publishing graphs in migrations

jacobtylerwalls:
### Description of Change
Using a proxy model in a migration can create drift. Better to accomplish this with a `publish` command inside the `updateproject` command.

### Issues Solved
Follow-up to #9333

### Checklist

### DIFF
--- a/arches/management/commands/updateproject.py
+++ b/arches/management/commands/updateproject.py
@@ -1,5 +1,11 @@
+import arches
+import os
+import shutil
+
 from django.core.management.base import BaseCommand
 from django.core import management
+from django.db import connection
+from arches.app.models.system_settings import settings
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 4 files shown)


arches/management/commands/updateproject.py
```diff
@@ -64,3 +65,8 @@ def update_to_v7(self):
                         graphs.append(graphid)
                 management.call_command("graph", operation="publish", update_instances=True, graphs=",".join(graphs))
                 cursor.execute("drop table if exists temp_graph_status")
+
+    def update_to_v7_5(self):
+        # Republish to pick up CardsXNodeXWidget changes from
+        # migration 10150_add_default_value_file_list.py
+        management.call_command("graph", operation="publish")
```
chiatt, line 72: When I tested this it ended up publishing even unpublished graphs. There could be a reasons not to update the existing publications, but if we pass in the update flag we can prevent publishing unpublished models. 

I tried that just to confirm and found I got an integrity error if the graph was not already published, so I had to wrap it in a try block. 
```
from django.db.utils import IntegrityError

try:
    management.call_command("graph", operation="publish", update=True)
except IntegrityError:
    pass
```

jacobtylerwalls: Ah got it, thanks. Added `--update` flag.

Re: the IntegrityError, I'd expect that to roll back the entire transaction. I'll see if I can do something about the root cause tomorrow?

jacobtylerwalls:
I noticed the graph command only works on resource models, not branches. Is that okay here?

arches/management/commands/updateproject.py
jacobtylerwalls: #10410

chiatt (approved):
Looks good!

Issue #10457, opened 2023-12-19
Add default timewheel cache timeout 


chiatt:
Caching the timewheel config can result in a significant performance boost in search. Currently we allow admins to configure the cache timeout only by user. This is because the shape of the config can be different for each user based on permissions. In many cases admins aren't going to list different timeout periods for each user, so should provide a default that can apply to any user.

johnatawnclementawn:
Resolved in #10461

PR #10461, opened 2023-12-19
Adds v5 timewheel cache default

chiatt:
### Types of changes

-   [x] Performance improvement

### Description of Change
Adds a default cache timeout to the CACHE_BY_USER setting so that the timewheel config can be cached for all users.

### Issues Solved

#10457


johnatawnclementawn (approved):
Looks good to me!

PR #10462, opened 2023-12-19
Replace `strtobool()` #10460

### LINKED ISSUE #10460: Python 3.12 compat: Replace `strtobool()`
The boolean datatype uses `strtobool()` in a couple places, which comes from earlier versions of `setuptools`, which we were pulling in as a transitive dependency from `mapbox-vector-tile`.

Noticed on a fresh install of Python 3.12 (which no longer has `distutils` anyway).

We can write a teeny replacement to keep the existing functionality. Suggest upgrading `mapbox-vector-tile` so that fresh installs don't pull in `setuptools` and the failure can be seen more clearly.

jacobtylerwalls:
### Types of changes

### Description of Change
Depends on archesproject/arches#10456

On python 3.12, Arches had an undeclared dependency on a util (`strtobool()`) from older versions of `setuptools` which still vendored distutils, since `distutils` is no longer in the stdlib. Instead just reimplement the functionality to avoid pulling in setuptools or anything else as a dep.

~Bumps `mapbox-vector-tile` to a version that doesn't pull in `setuptools` as a dependency (and a version that explicitly supports python 3.11).~ EDIT: removes it altogether.

### Issues Solved
Closes #10460

### Checklist

### DIFF
--- a/arches/app/utils/string_utils.py
+++ b/arches/app/utils/string_utils.py
@@ -1,0 +1,7 @@
+def str_to_bool(value):
+    match value:
+        case "y" | "yes" | "t" | "true" | "on" | "1":
+            return True
+        case "n" | "no" | "f" | "false" | "off" | "0":
+            return False
+    raise ValueError
--- a/arches/install/requirements.txt
+++ b/arches/install/requirements.txt
@@ -7,7 +7,6 @@ django-guardian==2.4.0
 python-memcached==1.59
 celery==5.3.6
 django-celery-results==2.5.1
-mapbox-vector-tile==1.2.0
 SPARQLWrapper==1.8.5
 django-recaptcha==3.0.0
 edtf==4.0.1
--- a/tests/utils/datatype_tests.py
+++ b/tests/utils/datatype_tests.py
@@ -26,6 +26,36 @@ from tests.base_test import ArchesTestCase
 # python manage.py test tests/utils/datatype_tests.py --pattern="*.py" --settings="tests.test_settings"
 
 
+class BooleanDataTypeTests(ArchesTestCase):
+    def test_validate(self):
+        boolean = DataTypeFactory().get_instance("boolean")
+
+        for good in ["true", "false", "yes", "no", None]:
+            with self.subTest(input=good):
+                no_errors = boolean.validate(good)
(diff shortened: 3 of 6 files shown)


tests/utils/datatype_tests.py
```diff
@@ -26,6 +26,36 @@
...
+class BooleanDataTypeTests(ArchesTestCase):
+    def test_validate(self):
+        boolean = DataTypeFactory().get_instance("boolean")
+
+        for good in ["true", "false", "yes", "no", None]:
+            with self.subTest(input=good):
+                no_errors = boolean.validate(good)
+                self.assertEqual(len(no_errors), 0)
+
+        for bad in ["garbage", "True", "False", "None"]:
```
jacobtylerwalls, line 38: We reject upper cased True/False, so I made sure to surface that here.

arches/app/utils/string_utils.py
```diff
@@ -0,0 +1,7 @@
+def str_to_bool(value):
+    match value:
+        case "y" | "yes" | "t" | "true" | "on" | "1":
+            return True
+        case "n" | "no" | "f" | "false" | "off" | "0":
+            return False
+    raise ValueError
```
jacobtylerwalls, line 7: Reference: https://docs.python.org/3.11/distutils/apiref.html?highlight=strtobool#distutils.util.strtobool

arches/install/requirements.txt
```diff
@@ -7,7 +7,7 @@ django-guardian==2.4.0
 python-memcached==1.59
 celery==5.3.6
 django-celery-results==2.5.1
-mapbox-vector-tile==1.2.0
+mapbox-vector-tile==2.0.1
```
chiatt, line 10: I wonder if we need this package at all anymore.

jacobtylerwalls: Oh good call. Looks like TileStache was removed in arches 5. Maybe it was here for that?

chiatt: Yeah - that must be the case. Arches used to ship with a tile server using TileStache, but now we just let people implement their own and proxy tile requests through Arches.

chiatt (approved):

PR #10473, opened 2023-12-21
Handle empty node values in map report #10429

### LINKED ISSUE #10429: Empty string node not handled in Map Header Template
Currently on Arches 7.4.3 and found that whenever we are updating an existing Geospatial Coordinates on the resource, it just goes to a white screen of death.
image

I have also did an inspect on the web browser and this is what the errors are
image

To Reproduce:
1. Create a model with a node group with geojson node and string children nodes, this model should use the **map header report template**. 
2. Create a new resource for that model and create a geometry, but leave the string node blank
3. Leave the editor, then return to the edit this resource.
4. Make a change to the geometry, but maintain the string node as blank
5. ** If data entered into string node, UI is not blocked **

johnatawnclementawn:
### Types of changes

### Description of Change

Handles null values in the map header report template

### Issues Solved

Resolves #10429

### DIFF
--- a/arches/app/media/js/reports/map-header.js
+++ b/arches/app/media/js/reports/map-header.js
@@ -16,7 +16,7 @@ define([
             var features = [];
             ko.unwrap(params.tiles).forEach(function(tile) {
                 _.each(tile.data, function(val) {
-                    if ('features' in val) {
+                    if (val?.features) {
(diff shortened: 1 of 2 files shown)


jacobtylerwalls (approved):
Tested, fix works!

PR #10475, opened 2023-12-21
Stop map buttons from appearing over modals and help #10432

### LINKED ISSUE #10432: z-index issue: When Adding Geometry, The Edit toolbar is still visible when opening the Help Option
Hi Team,
Currently running 7.4.3 version but have noticed this on previous version too:

The EDIT toolbar seems to be visible when the HELP option has been toggled On.
image

jacobtylerwalls:
### Types of changes

### Description of Change
Prevent card help or graph publication modals from appearing behind the map workbench buttons.

### Issues Solved
Fixes #10432

#### Ticket Background
*   Sponsored by: 
*   Found by: @JPdelaRosa83
*   Tested by: @ 
*   Designed by: @ 

### Further comments
The git history shows this change (200 -> 1000) was made for RTL, so a quick check that all is good in RTL at 500 would be wise. (Looked OK to me.)

### DIFF
--- a/arches/app/media/css/arches.scss
+++ b/arches/app/media/css/arches.scss
@@ -1991,7 +1991,7 @@ div.final-cons-step-separator>h4 {
     width: 75px;
     background-color: #f1f1f1;
     border-left: 1px solid #ddd;
-    z-index: 1000;
+    z-index: 500;
 }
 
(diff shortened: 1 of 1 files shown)


johnatawnclementawn (approved):
Looks good!

PR #10482, opened 2023-12-29
fix issue with concept pagination for offset 0

### LINKED ISSUE #10466: Prevent duplicate options from appearing in concept dropdown
Duplicate entries can appear in the concept widget dropdown if the number of concepts exceeds 50.

More details on this forum post:
https://community.archesproject.org/t/duplicated-entry-in-dropdown/2188/5

This issue was identified in 7.4.3, but we should see if it exists in 6.2 as well.

aarongundel:
Fixes #10466

Issue is that the comparison to offset 0 results in a falsy value, which causes the pagination limit to not be applied to the query.  Comparing with "None" instead resolves the issue. - L511/512

Other changes are black styleguide based formatting mostly based on arches predefined line width.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,9 +28,9 @@ jobs:
       - name: Install Java, GDAL, and other system dependencies
         run: |
           sudo apt update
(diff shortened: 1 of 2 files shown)


chiatt (approved):
Looks good to me!

PR #10487, opened 2023-12-29
Use widget label in TileValidationError #10486

### LINKED ISSUE #9280: Required node error is unclear
It appears the required node error needs to be updated to support i18n

### LINKED ISSUE #10486: Use widget label in required node errors
Before arches 7.4, the error when required nodes were missing used to use the widget label if available. Now it always uses the node name.

jacobtylerwalls:
### Types of changes

### Description of Change
In the if block in this diff, the usual code path is to fetch from the Nodes table because `node` was a `SimpleNamespace` object from the serialized graph. Then, the cardxnodewidgets was never fetched, so we defaulted to the node name, even if a widget label was available.

Then, this uncovered #9280 again, so I fixed that also.

### Issues Solved

Closes #10486
Closes #9280

### Checklist

### DIFF
--- a/arches/app/models/tile.py
+++ b/arches/app/models/tile.py
@@ -308,7 +308,7 @@ class Tile(models.TileModel):
 
                     first_card_x_node_x_widget = node.cardxnodexwidget_set.first()
                     if first_card_x_node_x_widget:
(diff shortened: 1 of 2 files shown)


aarongundel (approved):
Looks good; works as specified.

PR #10504, opened 2024-01-04
update workflow component abstract to default to null value

aarongundel:
Fixes arches for science workflow issue where default step value of undefined is not equal to resource instance select empty value of null.


chiatt (approved):
Looks good!

PR #10506, opened 2024-01-04
fixes workflow steps not respecting 'locked' value

aarongundel:
Fixes issue where "locked" value would not be respected in workflow step, then would immediately reset the "locked" value in the database.

### DIFF
--- a/arches/app/media/js/viewmodels/workflow-step.js
+++ b/arches/app/media/js/viewmodels/workflow-step.js
@@ -98,10 +98,10 @@ define([
             }
 
             /* cached workflowComponentAbstract logic */ 
-            if (config.workflowHistory.stepData) {
-                const stepData = config.workflowHistory[ko.unwrap(self.name)];
-                self.componentIdLookup(stepData[COMPONENT_ID_LOOKUP_LABEL]);
-                self.locked(config.workflowHistory.stepdata.locked);
+            if (config.workflowHistory.stepdata) {
+                const stepData = config.workflowHistory.stepdata[ko.unwrap(self.name)];
+                self.componentIdLookup(stepData?.[COMPONENT_ID_LOOKUP_LABEL]);
+                self.locked(stepData?.locked);
             }
     
             /* cached informationBox logic */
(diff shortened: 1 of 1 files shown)


arches/app/media/js/viewmodels/workflow-step.js
```diff
@@ -98,10 +98,10 @@ define([
...
 
             /* cached workflowComponentAbstract logic */ 
-            if (config.workflowHistory.stepData) {
-                const stepData = config.workflowHistory[ko.unwrap(self.name)];
-                self.componentIdLookup(stepData[COMPONENT_ID_LOOKUP_LABEL]);
-                self.locked(config.workflowHistory.stepdata.locked);
+            if (config.workflowHistory.stepdata) {
+                const stepData = config.workflowHistory.stepdata[ko.unwrap(self.name)];
+                self.componentIdLookup(stepData?.[COMPONENT_ID_LOOKUP_LABEL]);
+                self.locked(stepData?.locked);
```
jacobtylerwalls, line 104: This works, but I kind of like the idea of removing both `?` so that if something is missing, we don't silently update `componentIdLookup` with garbage -- and instead fail more obviously. Or keep as is. Up to you.

arches/app/media/js/viewmodels/workflow-step.js
```diff
@@ -132,8 +132,8 @@ define([
             var workflowComponentAbstractLookup = self.workflowComponentAbstractLookup();
             var workflowComponentAbstractId = self.id();
 
-            if (config.workflowHistory.stepData) {
-                const componentIdLookup = config.workflowHistory.stepData[COMPONENT_ID_LOOKUP_LABEL];
+            if (config.workflowHistory.stepdata) {
+                const componentIdLookup = config.workflowHistory.stepdata?.[ko.unwrap(self.name)]?.[COMPONENT_ID_LOOKUP_LABEL];
```
jacobtylerwalls, line 136: Thanks, good example of something typescript would have caught.

jacobtylerwalls (approved):

arches/app/media/js/viewmodels/workflow-step.js
aarongundel: @jacobtylerwalls you're right on the approach here, but these were added because the conditional steps of xy upload will fail if this isn't added.  It'll look up select-dataset-files-step (which is part of the non-default workflow path) and then it will blow up, because that's not part of the default flow.  I'll keep as is for now but add a ticket to deal with that issue in the future.

aarongundel: https://github.com/archesproject/arches/issues/10527

PR #10510, opened 2024-01-08
adds frontend i18n endpoint and vue wrapper #10501

chrabyrd:
### Types of changes

### Issues Solved

#10501


apeters (changes requested):
`gettext.config.js` and `yarn.rc` files aren't getting interpolated properly during app creation.

apeters (changes requested):
it would be nice if a compiled translation file isn't required in the api call `GetFrontendI18NData`

arches/app/views/api.py
```diff
@@ -110,6 +107,37 @@ def dispatch(self, request, *args, **kwargs):
         return super(APIBase, self).dispatch(request, *args, **kwargs)
 
 
+class GetFrontendI18NData(APIBase):
+    def get(self, request):
+        origin_entity_name = request.GET.get('originEntityName')
+
+        user_language = get_language()
+        if user_language is None:
+            user_language = "en"
```
apeters, line 116: I think this can be omitted as `get_language()` will always return a valid language code

apeters (changes requested):

apeters (changes requested):
the following files can safely be removed from both the project and app templates:

- en.json
- en/app.po (as well as the folder)
- messages.pot

arches/app/media/js/utils/create-vue-application.js
```diff
@@ -0,0 +1,38 @@
...
+     * we'd like to use across the Arches ecosystem will be available. This also
+     * Vue apps more easily extensible if we choose to add plugins or logic in
+     * the future.
+     **/ 
+
+    /**
+     * TODO: cbyrd #10501 - we should add an event listener that will re-fetch i18n data
+     * and rebuild the app when a specific event is fired from the LanguageSwitcher component.
+    **/ 
+    return fetch(`/api/get_frontend_i18n_data?originEntityName=${originEntityName}`).then(resp => {
```
apeters, line 20: we should remove the leading slash from the fetch call so that the url isn't rooted at the base host.  This then allows for urls that include the language code in the url itself (which is one of the ways in which the get_language() method picks up the language code).

apeters (changes requested):

arches/app/views/api.py
chrabyrd: fixed in latest

chrabyrd:
> the following files can safely be removed from both the project and app templates:
> 
> * en.json
> * en/app.po (as well as the folder)
> * messages.pot

I believe `messages.pot` must stay, but the other 2 have been removed.

arches/app/media/js/utils/create-vue-application.js
chrabyrd: fixed in latest

arches/locale/en/app.po
apeters, line 1: I think we can safely remove this file as well as enclosing folder as well.

apeters (changes requested):

arches/locale/en/app.po
chrabyrd: fixed in latest

apeters:
> `gettext.config.js` and `yarn.rc` files aren't getting interpolated properly during app creation.

this is still an issue for app creation

chrabyrd:
> > `gettext.config.js` and `yarn.rc` files aren't getting interpolated properly during app creation.
> 
> this is still an issue for app creation

this should be fixed, as a sanity check would you mind re-installing arches core in your virtual env?

apeters (approved):
this works as expected

PR #10513, opened 2024-01-09
fix function load

aarongundel:
Various black fixes, fixes package function load - if it's an application, it does not copy the function to the project but it still registers the function.  

Expects that the application's function location is registered in settings.


chiatt (approved):
Looks good to me!

Issue #10476, opened 2023-12-22
[rdm/bulk-data-manager] run etl from the command line 


njkim:
It would be great if the admin user can run the etl module from command line.

PR #10519, opened 2024-01-11
Run single csv importer from command line, #10476

njkim:
### Types of changes

### Issues Solved

#

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/etl_modules/branch_excel_importer.py
```diff
@@ -22,14 +22,21 @@
 
 
 class BranchExcelImporter(BaseImportModule):
-    def __init__(self, request=None, loadid=None, temp_dir=None):
-        self.request = request if request else None
-        self.userid = request.user.id if request else None
+    def __init__(self, request=None, loadid=None, temp_dir=None, params=None):
+        self.loadid = request.POST.get("load_id") if request else loadid
+        self.userid = request.user.id if request else 1
```
chiatt, line 27: We can't count on there being a user 1. In fact, it's probably wise for admins to remove that user for security reasons.

arches/app/etl_modules/base_import_module.py
```diff
@@ -287,10 +286,86 @@ def write(self, request):
...
+            return {"success": False, "data": {"title": _("Error"), "message": written["data"]["message"]}}
+
+    def read_for_cli(self, source):
+        self.cumulative_excel_files_size = 0
+        self.temp_dir = os.path.join(settings.UPLOADED_FILES_DIR, "tmp", self.loadid)
+        try:
+            self.delete_from_default_storage(self.temp_dir)
+        except (FileNotFoundError):
+            pass
+        file_name = os.path.basename(source).split('/')[-1]
```
chiatt, line 320: `os.path.basename` returns the name of a file, so this seems like it should just be 
```
file_name = os.path.basename(source)
```
I could be missing something, but I don't think the `.split('/')[-1]` is doing anything because there shouldn't be any slashes in the file's basename.

arches/app/etl_modules/save.py
```diff
@@ -19,8 +19,8 @@ def save_to_tiles(userid, loadid, multiprocessing=False):
         if error_saving_tiles:
             return error_saving_tiles
 
-        _post_save_edit_log(cursor, userid, loadid, multiprocessing)
-
+        response = _post_save_edit_log(cursor, userid, loadid, multiprocessing)
+    return response
```
chiatt, line 23: Maybe better just to return `_post_save_edit_log(cursor, userid, loadid, multiprocessing)` rather than to assign the result to a variable and then return it.

arches/app/etl_modules/base_import_module.py
```diff
@@ -367,27 +370,147 @@ def start(self, request):
...
+                    self.validate_uploaded_file(workbook)
+                    has_valid_excel_file = True
+                except:
+                    pass
+        if not has_valid_excel_file:
+            title = _("Invalid Uploaded File")
+            message = _(
+                "This file has missing information or invalid formatting. Make sure the file is complete and in the expected format."
+            )
+            return {"success": False, "data": {"title": title, "message": message}}
```
chiatt, line 510: If a user were to upload multiple excel files, one valid and one invalid, this would pass without an error. It seems like we should at least provide a message letting the user know that one of their files is invalid if not fail the load entirely.

arches/app/etl_modules/base_import_module.py
```diff
@@ -367,27 +370,147 @@ def start(self, request):
...
+                "files": {},
+            }
+        }
+        validator = FileValidator()
+        if len(validator.validate_file_type(source)) > 0:
+            return {
+                "status": 400,
+                "success": False,
+                "title": _("Invalid excel file/zip specified"),
+                "message": _("Upload a valid excel file"),
```
chiatt, line 454: Could this message be "Upload a valid .xlsx or .zip file"? I think we don't accept .xls files, so it might be better to explicitly state `.xlsx` rather than 'excel'.

arches/app/etl_modules/base_import_module.py
```diff
@@ -367,27 +370,147 @@ def start(self, request):
...
+
+        if written["success"]:
+            return {"success": True, "data": "done"}
+        else:
+            return {
+                "success": False,
+                "data": {"title": _("Error"), "message": written["data"]["message"]},
+            }
+
+    def read_for_cli(self, source):
```
chiatt, line 431: This method is pretty long and looks like it shares some code with the `read` method. I'm wondering if some of it could be pulled out and shared between methods.

arches/app/etl_modules/import_single_csv.py
```diff
@@ -141,12 +228,20 @@ def write(self, request):
...
         Move the records from load_staging to tiles table using db function
         """
-
+        # # if self.params:
+        # #     graphid = self.params["graphid"]
+        # #     has_headers = bool(self.params["has_headers"])
+        # #     fieldnames = self.params["fieldnames"]
+        # #     csv_mapping = self.params["csv_mapping"]
+        # #     csv_file_name = self.params["csv_file_name"]
+        # else:
```
chiatt, line 237: A little cruft

chiatt (changes requested):

arches/app/etl_modules/save.py
```diff
@@ -11,16 +11,16 @@
...
         disable_tile_triggers(cursor, loadid)
         error_saving_tiles = _save_to_tiles(cursor, loadid)
         reenable_tile_triggers(cursor, loadid)
         if error_saving_tiles:
             return error_saving_tiles
 
-        _post_save_edit_log(cursor, userid, loadid)
-
+        response = _post_save_edit_log(cursor, userid, loadid, multiprocessing)
+    return response
```
jacobtylerwalls, line 23: Thanks @njkim for catching this -- can you also use this opportunity to fix JSONLDImporter.save_to_tiles to also return the result of _post_save_edit_log()?

arches/app/etl_modules/import_single_csv.py
```diff
@@ -71,6 +80,84 @@ def get_node_lookup(self, graphid):
...
+        try:
+            self.delete_from_default_storage(temp_dir)
+        except FileNotFoundError:
+            pass
+
+        csv_file_name = None
+        if source.split(".")[-1].lower() == "csv":
+            csv_file_name = os.path.basename(source).split("/")[-1]
+            csv_file_path = os.path.join(temp_dir, csv_file_name)
+            default_storage.save(csv_file_path, File(open(source, "r")))
```
jacobtylerwalls, line 126: Can we use a `with` statement to avoid leaving this file handle open?

arches/app/etl_modules/import_single_csv.py
```diff
@@ -71,6 +80,84 @@ def get_node_lookup(self, graphid):
...
+        else:
+            return {
+                "success": False,
+                "data": {"title": _("Error"), "message": written["message"]},
+            }
+
+    def read_for_cli(self, source):
+        """
+        Reads added csv file and returns all the rows
+        If the loadid already exsists also returns the load_details
```
jacobtylerwalls, line 113: spelling

arches/app/etl_modules/base_import_module.py
```diff
@@ -367,27 +370,147 @@ def start(self, request):
...
+                            result["summary"]["files"][file.filename] = {
+                                "size": (self.filesize_format(file.file_size))
+                            }
+                            result["summary"][
+                                "cumulative_files_size"
+                            ] = self.cumulative_files_size
+                        default_storage.save(
+                            os.path.join(self.temp_dir, file.filename),
+                            File(zip_ref.open(file)),
+                        )
```
jacobtylerwalls, line 474: Looks like this executes even if the file.is_dir() returns true. Is that wanted? Maybe the two if's for MACOSX and is_dir() can be combined?

jacobtylerwalls: Also, we can make sure to use `with` to open the file so we don't leave it open?

arches/app/etl_modules/base_import_module.py
```diff
@@ -367,27 +370,147 @@ def start(self, request):
...
+        if len(validator.validate_file_type(source)) > 0:
+            return {
+                "status": 400,
+                "success": False,
+                "title": _("Invalid excel file/zip specified"),
+                "message": _("Upload a valid excel file"),
+            }
+
+        if source.split(".")[-1].lower() == "zip":
+            with zipfile.ZipFile(source, "r") as zip_ref:
```
jacobtylerwalls, line 458: Around JSONLDImporter.read_zip_file() I'm catching `BadZipFile`, do you think we should do that here also?

jacobtylerwalls (commented):
Hey @njkim thanks for catching the regression I introduced with the missing return value in _post_save_edit_log(). I was curious for impact on the JSONLDImporter so I gave this a quick look and have a couple suggestions.

arches/app/etl_modules/branch_excel_importer.py
```diff
@@ -24,15 +24,27 @@
...
 
 class BranchExcelImporter(BaseImportModule):
-    def __init__(self, request=None, loadid=None, temp_dir=None):
-        self.request = request if request else None
-        self.userid = request.user.id if request else None
+    def __init__(self, request=None, loadid=None, temp_dir=None, params=None):
+        self.loadid = request.POST.get("load_id") if request else loadid
+        self.userid = (
+            request.user.id if request else User.objects.filter(is_superuser=True)[0].id
+        )
```
chiatt, line 31: I think instead of finding the first superuser, it would be better to require that a username be passed into the command - maybe like psql (-U, --user).

chiatt (changes requested):

arches/app/etl_modules/branch_excel_importer.py
chiatt: For the time being you could also just use ```settings.DEFAULT_RESOURCE_IMPORT_USER["userid"]```, but eventually it would be nice to optionally pass a user to the command.

arches/app/etl_modules/import_single_csv.py
```diff
@@ -72,13 +96,57 @@ def get_node_lookup(self, graphid):
...
+                "data": {"title": _("Error"), "message": read["message"]},
+            }
+
+        if written["success"]:
+            return {"success": True, "data": "done"}
+        else:
+            return {
+                "success": False,
+                "data": {"title": _("Error"), "message": written["message"]},
+            }
```
chiatt, line 124: `read` and `written` shouldn't be defined in `if` statements because this could fail with undefined errors.

arches/app/etl_modules/import_single_csv.py
```diff
@@ -72,13 +96,57 @@ def get_node_lookup(self, graphid):
...
 
-        content = request.FILES.get("file")
+        if request:
+            content = request.FILES.get("file")
+        else:
+            if source.split(".")[-1].lower() == "csv":
+                file_type = "text/csv"
+            elif source.split(".")[-1].lower() == "zip":
+                file_type = "application/zip"
+            file = open(source, "rb")
```
chiatt, line 139: Seems like this should be in a context statement: ```with open(source, "rb") as file:```

arches/app/etl_modules/branch_excel_importer.py
```diff
@@ -24,15 +24,39 @@
 
 
 class BranchExcelImporter(BaseImportModule):
-    def __init__(self, request=None, loadid=None, temp_dir=None):
-        self.request = request if request else None
-        self.userid = request.user.id if request else None
+    def __init__(self, request=None, loadid=None, temp_dir=None, params=None):
```
chiatt, line 27: Certainly not necessary for this PR, but we should probably stop passing request objects to the importer, but rather the data needed from the request object.

arches/app/etl_modules/import_single_csv.py
njkim: The variables are now defined beforehand. The unexpected errors are handled, too.

arches/app/etl_modules/branch_excel_importer.py
njkim: Handling the request separately from the module so that the module can be used without the request... something like that? Yeah, I was thinking the same thing.

arches/app/etl_modules/import_single_csv.py
njkim: It turns out that the file needs to open a little bit longer so the opened is manually closed when it is done, now.

arches/app/etl_modules/base_import_module.py
```diff
@@ -367,27 +389,76 @@ def start(self, request):
...
+
+        if read["success"]:
+            try:
+                self.request.POST.__setitem__(
+                    "load_details", json.dumps({"result": read["data"]})
+                )
+                written = self.write(self.request)
+            except Exception as e:
+                return return_with_error(
+                    _("Unexpected Error during saving the tiels: {}").format(e)
```
chiatt, line 452: Looks like there is a typo in the word "files". Because there is a bit more going on with this step than just saving, I'd change this to `Unexpected error while processing files:`.

arches/app/etl_modules/base_import_module.py
```diff
@@ -367,27 +389,76 @@ def start(self, request):
...
+        written = {"success": False, "message": ""}
+
+        initiated = self.start(self.request)
+
+        if initiated["success"]:
+            try:
+                read = self.read(source=source)
+            except Exception as e:
+                return return_with_error(
+                    _("Unexpected Error during reading the files: {}").format(e)
```
chiatt, line 439: Because there could be only one file, could this be changed to `Unexpected error while reading file(s):`

chiatt (approved):

PR #10523, opened 2024-01-12
Fix display of map coord editor #9145

chiatt:
### Types of changes

### Description of Change
Fix display of map coord editor #9145

### Issues Solved

#9145

To test this you'll need to verify that scrolling works properly in both the map widget and the map card.


jacobtylerwalls (approved):
Looks good! There's still a little extra vertical scroll in the map widget (workbench-card-container), but that looks like it predates this PR.

Issue #10538, opened 2024-01-18
Workflow History doesn't reload every tile


teenorth:
I'm working on a project and implementing a system that will allow workflows to be reopened. I had a working system on an older version of Arches when it utilised local storage.

I noticed the newly implemented "Workflow History" which stores the workflows in the database! This will make creating the feature far easier.

There was one problem I noticed with the "Workflow History" code. When you refresh in a workflow that has had data saved to the workflow history table. It will only save one tile per step.

## Broken example

Problems:
- Only two tiles exist in component data, there should be 4
- The `activity-details` step's component lookup ids are all identical
- The id being repeated in `activity-details` is the step id

The lookup ids are used to insert and get the data for the card from `componentdata`. With them all being identical every time a card attempts to save it's data to the workflow history it will overwrite the previous component data that was there.

GET http://localhost:8000/workflow_history/1077f7b7-74ae-4e9a-a820-55494d0dc83e

![broken-workflow-history](https://github.com/archesproject/arches/assets/62624804/206211ee-7af4-45fa-8c95-3bd7813bd694)

```json
{
    "completed": false,
    "componentdata": {
        "4b941615-12ba-49e4-ab3d-dc3d37c39276": {
            "value": [
                {
                    "nodegroupId": "4f5ec415-993e-11ea-bab0-f875a44e0e11",
                    "resourceInstanceId": "0d83a9e2-0ac0-4e5b-8289-d40f4d59b539",
                    "tileData": "{\"4f5eeb26-993e-11ea-b156-f875a44e0e11\":\"59a0c936-3f60-44f9-9e37-67b6ed0f4fa4\",\"4f5eeb27-993e-11ea-b9f7-f875a44e0e11\":{\"en\":{\"direction\":\"ltr\",\"value\":\"18-10-2024\"}},\"4f5eeb28-993e-11ea-b5c1-f875a44e0e11\":\"2024-10-18\",\"4f5eeb29-993e-11ea-8078-f875a44e0e11\":\"2024-10-18\",\"4f5eeb2a-993e-11ea-a2db-f875a44e0e11\":\"586a2776-7f4a-4be9-9219-3a9896c7f965\"}",
                    "tileId": "cf76479f-ec36-45c6-a2be-b570353b7038"
                }
            ]
        },
        "9726fa1e-8c07-4c66-8f3d-256fc845dc6d": {
            "value": [
                {
                    "nodegroupId": "e7d695ff-9939-11ea-8fff-f875a44e0e11",
[...]

PR #10539, opened 2024-01-18
Workflow History doesn't reload every tile #10538

teenorth:
### Types of changes

### Description of Change

Description included in issue.

### Issues Solved

#10538 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @teenorth
*   Tested by: @teenorth
*   Designed by: @

### DIFF
--- a/arches/app/media/js/viewmodels/workflow-step.js
+++ b/arches/app/media/js/viewmodels/workflow-step.js
@@ -214,7 +214,7 @@ define([
                     [ko.unwrap(self.name)]: {
                         [key]: value,
                         locked: self.locked(),
-                        stepId: self.id()
+                        stepId: self.id(),
                     },
                 },
             };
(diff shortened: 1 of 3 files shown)


jacobtylerwalls (changes requested):
Thanks for the fix and helpful diagnosis!

I missed these because I happened to be testing with workflows that had at most one card per step. Great catch.

Thanks for finding a solution that doesn't present a backwards-compatibility problem. I tested with a workflow created on 7.5.0 and it restores on this branch just fine. (Tested with Arches for Science.)

This is a bug in a new feature, so I think we should target 7.5. Could you rebase on dev/7.5.x? There's likely to be a 7.5.1 release on Thursday, so I would very much like to get this in before then. Thanks.

jacobtylerwalls (approved):
Thanks for the updates!

teenorth:
Sure! I've doing quite a lot of work around the workflows lately and I'm currently setting up a reload/edit/open previous workflow for a resource. I had a working version on an older version of Arches that used the local storage but it required a complex setup function. Having the workflow data stored in the database is going to be a massive time save and smoother implementation!

Issue #10566, opened 2024-01-29
Prepend slash to i18n fetch call in `create-vue-application.js`


chrabyrd:

chrabyrd:
resolved in #10567

PR #10567, opened 2024-01-29
updates i18n call in create-vue-application component #10566

chrabyrd:
### Types of changes

### Issues Solved

#10566


apeters (approved):
this seems to be working now, so I'll approve it, we should keep an eye on this though

Issue #10552, opened 2024-01-22
Add reference datatype


jacobtylerwalls:
Analogous to the concept-list datatype, create a datatype that resolves against the `ListItem`s managed in #10544.

Consider whether to have it "look like JSON" or just use the pk of the `ListItem`.

PR #10572, opened 2024-02-01
Add reference datatype

chiatt:
### Types of changes

### Description of Change
Stubs out the reference data type with a basic widget and node config.

### Issues Solved

#10552

### For reviewer:
1. After running migrations you can test using lists in the test data: ```PYTHONPATH="../arches" python3 manage.py shell -c "from tests.views.controlled_lists_tests import ControlledListTests; ControlledListTests.setUpTestData()"```

You could also use the following which has some pref labels in German for localization testing :
```
insert into controlled_lists (id, name, dynamic) values ('06b58ca6-bcb6-46b3-8aaa-64cca94ab8f8', 'menu', FALSE); 
insert into controlled_list_items values ('1ea5f71d-7d56-42d4-8d71-c806b0766d29', 'https://www.subec.com/lunch', 0, '06b58ca6-bcb6-46b3-8aaa-64cca94ab8f8', NULL);
insert into controlled_list_items values ('fd9508dc-2aab-4c46-85ae-dccce1200035', 'https://www.subec.com/dinner', 2, '06b58ca6-bcb6-46b3-8aaa-64cca94ab8f8', NULL);
insert into controlled_list_item_labels values ('26c6a505-3d20-4fa3-9a42-44678a754164', 'lunch', '1ea5f71d-7d56-42d4-8d71-c806b0766d29', 'en', 'prefLabel');
insert into controlled_list_item_labels values ('ca1a40cf-8642-48cb-8fdb-c4cdab0b711d', 'Mittagessen', '1ea5f71d-7d56-42d4-8d71-c806b0766d29', 'de', 'prefLabel');
insert into controlled_list_item_labels values ('23b4efbd-2e46-4b3f-8d75-2f3b2bb96af2', 'dinner', 'fd9508dc-2aab-4c46-85ae-dccce1200035', 'en', 'prefLabel');
insert into controlled_list_item_labels values ('5ce4841d-e543-42f2-b47b-1fbca21f6a04', 'supper', 'fd9508dc-2aab-4c46-85ae-dccce1200035', 'en', 'altLabel');
insert into controlled_list_item_labels values ('e8676242-f0c7-4e3d-b031-fded4960cd86', 'Abendessen', 'fd9508dc-2aab-4c46-85ae-dccce1200035', 'de', 'prefLabel');

insert into controlled_list_items values ('228b53af-80ba-4bb2-ad46-1d873c2ff4ec', 'https://www.subec.com/sandwich', 1, '06b58ca6-bcb6-46b3-8aaa-64cca94ab8f8', '1ea5f71d-7d56-42d4-8d71-c806b0766d29');
insert into controlled_list_items values ('a050ddbd-a1d9-42a6-a0b5-101683b0a0c9', 'https://www.subec.com/salmon', 3, '06b58ca6-bcb6-46b3-8aaa-64cca94ab8f8', 'fd9508dc-2aab-4c46-85ae-dccce1200035');

insert into controlled_list_item_labels values ('2dbe4f9d-b16d-4ba7-8fbb-7b56dc9f334d', 'sandwich', '228b53af-80ba-4bb2-ad46-1d873c2ff4ec', 'en', 'prefLabel');
insert into controlled_list_item_labels values ('27524045-e76c-4698-9cb6-8e567b25a586', 'salmon', 'a050ddbd-a1d9-42a6-a0b5-101683b0a0c9', 'en', 'prefLabel');
insert into controlled_list_item_labels values ('3583c278-b953-40c4-8c09-8ad62c0a93f5', 'lox', 'a050ddbd-a1d9-42a6-a0b5-101683b0a0c9', 'en', 'altLabel');
insert into controlled_list_item_labels values ('c18e6183-3d95-4c1c-909e-a13b946ae2d9', 'Sandwich', '228b53af-80ba-4bb2-ad46-1d873c2ff4ec', 'de', 'prefLabel');
insert into controlled_list_item_labels values ('8ce82adc-ce0f-4968-8e67-acd1d942cf23', 'Lachs', 'a050ddbd-a1d9-42a6-a0b5-101683b0a0c9', 'de', 'prefLabel');
```
2.  Create a node with using the `reference` datatype
3.  Leave `multiple values` off. (The multiple values parameter controls whether the node will accept only one list item value or many. If `false`, the widget will display a concept dropdown for a single item. If `true` the widget will support multiple selections.
4. Create a tile ensuring that the concept widget only supports selection of one value.
5. Go back to the graph designer and switch `multiple values` to `true`. Refresh your card in the resource editor. The widget should now support multiple values.
6. Data saved to the tile should have this structure:
```[
    {
      "uri": "https://www.subec.com/dinner",
      "labels": [
        {
          "id": "23b4efbd-2e46-4b3f-8d75-2f3b2bb96af2",
          "value": "dinner",
          "language": "en",
          "valuetype": "prefLabel"
        },
        {
          "id": "5ce4841d-e543-42f2-b47b-1fbca21f6a04",
          "value": "supper",
          "language": "en",
          "valuetype": "altLabel"
        },
        {
          "id": "e8676242-f0c7-4e3d-b031-fded4960cd86",
          "value": "Abendessen",
          "language": "de",
          "valuetype": "prefLabel"
        }
      ],
      "listid": "fd9508dc-2aab-4c46-85ae-dccce1200035"
    }
  ]
```
7. Change your system language to `de` and ensure the widget localizes the options and selected values.


arches/app/media/js/viewmodels/reference-select.js
```diff
@@ -0,0 +1,158 @@
...
+        this.selectionValue = ko.observable([]); // formatted version of this.value that select2 can use
+        this.activeLanguage = arches.activeLanguage;
+
+        WidgetViewModel.apply(this, [params]);
+
+        this.getPrefLabel = function(labels){
+            return koMapping.toJS(labels)?.find(label => label.language===arches.activeLanguage && label.valuetype === 'prefLabel').value;
+        }; 
+
+        this.flattenItems = function(items, flatList, uniqueItems){
```
jacobtylerwalls, line 24: Trying to remember what we said about this. Were you planning to just use the flat representation from the endpoint instead of flattening here?

arches/app/media/js/viewmodels/reference-select.js
```diff
@@ -0,0 +1,158 @@
...
+        this.multiple = !!ko.unwrap(params.node.config.multiValue);
+        this.displayName = ko.observable('');
+        this.selectionValue = ko.observable([]); // formatted version of this.value that select2 can use
+        this.activeLanguage = arches.activeLanguage;
+
+        WidgetViewModel.apply(this, [params]);
+
+        this.getPrefLabel = function(labels){
+            return koMapping.toJS(labels)?.find(label => label.language===arches.activeLanguage && label.valuetype === 'prefLabel').value;
+        };
```
jacobtylerwalls, line 22: Need to handle a lack of label in the active language

arches/app/media/js/viewmodels/reference-select.js
```diff
@@ -0,0 +1,158 @@
...
+
+        this.select2Config = {
+            value: self.selectionValue,
+            clickBubble: true,
+            multiple: this.multiple,
+            closeOnSelect: true,
+            placeholder: self.placeholder,
+            allowClear: true,
+            ajax: {
+                url: arches.urls.controlled_lists,
```
jacobtylerwalls, line 70: Now's a good time to rebase on the latest and use the GET for a single list, see 0b43664954817dd6c42a4909373bc22247717a6d

arches/app/templates/views/components/widgets/reference-select.htm
```diff
@@ -0,0 +1,55 @@
+{% extends "views/components/widgets/base.htm" %}
+{% load i18n %}
+
+{% block form %}
+<div class="row widget-wrapper" data-bind="class: nodeCssClasses">
+    <div class="form-group">
+        <span class="control-label widget-input-label" data-bind="text:label, class: nodeCssClasses"></span>
+        <!-- ko if: node -->
+        <i data-bind="css: {'ion-asterisk widget-label-required': node.isrequired}"></i>
```
jacobtylerwalls, line 9: I know they're all like this, but let's use this opportunity to get an aria-label on the icon.

jacobtylerwalls (commented):
Thanks, Cyrus! I left some initial comments, and I'll look forward to giving this a functional test next week.

What do you think about taking the opportunity to write one sanity-check unit test in `datatype_tests.py`?

arches/app/media/js/viewmodels/reference-select.js
chiatt: Good question - I just left this as it was, but I could switch to the flat list as long as we are confident that the list is ordered with children always following their parents.

arches/app/media/js/viewmodels/reference-select.js
chiatt: Ah, right. I did mean to make that change.

arches/app/media/js/viewmodels/reference-select.js
jacobtylerwalls: Yeah, we should be alright. I added a [comment](https://github.com/archesproject/arches/commit/7b655e0ab9f7319af2b7a8de9a9022e4d38ed393) next to the validation on sortorder -- sortorder concerns the flattened list as a whole. And the endpoint sorts on it.

arches/app/media/js/viewmodels/reference-select.js
```diff
@@ -0,0 +1,158 @@
...
+                self.value(tileReady);
+            }
+        });
+
+        this.select2Config = {
+            value: self.selectionValue,
+            clickBubble: true,
+            multiple: this.multiple,
+            closeOnSelect: true,
+            placeholder: self.placeholder,
```
jacobtylerwalls, line 67: Maybe this is a future card/not important, but we don't have any validation on invalid placeholders.

arches/app/media/js/viewmodels/reference-select.js
```diff
@@ -0,0 +1,158 @@
...
+                    callback(valueData);
+                };
+
+                if (self.value()?.length) {
+                    setSelectionData();
+                } else {
+                    callback([]);
+                }
+
+            }
```
jacobtylerwalls, line 152: In this scenario:
- node allows multiple references
- tile is saved with 2 references
- click to "X" the 2nd entry
- click "Cancel Edit"


The original entry is not restored.

arches/app/templates/views/components/datatypes/reference.htm
```diff
@@ -0,0 +1,41 @@
...
+    <div class="col-xs-12 control-label widget-input-label">
+        <span data-bind="text: $root.translations.controlledList"></span>
+    </div>
+    <!-- ko if: $data.controlledLists() -->
+        <select 
+            data-bind="
+                options: $data.controlledLists, 
+                optionsText: 'name', 
+                optionsValue: 'id', 
+                select2Query: {
```
jacobtylerwalls, line 15: Would be good to have some validation against saving the node in the graph editor without a controlled list, which causes frontend TypeErrors later in the widgets.

arches/app/media/js/viewmodels/reference-select.js
```diff
@@ -0,0 +1,158 @@
...
+        });
+
+        this.select2Config = {
+            value: self.selectionValue,
+            clickBubble: true,
+            multiple: this.multiple,
+            closeOnSelect: true,
+            placeholder: self.placeholder,
+            allowClear: true,
```
jacobtylerwalls, line 68: When clearing an entry saved to a tile (in single select mode), if the node has no placeholder configured, I cannot:

```js
selectWoo.full.js:1928 Uncaught TypeError: Cannot read properties of undefined (reading 'id')
    at AllowClear._handleClear (selectWoo.full.js:1928:40)
```

arches/app/templates/views/components/widgets/reference-select.htm
```diff
@@ -0,0 +1,55 @@
...
+    <dd data-bind="class: $parent.nodeCssClasses">
+        <a data-bind="attr:{'href': uri}, text: $parent.getPrefLabel(labels), style: {'color': 'steelblue'}"></a>
+    </dd>
+<!-- /ko -->
+<!-- ko ifnot: value() -->
+<dd data-bind="class: nodeCssClasses">
+    <span data-bind="text: $root.translations.none"></span>
+</dd>
+<!-- /ko -->
+{% endblock report %}
```
jacobtylerwalls, line 55: What is the expected behavior when deleting the tile data without deleting the resources? My understanding is that the tile remains, and we just log out this error to arches.log:

```py
2024-02-05 22:32:43,918 arches.app.datatypes.base ERROR    Tile has no authoritative or provisional data
NoneType: None
```

Is that right?

arches/app/templates/views/components/widgets/reference-select.htm
```diff
@@ -0,0 +1,55 @@
...
+        >
+    </div>
+</div>
+{% endblock config_form %}
+
+{% block report %}
+<dt data-bind="text: label, class: nodeCssClasses"></dt>
+<!-- ko foreach: value -->
+    <dd data-bind="class: $parent.nodeCssClasses">
+        <a data-bind="attr:{'href': uri}, text: $parent.getPrefLabel(labels), style: {'color': 'steelblue'}"></a>
```
jacobtylerwalls, line 47: Directing this to a new tab would be nice.

jacobtylerwalls (changes requested):

arches/app/media/js/viewmodels/reference-select.js
```diff
@@ -0,0 +1,158 @@
...
+            items.forEach(item=>{
+                if (uniqueItems.includes(item.uri) === false) {
+                    uniqueItems.push(item.uri);
+                    flatList.push(item);
+                    if (item.children.length) {
+                        self.flattenItems(item.children, flatList, uniqueItems);
+                    }
+                }
+            });
+            flatList.forEach(subItem=>delete subItem.children); //Prevent select2 from symbolizing parent as grouping reference
```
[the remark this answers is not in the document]
jacobtylerwalls: Feeling like that could be an important performance consideration.

arches/app/models/migrations/10552_add_reference_datatype.py
```diff
@@ -0,0 +1,57 @@
...
+                'views/components/widgets/reference-select',
+                'reference',
+                '{"i18n_properties": ["placeholder"]}'
+            );
+
+            """,
+            reverse_sql="""
+            DELETE FROM widgets WHERE widgetid = '19e56148-82b8-47eb-b66e-f6243639a1a8';
+            DELETE FROM d_data_types WHERE datatype = 'reference';
+            """,
```
jacobtylerwalls, line 55: Just noticed this failure trying to reverse migrate:

```py
django.db.utils.IntegrityError: update or delete on table "d_data_types" violates foreign key constraint "nodes_ddatatypes_fk" on table "nodes"
DETAIL:  Key (datatype)=(reference) is still referenced from table "nodes".
```

arches/app/media/js/viewmodels/reference-select.js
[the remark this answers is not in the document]
chiatt: Yeah - I think that would be a nice change

arches/app/models/migrations/10552_add_reference_datatype.py
chiatt: This is pretty tricky. I pushed a fix that will allow reverse migration. I tried to clean out all the objects that might have used a node with the reference datatype, but in some cases it leaves some reference datatype cruft behind that I still have to track down.

arches/app/templates/views/components/widgets/reference-select.htm
chiatt: Do you mean deleting the a tile's data without deleting the tile, or all of a resources tiles without deleting the resource?

arches/app/media/js/viewmodels/reference-select.js
chiatt: Adding a default placeholder to the default config seemed to fix this, but you would have to re-run the migration to get the change, or you could run 
```update widgets set defaultconfig = '{"placeholder": "Select an option","i18n_properties": ["placeholder"]}' where widgetid = '19e56148-82b8-47eb-b66e-f6243639a1a8'```

File: arches/app/templates/views/components/widgets/reference-select.htm (hunk 9 above)
Comment (jacobtylerwalls) [reply arches/app/templates/views/components/widgets/reference-select.htm:58]: Sorry -- deleting tile data. Basically using the widget to clear a saved value, then clicking save edit.

File: tests/utils/datatype_tests.py [hunk 13]
```diff
@@ -104,3 +104,84 @@ def test_tile_clean(self):
...
+
+class ReferenceDataTypeTests(ArchesTestCase):
+    def test_validate(self):
+        reference = DataTypeFactory().get_instance("reference")
+
+        for value in [
+                "", 
+                [], 
+                [{}], # reference has no 'uri'
+                [{"uri":""}], # reference uri is empty
```
jacobtylerwalls, line 117: The mockups I'm working from suggest that a URI is optional. We might need more direction here.

tests/utils/datatype_tests.py
```diff
@@ -104,3 +104,84 @@ def test_tile_clean(self):
...
+            ]:
+            errors = reference.validate(value="")
+            self.assertIsNotNone(errors)
```
jacobtylerwalls, line 120: So that all four tests run and tell you which ones specifically failed:
```suggestion
            with self.subTest(whatever_label_you_want_in_the_output=value):
                errors = reference.validate(value="")
                self.assertIsNotNone(errors)
```

arches/app/media/js/viewmodels/reference-select.js
jacobtylerwalls: With the config change I'm still seeing:
```
Unable to save. Please verify your input is valid
TileValidationError: Reference value must be a list of reference objects
```
when removing a value for a node without a placeholder, or upon saving the node with the placeholder

tests/utils/datatype_tests.py
robgaston: URIs are required for references in business data (ie the URI is the thing being referenced), but should be optional for creating list items.  when they are not provided in the creation of list items we should generate a URI, but what that entails exactly still needs to be defined.

arches/app/media/js/viewmodels/reference-select.js
jacobtylerwalls: I'm still getting "undefined", but for consistency with the controlled list manager we could use "Unlabeled Item".

arches/app/media/js/viewmodels/reference-select.js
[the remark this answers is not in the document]
jacobtylerwalls: See [a6c923b](https://github.com/archesproject/arches/pull/10572/commits/a6c923b2778d012e3a75db8d25fe53574b181b6e)

arches/app/templates/views/components/widgets/reference-select.htm
chiatt: Yeah - that's right then. The tile persists even if you used the widget to clear out all of the data on all of the nodes of the tile.

jacobtylerwalls: Thanks, opened #10607 for it.

arches/app/templates/views/components/datatypes/reference.htm
chiatt: I added a method to the datatype to validate the config which gets called in graph.validate. This seems like something that we should be doing for other datatypes as well.

jacobtylerwalls (commented):
I found that if the node is required, in multi-select mode, and I configure a placeholder, the placeholder appears in the widget, but "Save edit" fails with:

TileValidationError: This card requires values for the following: New Node

This may have something to do with the fact that the placeholder doesn't appear to look like it should, e.g the string "lunch" instead of the badge "X lunch" etc.

jacobtylerwalls:
We could show the label instead of the JSON for uniqueness errors:

Screenshot 2024-02-22 at 12 42 16 PM

arches/app/templates/views/components/widgets/reference-select.htm
```diff
@@ -0,0 +1,58 @@
...
+        >
+    </div>
+</div>
+{% endblock config_form %}
+
+{% block report %}
+<dt data-bind="text: label, class: nodeCssClasses"></dt>
+<!-- ko foreach: value -->
+    <dd data-bind="class: $parent.nodeCssClasses">
+        <a data-bind="attr:{'href': uri, 'target': '_blank'}, text: $parent.getPrefLabel(labels), style: {'color': 'steelblue'}"></a>
```
jacobtylerwalls, line 50: This link doesn't seem to work when the report is viewed from Search Results > Details. Have you seen this before?

arches/app/media/js/viewmodels/reference-select.js
```diff
@@ -45,17 +30,37 @@ define([
...
-                        "labels": NAME_LOOKUP[uri].labels
-                    };
-                });
-                self.value(tileReady);
+        this.valueAndSelectionDiffer = function(value, selection) {
+            if (!(ko.unwrap(value) instanceof Array)) {
+                return true;
+            }
+            const valueUris = ko.unwrap(value).map(val=>ko.unwrap(val.uri));
+            return (JSON.stringify(selection) != JSON.stringify(valueUris))
```
jacobtylerwalls, line 38:
```suggestion
            return (JSON.stringify(selection) !== JSON.stringify(valueUris))
```

arches/app/media/js/viewmodels/reference-select.js
```diff
@@ -45,17 +30,37 @@ define([
...
-                const tileReady = val.map(uri => {
-                    return {
-                        "uri": uri,
-                        "listid": NAME_LOOKUP[uri]["listid"],
-                        "labels": NAME_LOOKUP[uri].labels
-                    };
-                });
-                self.value(tileReady);
+        this.valueAndSelectionDiffer = function(value, selection) {
+            if (!(ko.unwrap(value) instanceof Array)) {
```
jacobtylerwalls, line 34: I found that `selection` also needed to be unwrapped in the case of re-selecting the same value.

arches/app/models/migrations/10552_add_reference_datatype.py
```diff
@@ -0,0 +1,64 @@
...
+            );
+
+            """,
+            reverse_sql="""
+            DELETE FROM cards_x_nodes_x_widgets
+                WHERE nodeid IN (SELECT nodeid FROM nodes WHERE datatype = 'reference'); 
+            DELETE FROM edges
+                WHERE domainnodeid IN (SELECT nodeid FROM nodes WHERE datatype = 'reference') 
+                OR rangenodeid IN (SELECT nodeid FROM nodes WHERE datatype = 'reference');
+            DELETE FROM nodes WHERE datatype = 'reference';
```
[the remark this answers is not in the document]
jacobtylerwalls: If we develop another datatype soon, it might be nice to have this pulled out into a util, and in that case doing it with Django would be easier I imagine, but no worries for now.

arches/app/datatypes/datatypes.py
```diff
@@ -2570,22 +2570,59 @@ def get_value_from_jsonld(json_ld_node):
...
+                    else:
+                        errors.append(
+                            {
+                                "type": "ERROR",
+                                "message": _("Reference objects require a 'uri' property and corresponding value"),
+                                "title": title,
+                            }
+                        )
+                    if "labels" in reference:
+                        pref_label_languages = []
```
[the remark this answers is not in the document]
chiatt: Nice catch. I must have gotten distracted and left that unfinished.

arches/app/datatypes/datatypes.py
```diff
@@ -2570,22 +2570,59 @@ def get_value_from_jsonld(json_ld_node):
 class ReferenceDataType(BaseDataType):
     def validate(self, value, row_number=None, source="", node=None, nodeid=None, strict=False, **kwargs):
         errors = []
+        title = _("Invalid Reference Datatype Value")
+        if value != None:
```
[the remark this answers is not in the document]
chiatt: Yeah - that would be a bit more readable

arches/app/templates/views/components/widgets/reference-select.htm
chiatt: This is really puzzling. I added a click handler to that element and clicks are landing as expected. It must be related to the css of the summary(details) report. I took a look at the url datatype and see the same behavior.

jacobtylerwalls: Thanks for looking into it #10639

arches/app/datatypes/datatypes.py
```diff
@@ -2622,6 +2622,16 @@ def clean(self, tile, nodeid):
...
         new_values = value
         return ",".join(new_values)
+    
+    def get_display_value(self, tile, node, **kwargs):
+        labels = []
+        requested_language = kwargs.pop("language", None)
+        current_language = requested_language or get_language()
+        for item in self.get_tile_data(tile)[str(node.nodeid)]:
+            for label in item["labels"]:
+                if label["language"] == current_language and label["valuetype"] == "prefLabel":
```
jacobtylerwalls, line 2636: This is okay for now, but in the next iteration we could probably use `rank_label()` so that we have good fallbacks. Right now it will fallback to nothing.

arches/app/datatypes/datatypes.py
```diff
@@ -2571,43 +2571,47 @@ class ReferenceDataType(BaseDataType):
...
-                                        {
-                                            "type": "ERROR",
-                                            "message": _("A reference can have only one prefLabel per language"),
-                                            "title": title,
-                                        }
-                                    )
-            else:
-                errors.append({"type": "ERROR", "message": _("Reference value must be a list of reference objects"), "title": title})
+        else:
+            errors.append({"type": "ERROR", "message": _("Reference value must be a list of reference objects"), "title": title})
```
jacobtylerwalls, line 2614: Isn't `None` a valid value, since we aren't cleaning None -> `[]`?

arches/app/media/js/viewmodels/reference-select.js
jacobtylerwalls: I still need a little help with this one :pray:

`this.valueAndSelectionDiffer` is called twice:
- one caller (selectionValue subscription) provides `selection` a list of strings 
- the other (value subscription) provides `selection` an observable

JSON.stringify'ing an observable will always be undefined, so why is this the value subscription calling this if it's always going to return false?

jacobtylerwalls (changes requested):
Updates looks great! One last Q for this re: the `valueAndSelectionDiffer()`, and then I'll merge. 👍

arches/app/models/migrations/10552_add_reference_datatype.py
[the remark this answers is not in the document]
jacobtylerwalls: Added the constraints_x_nodes in 2703531383f810b596a6b92ffde38dbcd33c7b1c

arches/app/models/migrations/10552_add_reference_datatype.py
apeters: @chiatt for the reasons I laid out [here](https://github.com/archesproject/arches/pull/10641#discussion_r1512042566) you might consider not having a reverse migration at all.

chiatt:
> Updates looks great! One last Q for this re: the `valueAndSelectionDiffer()`, and then I'll merge. 👍

Nice catch - thanks for looking into that. I'm not sure why I decided I needed to call that method in the value subscription. I've removed it and made a couple minor changes to ensure the dirty state clears properly if a user choses the same options with which the tile loaded (as opposed to just clicking 'Cancel Edit').

arches/app/media/js/viewmodels/reference-select.js
```diff
@@ -0,0 +1,160 @@
...
+            }
+            return val ? name : null;
+        });
+
+        this.valueAndSelectionDiffer = function(value, selection) {
+            if (!(ko.unwrap(value) instanceof Array)) {
+                return true;
+            }
+            const valueUris = ko.unwrap(value).map(val=>ko.unwrap(val.uri));
+            return (JSON.stringify(selection) !== JSON.stringify(valueUris))
```
jacobtylerwalls, line 38: > a couple minor changes to ensure the dirty state clears properly if a user choses the same options with which the tile loaded

Thanks, I'll be sure to update that again after #10602, since that adds more fields.

Your point reminds me that .stringify isn't the best choice here, as JSON object keys are unordered. Python dicts are ordered (now), but nothing on the JS side is. I can open a follow-up issue if you prefer.

jacobtylerwalls: I misstated things, looks like `JSON.stringify()` has a deterministic ordering as of ES2015, not sure about response.json() though. Maybe we should just sort right before this comparison.

arches/app/models/migrations/10552_add_reference_datatype.py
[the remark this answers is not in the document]
jacobtylerwalls: Sounds like we're going to use ` RunSQL.noop` here instead.

arches/app/media/js/viewmodels/reference-select.js
chiatt: Yeah - a quick sort would be the safe thing to do.

jacobtylerwalls: Based on the experience with the resource instance widget this is pretty hypothetical

jacobtylerwalls (approved):
Thanks, looking great!

Issue #10103, opened 2023-10-02
Running command `python manage.py packages` defaults to resetting database


SDScandrettKint:
**Arches Version/Branch:** All versions.

**Describe the bug:**
When running the command `python manage.py packages` (with nothing after and no `-o` flag) the default action is to run the setup_db function.
Potentially caused by this line: https://github.com/archesproject/arches/blob/6346ab34e10e5cf2331108947e097d8285c183cc/arches/management/commands/packages.py#L82C1-L82C29 as changing it to a different function runs the changed one.

The intended outcome should be to show the same help message as when running `python manage.py packages -o`.

**Example**
```
(env) samuel@DESKTOP:~/testing/v7test/v7test$ python manage.py packages
operation: setup
Drop and recreate the database...

SELECT pg_terminate_backend(pid) FROM pg_stat_activity
    WHERE datname IN ('v7test', 'template_postgis');

DROP DATABASE IF EXISTS v7test;

CREATE DATABASE v7test
    WITH OWNER = postgres
        ENCODING = 'UTF8'
        CONNECTION LIMIT=-1
        TEMPLATE = template_postgis;

deleting index : v7test_terms
... 
```

jacobtylerwalls:
Fixed in #10577

PR #10577, opened 2024-02-02
Remove setup as default from `python manage.py packages` to prevent fatal mistakes

SDScandrettKint:
### Types of changes

### Description of Change

Replaces the default `python manage.py packages` command with a helpful prompt to use --help to learn more about the command rather than resetting the database.  

Could replace `setup` with another of the commands, but a non-harming message seemed better even though it isn't in the choices list.  If the default value could call --help that would be better, but I'm not sure if that is possible

### Issues Solved

#10103 

### Checklist

#### Ticket Background
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/management/commands/packages.py
+++ b/arches/management/commands/packages.py
@@ -73,7 +73,6 @@ class Command(BaseCommand):
             "--operation",
             action="store",
             dest="operation",
-            default="Use the -h/--help argument to learn about this command",
             choices=[
                 "setup",
                 "install",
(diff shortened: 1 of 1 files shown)


jacobtylerwalls (changes requested):
Nice idea to just print help. I was able to get the help printed like this:

```diff
diff --git a/arches/management/commands/packages.py b/arches/management/commands/packages.py
index 3797afa17..43904d006 100644
--- a/arches/management/commands/packages.py
+++ b/arches/management/commands/packages.py
@@ -73,7 +73,6 @@ class Command(BaseCommand):
             "--operation",
             action="store",
             dest="operation",
-            default="setup",
             choices=[
                 "setup",
                 "install",
@@ -263,6 +262,10 @@ class Command(BaseCommand):
         parser.add_argument("--languages", action="store", dest="languages", help="languages desired as a comma separated list")
 
     def handle(self, *args, **options):
+        if options["operation"] is None:
+            self.print_help("manage.py", "packages")
+            return
+
         print("operation: " + options["operation"])
         package_name = settings.PACKAGE_NAME
         celery_worker_running = task_management.check_if_celery_available()
```

Issue #10027, opened 2023-09-11
Unsaved card edits no longer persist in front-end


SDScandrettKint:
In Arches V6.2, entering data into a resource card and navigating away to another card would result in the initial inputted data being in an "unsaved edits" state. The UI would represent this with yellow highlighting and navigating back would return the original inputted data:
![unsaved_edits_v6 2_cropped](https://github.com/archesproject/arches/assets/119508494/13294477-849c-4aa9-8ba0-3be3001c12f5)

In V7, this feature doesn't appear to be working as the yellow highlighting is missing and the data entered doesn't show up:
![unsaved_edits_v7_cropped](https://github.com/archesproject/arches/assets/119508494/28c88c97-6c20-4e4e-aee9-0009ff6c6b61)

SDScandrettKint:
Update: The highlighting doesn't appear for all datatypes because of the missing CSS found [here from V6](https://github.com/archesproject/arches/blob/b09dc9c3875c13f5bd44dc8431065d88d8a533d8/arches/app/media/css/tree/tree.css#L209C1-L209C32), but the issue of data not persisting is only for the string and URL datatypes.

SDScandrettKint:
Update:

This bug occurs for three datatypes:
- String 
- URL
- Node Value

PR #10581, opened 2024-02-05
Fix for cards losing their values when in "dirty" state & adding back highlight

SDScandrettKint:
### Types of changes

### Description of Change
The main bug linked shows that some datatypes do not persist card data that is entered but not saved, a feature that almost all datatypes have.  This bug is present for three datatypes: String, URL and Node Value.
This PR fixes all three (details on each fix below).

### Issues Solved

Main issue: #10027 
Datatype Issues: #10592 & #8451

### Checklist

#### Ticket Background

### Further comments
String dtype:
This bug was caused by the `defaultValue` of the string datatype being set to the i18n dictionary of language and value, causing the data entered in the string card to be reset by the empty default value dictionary.  The change in this PR checks whether the active language value is an empty string, and if so then sets the default value as an empty string, allowing the data to persist once again.

URL dtype:
The values `this.url` and `this.url_label` are used in the front end HTM for the datatype, but the variable changed when entering data is `this.value().url` & `url_label`.  The two were not visually synced, as navigating back and forth from the card would reset the UI (`this.url` & `this.url_label`) data but not the 'this.value().X' data.  This fix updates the visual variables with what has been entered.
This PR also includes a fix for #10592 by raising an error if the URL label has been inputted without the URL itself.
It also includes a fix for #8451 by adding the url_label to the object with the value of an empty string.

Node Value dtype:
When a node value is selected and navigated from and to, the `$(el).val()` in `select2-query` is not retained unlike the other select widgets (concept, resource-instance etc).  Similarly, the `initSelection` function within `select2config` id value was reset each time the node was navigated away from.  Assigning the `self.value()` to the id allows data to be retained.

### DIFF
--- a/arches/app/datatypes/url.py
+++ b/arches/app/datatypes/url.py
@@ -70,6 +70,7 @@ class URLDataType(BaseDataType):
 
     def validate(self, value, row_number=None, source=None, node=None, nodeid=None, strict=False, **kwargs):
         errors = []
+
         if value is not None:
             try:
                 if value.get("url") is not None:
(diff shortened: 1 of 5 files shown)


chiatt (commented):
Because this PR addresses bugs, it needs to be made against 7.5 rather than 7.6. Ideally the fixes that resolve #8451 should go into a second PR to be merged into 6.2.x.

SDScandrettKint:
Replaced by #10642 for 7.5.x and #10643 for cherry picked commits into 6.2.x

PR #10607, opened 2024-02-13
Avoid logging exceptions for tiles with only None values

jacobtylerwalls:
### Types of changes

### Description of Change
@chiatt [points out](https://github.com/archesproject/arches/pull/10572#discussion_r1486902445) that a tile can have only None saved to it after saved data is deleted, so this shouldn't emit a logged exception.

Then, remove release note duplicated from 7.5.0 in 7.5.1.

### History
4b02d35d097230f273d34e8aac86204efe5d2057 changed the if/else logic so that tiles with only falsy values reach the else catchall and log an exception (7.5.0)
e4a519e531a tightened that up to "only None values" to exempt falsy values like `[]` in the file-list datatype. (7.5.0)


chiatt (approved):
Looks good to me!

PR #10611, opened 2024-02-15
Update 7.5 breaking changes with `<select>` admonition

jacobtylerwalls:
Came up in an Arches implementation upgrading to 7.5.


johnatawnclementawn (approved):
Looks good 👍🏼 , I added a comment about adding an aria-label by using `data-label`. That usage only applies to selectWoo elements

Issue #10619, opened 2024-02-20
bulk_update() on I18n_JSON fields fails with heterogenous values


jacobtylerwalls:
See skipped unittest `test_bulk_update_heterogenous_values()` from #10620:

https://github.com/archesproject/arches/blob/44127d53dc8aa2bcccc8170956dc5981de618a2f/tests/localization/field_tests.py#L381-L402

Fails with:

```py
======================================================================
FAIL: test_bulk_update_heterogenous_values (tests.localization.field_tests.I18nJSONFieldBulkUpdateTests.test_bulk_update_heterogenous_values) (new_config='{}')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/jwalls/prj/arches/tests/localization/field_tests.py", line 399, in test_bulk_update_heterogenous_values
    self.assertEqual(str(obj.defaultconfig), new_config_as_string)
AssertionError: '{"en": "some", "zh": "json"}' != '{}'
- {"en": "some", "zh": "json"}
+ {}


======================================================================
FAIL: test_bulk_update_heterogenous_values (tests.localization.field_tests.I18nJSONFieldBulkUpdateTests.test_bulk_update_heterogenous_values) (new_config='None')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/jwalls/prj/arches/tests/localization/field_tests.py", line 399, in test_bulk_update_heterogenous_values
    self.assertEqual(str(obj.defaultconfig), new_config_as_string)
AssertionError: '{"en": "some", "zh": "json"}' != 'None'
- {"en": "some", "zh": "json"}
+ None


----------------------------------------------------------------------
Ran 19 tests in 2.432s

FAILED (failures=2)
```

jacobtylerwalls:
This might be something that goes away when the [underlying issue](https://code.djangoproject.com/ticket/35167) in Django is solved?

jacobtylerwalls:
Fixed in 28bb2e5401aa6b75263fdd448c85a8be48c62654

PR #10620, opened 2024-02-20
Fix bulk_update() on I18n_JSON fields for homogenous values #10618

### LINKED ISSUE #10618: bulk_update() not allowed for I18n_JSON fields
A [regression](https://code.djangoproject.com/ticket/35167) in Django 4.2 broke bulk_updates for I18n_JSON fields, with a silent failure in arches. Will open a patch for arches with a test shortly.

(A Django patch is already suggested at that link, so it should be easy for someone to pick up if they want to get a first contribution under their belts.)

jacobtylerwalls:
### Types of changes

### Description of Change
Before, there was silent data loss when using `bulk_update()` on I18n_JSON fields such as node, datatype, widget, plugin configs.

Now, we at least allow bulk_updates to homogenous values. Heterogenous values (not yet supported) throw an error to prevent a silent failure.

The situation may improve in Django 5.1 if someone submits a PR with the patch identified in [this ticket](https://code.djangoproject.com/ticket/35167). Maybe that someone is one of us? :D

### Issues Solved
Closes #10618 

Opened a followup #10619 for the issue demo'd in the skipped test.

### Checklist

#### Ticket Background
*   Sponsored by: GCI

### DIFF
--- a/arches/app/models/fields/i18n.py
+++ b/arches/app/models/fields/i18n.py
@@ -5,6 +5,7 @@ from arches.app.utils import import_class_from_string
 from django.utils.translation import gettext_lazy as _
 from django.db.migrations.serializer import BaseSerializer, Serializer
 from django.db.models import JSONField
+from django.db.models.functions.comparison import Cast
 from django.db.models.sql.compiler import SQLInsertCompiler
 from django.db.models.sql.where import NothingNode
 from django.utils.translation import get_language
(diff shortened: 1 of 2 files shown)


apeters (commented):
looks good to me.  How should we remind ourselves to remove this code once we move to Django 5.1?  A deprecation warning?

jacobtylerwalls:
Good call. I guess I'm hopeful it will be fixed in 5.1, but we don't know.

I can take the test that currently has an expected failure decorator and make it more specific to assert against `NotImplementedError`, and then if the underlying issue of sending `Cast` objects is fixed, that test will start failing because we won't reach that code block, and I'll leave a comment telling future us what to do.

jacobtylerwalls:
Retargeted 7.6, as this isn't really important enough to go back to 7.5, just adding safety for development.

PR #10622, opened 2024-02-20
Add primevue services to vue app wrapper #10621

### LINKED ISSUE #10621: Add some PrimeVue services to `createVueApplication`
> This opens up an interesting discussion around what should be included in in `createVueApplication`. Presumably we should include a subset of the PrimeVue library instead of the library itself but what heuristic should we use to define the subset?
Ideally this would grow organically over time but since the styleguide directs using `createVueApplication` we want to get it as fleshed-out as possible for external consumers

_Originally posted by @chrabyrd in https://github.com/archesproject/arches/pull/10569#discussion_r1495129798_

jacobtylerwalls:
### Types of changes

### Description of Change
Add ~toast (error message) and tooltip~all services and directives from PrimeVue to the `createVueApplication()` util so that they're available out of the 🎁 .

### Issues Solved
Closes #10621

### DIFF
--- a/arches/app/media/js/utils/create-vue-application.js
+++ b/arches/app/media/js/utils/create-vue-application.js
@@ -1,4 +1,9 @@
 import PrimeVue from 'primevue/config';
+import AnimateOnScroll from 'primevue/animateonscroll';
+import ConfirmationService from 'primevue/confirmationservice';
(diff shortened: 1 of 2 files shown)


chrabyrd (approved):
lgtm!

Issue #10624, opened 2024-02-22
The root ontology classes presented for the root node of a graph aren't filtered by the selected ontology


apeters:
If a user has loaded multiple ontologies into Arches, then the root class displays all ontology classes for the root node rather than filtering them by the selected Ontology (eg: Linked.Art in the screen shot below).

![Screenshot 2024-02-21 at 4 14 58 PM](https://github.com/archesproject/arches/assets/823773/280c9229-2f82-4f94-a7b6-2cc31708f878)

PR #10625, opened 2024-02-22
filter root ontology class by selected ontology, re #10624

apeters:
re #10624

### DIFF
--- a/arches/app/media/js/views/graph-designer.js
+++ b/arches/app/media/js/views/graph-designer.js
@@ -34,7 +34,9 @@ define([
             viewModel.graphSettingsVisible = ko.observable(false);
             viewModel.graph = koMapping.fromJS(data['graph']);
             viewModel.ontologies = ko.observable(data['ontologies']);
-            viewModel.ontologyClasses = ko.observable(data['ontologyClasses']);
+            viewModel.ontologyClasses = ko.pureComputed(function(){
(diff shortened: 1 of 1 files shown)


arches/app/media/js/views/graph-designer.js
```diff
@@ -34,7 +34,9 @@ define([
             viewModel.graphSettingsVisible = ko.observable(false);
             viewModel.graph = koMapping.fromJS(data['graph']);
             viewModel.ontologies = ko.observable(data['ontologies']);
-            viewModel.ontologyClasses = ko.observable(data['ontologyClasses']);
+            viewModel.ontologyClasses = ko.pureComputed(function(){
```
chiatt, line 37: I have Linked Art loaded, but this change doesn't seem to affect the available options.

apeters: This only "works" if you have more than one ontology loaded.  Eg, both CIDOC CRM and Linked.Art  
I used SHPO which has both of those ontologies.

I just tested again and this works. 

**Searching for String with CIDOC CRM**
Screenshot 2024-04-05 at 9 41 52 AM

**Searching for String with Linked.Art**
Screenshot 2024-04-05 at 9 41 36 AM

chiatt (approved):
Looks good!

PR #10569, opened 2024-02-01
Implement controlled list manager and reference datatype #10541

jacobtylerwalls:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
Implement the initial features for the Controlled List Manager.

TODOs:
- [x] wire up last bits of item editor
- [x] new i18n pattern for vue
- [x] refactor to typescript
- [x] make design changes per vue style guide discussions

Run migrations; then, to import some sample test data into your local db:
```py
PYTHONPATH="../arches" python3 manage.py shell -c "from tests.views.controlled_lists_tests import ControlledListTests; ControlledListTests.setUpTestData()"
```

It's likely you'll need a `manage.py updateproject` also.

Give your test user the `RDM_Administrator` group.

### Issues Solved
Closes #10541
Closes #10556

### Checklist

-   [x] Unit tests pass locally with my changes
-   [x] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: GCI

#### Demo

<details>
<summary>Component breakdown</summary>

</details>

PR #10626, opened 2024-02-22
Adjustments to eslint / typescript configs

jacobtylerwalls:
### Types of changes

### Description of Change
I found these small adjustments to be necessary when dev'ing #10569.

- [Declare primevue submodules](https://github.com/archesproject/arches/commit/bcd6f8e6c3a5d5e38ed30b97d5e4fbd3106304e5): was getting:
```
arches/app/src/plugins/ControlledListManager.vue:4:19 - error TS2307: Cannot find module 'primevue/toast' or its corresponding type declarations.

4 import Toast from "primevue/toast";
                    ~~~~~~~~~~~~~~~~
```
- [Turn off typescript in vue templates (for now)](https://github.com/archesproject/arches/commit/92123cdf9d85576f3c0ffaf5e74cf2cb7bd63313): We weren't getting meaningful type-checking in the templates, and moreover we were getting friction from having to use `this` in the templates to avoid false positives, which is discouraged, not ergonomic, and can cause behavior changes. More info [here](https://github.com/archesproject/arches/pull/10569#issuecomment-1924787017).
- [Enable vue/this-in-template](https://github.com/archesproject/arches/commit/0e0ca89fe585cf4daa79034848365af7b9ec97e2) re: previous commit
- [Remove trailing commas in tsconfig](https://github.com/archesproject/arches/commit/854f5e52e32d063fdc9996ae26cf7384a07c328a) - cosmetic change noticed while doing above (trailing commas not allowed in json, but clearly microsoft added some syntactic sugar around this? anyway, if we rig up prettier to format json, it'll do this anyway...)

### Future work
It would be great to find out why we're not getting meaningful type-checking in the templates, but it shouldn't block roll out of typescript (typing of `<script setup>` is better than nuthin'). I can open a follow-up issue.

- [ ] docs update after

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -41,7 +41,6 @@ module.exports = {
             "files": [ "*.vue" ],
             "rules": {
                 "vue/html-indent": [2, 4],
-                "vue/this-in-template": "off"
             }
         },
         {
--- a/arches/app/src/declarations.d.ts
+++ b/arches/app/src/declarations.d.ts
@@ -54,6 +54,7 @@ declare module "moment-timezone";
 declare module "nouislider";
 declare module "numeral";
 declare module "primevue";
+declare module "primevue/*";
 declare module "proj4";
 declare module "regenerator-runtime";
 declare module "requirejs";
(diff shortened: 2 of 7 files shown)


chrabyrd (approved):
sounds reasoning && good changes all around

PR #10631, opened 2024-02-22
Collect nodes using controlled lists #10554

### LINKED ISSUE #10554: [controlled-lists] Wire up "List used by these nodes"
Blocked by #10552

Include link to edit node

jacobtylerwalls:
### Types of changes

### Description of Change
Include information about nodes referencing controlled lists in the call to the controlled lists endpoint.

Front-end integration will be handled in the [next branch](https://github.com/archesproject/arches/tree/controlled-list-item-editor-second-iteration), you can see it working here: ba6c5f4221de5e2a6ceee17c4b33b5344dd9cb3d

Screenshot 2024-02-23 at 3 43 34 PM

### Issues Solved
Closes #10554

### Checklist

#### Ticket Background
*   Sponsored by: GCI


arches/app/models/models.py
```diff
@@ -725,6 +727,12 @@ class Meta:
             models.UniqueConstraint(fields=["name", "nodegroup"], name="unique_nodename_nodegroup"),
             models.UniqueConstraint(fields=["alias", "graph"], name="unique_alias_graph"),
         ]
+        indexes = [
+            models.Index(
+                Cast(KT("config__controlledList"), output_field=models.UUIDField(null=True)),
```
chiatt, line 732: I didn't realize this was available from the ORM. It seems worth looking into other places in Arches where we should be adding JSON indexes.

I did a little test to confirm it's being used (and it is). 
```
EXPLAIN(SELECT n."nodeid" 
FROM "nodes" n 
WHERE ((n."config" ->> 'controlledList'))::uuid = ('06b58ca6-bcb6-46b3-8aaa-64cca94ab8f8'))
```

chiatt (approved):
This looks great. A list of nodes accompanies each controlled list. Maybe it's beyond the scope of this PR, but should the ControlledList model prevent a user from deleting a list if it's in use?

arches/app/models/models.py
jacobtylerwalls: Smart! (One thing I've done recently is to just call `.explain()` on the queryset and log out the result.)

chiatt: > Smart! (One thing I've done recently is to just call `.explain()` on the queryset and log out the result.)

Oh, that's really cool. I think @apeters would definitely appreciate that trick.

arches/app/datatypes/datatypes.py
```diff
@@ -1886,7 +1886,7 @@ def get_localized_option_text(self, node, option_id, return_lang=False):
         for option in node.config["options"]:
             if option["id"] == option_id:
                 return get_localized_value(option["text"], return_lang=return_lang)
-        raise Exception(_("No domain option found for option id {0}, in node conifg: {1}".format(option_id, node.config["options"])))
+        raise Exception(_("No domain option found for option id {0}, in node config: {1}".format(option_id, node.config["options"])))
```
chiatt, line 1889: Nice catch

chiatt (approved):
Looks good to me!

arches/app/models/models.py
apeters: That is cool!   🥇

PR #10632, opened 2024-02-26
Include .ts files in gettext extraction

jacobtylerwalls:
Follow-up to 01ffc0e5214940455c8418eeff8c1277ef0bde91

I guess we may hold this if we decide not to allow typescript functions to take `$gettext` callbacks, but I'm currently doing it, and figure would be nice to support:

https://github.com/archesproject/arches/blob/c677d557791c678cad03b51bc969964cc506f086/arches/app/src/components/ControlledListManager/api.ts#L38-L42

### DIFF
--- a/arches/install/arches-app-templates/gettext.config.js
+++ b/arches/install/arches-app-templates/gettext.config.js
@@ -2,7 +2,7 @@
 module.exports = {
     input: {
       path: "./{{ app_name }}/src", // only files in this directory are considered for extraction
-      include: ["**/*.vue"], // glob patterns to select files for extraction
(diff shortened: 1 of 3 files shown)


chrabyrd (commented):
Overall looks good, just raises the question of whether we should allow it or not. I'll bring it up for discussion in MMM

chrabyrd:
lgtm -- after chatting with @apeters landed on the idea that we support this but update the styleguide to urge developers to not use i18n in `.ts` files and instead have the error messages live as close to the callsites as possible 👍

PR #10641, opened 2024-03-01
9191 non localized string datatype

apeters:
### Types of changes

### Issues Solved

#

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/models/migrations/9191_string_nonlocalized.py
```diff
@@ -0,0 +1,62 @@
...
+                    "defaultValue": ""
+                }'
+            );
+            """,
+            reverse_sql = """
+            DELETE FROM d_data_types
+                WHERE datatype = 'non-localized-string';
+            DELETE from widgets
+                WHERE widgetid = '46ef064b-2611-4708-9f52-60136bd8a65b';
+            """
```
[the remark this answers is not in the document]
apeters: Although I haven't tested this explicitly I'm leery of deleting items we didn't explicitly add.  For example deleting a node/edge in a graph just because it references a datatype that now doesn't exist.  I'm pretty sure that's going to blow up the graph and at least will orphan any child nodes of the deleted node.  That seems unexpected.  If I was a graph designer I would be pretty bummed, and there would be no way to recover from that without going to the db and even then it would be tough to do.  It would be nice if instead we provide a sensible default datatype for those nodes (maybe semantic).  Or maybe leave it as-is.   I'm going to test this first though.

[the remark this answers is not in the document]
jacobtylerwalls: I like the idea of a fallback. I'm sure you've noticed this already, but the reason Cyrus was deleting nodes was because the reverse migration failed to delete the datatype [like this](https://github.com/archesproject/arches/pull/10572#discussion_r1482201070)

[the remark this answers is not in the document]
apeters: I tired adding the code that @chiatt  created above and it made my whole graph unavailable.  I then thought about changing the datatype to a localized string, but that also has substantial pitfalls (not the least of which is the ES mappings). 
The more I think about this the more I think that we shouldn't have a reverse migration at all.  That we should essentially leave the "new" datatype in place, even if a user wants to reverse the migration.  For one thing it won't hurt to leave it in.  For another, this is a new feature that we want users to have.  It causes a bunch of chaos if a user has started to use it and then for whatever reason wants to undo the migration.  Leaving it in place allows a user to manually undo the work **if they want to**, creating a migration path for their own data if they want by using existing methods built in to Arches.  
In this case at least, if we could re-imagine our whole "upgrade" to the localized  string datatype, we would have left in a non-localized version.

arches/app/templates/views/components/widgets/non-localized-text.htm
```diff
@@ -0,0 +1,46 @@
+{% extends "views/components/widgets/base.htm" %}
+{% load i18n %}
+
+{% block form %}
+<div class="row widget-wrapper">
+    <div class="form-group">
+        <label class="control-label widget-input-label" for="" data-bind="text:label"></label>
+        <!-- ko if: node -->
+        <i data-bind="css: {'ion-asterisk widget-label-required': node.isrequired}"></i>
```
jacobtylerwalls, line 9:
```suggestion
        <i data-bind="
            css: {'ion-asterisk widget-label-required': node.isrequired},
            attr: {'aria-label': $root.translations.required}        
        "></i>
```

arches/app/templates/views/components/widgets/non-localized-text.htm
```diff
@@ -0,0 +1,46 @@
...
+    </div>
+</div>
+{% endblock form %}
+
+{% block config_form %}
+<div class="control-label">
+    {% trans "Placeholder" %}
+</div>
+<div class="col-xs-12 pad-no crud-widget-container">
+    <input type="" placeholder="{% trans "Placeholder" %}" id="" class="form-control input-md widget-input" data-bind="textInput: placeholder">
```
jacobtylerwalls, line 23: Is there a reason for empty `id` attrs here and below?

tests/utils/datatype_tests.py
```diff
@@ -23,7 +23,7 @@
 
 
 # these tests can be run from the command line via
-# python manage.py test tests/utils/datatype_tests.py --pattern="*.py" --settings="tests.test_settings"
+# python manage.py test tests.utils.datatype_tests.py --pattern="*.py" --settings="tests.test_settings"
```
jacobtylerwalls, line 26: .py also needs to be stripped. I'll fix this in a separate PR

arches/app/templates/javascript.htm
```diff
@@ -326,7 +324,6 @@
         date-format='{% trans "Date Format" as dateFormat %} "{{ dateFormat|escapejs }}"'
         has-no-value='{% trans "Has no value" as hasNoValue %} "{{ hasNoValue|escapejs }}"'
         has-any-value='{% trans "Has any value" as hasAnyValue %} "{{ hasAnyValue|escapejs }}"'
-        default-value='{% trans "Default Value" as defaultValue %} "{{ defaultValue|escapejs }}"'
```
jacobtylerwalls, line 329: Still seems to be in use in several places:
https://github.com/archesproject/arches/blob/a1c2b429f5aaa54af95a4111862eb87d56709107/arches/app/templates/views/components/widgets/datepicker.htm#L83-L85

arches/app/templates/javascript.htm
```diff
@@ -230,8 +230,6 @@
         continue='{% trans "Continue" as continue %} "{{ continue|escapejs }}"'
         value='{% trans "Value" as value %} "{{ value|escapejs }}"'
         disabled='{% trans "Disabled" as disabled %} "{{ disabled|escapejs }}"'
-        disable-editing='{% trans "Disable Editing" as disableEditing %} "{{ disableEditing|escapejs }}"'
-        prevent-user-from-editing-value='{% trans "Prevent users from editing value" as preventUsersFromEditingValue %} "{{ preventUsersFromEditingValue|escapejs }}"'
```
jacobtylerwalls, line 234: Same

arches/app/templates/views/components/widgets/non-localized-text.htm
```diff
@@ -0,0 +1,69 @@
...
+<div class="col-xs-12 pad-no crud-widget-container">
+    <input class="form-control input-md widget-input"
+        data-bind="
+            attr: {placeholder: $root.translations.maxLength, 'aria-label': $root.translations.maxLength},
+            textInput: maxLength
+        "
+    >
+</div>
+
+<div class="control-label" >
```
jacobtylerwalls, line 39:
```suggestion
<div class="control-label">
```

arches/app/templates/javascript.htm
apeters: These were dupes.

arches/app/templates/javascript.htm
apeters: More dupes.

Issue #10592, opened 2024-02-08
URL datatype able to save label without URL resulting in JS errors


SDScandrettKint:
Found on Arches v6 and v7

Entering a URL label without the URL allows the card to be saved and produces the following visual error:
![image](https://github.com/archesproject/arches/assets/119508494/1f76f0e2-4e25-460c-bffd-5d4043754f36)

PR #10642, opened 2024-03-01
 Fix for cards losing their values when in "dirty" state & adding back highlight - 7.5.x 

SDScandrettKint:
### Types of changes

### Description of Change
The main bug linked shows that some datatypes do not persist card data that is entered but not saved, a feature that almost all datatypes have.  This bug is present for three datatypes: String, URL and Node Value.
This PR fixes all three (details on each fix below).

### Issues Solved

Main issue: #10027 
URL datatype issues: #10592, #8451

### Checklist

#### Ticket Background

### Further comments
String dtype:
This bug was caused by the `defaultValue` of the string datatype being set to the i18n dictionary of language and value, causing the data entered in the string card to be reset by the empty default value dictionary.  The change in this PR checks whether the active language value is an empty string, and if so then sets the default value as an empty string, allowing the data to persist once again.

URL dtype:
The values `this.url` and `this.url_label` are used in the front end HTM for the datatype, but the variable changed when entering data is `this.value().url` & `url_label`.  The two were not visually synced, as navigating back and forth from the card would reset the UI (`this.url` & `this.url_label`) data but not the 'this.value().X' data.  This fix updates the visual variables with what has been entered.
This PR also includes a fix for #10592 by raising an error if the URL label has been inputted without the URL itself.
It also includes a fix for #8451 by adding the url_label to the object with the value of an empty string.
Also found during testing this PR was #10732.

Node Value dtype:
When a node value is selected and navigated from and to, the `$(el).val()` in `select2-query` is not retained unlike the other select widgets (concept, resource-instance etc).  Similarly, the `initSelection` function within `select2config` id value was reset each time the node was navigated away from.  Assigning the `self.value()` to the id allows data to be retained.

### DIFF
--- a/arches/app/datatypes/url.py
+++ b/arches/app/datatypes/url.py
@@ -70,6 +70,7 @@ class URLDataType(BaseDataType):
 
     def validate(self, value, row_number=None, source=None, node=None, nodeid=None, strict=False, **kwargs):
         errors = []
+
         if value is not None:
             try:
                 if value.get("url") is not None:
@@ -82,6 +83,18 @@ class URLDataType(BaseDataType):
                 title = _("Invalid HTTP/HTTPS URL")
                 error_message = self.create_error_message(value, source, row_number, message, title)
                 errors.append(error_message)
+            
+            #10592 raise error if label added without URL 
+            if value.get("url_label") and not value.get("url"):
+                message = _("URL label cannot be saved without a URL")
+                title = _("No URL added")
+                error_message = self.create_error_message(value, source, row_number, message, title)
+                errors.append(error_message)
+
+            #10593 adds url_label to JSON if no label added
+            if "url_label" not in value:
+                value["url_label"] = ""
+                
         return errors
 
     def transform_value_for_tile(self, value, **kwargs):
--- a/arches/app/datatypes/url.py
+++ b/arches/app/datatypes/url.py
@@ -266,8 +266,5 @@ class URLDataType(BaseDataType):
         }
 
     def pre_tile_save(self, tile, nodeid):
-        try:
-            if "url_label" not in tile.data[nodeid]:
-                tile.data[nodeid]["url_label"] = ""
-        except AttributeError:
-            pass
+        if tile and "url_label" not in tile.data[nodeid]:
+            tile.data[nodeid]["url_label"] = ""
--- a/arches/app/datatypes/url.py
+++ b/arches/app/datatypes/url.py
@@ -266,5 +266,6 @@ class URLDataType(BaseDataType):
         }
 
     def pre_tile_save(self, tile, nodeid):
-        if tile and "url_label" not in tile.data[nodeid]:
-            tile.data[nodeid]["url_label"] = ""
+        if tile.data[nodeid]:
+            if tile and "url_label" not in tile.data[nodeid]:
+                tile.data[nodeid]["url_label"] = ""
--- a/arches/app/datatypes/url.py
+++ b/arches/app/datatypes/url.py
@@ -73,7 +73,7 @@ class URLDataType(BaseDataType):
 
         if value is not None:
             try:
-                if value.get("url") is not None:
+                if value.get("url"):
                     # check URL conforms to URL structure
                     url_test = self.URL_REGEX.match(value["url"])
                     if url_test is None:
--- a/arches/app/datatypes/url.py
+++ b/arches/app/datatypes/url.py
@@ -266,6 +266,5 @@ class URLDataType(BaseDataType):
         }
 
     def pre_tile_save(self, tile, nodeid):
-        if tile.data[nodeid]:
-            if tile and "url_label" not in tile.data[nodeid]:
-                tile.data[nodeid]["url_label"] = ""
+        if (tile_val := tile.data[nodeid]) and "url_label" not in tile_val:
+            tile_val["url_label"] = ""
--- a/arches/app/datatypes/url.py
+++ b/arches/app/datatypes/url.py
@@ -90,9 +90,6 @@ class URLDataType(BaseDataType):
                 title = _("No URL added")
                 error_message = self.create_error_message(value, source, row_number, message, title)
                 errors.append(error_message)
-
-            if "url_label" not in value:
-                value["url_label"] = ""
                 
         return errors
 
@@ -267,3 +264,10 @@ class URLDataType(BaseDataType):
                 },
             }
         }
+
+    def pre_tile_save(self, tile, nodeid):
+        try:
+            if "url_label" not in tile.data[nodeid]:
+                tile.data[nodeid]["url_label"] = ""
+        except AttributeError:
+            pass
--- a/arches/app/datatypes/url.py
+++ b/arches/app/datatypes/url.py
@@ -268,3 +268,15 @@ class URLDataType(BaseDataType):
     def pre_tile_save(self, tile, nodeid):
         if (tile_val := tile.data[nodeid]) and "url_label" not in tile_val:
             tile_val["url_label"] = ""
+
+    def clean(self, tile, nodeid):
+        if (data := tile.data[nodeid]):
+            try:
+                if not any([val.strip() for val in data.values()]):
+                    tile.data[nodeid] = None
+            except:
+                pass # Let self.validate handle malformed data
+
+    def pre_structure_tile_data(self, tile, nodeid, **kwargs):
+        if (tile_val := tile.data[nodeid]) and "url_label" not in tile_val:
+            tile_val["url_label"] = ""
--- a/arches/app/datatypes/url.py
+++ b/arches/app/datatypes/url.py
@@ -84,14 +84,13 @@ class URLDataType(BaseDataType):
                 error_message = self.create_error_message(value, source, row_number, message, title)
                 errors.append(error_message)
             
-            #10592 raise error if label added without URL 
+            # raise error if label added without URL (#10592)
             if value.get("url_label") and not value.get("url"):
                 message = _("URL label cannot be saved without a URL")
                 title = _("No URL added")
                 error_message = self.create_error_message(value, source, row_number, message, title)
                 errors.append(error_message)
 
-            #10593 adds url_label to JSON if no label added
             if "url_label" not in value:
                 value["url_label"] = ""
                 
--- a/arches/app/media/js/viewmodels/node-value-select.js
+++ b/arches/app/media/js/viewmodels/node-value-select.js
@@ -121,6 +121,11 @@ define([
             initSelection: function(element, callback) {
                 var id = $(element).val();
                 var tiles = self.tiles();
+                
+                if (self.value()) {
+                    id = self.value()
+                }
+                
                 if (id !== "") {
                     var setSelection = function(tiles, callback)   {
                         var selection =  _.find(tiles, function(tile) {
--- a/arches/app/media/js/viewmodels/node-value-select.js
+++ b/arches/app/media/js/viewmodels/node-value-select.js
@@ -122,6 +122,7 @@ define([
                 var id = $(element).val();
                 var tiles = self.tiles();
                 
+                // fixes #10027 where inputted values will be reset after navigating away  
                 if (self.value()) {
                     id = self.value()
                 }
--- a/arches/app/media/js/viewmodels/node-value-select.js
+++ b/arches/app/media/js/viewmodels/node-value-select.js
@@ -124,7 +124,7 @@ define([
                 
                 // fixes #10027 where inputted values will be reset after navigating away  
                 if (self.value()) {
-                    id = self.value()
+                    id = self.value();
                 }
                 
                 if (id !== "") {
--- a/arches/app/media/js/views/components/widgets/text.js
+++ b/arches/app/media/js/views/components/widgets/text.js
@@ -21,6 +21,14 @@ define([
     const viewModel = function(params) {
         params.configKeys = ['placeholder', 'width', 'maxLength', 'defaultValue', 'uneditable'];
 
+        // If the defaultValue of the selected language is empty set to ""
+        // fixes #10027 where inputted values will be reset after navigating away
+        if (params?.config?.()?.defaultValue != "") {
+            if (params?.config?.()?.defaultValue[arches.activeLanguage]["value"] == '') {
+                params.config().defaultValue = ""
+            }
+        }
+
         WidgetViewModel.apply(this, [params]);
         const self = this;
 
--- a/arches/app/media/js/views/components/widgets/text.js
+++ b/arches/app/media/js/views/components/widgets/text.js
@@ -21,14 +21,6 @@ define([
     const viewModel = function(params) {
         params.configKeys = ['placeholder', 'width', 'maxLength', 'defaultValue', 'uneditable'];
 
-        // If the defaultValue of the selected language is empty set to ""
-        // fixes #10027 where inputted values will be reset after navigating away
-        if (params?.config?.()?.defaultValue != "") {
-            if (params?.config?.()?.defaultValue?.[arches.activeLanguage]?.["value"] == '') {
-                params.config().defaultValue = ""
-            }
-        }
-
         WidgetViewModel.apply(this, [params]);
         const self = this;
 
@@ -187,6 +179,10 @@ define([
             self.currentDirection(koMapping.toJS(self.value)[currentLanguage.code]?.direction);
 
         });
+
+        if (self.currentDefaultText() === "") {
+            self.defaultValue("")
+        }
 
     };
 
--- a/arches/app/media/js/views/components/widgets/text.js
+++ b/arches/app/media/js/views/components/widgets/text.js
@@ -181,7 +181,7 @@ define([
         });
 
         if (self.currentDefaultText() === "") {
-            self.defaultValue("")
+            self.defaultValue("");
         }
 
     };
--- a/arches/app/media/js/views/components/widgets/text.js
+++ b/arches/app/media/js/views/components/widgets/text.js
@@ -24,7 +24,7 @@ define([
         // If the defaultValue of the selected language is empty set to ""
         // fixes #10027 where inputted values will be reset after navigating away
         if (params?.config?.()?.defaultValue != "") {
-            if (params?.config?.()?.defaultValue[arches.activeLanguage]["value"] == '') {
+            if (params?.config?.()?.defaultValue?.[arches.activeLanguage]?.["value"] == '') {
                 params.config().defaultValue = ""
             }
         }
--- a/arches/app/media/js/views/components/widgets/urldatatype.js
+++ b/arches/app/media/js/views/components/widgets/urldatatype.js
@@ -11,6 +11,18 @@ define([
          
         WidgetViewModel.apply(this, [params]);
 
+
+        // #10027 assign this.url & this.url_label with value versions for updating UI with edits
+        if (typeof (this.value) === "function") {
+            if (this.value()) {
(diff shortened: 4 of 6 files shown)


jacobtylerwalls:
Hey @SDScandrettKint thanks for this. I noticed there wasn't optional chaining after accessing `arches.activeLanguage`, so I gave it a spin and indeed can break the widget in this workflow:

- save string node
- save text widget
- switch language while in graph designer
- view text widget config


https://github.com/archesproject/arches/assets/38668450/f57210e5-7510-4be9-b421-7f25af6a1918

SDScandrettKint:
Thanks for pointing this out @jacobtylerwalls !  I've added optional chaining to the defaultValue `arches.activeLanguage` value.

I was doing some debugging to figure out why the changed language default value wasn't working, and it left me a bit confused.  In the screenshot below, logging out `params?.config?.()?.defaultValue` (1) shows the "de" default value in the object, the `arches.activeLangauge` is "de" but the `params?.config?.()?.defaultValue[arches.activeLanguage]` doesn't exist (3) which somewhat contradicts the first.  
image

When the active langauge is "en", `params?.config?.()?.defaultValue[arches.activeLanguage]` works fine.  Does this only work for the default language i.e "en"?

jacobtylerwalls:
Interesting, so it sounds like it's mutating in between your breakpoints. I had a couple items I was wondering if you gave some thought to:

1. Your diff sets `defaultValue = ""`, which I think will only create an entry for the active language. Is that why you're seeing a different result for English only? What happens if you set all languages to the empty string there?
2. There's code below this in `init()` that makes determinations about the default value for all languages. Is there a deeper source for this bug in that logic?

SDScandrettKint:
I have taken another look at it and believe the latest commit is a much better, simpler and overall nicer looking fix for the issue.  It uses the `currentDefaultText` variable that already exists and interacts with the default value, and checks if that is a value or an empty string.  If `currentDefaultText` is empty then `self.defaultValue()` can be made empty too, as otherwise it is remaining as populated with the i18n empty dictionary, which causes the original issue of the entered data not being retained.

This improved fix also means there is no need to access the current language, as this is already done in `currentDefaultText`, allowing for the value data to be looked at directly.

I have tested switching between languages, as demonstrated in your video, and that error no longer occurs.

arches/app/datatypes/url.py
```diff
@@ -82,6 +83,17 @@ def validate(self, value, row_number=None, source=None, node=None, nodeid=None,
...
                 errors.append(error_message)
+            
+            # raise error if label added without URL (#10592)
+            if value.get("url_label") and not value.get("url"):
+                message = _("URL label cannot be saved without a URL")
+                title = _("No URL added")
+                error_message = self.create_error_message(value, source, row_number, message, title)
+                errors.append(error_message)
+
+            if "url_label" not in value:
```
chiatt, line 94: The validate method shouldn't modify data. It should only check if a value is valid. The pre_tile_save datatype method is intended for this kind of change.

SDScandrettKint: I noticed that the urldatatype.py didn't have a `pre_tile_save` function, so I have added one and moved the fix from validate.  Will reflect the changes in #10643

One thing I have noticed is that this fix helps to validate all future urldatatype additions, however, all of the url datatype tiles saved prior to this fix will still lack the `"url_label": ""` from their data, and thus the UI will not allow them to add a label unless the tile is deleted and data re-populated.

Is there some way we can check and correct all previously saved data that doesn't contain "url_label" and insert the key with an empty string value (`"url_label": ""`) like we are doing here?

arches/app/datatypes/url.py
```diff
@@ -267,3 +264,10 @@ def default_es_mapping(self):
                 },
             }
         }
+
+    def pre_tile_save(self, tile, nodeid):
+        try:
+            if "url_label" not in tile.data[nodeid]:
+                tile.data[nodeid]["url_label"] = ""
+        except AttributeError:
```
chiatt, line 272: Is the AttributeError exception because `tile` could be None? If so, I think this would be clearer:
```
if tile and "url_label" not in tile.data[nodeid]:
```

SDScandrettKint: Fixed in https://github.com/archesproject/arches/pull/10642/commits/aa9e9d9f672af6de7c6924c2e80ab86541676adf

arches/app/media/js/views/components/widgets/urldatatype.js
```diff
@@ -11,6 +11,18 @@ define([
          
         WidgetViewModel.apply(this, [params]);
 
+
+        // #10027 assign this.url & this.url_label with value versions for updating UI with edits
+        if (typeof (this.value) === "function") {
+            if (this.value()) {
```
jacobtylerwalls, line 17: Could just be
```suggestion
        if (ko.isObservable(this.value) && this.value()) {
```

SDScandrettKint: I much prefer this.  Applied in [17bfad7](https://github.com/archesproject/arches/pull/10642/commits/17bfad73e18a6dce480859f3a677a4cf6b7e2171).

arches/app/datatypes/url.py
```diff
@@ -255,3 +264,7 @@ def default_es_mapping(self):
                 },
             }
         }
+
+    def pre_tile_save(self, tile, nodeid):
+        if tile and "url_label" not in tile.data[nodeid]:
```
jacobtylerwalls, line 269: Will also need to guard against `tile.data[nodeid]` being None:

```
{
    "message": "argument of type 'NoneType' is not iterable",
    "status": "false",
    "success": false,
    "title": "Unable to save."
}
```

arches/app/media/js/viewmodels/node-value-select.js
```diff
@@ -121,6 +121,12 @@ define([
             initSelection: function(element, callback) {
                 var id = $(element).val();
                 var tiles = self.tiles();
+                
+                // fixes #10027 where inputted values will be reset after navigating away  
+                if (self.value()) {
+                    id = self.value()
```
jacobtylerwalls, line 127: And elsewhere in this PR, if you don't mind please add trailing semicolons, thanks!
```suggestion
                    id = self.value();
```

arches/app/media/js/views/components/widgets/urldatatype.js
```diff
@@ -11,6 +11,16 @@ define([
          
         WidgetViewModel.apply(this, [params]);
 
+
+        // #10027 assign this.url & this.url_label with value versions for updating UI with edits
```
jacobtylerwalls, line 15: Noticed that with this change, if I save a tile with a URL and without a label, the dirty state is still active, i.e. "cancel edit" and "save" buttons still present.

This is tricky because the old behavior is definitely more buggy, but would be nice to leave the situation improved in all respects.

jacobtylerwalls (changes requested):

arches/app/datatypes/url.py
jacobtylerwalls: > and thus the UI will not allow them to add a label unless the tile is deleted and data re-populated.

Indeed, when I try to do this, I see two things:
1. No dirty state detected
2. TypeError:
```js
urldatatype.js:17 Uncaught TypeError: this.url_label is not a function
    at viewModel.eval (urldatatype.js:17:36)
```

I think we need to find a way to guard this on the frontend.

> Is there some way we can check and correct all previously saved data that doesn't contain "url_label" and insert the key with an empty string value ("url_label": "") like we are doing here?

We could add a data migration, but then we'd want to target 7.6. We should probably just be more permissive in the UI.

arches/app/datatypes/url.py
SDScandrettKint: Ah, this occurs when an empty url datatype tile is saved, something I forgot to test.  Great spot.  

I have added an `if tile.data[nodeid]:` in [2aedaa7](https://github.com/archesproject/arches/pull/10642/commits/2aedaa718b484c95f92de69003cb69747536ab9f), would this suffice?

arches/app/media/js/viewmodels/node-value-select.js
SDScandrettKint: Added in [2e569c0](https://github.com/archesproject/arches/pull/10642/commits/2e569c0896444644c3594f093394578be5362d62)

arches/app/media/js/views/components/widgets/urldatatype.js
SDScandrettKint: I'm not sure I can reproduce this, once saving the URL tile (with no label) the edit and save buttons are gone for me. 
https://github.com/archesproject/arches/assets/119508494/50376937-7c41-44e8-bbf5-df1fe7a7c40f

Am I trying the wrong thing, or could it be something wrong with my testing environment?

SDScandrettKint:
One extra bug I've noticed whilst testing is that "Cancel edits" doesn't work correctly in the URL datatype for an **unsaved** tile edit.  Unlike the other datatypes, clicking "Cancel edits" does not remove the edits and revert to the an empty text field.  "Cancel edits" does work once the tile has been initially saved.

https://github.com/archesproject/arches/assets/119508494/cda8d6cf-8742-45e1-a34c-c418c46f896e

I have created an issue for this bug (#10732), but I think it seems relevant to look for a fix as a part of this PR.  Just thought I'd make a note of it here.

arches/app/media/js/views/components/widgets/urldatatype.js
jacobtylerwalls: I cannot reproduce this problem any longer. Thanks for bearing with me.

arches/app/datatypes/url.py
```diff
@@ -255,3 +264,8 @@ def default_es_mapping(self):
                 },
             }
         }
+
+    def pre_tile_save(self, tile, nodeid):
+        if tile.data[nodeid]:
+            if tile and "url_label" not in tile.data[nodeid]:
```
jacobtylerwalls, line 270: Now `if tile` is redundant, as we've already checked something more specific.
```suggestion
            if "url_label" not in tile.data[nodeid]:
```

Probably could just be consolidated, at your option:
```py
    if (tile_val := tile.data[nodeid]) and "url_label" not in tile_val:
        tile_val["url_label"] = ""
```

arches/app/datatypes/url.py
```diff
@@ -82,6 +83,14 @@ def validate(self, value, row_number=None, source=None, node=None, nodeid=None,
                 title = _("Invalid HTTP/HTTPS URL")
                 error_message = self.create_error_message(value, source, row_number, message, title)
                 errors.append(error_message)
+            
+            # raise error if label added without URL (#10592)
+            if value.get("url_label") and not value.get("url"):
+                message = _("URL label cannot be saved without a URL")
+                title = _("No URL added")
+                error_message = self.create_error_message(value, source, row_number, message, title)
+                errors.append(error_message)
```
jacobtylerwalls, line 92: @chiatt mentioned at the developer meeting last month (btw, great to meet you there!) that we're trying to improve the test coverage of the python backend. There's a test file for the datatype validation methods [here](https://github.com/archesproject/arches/blob/0c89e7d117f4c639f1ce706f3dfaa82462fb1ca4/tests/utils/datatype_tests.py#L4).

Would you be able to add a `URLDataTypeTests.test_validate()` method to cover your changes here?

jacobtylerwalls:
> I have created an issue for this bug (https://github.com/archesproject/arches/issues/10732), but I think it seems relevant to look for a fix as a part of this PR. Just thought I'd make a note of it here.

This PR is already addressing several issues, so if the fix turns out to be cleanly segregated from your changes here, it would be better to have a new PR for it. (It may be that it doesn't qualify for a backport to 7.5 if it's not a recent regression or related to data loss.)

arches/app/datatypes/url.py
SDScandrettKint: Great to meet you there too! 

I agree and was thinking about how this PR needed some tests during that presentation at the developer meetings.  I'll start working on writing some.

SDScandrettKint:
> > I have created an issue for this bug (#10732), but I think it seems relevant to look for a fix as a part of this PR. Just thought I'd make a note of it here.
> 
> This PR is already addressing several issues, so if the fix turns out to be cleanly segregated from your changes here, it would be better to have a new PR for it. (It may be that it doesn't qualify for a backport to 7.5 if it's not a recent regression or related to data loss.)

After a little bit of investigating this issue, I believe that it belongs in a separate PR as it seems like it can be cleanly separated from the work in this PR.

arches/app/datatypes/url.py
SDScandrettKint: Added in [685a5c6](https://github.com/archesproject/arches/pull/10642/commits/685a5c682bbac3614a4d44a789768440c4f755d4)

arches/app/datatypes/url.py
SDScandrettKint: I've written some tests for the url datatype in `datatype_tests.py` under the class name `URLDataTypeTests`, testing validation and pre_tile_save.  They are in the latest commit [685a5c6](https://github.com/archesproject/arches/pull/10642/commits/685a5c682bbac3614a4d44a789768440c4f755d4).

This was my first time writing django tests so please let me know if things need changing/adding @jacobtylerwalls @chiatt

SDScandrettKint: In the latest commit ([fd3c389](https://github.com/archesproject/arches/pull/10642/commits/fd3c3890bd0a5a405b1e287d892f9185f7de1baa)) I changed the line https://github.com/archesproject/arches/blob/0c89e7d117f4c639f1ce706f3dfaa82462fb1ca4/arches/app/datatypes/url.py#L75 to https://github.com/archesproject/arches/blob/fd3c3890bd0a5a405b1e287d892f9185f7de1baa/arches/app/datatypes/url.py#L76
This was because the test for a URL with an empty string and no label (e.g. `{"url": "", "url_label": "Label"}` was showing two errors, one for the fact a label couldn't be saved without a URL, and a second for the URL being invalid http/https.  This seemed wrong to me, as an empty string shoudn't be tested as a valid URL just because it isn't `None`.  Thoughts?

arches/app/datatypes/url.py
jacobtylerwalls: Hey @SDScandrettKint I just wanted to make sure this comment wasn't lost. Do you have a view about how to proceed here? Should we guard for this in the UI and just create the missing key on demand?

arches/app/datatypes/url.py
jacobtylerwalls: I suppose the problem with that is that in the case of `{'url': '', 'url_label': ''}`, which you can have if you type into the fields and then backspace your entries before clicking save, the validation will now pass (when it shouldn't).

I think having both errors isn't hurting anything. The UI in the resource editor only presents the more useful error about the invalid URL.

Happy to hear @chiatt's thoughts too.

jacobtylerwalls: Suggest adding `{'url': '', 'url_label': ''}` as a test case once we know how it should behave.

tests/utils/datatype_tests.py
```diff
@@ -74,3 +74,53 @@ def test_tile_clean(self):
         string.clean(tile2, nodeid)
 
         self.assertIsNotNone(tile2.data[nodeid])
+
+
+class URLDataTypeTests(ArchesTestCase):
+    def test_validate(self):
```
jacobtylerwalls, line 80: Nice work with the tests; they're clear and they fail on the target branch.

One thing that can improve the developer experience when tests fail is when all of the assertions have a chance to fail. For example here, when I check them on the target branch, the test stops at the first failed assertion.

If you find yourself doing something repetitive, like calling `url.validate()`, you can batch them up and call them under `with self.subTest()` so that the runner will actually assert everything instead of failing fast and exiting.

Here's an example, where the left side `input=` is an arbitrary label that I want to appear in my test output:

https://github.com/archesproject/arches/blob/e4145e4cd546210c1bde18a04634fb36ec8d548b/tests/utils/datatype_tests.py#L33-L42

If you play with that example and change the expected result, you'll get all 9 failures with the info about the specific way each one failed. Versus if you remove the subTest() statement, you only get 1 failure.

We can merge this work without adding subTest IMO, but I just wanted to mention it in case you find this useful when writing other unit tests.

arches/app/datatypes/url.py
chiatt: It seems like what we need in the case of `{'url': '', 'url_label': ''}` is a `clean` method to return the value back to `None` on save. Perhaps something like:

```
def clean(self, tile, nodeid):
    if (data := tile.data[nodeid]):
        try:
            if not any([val.strip() for val in data.values()]):
                tile.data[nodeid] = None
        except Exception:
            pass # Let self.validate handle malformed data
```

arches/app/datatypes/url.py
SDScandrettKint: Thanks for bringing this to my attention, you were right it was lost - I somehow didn't see this comment.
I also see the lack of dirty state and an error in the browser console.

I've been looking at this for a few days, trying to get a dirty state in the UI, but this approach failed as it doesn't seem to be possible to turn this.tile.dirty(true).  Similarly, modifying this.url_label also didn't affect the dirty tile state.  After speaking with @chiatt I've changed the approach and added the `pre_structure_tile_data` function which uses the code from validate to add the "url_label" key and allow editing of the label.  Then upon save the key will be saved to the JSON.  Though this approach doesn't prompt a dirty state, it does avoid the need for a migration.

This change is in the latest commit [60c1af4](https://github.com/archesproject/arches/pull/10642/commits/60c1af4d61441bf1517777dc0ba9999a1260ec52)

arches/app/datatypes/url.py
SDScandrettKint: Thanks for this function @chiatt , it works well!  I've added it to the latest commit ([60c1af4](https://github.com/archesproject/arches/pull/10642/commits/60c1af4d61441bf1517777dc0ba9999a1260ec52)) and added the test_clean function in datatype_tests.py to test that None results from both empty URL and label values.

arches/app/datatypes/url.py
jacobtylerwalls: Great. I retested and dirty state works fine now, too.

jacobtylerwalls:
After `''` is cleaned to None, the dirty state isn't as I would expect. But I can imagine futuring this to a new issue if it's not straightforward.


https://github.com/archesproject/arches/assets/38668450/66a17458-6566-4ad6-ade5-16dedd0b578b

jacobtylerwalls (approved):
Thanks, fixes look great!

arches/app/media/js/views/components/widgets/text.js
```diff
@@ -180,6 +180,10 @@ define([
 
         });
 
+        if (self.currentDefaultText() === "") {
+            self.defaultValue("");
```
bferguso, line 184: @SDScandrettKint, @jacobtylerwalls - I was looking at this as it was causing me issues (of my own making) and in look at this I was wondering if line 184 should be:
` self.defaultValue(initialDefault);`
rather than setting the value to an empty string. I think (?) that the value and defaultValues should be the i18n object unless I've got that wrong... If my assumption is correct, not sure if this might bite us somewhere down the road...

jacobtylerwalls: I was wondering along the same lines [here](https://github.com/archesproject/arches/pull/10642#issuecomment-1979332998), but I remember being less concerned about it when I saw other logic that watches the value and supplies the missing language keys (I think?)

This whole viewmodel is pretty involved, so I'm hoping we can kind of just make it to the Vue cutover without having to refactor it.

PR #10654, opened 2024-03-05
Add regression test for #9198

### LINKED ISSUE #9198: Resources that have a resource-instance reference to themselves cause a python stack overflow
**Arches Version/Branch:** stable/7.2.1

**Describe the bug and how to reproduce it:**

When I try to create a record that has a resource-instance reference to itself via the JSON-LD PUT API, the code ends up in an infinite loop which (when using runserver) causes python to stack overflow and die.  This happens in the primary descriptor function code, so if there aren't functions defined for the model, then it's okay ... but they're pretty important for the UI obviously.

My only function is the Display Name, which is configured to get the `content` of the (first) Name card.

```
Fatal Python error: _Py_CheckRecursiveCall: Cannot recover from stack overflow.
Python runtime state: initialized

Current thread 0x0000700011713000 (most recent call first):
...
  File "/Users/rs2668/Development/ENV/lib/python3.9/site-packages/django/db/models/query.py", line 1324 in _fetch_all
  File "/Users/rs2668/Development/ENV/lib/python3.9/site-packages/django/db/models/query.py", line 262 in __len__
  File "/Users/rs2668/Development/arches/arches/app/functions/primary_descriptors.py", line 44 in get_primary_descriptor_from_nodes
  **File "/Users/rs2668/Development/arches/arches/app/models/resource.py", line 97 in get_descriptor**
  File "/Users/rs2668/Development/arches/arches/app/models/resource.py", line 114 in displayname
  File "/Users/rs2668/Development/arches/arches/app/datatypes/datatypes.py", line 2162 in get_display_value
  File "/Users/rs2668/Development/arches/arches/app/functions/primary_descriptors.py", line 62 in get_primary_descriptor_from_nodes
  **File "/Users/rs2668/Development/arches/arches/app/models/resource.py", line 97 in get_descriptor**
...
```

jacobtylerwalls:
### Description of Change
The recursion error described in #9198 for 7.2 was fixed in 7.4 by de4b9d2, which isolated the resource descriptor getting from resource descriptor calculating. The resource-instance datatype calls `displayname()`, but in 7.4 that just fetches--it doesn't recalculate. Before, that call to displayname() would recalculate, but the caller was already trying to recalculate.

Added a regression test that fails on 7.2:
```py
FAIL: test_self_referring_resource_instance_descriptor (tests.models.resource_test.ResourceTests)
  File "/Users/jwalls/prj/arches/tests/models/resource_test.py", line 317, in test_self_referring_resource_instance_descriptor
    self.assertEqual(r.displayname(), "test value")
AssertionError: 'test value test value test value test value test va[2530 chars]lue ' != 'test value'
- test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value test value 
+ test value

```
### Issues Solved
Closes #9198

### Checklist

#### Ticket Background
*   Sponsored by: Yale
*   Found by: @azaroth42

#### Additional Info
We have a wide catch that was trapping RecursionError, so it was only on @azaroth42's architecture that it blew up more spectacularly, otherwise on my architecture it was failing silently, e.g. only with an INFO message:
https://github.com/archesproject/arches/blob/a1c2b429f5aaa54af95a4111862eb87d56709107/arches/app/datatypes/datatypes.py#L2315-L2316

### DIFF
--- a/tests/models/resource_test.py
+++ b/tests/models/resource_test.py
@@ -283,3 +283,75 @@ class ResourceTests(ArchesTestCase):
             q for q in queries if q['sql'].startswith('SELECT "functions_x_graphs"."id"')
         ]
         self.assertEqual(len(function_x_graph_selects), 1)
+
+    def test_self_referring_resource_instance_descriptor(self):
+        # Create a nodegroup with a string node and a resource-instance node.
+        graph = Graph.new(name="Self-referring descriptor test", is_resource=True)
+        node_group = models.NodeGroup.objects.create()
+        string_node = models.Node.objects.create(
+            graph=graph,
+            nodegroup=node_group,
+            name="String Node",
+            datatype="string",
+            istopnode=False,
+        )
+        resource_instance_node = models.Node.objects.create(
+            graph=graph,
+            nodegroup=node_group,
+            name="Resource Node",
+            datatype="resource-instance",
+            istopnode=False,
+        )
+
+        # Configure the primary descriptor to use the string node
+        models.FunctionXGraph.objects.create(
+            graph=graph,
+            function_id="60000000-0000-0000-0000-000000000001",
+            config={
+                "descriptor_types": {
+                    "name": {
+                        "nodegroup_id": str(node_group.nodegroupid),
+                        # The bug report did not have <Resource Node> in the descriptor
+                        # template, but including it here to allow the assertion to fail
+                        "string_template": "<String Node> <Resource Node>",
+                    },
+                    "map_popup": {
+                        "nodegroup_id": None,
+                        "string_template": "",
+                    },
+                    "description": {
+                        "nodegroup_id": None,
+                        "string_template": "",
+                    },
+                },
+            },
+        )
+
+        # Create a tile that references itself
+        resource = models.ResourceInstance.objects.create(graph=graph)
+        tile = models.TileModel.objects.create(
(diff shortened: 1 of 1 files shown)


apeters (approved):
this looks great!  Love having a new unit test for this.

PR #10659, opened 2024-03-06
Add raise_exception arg to @group_required #10658

### LINKED ISSUE #10658: Add `raise_exception` arg to `@group_required`
The `@group_required` decorator for permission checking will, when the user fails the test, redirect to nowhere with status 302.

This is fine for requests to views that render templates. But for views that return JSON payloads, I'm suggesting that better than a 302 redirect is a 403 permission denied code.

We found in AfS that returning 3xx instead of 4xx error codes [resulted in no error banner in knockout components](https://github.com/archesproject/arches/pull/10096#discussion_r1361310939).

Will open a PR to illustrate the idea.

jacobtylerwalls:
### Types of changes

### Description of Change
Allows raising a 403 Forbidden for API-style requests, instead of a 302 redirect to nowhere.

More info about the use case in the linked issue.

### Issues Solved
Closes #10658

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/views/workflow_history.py
+++ b/arches/app/views/workflow_history.py
@@ -1,22 +1,23 @@
 import json
 
+from django.core.exceptions import PermissionDenied
 from django.db import transaction
 from django.db.models import F, JSONField, Value
 from django.db.models.expressions import CombinedExpression
 from django.db.utils import IntegrityError
+from django.utils.decorators import method_decorator
 from django.utils.translation import gettext as _
 from django.views.generic import View
 
-from arches.app.utils.permission_backend import user_is_resource_editor
+from arches.app.utils.decorators import group_required
 from arches.app.utils.response import JSONErrorResponse, JSONResponse
 from arches.app.models import models
 
 
+@method_decorator(group_required("Resource Editor", raise_exception=True), name="dispatch")
 class WorkflowHistoryView(View):
 
     def get(self, request, workflowid):
-        if not user_is_resource_editor(request.user):
-            return JSONErrorResponse(_("Request Failed"), _("Permission Denied"), status=403)
         try:
             if request.user.is_superuser:
                 workflow_history = models.WorkflowHistory.objects.get(workflowid=workflowid)
@@ -28,9 +29,6 @@ class WorkflowHistoryView(View):
         return JSONResponse(workflow_history, status=200)
 
     def post(self, request, workflowid):
-        if not user_is_resource_editor(request.user):
-            return JSONErrorResponse(_("Request Failed"), _("Permission Denied"), status=403)
-
         data = json.loads(request.body)
         stepdata = data.get("stepdata", {})
         componentdata = data.get("componentdata", {})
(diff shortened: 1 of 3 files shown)


tests/views/workflow_tests.py
```diff
@@ -83,13 +86,6 @@ def test_get_workflow_history(self):
 
     def test_post_workflow_history(self):
         """Partial updates of componentdata and stepdata are allowed."""
-        self.client.force_login(self.anonymous)
-        response = self.client.post(reverse("workflow_history", kwargs={"workflowid": str(self.history.workflowid)}))
```
jacobtylerwalls, line 87: Noticed this wasn't testing what it I had intended (because the user wasn't even a resource editor), so I added a comment and adjusted the request so that it's an editor user and actually has some data.

jacobtylerwalls:
I borrowed the idea from [@permission_required](https://docs.djangoproject.com/en/dev/topics/auth/default/#django.contrib.auth.decorators.permission_required).

arches/app/views/workflow_history.py
```diff
@@ -60,8 +58,8 @@ def post(self, request, workflowid):
                         workflowid = workflowid,
                     )
                 else:
-                    return JSONErrorResponse(_("Request Failed"), _("Permission Denied"), status=403)
-            
+                    raise PermissionDenied
```
apeters, line 61: Will the UI be able to handle the fact that we used to return a JSON object and are now returning a python exception?

apeters (commented):
can we add the line at the top of the workflow_tests.py file that specifies the command to run this test file individually?

arches/app/views/workflow_history.py
jacobtylerwalls: Good question. I just retested. No change on the frontend because the frontend doesn't even check the response for success:

Notice no `try/catch` here:

https://github.com/archesproject/arches/blob/a1c2b429f5aaa54af95a4111862eb87d56709107/arches/app/media/js/views/components/workflows/workflow-component-abstract.js#L776

This is probably because you shouldn't be able to end up in this scenario (editing someone else's workflow) through the UI. You'd need to be trying to hack the endpoint with your own POST requests to reach this. I tested by leaving the UI open and then changing the owner of the workflow in the database.

In terms of the motivation for this change, I figured it was better to have all the permission failure cases return the same shape of a response so that we don't expose any information about what workflows exist.

apeters: what about here?
https://github.com/archesproject/arches/blob/dev/7.6.x/arches/app/media/js/viewmodels/workflow.js#L371-L394

jacobtylerwalls: Retested this one also. Same. Both 7.6 and this PR show the same in the UI:

Screenshot 2024-03-14 at 8 26 02 PM

The fact that the response changes from:

dev/7.6.x
```
{
    "message": "Permission Denied",
    "status": "false",
    "success": false,
    "title": "Request Failed"
}
```
to
```
<!doctype html>
<html lang="en">
<head>
  <title>403 Forbidden</title>
</head>
<body>
  <h1>403 Forbidden</h1><p></p>
</body>
</html>
```

doesn't seem to matter.

jacobtylerwalls: All we're using is the statusText which remains the same.

apeters (approved):
looks good to me

Issue #9769, opened 2023-07-12
ActivityStream is out of order -- Updates can't be before Creates


azaroth42:
For example:

![Screenshot 2023-07-12 at 2 00 51 PM](https://github.com/archesproject/arches/assets/871868/2aae3450-6e14-4c75-a4d8-b5abe6be96ee)

The Update to 53de8bae cannot happen before it was Created. Updates can only be after Creates.

azaroth42:
I think the fix is in `app/models/resource.py` by moving the `self.save_edit()` line  in `save()` to before the `tile.save()` line ... so would become line 228.

Issue #9768, opened 2023-07-12
ActivityStream is swamped by tiles during resource creation


azaroth42:
When creating a resource, every tile that gets created during the process spawns an entry in the activity stream at /history. With only 19 instances, my stream is already at 1200 entries, making it very expensive to consume.

Instead, there should be a Create entry that subsumes all of the subsequent tile edit Update entries.

azaroth42:
Propose that resource.save() should send a flag to the tile.save() function to prevent the creation of the edit_log entry for each tile. Tile edits don't go through resource.save(), so the updates still end up in the history.

Any downsides to this?

apeters:
Maybe the activity stream could filter out the tile create entries?

azaroth42:
Got it, thanks Cyrus! For our current usage, I don't need bulk loading or backing out transactions so I don't think we're blocked in our current hacky workaround, but it would be good to discuss the right solution that allows both to co-exist. Per Alexei's response, having a filter that could distinguish internal tile creation within the resource creation would be great...

https://github.com/archesproject/arches/blob/03b0b1d6211004fd353952a07473ed52652f00ad/arches/app/views/resource.py#L669 

But I don't know how to do that :)

apeters:
@azaroth42 Just to be clear it sounds like it's not simply tile create events that are the issue (because those could easily be filtered with `edittype != "tile create"`).  What it sounds like you want to exclude is tiles created as part of resource creation, correct?  Subsequent tile creation is essentially treated as an `update` to the resource and is a desired entry in the ActivityStream.  Is that right?

azaroth42:
Right!

If this line passed /something/ that could then be filtered on when constructing the list:

https://github.com/archesproject/arches/blob/master/arches/app/models/resource.py#L228 

but rolled together for the transaction / bulk edit usage.

"import tile create" or somesuch?

chiatt:
The 'note' column is unused. We could document the context of the tile creation there.

PR #10662, opened 2024-03-06
Update activity stream, #9768, #9769

njkim:
### Types of changes

### Description of Change

Update the active stream view not to returns "Update" (such as tile create) when the tiles are created with the resource creation.

### Issues Solved

#9768, #9769

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/models/resource.py
+++ b/arches/app/models/resource.py
@@ -228,6 +228,8 @@ class Resource(models.ResourceInstance):
         context = kwargs.pop("context", None)
         transaction_id = kwargs.pop("transaction_id", None)
         super(Resource, self).save(*args, **kwargs)
+        self.save_edit(user=user, edit_type="create", transaction_id=transaction_id)
+
         for tile in self.tiles:
             tile.resourceinstance_id = self.resourceinstanceid
             tile.save(request=request, index=False, transaction_id=transaction_id, context=context)
@@ -243,7 +245,6 @@ class Resource(models.ResourceInstance):
         except NotUserNorGroup:
             pass
 
-        self.save_edit(user=user, edit_type="create", transaction_id=transaction_id)
         if index is True:
             self.index(context)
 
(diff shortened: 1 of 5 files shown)


arches/app/views/resource.py
```diff
@@ -637,10 +637,17 @@ def get(self, request, page=None):
...
+            .exclude(note="resource creation")
+            .count()
+        )
         edits = (
-            models.EditLog.objects.all().exclude(resourceclassid=settings.SYSTEM_SETTINGS_RESOURCE_MODEL_ID).order_by("timestamp")[st:end]
+            models.EditLog.objects.all()
+            .exclude(resourceclassid=settings.SYSTEM_SETTINGS_RESOURCE_MODEL_ID)
+            .exclude(note="resource creation")
+            .order_by("timestamp")[st:end]
         )
```
jacobtylerwalls, line 651: Hey @njkim :wave:

Wondering if you considered consolidating this into a single query?

jacobtylerwalls: My bad -- didn't notice the slice at the end. Resolving.

arches/app/models/resource.py
```diff
@@ -228,9 +228,11 @@ def save(self, *args, **kwargs):
         context = kwargs.pop("context", None)
         transaction_id = kwargs.pop("transaction_id", None)
         super(Resource, self).save(*args, **kwargs)
+        self.save_edit(user=user, edit_type="create", transaction_id=transaction_id)
```
jacobtylerwalls, line 231: I think as the ticket author pointed out it's possible that moving this call to solve #9769 is not necessary if we do the filtering in #9768 as in this PR, but I still like the idea of doing it in case the another caller somewhere else in the application chooses not to filter in the same precise way.

jacobtylerwalls (approved):
+1 tested. Checked importing business data on the CLI, some AfS workflows. Looks good!

Issue #10452, opened 2023-12-17
Don't query for mvt tiles we know don't exist


chiatt:
We should cache whether a query for a vector tile has recently returned a 404.

We do this query, even if the vector tile will return a 404: https://github.com/archesproject/arches/blob/24c781a960d9348e328306e7a84e8d60cb8f5571/arches/app/views/api.py#L256-L267

We have the nodeid, so we can check the cache for a 404 even before we go to the database for 'viewable_nodegroups'.
https://github.com/archesproject/arches/blob/24c781a960d9348e328306e7a84e8d60cb8f5571/arches/app/views/api.py#L246-L255

whatisgalen:
FWIW I made this modification a while ago to optimize the cluster SQL query by first identifying any geoms within the bounding box (which is a significantly faster query):

`with connection.cursor() as cursor:
                if int(zoom) <= int(config["clusterMaxZoom"]):
                    arc = self.EARTHCIRCUM / ((1 << int(zoom)) * self.PIXELSPERTILE)
                    distance = arc * float(config["clusterDistance"])
                    min_points = int(config["clusterMinPoints"])
                    distance = settings.CLUSTER_DISTANCE_MAX if distance > settings.CLUSTER_DISTANCE_MAX else distance

                    # potential_geometries_query = """
                    #     SELECT ST_AsText(geom) AS geom
                    #     FROM geojson_geometries
                    #     WHERE nodeid = %s and resourceinstanceid not in %s
                    # """
                    # SELECT m.*,
                    #             ST_ClusterDBSCAN(geom, eps := %s, minpoints := %s) over () AS cid
                    #             FROM (
                    #                 SELECT tileid,
                    #                     resourceinstanceid,
                    #                     nodeid,
                    #                     geom
                    #                 FROM geojson_geometries
                    #                 WHERE nodeid = %s and resourceinstanceid not in %s
                    #                 AND ST_Intersects(geom, ANY(ARRAY[%s]))
                    #             ) m

                    pre_query = """
                    SELECT id FROM geojson_geometries
                    WHERE
                    ST_Intersects(geom, TileBBox(%s, %s, %s, 3857))
                    AND
                    nodeid = %s and resourceinstanceid not in %s
                    """

                    # Execute the potential_geometries_query to get the list of geometries
                    cursor.execute(pre_query, [zoom, x, y, nodeid, restricted_resource_ids])
                    search_geometry = tuple(str(row[0]) for row in cursor.fetchall())

[...]

PR #10663, opened 2024-03-07
pre-queries for geojson geoms in bounding box before cluster query re…

whatisgalen:
… #10452

### Types of changes

### Description of Change

This PR aims to do a much simpler `count_query` that, if the count is 0, skips a more complex query, caches an empty tile, and raises a 404 for the MVT.

The `count_query` first counts how many geojson geoms for the requested node actually lie within the tile bounding box. If count >= the `min_points` parameter for that node config, it proceeds to the clustering query which further narrows its query with the same subquery as the `count_query` to speed things up. If 0 < count < `min_points` it proceeds to render the unclustered mvt geometry tile.

This PR also creates a method to create a MVT cache key incorporating...
- nodeid, x, y, and zoom
- the userid (proxy for `viewable_nodegroups`)
- a snapshot of the database in the form of the editlog count for the nodegroup linked to the request node

...so that stale MVT/data won't override more recent MVT/data for that node/zoom/x/y/user.

Anecdotal benchmarks of this branch vs the target branch show up to a 30% speed improvement for rendering the largest clustered MVTs.

### Testing This PR
- To test the query-related performance enhancement, you'll want to set `tile = None` between lines 285 and 286 to temporarily disable MVT caching.
- To test the caching of 404's, you'll want to clear out your cache

### Issues Solved

#10452 

### Checklist

#### Ticket Background
*   Sponsored by: @[scholiumtech](https://github.com/scholiumtech) & Los Angeles OHR
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @whatisgalen

### DIFF
--- a/arches/app/views/api.py
+++ b/arches/app/views/api.py
@@ -295,8 +295,8 @@ class MVT(APIBase):
                     min_points = int(config["clusterMinPoints"])
                     distance = settings.CLUSTER_DISTANCE_MAX if distance > settings.CLUSTER_DISTANCE_MAX else distance
 
-                    pre_query = """
-                    SELECT id FROM geojson_geometries
+                    count_query = """
+                    SELECT count(*) FROM geojson_geometries
                     WHERE
                     ST_Intersects(geom, TileBBox(%s, %s, %s, 3857))
                     AND
--- a/arches/app/views/api.py
+++ b/arches/app/views/api.py
@@ -279,7 +279,7 @@ class MVT(APIBase):
             node = models.Node.objects.get(nodeid=nodeid, nodegroup_id__in=viewable_nodegroups)
         except models.Node.DoesNotExist:
             raise Http404()
-        search_geometries = []
+        search_geom_count = 0
         config = node.config
         cache_key = MVT.create_mvt_cache_key(node, zoom, x, y, request.user)
         tile = cache.get(cache_key)
@@ -303,12 +303,11 @@ class MVT(APIBase):
                     nodeid = %s and resourceinstanceid not in %s
                     """
 
-                    # Execute the potential_geometries_query to get the list of geometries
-                    cursor.execute(pre_query, [zoom, x, y, nodeid, resource_ids])
-                    search_geometries = tuple(str(row[0]) for row in cursor.fetchall())
+                    # get the count of matching geometries
+                    cursor.execute(count_query, [zoom, x, y, nodeid, resource_ids])
+                    search_geom_count = cursor.fetchone()[0]
 
-                    # unclear if the pre_query could be inserted into the clustering query as a subquery
-                    if len(search_geometries):
+                    if search_geom_count:
                         cursor.execute(
                             """WITH clusters(tileid, resourceinstanceid, nodeid, geom, cid)
                             AS (
--- a/arches/app/views/api.py
+++ b/arches/app/views/api.py
@@ -402,11 +402,19 @@ class MVT(APIBase):
 
     def create_mvt_cache_key(node, zoom, x, y, user, mvt_snapshot=None):
         if not mvt_snapshot:
-            mvt_snapshot = (
-                cache.get(f'editlog_nodegroupid_{str(node.nodegroup_id)}_count')
-                or
-                models.EditLog.objects.filter(nodegroupid=str(node.nodegroup_id)).count()
+            mvt_snapshot = cache.get(
+                f'editlog_nodegroupid_{str(node.nodegroup_id)}_count'
             )
+            if not mvt_snapshot:
+                mvt_snapshot = models.EditLog.objects.filter(
+                    nodegroupid=str(node.nodegroup_id)
+                ).count()
+                cache.set(
+                    f'editlog_nodegroupid_{str(node.nodegroup_id)}_count',
+                    mvt_snapshot, 
+                    settings.CACHE_BY_USER["editlog"]
+                )
+
         return f"mvt_{str(node.nodeid)}_{zoom}_{x}_{y}_{mvt_snapshot}_{user.id}"
 
 @method_decorator(csrf_exempt, name="dispatch")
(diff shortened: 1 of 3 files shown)


chiatt:
While this looks like a nice performance improvement, it doesn't actually address #10452. A solution to #10452 would cache the fact that the database query was just performed and returned a 404, so that a subsequent request would skip any database query altogether.

whatisgalen:
@chiatt we can do that, but I would strongly advise adding in some logic that embeds the state of the db into the cache key so that we avoid a case where a 404 was cached, then a new geom was created but it doesn't get included in a tile because the cache hasn't timed out yet.

For example:
```
cache_key = create_mvt_cache_key(node, zoom, x, y)
...
def create_mvt_cache_key(node, zoom, x, y, mvt_snapshot=None):
    if not mvt_snapshot:
        mvt_snapshot = int(models.EditLog.objects.filter(nodegroupid=str(node.nodegroup_id)).count())
    return f"mvt_{str(node.nodeid)}_{zoom}_{x}_{y}_{mvt_snapshot}"
```

whatisgalen:
still testing a couple alternative approaches here

arches/app/views/api.py
```diff
@@ -293,57 +294,91 @@ def get(self, request, nodeid, zoom, x, y):
...
-                        ) as tile;""",
-                        [distance, min_points, nodeid, resource_ids, nodeid, zoom, x, y, zoom, x, y],
-                    )
+                                1 AS total
+                            FROM geojson_geometries
+                            WHERE nodeid = %s and resourceinstanceid not in %s) AS tile;""",
+                            [nodeid, zoom, x, y, nodeid, resource_ids],
+                        )
+                    else:
+                        tile = ""
```
chiatt, line 380: It seems like this alone will prevent a requery for the MVT because the value saved to the cache isn't None which is what is being checked for prior to the query.

arches/app/views/api.py
```diff
@@ -279,8 +279,9 @@ def get(self, request, nodeid, zoom, x, y):
             node = models.Node.objects.get(nodeid=nodeid, nodegroup_id__in=viewable_nodegroups)
         except models.Node.DoesNotExist:
             raise Http404()
+        search_geom_count = 0
         config = node.config
-        cache_key = f"mvt_{nodeid}_{zoom}_{x}_{y}"
+        cache_key = MVT.create_mvt_cache_key(node, zoom, x, y, request.user)
```
chiatt, line 284: Caching by user is a really good idea.

arches/app/views/api.py
```diff
@@ -365,6 +400,10 @@ def get(self, request, nodeid, zoom, x, y):
             raise Http404()
         return HttpResponse(tile, content_type="application/x-protobuf")
 
+    def create_mvt_cache_key(node, zoom, x, y, user, mvt_snapshot=None):
+        if not mvt_snapshot:
+            mvt_snapshot = models.EditLog.objects.filter(nodegroupid=str(node.nodegroup_id)).count()
```
chiatt, line 405: I'm not sure if this EditLog count is needed if you are setting the value of the tile in the cache to an empty string if no geometry was found (line 380), but maybe it has a different purpose that I'm overlooking.

chiatt (commented):
It really seems like this should offer a nice performance improvement, but when I look at request times in Silk, I'm not seeing any improvement. It might be that I'm not testing it with the right data because I'm sure the scale and distribution of points makes a difference. Also, the additional query to the edit log is probably increasing the request time by a wee bit as well. 

Also, it doesn't seem consequential, but I'm getting this 500 error:

```
  File "/usr/local/lib/python3.10/dist-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.10/dist-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.10/dist-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
  File "/web_root/arches/arches/app/views/api.py", line 107, in dispatch
    return super(APIBase, self).dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/django/views/generic/base.py", line 143, in dispatch
    return handler(request, *args, **kwargs)
  File "/web_root/arches/arches/app/views/api.py", line 396, in get
    tile = bytes(cursor.fetchone()[0])
TypeError: 'NoneType' object is not subscriptable
```

chiatt:
> @chiatt we can do that, but I would strongly advise adding in some logic that embeds the state of the db into the cache key so that we avoid a case where a 404 was cached, then a new geom was created but it doesn't get included in a tile because the cache hasn't timed out yet.
> 
> For example:
> 
> ```
> cache_key = create_mvt_cache_key(node, zoom, x, y, request.user)
> ...
> 
> def create_mvt_cache_key(node, zoom, x, y, user, mvt_snapshot=None):
>     if not mvt_snapshot:
>         mvt_snapshot = models.EditLog.objects.filter(nodegroupid=str(node.nodegroup_id)).count()
>     return f"mvt_{str(node.nodeid)}_{zoom}_{x}_{y}_{mvt_snapshot}_{user.id}"
> ```

If the value is different between what is saved to the cache when a user gets a 404 vs an MVT tile, then it seems ok to return no tile even if an editor has recently created a new geometry within those bounds. That's the price of relying on a cache, and the cache timeout can always be adjusted to an admin's preference for performance over accuracy.

whatisgalen:
@chiatt the `SELECT COUNT` query should leverage Postgres indexes and caching (vs a regular `SELECT` query) so while it does represent a query, it's still performant. If we want to basically rule out creating any new queries on principal, the alternative/workaround would be to cache a count of the all editlog objects every time a new editlog instance gets created. I would be a little surprised though if removing the `SELECT COUNT` query actually sped things up by more than a hundred milliseconds.

> It really seems like this should offer a nice performance improvement, but when I look at request times in Silk, I'm not seeing any improvement. It might be that I'm not testing it with the right data because I'm sure the scale and distribution of points makes a difference. Also, the additional query to the edit log is probably increasing the request time by a wee bit as well.
> 

How many geometries does the resource layer you're testing with have? I would expect little to no improvement for lower-density resource-layers but once you get into the tens of thousands I would expect noticeably faster responses for MVT requests.

chiatt (changes requested):

chiatt (approved):

Issue #10672, opened 2024-03-11
Knockout-less plugin template that doesn't extend base-manager.htm


jacobtylerwalls:
#### Use case
When developing a Vue plugin that has no need to integrate with Knockout functionality or even the rest of the existing Arches front-end, including anything in `javascript.htm`, it would be convenient to have a template that when rendered, provides just the bare minimum:

- a mount point for a vue app
- an initial page title?
- some `<meta>` boilerplate
- primevue stylesheet
- ...?

Currently, `plugin.htm` extends `base-manager.htm` and brings in the sidebars and other goodies.

#### Benefits
- no arches nav bar or sidebar (plugin has its own distinct feel)
- clarifies/declares there is no dependency on knockout or javascript.htm
- preserves the ability for a frontend not using any other part of arches to tree-shake it all out of the build (?)

#### Implementation notes:
- add a boolean field to the Plugin model?

cc @chrabyrd @robgaston

jacobtylerwalls:
Fixed in #10678

PR #10678, opened 2024-03-13
rough out plugin-standalone logic #10672

chrabyrd:
### Types of changes

### Description of Change

This allows for plugins to be declared as "standalone", free from `knockout` and Arches chrome. Standalone plugins do not have their own js/template files and should instead be vue-only

Some things to consider:
- currently this is reading from the `config` field, should `is_standalone` be its own column?
- this could be DRY-ed up by, say, creating something like `base_base.htm` and updating `base.htm` to inherit from it
- we could include support for standalone plugins to have their own `.ico` files
- `<title>` could be updated to not have the application name, and instead just the plugin name

### Issues Solved

#10672


arches/app/views/plugin.py
```diff
@@ -33,13 +33,25 @@ def get(self, request, pluginid=None, slug=None):
             plugin = models.Plugin.objects.get(slug=slug)
         else:
             plugin = models.Plugin.objects.get(pk=pluginid)
+
         if not request.user.has_perm("view_plugin", plugin):
```
jacobtylerwalls, line 37: In a follow-up, we might want to expose this as a config option as well. If my vue app desires to handle login itself, currently, this will redirect you to the arches auth page unnecessarily.

jacobtylerwalls (approved):
Great work on this. Nothing blocking a merge from my PoV. 👍

robgaston (approved):
@chrabyrd my quick thoughts re: DRYing up the templates: I think that it makes sense to have a separate template for this and not try to inherit what is shared.  I can imagine users wanting to customize this template specifically for standalone plugins (for example, changing favicons or themes) on an individual basis, which to me implies more drift in the shared parts of this and the `base-manager.htm` template.

arches/app/templates/views/plugin-standalone.htm
```diff
@@ -0,0 +1,85 @@
...
+        {% block title %} {{ app_settings.APP_NAME }} - {{ plugin.name }} {% endblock title %}
+    </title>
+
+        <!-- Meta -->
+        <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <meta name="description" content="">
+        <meta name="author" content="">
+    
+        <link rel="shortcut icon" href="{% webpack_static 'favicons/favicon.ico' %}" type="image/x-icon" />
```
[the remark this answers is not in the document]
chrabyrd: I'd caution against this for now. We'd need to remove support for using an `.ico` file and instead just drop a plain `<i>` tag in its place. Should be a good followup to allow users to upload a custom .ico defined in the Plugin db record.

arches/app/views/plugin.py
chrabyrd: Agreed 👍

jacobtylerwalls (approved):
Great!

PR #10688, opened 2024-03-15
Relax upper bound on Django patch releases #10650

### LINKED ISSUE #10650: Consider loosening Django pin
I've set a monthly calendar reminder to check for Django security patches, which usually come out the first Monday of the month. If we loosened the upper bound on Django to include micro releases, projects wouldn't need to wait for us to update the pin and then eventually issue a patch release.

I also think we should consider loosening the upper bound on major versions of Django, now that we run the Arches test suite with DeprecationWarnings enabled. (If you run tests with DeprecationWarnings, you can still jump from LTS to LTS as we do without missing any "news" about breaking changes.) Projects should be able to use new features in Django 5.0 if they wish without waiting for us to enforce it as the floor for the whole ecosystem.

cc/ @chiatt

jacobtylerwalls:
Closes #10650

There is no 4.3 of Django. 4.2 is the LTS; 5.0 is the current supported version; 5.1 comes out in August. We can probably loosen the upper bound to <5.2.0 in August when 5.1 comes out (in Arches dev/8.0.x).

### DIFF
--- a/arches/install/requirements.txt
+++ b/arches/install/requirements.txt
@@ -1,4 +1,4 @@
-Django==4.2.11
+Django>=4.2.11,<5.0.0
 psycopg2==2.9.9
 urllib3<2
 elasticsearch>=8.3.1,<9.0.0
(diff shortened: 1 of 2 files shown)


chiatt (approved):
Looks good!

Issue #10725, opened 2024-03-29
7.6 - ETL module import error (Bulk data mangement)


rayondemiel:
App: Arches 7.6 + AFS dev/1.1.x

`File "/usr/local/lib/python3.11/site-packages/arches/app/views/etl_manager.py", line 188, in post
response = import_function(request=request)`

ETL module import problems due to arguments in function. Unable to use bulk data management plugin functions.

rayondemiel:
There's also an indentation problem in create_tile_excel_workbook() in the etl_template module (cmd).

Issue #10614, opened 2024-02-16
openpyxl upgrade causes error in BDM template downloads


chiatt:
An error occurs when downloading branch excel and tile excel templates because openpyxl doesn't have a `save_virtual_workbook` method. This method was removed in openpyxl 3.1, which is the version we are using in Arches 7.6 (Arches 7.5 is still on openpyxl 3.0). It looks like the workaround is just to use a temporary file:

https://stackoverflow.com/a/75326043

PR #10726, opened 2024-03-29
Fixing bulk manager

rayondemiel:
### Types of changes

### Description of Change

Fixing problems with the ETL manager in relation to the Bulk Data Manager plugin. Failed to import module and indentation error to export xlsx templates. Deletion of the requestion option by simply passing the argument in import_function() into [etl_manager.py](https://github.com/archesproject/arches/compare/dev/7.6.x...rayondemiel:fixing_bulk_manager?expand=1#diff-24a293a1ba386be0486ab9504603c430fe3a10e3f4e4592e587099f7373031e7)

General replacement of the save_virtual_workbook decrepated function in the Openpyxl library, following the recommendations in the documentation [link](https://openpyxl.readthedocs.io/en/stable/tutorial.html?highlight=NamedTemporaryFile#saving-as-a-stream). Can be used to update the library Openpyxl.

### Issues Solved

 #10725 -> fix indentation + module import
 #10614  -> replace deprecated function (openpyxl)

### Checklist

#### Ticket Background
*   Sponsored by: Centre de la Recherche sur la Conservation (Paris, FR)
*   Found by: @rayondemiel
*   Tested by: @rayondemiel
*   Designed by:


chiatt (changes requested):
Thanks for submitting these improvements. I've tested it and it works as expected. Could you make just a couple of minor changes that I've mentioned inline?

rayondemiel:
I've added an exception to the read function in the import_single_csv.py module. Didn't give the right error if the loop couldn't find any csv file in a zip.

arches/app/etl_modules/branch_excel_importer.py
```diff
@@ -203,8 +203,11 @@ def download(self, request):
         format = request.POST.get("format")
         if format == "xls":
             wb = create_workbook(request.POST.get("id"))
-            response = HttpResponse(save_virtual_workbook(wb), content_type="application/vnd.ms-excel")
-            response["Content-Disposition"] = "attachment"
-            return {"success": True, "raw": response}
+            with NamedTemporaryFile(suffix='.xlsx', delete=False) as tmp:
```
jacobtylerwalls, line 206: Thanks for the `delete=False` update. I think to tidy up afterward you'll also want an `os.unlink(tmp)` after the with block (and before returning), as otherwise the file will remain.

(The intent of `delete=False` is not to leave the files behind afterward, but just to ensure `wb.save()` doesn't fail on Windows as described [here](https://github.com/archesproject/arches/pull/10322#issue-2021167595).)

arches/app/etl_modules/tile_excel_importer.py
```diff
@@ -211,8 +211,11 @@ def download(self, request):
         format = request.POST.get("format")
         if format == "xls":
             wb = create_tile_excel_workbook(request.POST.get("id"))
-            response = HttpResponse(save_virtual_workbook(wb), content_type="application/vnd.ms-excel")
-            response["Content-Disposition"] = "attachment"
-            return {"success": True, "raw": response}
+            with NamedTemporaryFile(suffix='.xlsx', delete=False) as tmp:
```
jacobtylerwalls, line 214: same

arches/install/requirements.txt
```diff
@@ -24,7 +24,7 @@ requests[security]>=2.31.0
 python-slugify==7.0.0
 pillow>=7.0.0
 arcgis2geojson==2.0.0
-openpyxl==3.0.10
+openpyxl==3.1.2
```
jacobtylerwalls, line 27: Would you mind also keeping the summary of upgraded versions in arches/releases/7.6.0.md up-to-date with this change?

jacobtylerwalls (commented):
:wave: Strolled by to approve the pending github action and noticed a couple minor points. Thanks for the updates!

arches/app/etl_modules/branch_excel_importer.py
rayondemiel: Do you think it's necessary to add try/except in the event of an error to be seen in the logs?

arches/install/requirements.txt
rayondemiel: Yes, of course.

arches/app/etl_modules/branch_excel_importer.py
chiatt: I'm not sure if it's necessary to add the try block, but I don't see any harm in it. In any case, it would still be important to add an `os.unlink(tmp)`. Sorry I didn't catch that detail earlier.

rayondemiel: I'll leave it as is in that case. It can be modified if necessary. I'm the one who didn't read the issue carefully enough.

chiatt (approved):

PR #10729, opened 2024-04-02
Create Arches test runners

chrabyrd:
### Types of changes

### Description of Change

New projects are now created with the ability to run tests, and perform the basic github actions that arches-core has.

### Issues Solved

#10842


arches/install/arches-templates/tests/test_settings.py-tpl
```diff
@@ -0,0 +1,111 @@
...
+    "user_permission": {
+        "BACKEND": "django.core.cache.backends.dummy.DummyCache",
+        "LOCATION": "user_permission_cache",
+    },
+}
+
+ELASTICSEARCH_PREFIX = "test"
+
+TEST_RUNNER = "tests.base_test.ArchesTestRunner"
+
```
jacobtylerwalls, line 67: WOOHOO to having a test_settings template!

I found recently when running the AfS tests that they were colliding with my arches test database because DATABASES (where the test database name is declared) was not re-defined. This really only matters if you're using `--keepdb` to speed up consecutive test runs, although it would still be good for clarity if you're not.

Suggest we do something like https://github.com/archesproject/arches-for-science/pull/1565/commits/b5e784801d442d24933b7f12dd0827a4b8fbcd02 to provide a DATABASES setting here that uses the app name.

arches/install/arches-templates/tests/test_settings.py-tpl
```diff
@@ -0,0 +1,111 @@
...
+# could add Chrome, PhantomJS etc... here
+LOCAL_BROWSERS = []  # ['Firefox']
+
+# these are set in Travis CI
+SAUCE_USERNAME = os.environ.get("SAUCE_USERNAME")
+SAUCE_ACCESS_KEY = os.environ.get("SAUCE_ACCESS_KEY")
```
jacobtylerwalls, line 73: I removed some of this cruft in #10711. In fact, perhaps you'd like to review that one? The two PRs touch several of the same files so more visibility is 🏅

arches/install/arches-templates/tests/run_tests.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+
+
+import sys
+import os
+from django.test.utils import get_runner
+from tests import test_settings
+
+
+def run_all(argv=None):
```
jacobtylerwalls, line 26: The CI workflow in this PR just uses `manage.py test` like AfS does, so this file isn't even used. Shall we remove it? (And from core, too?)

arches/install/arches-templates/tests/base_test.py
```diff
@@ -0,0 +1,155 @@
...
+        delete_search_index()
+
+        super().teardown_databases(old_config, **kwargs)
+
+class ArchesTestCase(TestCase):
+    def __init__(self, *args, **kwargs):
+        super(ArchesTestCase, self).__init__(*args, **kwargs)
+        if settings.DEFAULT_BOUNDS is None:
+            management.call_command("migrate")
+            with open(os.path.join("tests/fixtures/system_settings/Arches_System_Settings_Model.json"), "r") as f:
```
jacobtylerwalls, line 89: This may already be on your radar, but this path won't exist if your cwd is the project.

```py
  File "/Users/jwalls/newproj/tests/base_test.py", line 94, in __init__
    with open(
         ^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'tests/fixtures/system_settings/Arches_System_Settings_Model.json'
```

We probably don't even need `ArchesTestCase`, it's just here for some screwy workarounds with the testing of oauth in core. [AfS just uses plain TestCase from Django](https://github.com/archesproject/arches-for-science/blob/ba547c16a053d7b82be64455f5c3b8673b24dff8/arches_for_science/tests.py#L4).

What do you think about removing/simplifying this part of the template?

arches/test_settings.py
```diff
@@ -0,0 +1,122 @@
...
+# recommended that this setting only be set to True in tests/settings_local.py
+# and run in specific cases at the discretion of the developer.
+LOAD_V3_DATA_DURING_TESTS = False
+
+RESOURCE_GRAPH_LOCATIONS = (os.path.join(TEST_ROOT, "fixtures", "resource_graphs"),)
+
+ONTOLOGY_FIXTURES = os.path.join(TEST_ROOT, "fixtures", "ontologies", "test_ontology")
+ONTOLOGY_PATH = os.path.join(TEST_ROOT, "fixtures", "ontologies", "cidoc_crm")
+
+BUSISNESS_DATA_FILES = (
```
jacobtylerwalls, line 48: typo? worth fixing now IMO

arches/install/arches-templates/tests/test_settings.py-tpl
```diff
@@ -0,0 +1,111 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+"""
+
+from arches.settings import *
+import arches
```
jacobtylerwalls, line 20: Some [additional settings](https://github.com/archesproject/arches-for-science/blob/dev/1.1.x/arches_for_science/test_settings.py) we found to be necessary in AfS to get urls to reverse correctly in tests:

- APP_NAME
- APP_ROOT
- ROOT_URLCONF

For example, removing ROOT_URLCONF and adding a breakpoint in the afs tests gives:
```py
(Pdb) reverse('deletesamplearea')
*** django.urls.exceptions.NoReverseMatch: Reverse for 'deletesamplearea' not found. 'deletesamplearea' is not a valid view function or pattern name.
```

jacobtylerwalls (commented):
Awesome!

Forgive me if you had some more pending changes, but I just had been thinking about some similar issues and wanted to share my notes while they would be useful.

arches/install/arches-templates/tests/run_tests.py
chrabyrd: I don't think it's cruft in Arches core as it's used in `setup.py`, but happy to pull it out of project templates 👍

jacobtylerwalls (approved):
+1 tested. Thanks for the great work on this!

PR #10439, opened 2023-12-14
Recalculate descriptors when loading business data re #9758

jacobtylerwalls:
### Types of changes

-   [x] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
Resource descriptors weren't being saved when importing business data after changes in #9758.

#### Ticket Background
*   Sponsored by: Farallon
*   Found by: @jacobtylerwalls 
*   Tested by: @jacobtylerwalls 
*   Designed by: @

PR #10743, opened 2024-04-05
Prefetch graphs and tiles when indexing resources #10453

### LINKED ISSUE #10453: Prefetch descriptor functions at graph-handling time, not resource-handling time
#10439 applied in two more places the trick from #10312 to skip queries for descriptor functions on the assumption that resources are generally grouped by graph:

https://github.com/archesproject/arches/blob/1af6ea8ccaf66da85f0c46b087ef38fcb438adc6/arches/app/utils/data_management/resources/formats/archesfile.py#L393-L395

But we could get more savings by removing this "trick" and just prefetching the descriptor functions before we start iterating the resources.

That would go something like this:

- When querying for `resources`, use a `Prefetch()` object to do the filtering for the graph's descriptor function and save it to `descriptor_function` on each Resource.
- Then remove the "trick" linked above in both files it appears

Then, with the hood popped on this (looking at callers of `save_descriptors()`), there are a few more wins to be had:
- `index_resources_by_type()` refetches graph objects because the earlier call in `index_resources()` only gets a values array
- `Tile.delete()` refetches the resource instance
- `Tile.delete()` saves the resource descriptor once per node
- `import_business_data_without_mapping()` fetches graphs one-by-one

jacobtylerwalls:
### Types of changes

### Description of Change
Save 2 queries per resource when indexing by prefetching tiles and graphs.

Save an additional query every time a resource from a different graph is encountered. Improvement on the optimization added in #10439.

### Issues Solved

Closes #10453

### Checklist

#### Ticket Background
*   Sponsored by: Farallon
*   Designed by: @jacobtylerwalls


chiatt (approved):
I tested this with 75k resources. The database was definitely less stressed and the load was about 10% to 15% faster than previous loads. Out of curiosity I changed the batch size from `batch_size // 8` to `batch_size // 2` and the load was actually a little bit slower, so I think 8 is a good default.

jacobtylerwalls:
Thanks for the review!

PR #10753, opened 2024-04-11
Replace setup.py with pyproject.toml in project template #9941

### LINKED ISSUE #9941: Package arches with pyproject.toml?
> Out of scope for this PR, but this might be an appetizing chance to move from `setup.py` to `pyproject.toml`, which is where python packaging seems to be coalescing. If in future versions we want to start shipping any other configuration, including tool configuration (some of which can _only_ be written in pyproject.toml, see how that got us into #9937 ...) we might wish we'd taken this opportunity to start new projects off on pyproject.toml from the get-go.

I'll open a new issue for this.

_Originally posted by @jacobtylerwalls in https://github.com/archesproject/arches/pull/9426#discussion_r1298505730_

jacobtylerwalls:
### Types of changes

### Description of Change
Guide new arches users toward packaging their projects with pyproject.toml rather than setup.py, which is where the python packaging ecosystem is heading.

Older projects don't necessarily need to make this change, since it's unlikely that before the advent of "arches applications" that users were distributing their arches work on PyPI for others to consume. And those that are starting on that path now are likely to have looked at AfS as a model, which already uses pyproject.toml.

### Issues Solved

Closes #9941


chrabyrd (approved):
lgtm! Is there a plan to deprecate `setup.py` in arches-core as well?

PR #10754, opened 2024-04-11
Move the i18n/ url outside of the i18n_patterns function as per the Django documentation

jkemper-pca:
Putting the code `path("i18n/", include("django.conf.urls.i18n"))` within the `i18n_patterns(*urlpatterns)` causes issues when the app tries to switch languages. Setting up the urls this way is specifically warned against [in the Django documentation](https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#django.views.i18n.set_language).

### Types of changes

### Description of Change
The "i18n/" url has been moved outside of the `i18n_patterns()` function by removing it from the urls.py in arches file (because all urls there get passed to the `i18n_patterns()` function and is added instead in the project's urls.py file after the rest of the urls are passed to the function.

### Issues Solved
When trying to switch languages using the top-right language switcher, the behaviour was inconsistent and often reloaded the page but did not switch the language.

One way that I've found this happens is:
1. Two languages enabled (ex. "en"(default), "fr")
2. Navigate to the search page
3. Switch language to French using the drop-down in the top-right
4. It should switch correctly
5. Now try switching back to English
6. It will reload the page but will stay in French

With this code change, the language will continue to swap correctly

#

### Checklist

#### Ticket Background
*   Found by: @jkemper-pca 
*   Tested by: @jkemper-pca 
*   Designed by: @jkemper-pca 

### Further comments

There might be other ways to changing the code around to take the i18n path out of the i18n_patterns so if this PR is just used to prompt looking into the issue that's fine with me.

If this change is merged, the documentation will also need to be updated slightly because the code in urls.py is used in a snippet on the [Localizing Arches](https://arches.readthedocs.io/en/stable/developing/advanced/localizing-arches/#localizing-arches) page


jacobtylerwalls (changes requested):
Thanks @jkemper-pca. Verified this fix works.

I retargeted your PR to dev/7.6.x, since master isn't a branch we use for development. (We should probably delete it.)

Since this is a change that requires user action (for existing projects), could you update 7.6.0.md and make these changes?

- add a changelog
- add an upgrade instruction (probably inserting a step after "In settings.py" so that it becomes "In urls.py")

jacobtylerwalls (approved):
Great, thanks for the update!

PR #10762, opened 2024-04-12
Improve git blames by permitting to ignore formatting commits

jacobtylerwalls:
### Types of changes

### Description of Change
The git blame experience can be cumbersome when so many formatting commits need to be skipped over. A standard tool for this involves adding a dotfile to ignore them.

1-28: the relevant subset of git log --grep black
29-754: git log --grep 'Fixing style errors.'
    
To see the benefit: run
`git config blame.ignoreRevsFile .git-blame-ignore-revs`
from the root of the repo.

The use case for pre-commit and a corresponding github action is so that this file never grows :-)


jacobtylerwalls:
We might want to backport this to make blames after branch switching easier (so that the git config isn't looking for a missing file).

whatisgalen:
Really looking forward to this PR so my name isn't always in the Blame! 🙃

apeters (approved):
no more black commits!!!

PR #10773, opened 2024-04-15
Make 5613_notification_types reversible #10772

### LINKED ISSUE #10772: 5613_notification_type can't be reversed
After commenting out the last line of 8042_3_spatialview_db_functions.py, then
```
% python manage.py migrate models 5475
```
```
django.db.utils.OperationalError: cannot ALTER TABLE "notification_types" because it has pending trigger events
```

jacobtylerwalls:
### Types of changes

### Description of Change
The table referenced here is deleted in the same migration, so deleting an individual row is not needed. Prevents error about "pending trigger events".

### Issues Solved

Closes #10772

### DIFF
--- a/arches/app/models/migrations/5613_notification_type.py
+++ b/arches/app/models/migrations/5613_notification_type.py
@@ -70,9 +70,6 @@ class Migration(migrations.Migration):
                 true
             );
             """,
-            """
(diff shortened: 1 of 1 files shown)


chiatt (approved):
Looks good to me!

PR #10777, opened 2024-04-16
Prevent language code duplication during concept import; remove some languages in new projects #10121

### LINKED ISSUE #10121: Polish i18n logic
Currently there are a few areas that deal with i18n and could use a little polish:

- When a project is created it creates several `Language` database objects. Instead it should either create no objects or just the `en` language object.
- There appears to be an `en-US` language object as well as an `en-us` language object.
- `translation.get_language()` returns a lowercase language code. We should check all call sites to ensure this logic is handled for
- Currently `PublishedGraph` objects have a foreign key to the `Language` table. It may be preferable to instead turn this into a text field so a user could select languages they'd want to update the UI with but not necessarily collect data with.

There may be other items to add to this list. If you have any questions please reach out to me or @apeters

jacobtylerwalls:
### Types of changes

### Description of Change
Unintentional differences in casing (e.g. `en-us` vs. `en-US`) from concept import ~or from Django utils~ were causing duplicate Language entries in the db.

With these changes:
- only an "en" Language object is created on setup_db
- only an "en-US" Language object is created when importing concept data that use either "en-us" or "en-US"
- ~a database check constraint is added to reject additional language codes that differ only in case~
- ~a data migration is added to de-duplicate existing language codes~
- ~lookups throughout the application use `iexact` (case-insensitive) lookups for language codes~

### Issues Solved

Closes #10121

### Checklist

#### Ticket Background
*   Found by: @chrabyrd
*   Designed by: @ 

<details>
<summary>
Test failures on target branch
</summary>

```
% python3 manage.py test -k case --settings=tests.test_settings

FAIL: test_case_insensitive_language_codes (tests.importer.concept_import_tests.ConceptImportTests.test_case_insensitive_language_codes)
  File "/Users/jwalls/prj/arches/tests/importer/concept_import_tests.py", line 77, in test_case_insensitive_language_codes
    self.assertQuerySetEqual(regions_inexact, [{"code": "en-ZA"}])
  File "/Users/jwalls/arches75/lib/python3.11/site-packages/django/test/testcases.py", line 1346, in assertQuerySetEqual
    return self.assertEqual(list(items), values, msg=msg)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Lists differ: [{'code': 'en-ZA'}, {'code': 'en-za'}] != [{'code': 'en-ZA'}]

First list contains 1 additional elements.
First extra element 1:
{'code': 'en-za'}

- [{'code': 'en-ZA'}, {'code': 'en-za'}]
+ [{'code': 'en-ZA'}]

FAIL: test_language_differs_only_in_case (tests.models.mapped_csv_import_tests.mappedCSVFileImportTests.test_language_differs_only_in_case)
  File "/Users/jwalls/prj/arches/tests/models/mapped_csv_import_tests.py", line 84, in test_language_differs_only_in_case
    self.assertEqual(new_languages, ["en-ZA"])
AssertionError: Lists differ: ['en-za', 'en-ZA', 'en-us'] != ['en-ZA']

First differing element 0:
'en-za'
'en-ZA'

First list contains 2 additional elements.
First extra element 1:
'en-ZA'

- ['en-za', 'en-ZA', 'en-us']
+ ['en-ZA']

Ran 2 tests in 12.789s

FAILED (failures=2)
```
</details>

### Alternatives considered
Considered using a case-insensitive db_collation, but that requires dropping/recreating the index, and we have foreign keys to this field that caused headaches dealing with recreating the index.


jacobtylerwalls:
@apeters I tested prefLabels a bit. The frontend has validation against submitting a second prefLabel for a language. Then the python view for concept values normalizes languages to xx-XX. So you can actually submit unlimited prefLabels for en-us, because en-us is treated as distinct language on the frontend, allows you to submit, and the backend just saves another value for en-US.

So if the data migration I added here could potentially leave folks with multiple pref labels for a given language if they had values saved in both en-us and en-US, that's a situation that was already possible.

The controlled list item labels in v8 won't allow this at the database level.

apeters:
@jacobtylerwalls it's unclear, but is this trying to standardize on a specific format for language codes?  
ie: either `xx-XX` or `xx-xx`

jacobtylerwalls:
Not per se.

I wanted to take a light touch, so I kept the existing logic in the `Concept.save()` method that normalizes xx-xx to xx-XX, but ran it only for _new_ languages.  (And then used it in the location that was causing the duplicates -- in import's `scan_for_new_languages()`.)

The idea being that if someone has a xx-xx code in their database that they put there themselves, we should leave it alone and just always use case-insensitive lookups to access language codes.

Then the rest of this PR is just to add safety going forward:
- a database constraint to prevent people from getting mixed case language codes into their db's
- a migration to deduplicate existing mixed case language codes, preferring xx-XX if there was both xx-xx and xx-XX before, but leaving everything else (e.g. xx-xx by itself) alone

If the idea is that we _should_ normalize everything to xx-XX going forward, I'm not sure how we'd do that without a database trigger. And it might be too strict. [English (World) is en-001.
](https://www.localeplanet.com/icu/en-001/index.html)

Essentially, I'm just proposing to treat this kind of like you would treat emails -- case-insensitive lookups, disallow dupes that differ only in case, but preserve the casing the user entered.

jacobtylerwalls:
@apeters I think this will need some more thought. I didn't add anything around migrating localized string tile data. Tile values keyed on a language that is deduped will appear to be missing (even though they still exist).

We could add tile value merging just like concept values are corrected here, but we'd need to define what to do when a tile has `{"en-us": "foo"; "en-US": "bar"}`. We probably don't want to block people's upgrades to 7.6 on a data validation rabbit hole.

I'm thinking I should take a step back here and just:
- remove the validating/merging out to a `manage.py validate --fix` command.
- pare the core fixes back to just:
-- new projects, removing the unnecessary languages in the .sql fixture
-- the one fix in concept import that was creating the `en-us` duplicates in the first place. (Somewhere in this PR is the one-line fix for that, but I just don't recall where at the moment. 😅 )

***

I18n_text fields besides tile data would also be potentially problematic. There is code that expects that any key in a string tile corresponds to a real language in the db, although here I'm not too worried about this one case, it just does a sort:

https://github.com/archesproject/arches/blob/0c89e7d117f4c639f1ce706f3dfaa82462fb1ca4/arches/app/etl_modules/base_data_editor.py#L72

jacobtylerwalls:
@apeters this is ready for another look. The PR no longer tries to deduplicate anything, it just:
- removes some initial languages for new projects
- removes the US region from some concept definitions
- removes the duplicate language for en-us created during concept import.

releases/7.6.0.md
```diff
@@ -40,6 +40,7 @@ Arches 7.6.0 Release Notes
 - 10911 Styling fix in resource model manage menu
 - 10726 Upgrade openpyxl package to 3.1.2 and fixes ETL modules
 - 9191 Adds unlocalized string datatype
+- 10121 Prevent language code duplication during concept import; remove some languages in new projects
```
apeters, line 43: maybe we should update this

apeters (approved):
looks good!

Issue #10781, opened 2024-04-17
Graph Designer: Deleting Associated Instances should tell you how many were deleted


whatisgalen:
This would be a nice improvement as a sanity check that you deleted the number you thought you would. Ideally it would give you this count in advance, too.

PR #10782, opened 2024-04-17
10781 Directs graph.delete_instances to call bulk_delete methods

whatisgalen:
### Types of changes

### Description of Change

This PR removes bespoke deletion logic for `Graph.delete_instances()` and instead has it call the appropriate methods from `BulkDataDeleter`. It also returns a count of deleted resources to the frontend.

### Issues Solved

#10781 

### Checklist

#### Ticket Background
*   Found by: @whatisgalen 
*   Tested by: @jacobtylerwalls & @johnatawnclementawn 
*   Designed by: @whatisgalen


jacobtylerwalls:
`delete_instances()` does the very same count, wondering if it would be nice to just pass that back up? 🤔

whatisgalen:
> `delete_instances()` does the very same count, wondering if it would be nice to just pass that back up? 🤔

Good call!

arches/app/models/graph.py
```diff
@@ -570,12 +570,15 @@ def delete_instances(self, verbose=False):
         """
         if verbose is True:
             bar = pyprind.ProgBar(Resource.objects.filter(graph_id=self.graphid).count())
+        delete_count = 0
         for resource in Resource.objects.filter(graph_id=self.graphid):
             resource.delete()
+            delete_count += 1
```
jacobtylerwalls, line 576: While here, I'd prefer to get rid of the extra query we're doing to get the count.

IMO we should only define the queryset once, get its length, and then iterate over the already evaluated queryset.

whatisgalen: I had thought that django did some wizardry regarding optimized memory management in iterating through a queryset but after looking into it, that doesn't seem to be the case (i.e. nothing distinct from lazy query evaluation). Otherwise I would have just declared the queryset as a list from the get-go (which is what I've done in the latest commits)

jacobtylerwalls: > I had thought that django did some wizardry regarding optimized memory management in iterating through a queryset but after looking into it

That's right, it will reuse the objects if you iterate an evaluated queryset, but it only works if you reuse the queryset, meaning you store a reference to it in a variable. The issue (predating your changes here of course) was that when you build it again from scratch with .objects instead, it's a new query.

If I'm missing your point, though, say so! :D

jacobtylerwalls: Oh, if your point is: why not just always convert to list?

I guess if you want to do something else like batch delete or further filtering, you can leave it as a queryset and queryset.delete(), which we can't do here without removing the progress bar.

So no harm no foul to listify it here, but there is something nice about staying a little future-proof and leaving as a queryset if you can. QoL for the next person touching the code, potentially.

jacobtylerwalls (approved):
Tested, works well. The message wasn't localized before, so we can worry about localization/pluralization at some other time.

Nice improvement. Thanks for being so responsive with the suggestions.

arches/app/models/graph.py
```diff
@@ -568,14 +568,15 @@ def delete_instances(self, verbose=False):
...
         if verbose is True:
-            bar = pyprind.ProgBar(Resource.objects.filter(graph_id=self.graphid).count())
-        for resource in Resource.objects.filter(graph_id=self.graphid):
+            bar = pyprind.ProgBar(delete_count)
+        for resource in resources_to_delete:
             resource.delete()
             if verbose is True:
                 bar.update()
-        if verbose is True:
-            print(bar)
```
jacobtylerwalls, line 578: Sorry, do you mind sharing why this was removed?

jacobtylerwalls (changes requested):

arches/app/models/graph.py
jacobtylerwalls: I'm showing that update() only prints progress, not start/finish/elapsed like print() does.

whatisgalen: sorry that was my mistake, I thought it was just an older usage of progbar. I've reverted this commit and pushed to the branch.

arches/app/etl_modules/bulk_data_deletion.py
```diff
@@ -141,7 +142,10 @@ def delete_resources(self, userid, loadid, graphid, resourceids):
                 resources = Resource.objects.filter(pk__in=resourceids)
             for resource in resources.iterator(chunk_size=2000):
                 resource.delete(user=user, index=False, transaction_id=loadid)
+                deleted_count += 1
             result["success"] = True
+            result["deleted_count"] = deleted_count
+            result["message"] = _(f"Successfully deleted {deleted_count} resources")
```
jacobtylerwalls, line 160: I'm pretty certain this won't work -- by the time gettext is called, the variable will have already been interpolated and so it won't match any translations. I think the pattern just below this works fine.

arches/app/models/graph.py
```diff
@@ -563,22 +563,18 @@ def delete(self):
                 )
             )
 
-    def delete_instances(self, verbose=False):
+    def delete_instances(self, verbose=False, request=None):
```
jacobtylerwalls: Why is this an optional argument if we use it unconditionally? Can we just make the user id the argument?

arches/app/models/graph.py
```diff
@@ -563,22 +563,18 @@ def delete(self):
...
-        if verbose is True:
-            bar = pyprind.ProgBar(delete_count)
-        for resource in resources_to_delete:
-            resource.delete()
-            if verbose is True:
-                bar.update()
-        if verbose is True:
-            print(bar)
-        return delete_count
+        from arches.app.etl_modules.bulk_data_deletion import BulkDataDeletion
```
jacobtylerwalls, line 571: Does it cause a circular import if you move this to the top of the file?

jacobtylerwalls (changes requested):
Noticed a few things in the latest. Now that the scope is changing, I'd add a changelog and update the PR title. Thanks.

arches/app/etl_modules/bulk_data_deletion.py
```diff
@@ -129,19 +130,34 @@ def get_sample_data(self, nodegroup_id, resourceids):
...
 
-    def delete_resources(self, userid, loadid, graphid, resourceids):
+    def delete_resources(self, userid, loadid, graphid, resourceids, verbose=False):
         result = {"success": False}
-        user = User.objects.get(id=userid)
+        deleted_count = 0
+        try:
+            user = User.objects.get(id=userid)
+        except ValueError:
+            user = None
```
jacobtylerwalls, line 139: I would have expected User.DoesNotExist, not ValueError. Did you see this during development?

arches/app/models/graph.py
```diff
@@ -563,19 +563,18 @@ def delete(self):
                 )
             )
 
-    def delete_instances(self, verbose=False):
+    def delete_instances(self, userid=None, verbose=False):
```
jacobtylerwalls, line 566: I make a habit of adding new optional arguments to the end of signatures to improve backwards compatibility.

arches/app/etl_modules/bulk_data_deletion.py
```diff
@@ -129,19 +130,34 @@ def get_sample_data(self, nodegroup_id, resourceids):
...
         return sample_data[0:5]
 
-    def delete_resources(self, userid, loadid, graphid, resourceids):
+    def delete_resources(self, userid, loadid, graphid, resourceids, verbose=False):
         result = {"success": False}
-        user = User.objects.get(id=userid)
+        deleted_count = 0
+        try:
+            user = User.objects.get(id=userid)
+        except ValueError:
```
chiatt, line 138: If just confirming whether userid is None, it would be more explicit to avoid a try block and instead do a None check. Something like:
```
user = User.objects.get(id=userid) if userid else None
```

arches/app/etl_modules/bulk_data_deletion.py
```diff
@@ -129,19 +130,34 @@ def get_sample_data(self, nodegroup_id, resourceids):
...
                 resources = Resource.objects.filter(pk__in=resourceids)
+            if verbose is True:
+                deleted_count = resources.count()
+                bar = pyprind.ProgBar(deleted_count)
             for resource in resources.iterator(chunk_size=2000):
                 resource.delete(user=user, index=False, transaction_id=loadid)
+                if verbose is True:
+                    bar.update()
+                else:
+                    deleted_count += 1
```
chiatt, line 155: I'm wondering if it's really necessary to track the deleted_count. Couldn't resources.count() be used whether or not verbose is True?

arches/app/etl_modules/bulk_data_deletion.py
jacobtylerwalls: Would also save a query. But looks like the proxy model expects `{}` instead of None.

arches/app/etl_modules/bulk_data_deletion.py
whatisgalen: if it's passed `None` it will raise a `ValueError`, otherwise if wold raise `DoesNotExist`

arches/app/models/graph.py
whatisgalen: If they're both kwargs does order still matter?

arches/app/etl_modules/bulk_data_deletion.py
whatisgalen: fix in latest

arches/app/etl_modules/bulk_data_deletion.py
whatisgalen: fix in latest

arches/app/models/graph.py
whatisgalen: fix in latest

arches/app/models/graph.py
jacobtylerwalls: Yeah, unless they're keyword-only with `*,` to consume the positional args, which we don't have here. I floated starting to do this and got feedback that we don't do it anywhere else in Arches so not to bother with it.

arches/app/etl_modules/bulk_data_deletion.py
```diff
@@ -129,19 +130,32 @@ def get_sample_data(self, nodegroup_id, resourceids):
 
         return sample_data[0:5]
 
-    def delete_resources(self, userid, loadid, graphid, resourceids):
+    def delete_resources(self, userid, loadid, graphid, resourceids, verbose=False):
         result = {"success": False}
-        user = User.objects.get(id=userid)
+        deleted_count = 0
+        user = User.objects.get(id=userid) if userid else None
```
jacobtylerwalls, line 136: Passing None instead of `{}` seems to trigger the delete permissions checks:

https://github.com/archesproject/arches/blob/0c89e7d117f4c639f1ce706f3dfaa82462fb1ca4/arches/app/models/resource.py#L502-L508

Is that desirable here?

arches/app/models/graph.py
```diff
@@ -563,19 +564,18 @@ def delete(self):
...
-        for resource in Resource.objects.filter(graph_id=self.graphid):
-            resource.delete()
-            if verbose is True:
-                bar.update()
-        if verbose is True:
-            print(bar)
+
+        bulk_deleter = BulkDataDeletion()
+        loadid = uuid.uuid4()
+        resp = bulk_deleter.delete_resources(userid, loadid, self.graphid, verbose=verbose)
```
johnatawnclementawn, line 575: When invoked from the command line using something like `python manage.py resources remove_resources -g graphid-here`, the resourceids argument is not provided and prevents users from deleting resources for a given graph. This is not an issue if you do not provide a graphid and delete all resources wholesale.

johnatawnclementawn: This also appears to be an issue with the `index_resource_deletion` method below

arches/app/etl_modules/bulk_data_deletion.py
whatisgalen: I'm not sure why we're doing that particular check. I traced it to [this commit](https://github.com/archesproject/arches/commit/2e91ec8a8779732819e79aac65b9d13a27cbe04d) by @chiatt. Maybe it makes sense to set the default value of `user=None` and just check against that? But that's starting to sound like a separate PR.

arches/app/models/graph.py
whatisgalen: > When invoked from the command line using something like `python manage.py resources remove_resources -g graphid-here`, the resourceids argument is not provided and prevents users from deleting resources for a given graph. This is not an issue if you do not provide a graphid and delete all resources wholesale.

@johnatawnclementawn I'm not sure I understand. In the case you mention `remove_resources(graphid)` -> `graph.delete_instances()` -> `delete_resources(self.graphid)` and deletes all resources of that graph. Or perhaps I'm misreading your comment?

whatisgalen:
apparently this PR loses ~2% of coverage. I find that somewhat hard to believe, unless it's actually just exposing the pre-existing lack of coverage around `bulk_delete` (which sounds more plausible to me). Was there a consensus on what the resolution strategy would be for this kind of case? I.e. was there a determination that the onus would be on new PR writers given whatever code they happen to touch? Or is there some kind of concerted effort to add genuinely novel coverage which I'm unaware of?

arches/app/etl_modules/bulk_data_deletion.py
jacobtylerwalls: Yes that would sound like a separate PR, so I guess what I'm getting at is unless/until we do that, shouldn't we pass `{}` instead of `None`?

jacobtylerwalls:
I ran the coverage report with:
```py
coverage run manage.py test tests --settings=tests.test_settings
coverage report
coverage html
```

You're right that all three python files you touched lacked coverage. And on net you added code (+16 lines), and so you only just barely degraded overall coverage.

> Was there a consensus on what the resolution strategy would be for this kind of case? I.e. was there a determination that the onus would be on new PR writers given whatever code they happen to touch? 

I don't think there's been that kind of discussion because the check is still extremely lenient. I'd prefer that all of a diff is tested, not just by some percentage that happens to not degrade the whole. The PR checklist has for some time asked authors to add "tests that prove my fix is effective or that my feature works", and I recently added another checkbox "my test fails on the target branch", which is what I'm in the habit of checking manually when reviewing pull requests.

So I don't think there's really a "policy change" here, only rather a push to use CI to get back to the practice of being serious about what our documented practices. (And a unit test for this looks pretty trivial to write IMO.)

arches/app/models/graph.py
johnatawnclementawn: Sorry for the confusion @whatisgalen - I'm just seeing an issue of backwards compatibility with the `remove_resources` [CLI command](https://arches.readthedocs.io/en/stable/developing/reference/command-line-reference/#other-data-management-commands), specifically when providing a graphid. Here's the traceback I see when I run that command with a specified graphid:

`All resources associated with the 'Concept' Resource Model will be removed. continue? [Y/n] y
  File "/web_root/arches_rdm_example_project/manage.py", line 30, in <module>
    execute_from_command_line(sys.argv)
  File "/web_root/ENV/lib/python3.11/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/web_root/ENV/lib/python3.11/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/web_root/ENV/lib/python3.11/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/web_root/ENV/lib/python3.11/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/web_root/arches/arches/management/commands/resources.py", line 49, in handle
    self.remove_resources(force=options["yes"], graphid=options["graph"])
  File "/web_root/arches/arches/management/commands/resources.py", line 71, in remove_resources
    graph.delete_instances(verbose=True)
  File "/web_root/arches/arches/app/models/graph.py", line 575, in delete_instances
    resp = bulk_deleter.delete_resources(userid, loadid, self.graphid, verbose=verbose)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: BulkDataDeletion.delete_resources() missing 1 required positional argument: 'resourceids'`

arches/app/models/graph.py
```diff
@@ -563,19 +564,18 @@ def delete(self):
                 )
             )
 
-    def delete_instances(self, verbose=False):
+    def delete_instances(self, userid={}, verbose=False):
```
jacobtylerwalls, line 567: Seems like None would be better here. I was suggesting instead that the `else None` in `delete_resources` could be `else {}` until we do the follow-up work.

arches/app/models/graph.py
whatisgalen: @johnatawnclementawn  That's perfect, thanks for providing the clarification and stack trace. I've included a dummy value of `None` for that pos arg `resourceids` so let me know if you're still seeing an issue testing this. Thanks!

arches/app/models/graph.py
whatisgalen: fix in latest

arches/app/etl_modules/bulk_data_deletion.py
whatisgalen: @jacobtylerwalls just added #10844 - feel free to add

whatisgalen: fix in latest

arches/app/models/graph.py
```diff
@@ -563,19 +564,18 @@ def delete(self):
...
-        for resource in Resource.objects.filter(graph_id=self.graphid):
-            resource.delete()
-            if verbose is True:
-                bar.update()
-        if verbose is True:
-            print(bar)
+
+        bulk_deleter = BulkDataDeletion()
+        loadid = uuid.uuid4()
+        resp = bulk_deleter.delete_resources(userid, loadid, self.graphid, None, verbose=verbose)
```
jacobtylerwalls, line 575: Passing `None` by kwarg would be nice.

tests/bulkdata/bulk_data_delete_tests.py
```diff
@@ -0,0 +1,73 @@
+from django.contrib.auth.models import User
+from arches.app.models.resource import Resource
+from arches.app.models.tile import Tile
+from arches.app.models import models
+from arches.app.utils.data_management.resource_graphs.importer import import_graph as ResourceGraphImporter
+from arches.app.utils.betterJSONSerializer import JSONSerializer, JSONDeserializer
+from yourapp.models import BulkDataDeletion  # Adjust import as necessary
```
jacobtylerwalls, line 7: hah, did an LLM write this? 😉

tests/bulkdata/bulk_data_delete_tests.py
```diff
@@ -0,0 +1,73 @@
...
+from tests.arches_test_case import ArchesTestCase
+import os
+import uuid
```
jacobtylerwalls, line 10: We don't have a hook for isort, but I usually put stdlib imports first.

tests/bulkdata/bulk_data_delete_tests.py
```diff
@@ -0,0 +1,73 @@
...
+# these tests can be run from the command line via
+# python manage.py test tests.bulkdata.bulk_data_delete_tests --settings="tests.test_settings"
+
+class BulkDataDeletionTests(ArchesTestCase):
+    def setUpClass(self):
+        super().setUpClass()
```
jacobtylerwalls, line 16: I prefer [setUpTestData](https://docs.djangoproject.com/en/5.0/topics/testing/tools/#django.test.TestCase.setUpTestData)() for setting up test data, rather than setUpClass(), which is more for overriding settings, dealing with a selenium client, etc.

jacobtylerwalls, line 16: Either way, it needs to be a classmethod
```py
TypeError: BulkDataDeletionTests.setUpClass() missing 1 required positional argument: 'self'
```

tests/bulkdata/bulk_data_delete_tests.py
```diff
@@ -0,0 +1,73 @@
...
+
+        with open(os.path.join("tests/fixtures/resource_graphs/Search Test Model.json"), "r") as f:
+            archesfile = JSONDeserializer().deserialize(f)
+        ResourceGraphImporter(archesfile["graph"])
+        self.search_model_graphid = "d291a445-fa5f-11e6-afa8-14109fd34195"
+        self.search_model_name_nodeid = "2fe14de3-fa61-11e6-897b-14109fd34195"
+        # Further setup like creating resources and tiles that the test methods will interact with.
+
+    def test_get_number_of_deletions_with_no_resourceids(self):
+        # Test to ensure deletion count is calculated correctly.
```
jacobtylerwalls, line 29: I usually delete comments like this if they paraphrase the test title, or at the very least I strike "test that..."

tests/bulkdata/bulk_data_delete_tests.py
```diff
@@ -0,0 +1,73 @@
...
+
+class BulkDataDeletionTests(ArchesTestCase):
+    def setUpClass(self):
+        super().setUpClass()
+        # Setup test data here. You might want to create users, resources, tiles, etc.
+        self.user = User.objects.create(username='testuser', password='securepassword')
+        self.bulk_deleter = BulkDataDeletion()
+
+        with open(os.path.join("tests/fixtures/resource_graphs/Search Test Model.json"), "r") as f:
+            archesfile = JSONDeserializer().deserialize(f)
```
jacobtylerwalls, line 22: Did you run this test locally in isolation?

This failed for me without:
```py
LanguageSynchronizer.synchronize_settings_with_db()
```

tests/bulkdata/bulk_data_delete_tests.py
```diff
@@ -0,0 +1,73 @@
...
+        self.search_model_name_nodeid = "2fe14de3-fa61-11e6-897b-14109fd34195"
+        # Further setup like creating resources and tiles that the test methods will interact with.
+
+    def test_get_number_of_deletions_with_no_resourceids(self):
+        # Test to ensure deletion count is calculated correctly.
+        loadid = str(uuid.uuid4())
+        resourceids = create_test_resources_and_tiles(self.search_model_graphid, self.search_model_name_nodeid, loadid)
+        number_of_resource, number_of_tiles = self.bulk_deleter.get_number_of_deletions(self.search_model_graphid, self.search_model_name_nodeid, None)
+        
+        self.assertIsInstance(number_of_resource, list)
```
jacobtylerwalls, line 34: ?

jacobtylerwalls (changes requested):
Will take another look once the tests are passing.

tests/bulkdata/bulk_data_delete_tests.py
```diff
@@ -0,0 +1,73 @@
...
+        self.assertEqual(number_of_tiles, len(resourceids))
+
+    def test_delete_resources(self):
+        # Test to ensure resources are deleted properly.
+        loadid = str(uuid.uuid4())
+        resourceids = create_test_resources_and_tiles(self.search_model_graphid, self.search_model_name_nodeid, loadid)
+        result = self.bulk_deleter.delete_resources(self.user.id, loadid, self.search_model_graphid, resourceids, verbose=True)
```
jacobtylerwalls, line 41: Consider testing with verbose=False so that you don't have to capture the output with `with captured_stdout()` to prevent test clutter. (Testing `pyprind` itself falls under "don't test the framework" IMO).

arches/app/models/graph.py
whatisgalen: fix in latest

tests/bulkdata/bulk_data_delete_tests.py
whatisgalen: "Adjust[ing] import as necessary" 😄

tests/bulkdata/bulk_data_delete_tests.py
whatisgalen: fix in latest

tests/bulkdata/bulk_data_delete_tests.py
whatisgalen: fix in latest

tests/bulkdata/bulk_data_delete_tests.py
whatisgalen: fix in latest

arches/app/models/graph.py
whatisgalen: in my humblest of opinions on kwarg order, userid makes more sense as an earlier argument since it's more important. I'm not married to it but would be great to merge this sooner rather than later

tests/bulkdata/bulk_data_delete_tests.py
whatisgalen: fix in latest

tests/bulkdata/bulk_data_delete_tests.py
whatisgalen: fix in latest

jacobtylerwalls (changes requested):
Thanks for the test cleanup, they look great. ⭐ 

Heads up, I pushed two cleanup commits to your branch including the release note I mentioned, so you'll wanna pull.

Noticed a failure when performing a manual test:

1. Graph Designer > Delete Associated Instances
```py
2024-05-02 10:49:26   File "/web_root/arches/arches/app/views/graph.py", line 472, in delete
2024-05-02 10:49:26     resp = graph.delete_instances(userid=request.user.id)
2024-05-02 10:49:26   File "/web_root/arches/arches/app/models/graph.py", line 576, in delete_instances
2024-05-02 10:49:26     bulk_deleter.index_resource_deletion(loadid)
2024-05-02 10:49:26 TypeError: BulkDataDeletion.index_resource_deletion() missing 1 required positional argument: 'resourceids'
```

whatisgalen:
> Thanks for the test cleanup, they look great. ⭐
> 
> Heads up, I pushed two cleanup commits to your branch including the release note I mentioned, so you'll wanna pull.
> 
> Noticed a failure when performing a manual test:
> 
> 1. Graph Designer > Delete Associated Instances
> 
> ```python
> 2024-05-02 10:49:26   File "/web_root/arches/arches/app/views/graph.py", line 472, in delete
> 2024-05-02 10:49:26     resp = graph.delete_instances(userid=request.user.id)
> 2024-05-02 10:49:26   File "/web_root/arches/arches/app/models/graph.py", line 576, in delete_instances
> 2024-05-02 10:49:26     bulk_deleter.index_resource_deletion(loadid)
> 2024-05-02 10:49:26 TypeError: BulkDataDeletion.index_resource_deletion() missing 1 required positional argument: 'resourceids'
> ```

I neglected to refactor the signature for the bulk index method so that `resourceids` would be a kwarg defaulting to None. I think wherever we're checking (like [here](https://github.com/archesproject/arches/blob/dev/7.6.x/arches/app/etl_modules/bulk_data_deletion.py#L171-L172)) for whether an arg `is None` or is at least Falsey, my first inclination is to change it from positional arg to kwarg. That change is in my latest commit.

arches/app/models/graph.py
```diff
@@ -563,19 +564,18 @@ def delete(self):
...
-            resource.delete()
-            if verbose is True:
-                bar.update()
-        if verbose is True:
-            print(bar)
+
+        bulk_deleter = BulkDataDeletion()
+        loadid = uuid.uuid4()
+        resp = bulk_deleter.delete_resources(userid, loadid, graphid=self.graphid, verbose=verbose)
+        bulk_deleter.index_resource_deletion(loadid)
```
jacobtylerwalls, line 576: This will drive more traffic into #10183, but that's not something that needs to block this work. Just FYI.

jacobtylerwalls:
Taking a final look

arches/app/etl_modules/bulk_data_deletion.py
```diff
@@ -129,19 +130,34 @@ def get_sample_data(self, nodegroup_id, resourceids):
...
+    def delete_resources(self, userid, loadid, graphid=None, resourceids=None, verbose=False):
         result = {"success": False}
-        user = User.objects.get(id=userid)
+        deleted_count = 0
+        user = User.objects.get(id=userid) if userid else {}
         try:
-            if resourceids and graphid:
-                resources = Resource.objects.filter(graph_id=graphid).filter(pk__in=resourceids)
+            if resourceids:
+                resources = Resource.objects.filter(pk__in=resourceids)
```
jacobtylerwalls, line 139: I think this slightly degrades performance if both resources and graph are provided, so I don't see the readability tradeoff as worth it, but I don't see a current code path that exercises it, so I'm not too bothered.

jacobtylerwalls (approved):
Thanks for the updates!

arches/app/models/graph.py
jacobtylerwalls: Finally got around to the forum post: https://community.archesproject.org/t/preferring-keyword-only-parameters-when-adding-functionality-to-existing-functions/2538

Issue #10780, opened 2024-04-17
Feature: Guess Node Mappings in import-single-csv etl module


whatisgalen:
See video:

https://github.com/archesproject/arches/assets/8366985/82d2d870-ab9a-4e5f-b105-7565d32bea5b

PR #10783, opened 2024-04-17
10780 guess node mappings

whatisgalen:
### Types of changes

### Description of Change

https://github.com/archesproject/arches/assets/8366985/9f0191ca-a306-4408-a20d-99b3292b4892

### Issues Solved

#10780 

### Checklist

#### Ticket Background
*   Found by: @whatisgalen 
*   Tested by: @ 
*   Designed by: @whatisgalen


njkim (commented):
@whatisgalen It looks like a nice addition to the importer and it works great.
But it is little bit concerning that this is deprecated quite a while ago.
Do you think if there are any alternatives or other way to implement this?

njkim (commented):
@whatisgalen
Maybe my local arches is not properly set up for this but I am getting an error saying.
> TypeError: FuzzySet is not a function l.66

Some other minor suggestions are:
1. Can we consider something more authoritative than "Guess Mappable Nodes" (which is a little vague too)? Something like "Match Columns to Nodes" (nodes to columns? :) )?

2. Can we move the button to the right side of `the column names in the first row` so that the table can show more contents?

3. When the column names are not in the csv, the table automatically assigns the integer numbers to the column. This does not break the ui but there is an error when trying to convert to column name to the lowercase. One option is to hide button (or disable) when there are not column names in the csv, ie. `the column names in the first row` is off.

4. I am not sure if we need to add "fuzzyset" to javascipt.htm?

chiatt:
> @whatisgalen Maybe my local arches is not properly set up for this but I am getting an error saying.
> 
> > TypeError: FuzzySet is not a function l.66
> 
> Some other minor suggestions are:
> 
> 1. Can we consider something more authoritative than "Guess Mappable Nodes" (which is a little vague too)? Something like "Match Columns to Nodes" (nodes to columns? :) )?
> 2. Can we move the button to the right side of `the column names in the first row` so that the table can show more contents?
> 3. When the column names are not in the csv, the table automatically assigns the integer numbers to the column. This does not break the ui but there is an error when trying to convert to column name to the lowercase. One option is to hide button (or disable) when there are not column names in the csv, ie. `the column names in the first row` is off.
> 4. I am not sure if we need to add "fuzzyset" to javascipt.htm?

I agree with @njkim, 'Match Columns to Nodes' (or nodes to columns) is more explicit and although 'map' makes sense for more technical folks, 'match' would probably be clearer for most users.

whatisgalen:
@njkim @chiatt all requested changes pushed in latest

njkim (approved):
It works great.

Issue #10787, opened 2024-04-17
Search Export - export concept values as uuid


whatisgalen:
Concept preflabels are not unique across thesauri, and additionally for the use case of exporting data from one Arches instance (via search export) and importing it into another (via ETL modules) isn't really feasible with preflabels vs value uuids

PR #10789, opened 2024-04-18
10787 search export as system values

whatisgalen:
### Types of changes

### Description of Change

This PR enables exported search results to be rendered as their "system values", i.e. the node value comes from 
 `datatype.transform_export_values()` instead of `datatype.get_display_value()`. For exmaple concepts would render as their uuids instead of preflabels. It has a minor addition to the UI as well.

Screenshot 2024-04-17 at 6 30 47 PM

### Issues Solved

#10787 

### Checklist

#### Ticket Background
*   Found by: @whatisgalen 
*   Tested by: @chiatt 
*   Designed by: @whatisgalen


arches/app/search/search_export.py
```diff
@@ -50,6 +51,12 @@ def __init__(self, search_request=None):
         search_request.GET["export"] = True
         self.report_link = search_request.GET.get("reportlink", False)
         self.format = search_request.GET.get("format", "tilecsv")
+        self.concept_uuid_format = search_request.GET.get("conceptuuidformat", False)
+        if self.concept_uuid_format:
+            try:
+                self.concept_uuid_format = json.loads(self.concept_uuid_format)
+            except:
+                pass
```
chiatt, line 59: It would be better to avoid calling `json.loads` here because it's not actually json that is being loaded. Better to be explicit and do something like:

```
        use_system_value = search_request.GET.get("exportsystemvalue", False)
        self.export_system_value = False if use_system_value == 'false' else use_system_value
```

arches/app/search/search_export.py
```diff
@@ -314,7 +321,10 @@ def flatten_tiles(self, tiles, datatype_factory, compact=True, use_fieldname=Fal
                 node = self.get_node(nodeid)
                 if node.exportable:
                     datatype = datatype_factory.get_instance(node.datatype)
-                    node_value = datatype.get_display_value(tile, node)
+                    if 'concept' in node.datatype and self.concept_uuid_format:
+                        node_value = datatype.transform_export_values(value, concept_export_value_type='id')
+                    else:
+                        node_value = datatype.get_display_value(tile, node)
```
chiatt, line 327: This could just be: 

```
if self.export_system_value:
    node_value = datatype.transform_export_values(value, **{"concept_export_value_type": "id"})
else:
    node_value = datatype.get_display_value(tile, node)
```

chiatt (changes requested):
This is a good start, but would be much better concept specific but rather apply to all datatypes. That would eliminate the need for datatype checking and allow for future and custom datatypes to control how data is exported. 

The boolean variable could be 'Export System Value' and then in `search_export` the boolean would just determine whether 'get_display_value' or 'transform_export_values' is called.

arches/app/search/search_export.py
whatisgalen: It's not actually json but `json.loads()` still works in translating `'false' -> False` and `'true' -> True`. Explicitly checking the string value means we would have to do:
```
if user_system_value == 'false':
  user_system_value = False
elif user_system_value == 'true':
  user_system_value = True
```
... which feels a bit clunky

arches/app/search/search_export.py
whatisgalen: fix in latest

whatisgalen:
> This is a good start, but would be much better concept specific but rather apply to all datatypes. That would eliminate the need for datatype checking and allow for future and custom datatypes to control how data is exported.
> 
> The boolean variable could be 'Export System Value' and then in `search_export` the boolean would just determine whether 'get_display_value' or 'transform_export_values' is called.

fix in latest

arches/app/search/search_export.py
jacobtylerwalls: Agree with @chiatt that passing non-json to json.loads() increases comprehension burden.

What's clunky about this:
```py
new_var = (old_var == "true")
```

jacobtylerwalls: Alternatively, we have a `str_to_bool()` util for this purpose.

whatisgalen: fix in latest

whatisgalen:
FWIW coverage "failed" because of this delta:
`Coverage decreased from 65.89301327398458% to 65.87992212023681%`

tests/search/search_export_tests.py
```diff
@@ -132,7 +199,6 @@ def test_login_via_basic_auth_rate_limited(self):
         request = RequestFactory().get(
             reverse("api_export_results"),
             HTTP_AUTHORIZATION=auth_string,
-            # In reality this would be added by django_ratelimit.
```
jacobtylerwalls, line 135: Wondering if deleting this comment is intentional?

tests/search/search_export_tests.py
```diff
@@ -36,39 +46,54 @@ def setUpTestData(cls):
...
+        ret = skos.save_concepts_from_skos(rdf)
+        cls.valueid = "dadaeee5-57ef-409d-a6cf-98d332fdada8"
+        cultural_period_tile = Tile(
+            data={cls.search_model_cultural_period_nodeid: cls.valueid},
+            nodegroup_id=cls.search_model_cultural_period_nodeid,
+            resourceinstance_id=cls.test_resourceinstanceid,
+        )
+        cultural_period_tile.save(index=False)
+        cultural_period_tile.index()
+        sync_es(se)
```
jacobtylerwalls, line 90: Nice. I'm wondering if this belongs in setUp() so it runs before each test instead of once per class.

tests/search/search_export_tests.py
whatisgalen: whoops -- I had (wrongly) assumed that was an LLM artifact. Just restored it

tests/search/search_export_tests.py
whatisgalen: do you mean overriding setUp in the SearchExportTests class or in the base class method?

jacobtylerwalls: In SearchExportTests

jacobtylerwalls: Sorry Galen, I meant the regular instance method, not a classmethod. I'm not sure a classmethod like this is ever called?

whatisgalen: I just pushed the change you mentioned. I'm fairly certain the classmethod was being called as tests pass with it as a classmethod or instancemethod but fail without either.

arches/app/search/search_export.py
```diff
@@ -391,7 +394,12 @@ def flatten_tiles(self, tiles, datatype_factory, compact=True, use_fieldname=Fal
                 node = self.get_node(nodeid)
                 if node.exportable:
                     datatype = datatype_factory.get_instance(node.datatype)
-                    node_value = datatype.get_display_value(tile, node)
+                    if self.export_system_values:
+                        node_value = datatype.transform_export_values(
+                            value, **{"concept_export_value_type": "id"}
```
chiatt, line 399: Just a note - 'concept_export_value_type' is used here because that's the argument in both the domain and concept datatypes. Might be nice to use something more generic in the controlled list manager.

chiatt (approved):
Looks good to me!

PR #10792, opened 2024-04-18
Fix webpack failure for project names containing "js" #10791

### LINKED ISSUE #10791: webpack doesn't build if project name contains the string "js"
In setting up a new project to test #10642 this morning I was having trouble setting up a new project.

Traced it to my project name "prjsevenfive" containing the substring "js", which violates an assumption here:

https://github.com/archesproject/arches/blob/0c89e7d117f4c639f1ce706f3dfaa82462fb1ca4/arches/webpack/webpack-utils/build-javascript-filepath-lookup.js#L20

Sneaky one! 😉

jacobtylerwalls:
### Types of changes

### Description of Change
Prevents webpack failure seen when project name contains the substring "js"

### Issues Solved
Closes #10791

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @ 

### Further comments
To test:
```
arches-project create projs
cd projs
yarn build_development // with a django dev server running
```

### DIFF
--- a/arches/install/arches-templates/project_name/webpack/webpack-utils/build-css-filepath-lookup.js
+++ b/arches/install/arches-templates/project_name/webpack/webpack-utils/build-css-filepath-lookup.js
@@ -17,7 +17,7 @@ function buildCSSFilepathLookup(path, outerAcc, cssDirectoryPath) {
             );
         }
         else {
-            let subPath = Path.join(path, name).split(/css(.*)/s)[1];  // splits only on first occurance
+            let subPath = Path.join(path, name).split(/css(.*)/s)[1];  // splits only on first occurrence
             subPath = subPath.substring(1);
             const parsedPath = Path.parse(subPath);
 
--- a/arches/install/arches-templates/project_name/webpack/webpack-utils/build-css-filepath-lookup.js
+++ b/arches/install/arches-templates/project_name/webpack/webpack-utils/build-css-filepath-lookup.js
@@ -17,7 +17,8 @@ function buildCSSFilepathLookup(path, outerAcc, cssDirectoryPath) {
             );
         }
         else {
-            let subPath = Path.join(path, name).split(/\/css(.*)/s)[1];  // splits only on first occurrence
+            const regex = new RegExp(`${Path.sep}css(.*)`, 's');
+            let subPath = Path.join(path, name).split(regex)[1];  // splits only on first occurrence
             subPath = subPath.substring(1);
             const parsedPath = Path.parse(subPath);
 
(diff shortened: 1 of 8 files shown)


jacobtylerwalls:
All bets are still off if your project named exactly `/js/` or container under a folder named that way 🙄 , but at least now project names including the "js" substring even at the start, like `jstreet` will now build.

bferguso:
I'm seeing the same issue when using a symlink to a file that sits outside of the application root. I'm looking into an alternative solution for this. I think it is at least an issue when loading templates, and possibly other asset types as well. Perhaps this should be out of scope for this PR and addressed as a separate issue.

jacobtylerwalls:
Setting this to draft until @bferguso has an opportunity to flesh out an alternative solution.

bferguso:
I was wondering if another approach to building the asset URLs might be to check for file existence rather than pattern matching the file path strings against APP_ROOTs and ROOT_DIR.

The issue I’m having with with symlinks in the current approach is that the webpack process is picking up the linked file’s real path (which is outside of the APP_ROOT) instead of the symlink path in the project directory. For example, in my project:
APP_ROOT  is `/fs/u02/apps_ux/projects/custom_bcrhp/src/bcrhp/bcrhp/`
ROOT_DIR is `/fs/u02/apps_ux/projects/core_bcrhp/arches/app/`
and I have a symlink from `/fs/u02/apps_ux/projects/custom_bcrhp/src/bcrhp/bcrhp/emplates/views/components/search/map-filter.htm` ->
`/fs/u02/apps_ux/projects/custom_bcrhp/src/common/templates/views/components/search/map-filter.htm`

So the when current logic compares the symlink target file path `/fs/u02/apps_ux/projects/custom_bcrhp/src/common/templates/views/components/search/map-filter.htm`  to APP_ROOT, it doesn’t match and just assumes it’s an Arches core file in `/fs/u02/apps_ux/projects/core_bcrhp/arches/app/`, the string split fails and the process continues with the string `undefined`.

An alternative approach might be to strip the parent directories off the `resourcePath` (in this case `/fs/u02/apps_ux/projects/custom_bcrhp/src/common/templates/views/components/search/map-filter.htm` would become `templates/views/components/search/map-filter.htm`) and we would prepend the APP_ROOTs and ROOT_DIR in the same order as we're currently doing the path string comparison. The first file to match would be the one used in the webpack generation process.

I can't remember off the top if / where additional template/resource locations are specified but we may need to take that into consideration when we're building the path to check for file existence.

This should eliminate any current issues causing path mismatches when using symlinks to pull in front end assets. I'd also suggest failing early if a file isn't matched. Currently if the webpack log isn't inspected carefully it looks like the process succeeded and the first indication that something is wrong is when things look strange in the app.

jacobtylerwalls:
Hey @bferguso -- this issue bit me again today, so just wanted to circle back here. Did you think you'd have capacity to flesh out your alternative fix? If not, I'm wondering if it would make sense to get the band-aid fix into 7.6. Let me know!

bferguso:
@jacobtylerwalls - I'll see if I can have a look at this today or tomorrow.

jacobtylerwalls:
Superseded by #11014

Issue #10804, opened 2024-04-19
Discussion: Making Core Search logic more modular


whatisgalen:
As a developer, you can create or customize search components as you like, you just have to ensure the sortorder if one search filter/component has some kind of dependency on another. Where that customization ends is what I would call the "core search logic". The logic that exists outside of any search component governs the following:
- what properties of the document should be included
- what exactly should happen when the search query gets executed and should there be only 1 ([for example](https://github.com/archesproject/arches/blob/6db93307e383fd7d79ce7b64104a77bb775d3147/arches/app/search/components/results_related_resources_filter.py#L49-L88))
- how many results should be returned (which might then later get paginated by the `paging-filter`)
- how localized descriptors should be applied
- what properties are on the response object

It's conceivable that 1 or all of those could be implemented differently by a developer, but to do so requires overriding methods like `search_results()`, `export_results()` or the entire `SearchView`. It would then be easier and a more modular solution to include the core logic as essentially a core search component.

Here's some of that logic:

```
    dsl.include("graph_id")
    dsl.include("root_ontology_class")
    dsl.include("resourceinstanceid")
    dsl.include("points")
    dsl.include("permissions.users_without_read_perm")
    dsl.include("permissions.users_without_edit_perm")
    dsl.include("permissions.users_without_delete_perm")
    dsl.include("permissions.users_with_no_access")
    dsl.include("geometries")
    dsl.include("displayname")
    dsl.include("displaydescription")
    dsl.include("map_popup")
    dsl.include("provisional_resource")
    if load_tiles:
        dsl.include("tiles")
    if for_export or pages:
        results = dsl.search(index=RESOURCES_INDEX, scroll="1m")
        scroll_id = results["_scroll_id"]
        if not pages:
            if total <= settings.SEARCH_EXPORT_LIMIT:
                pages = (total // settings.SEARCH_RESULT_LIMIT) + 1
            if total > settings.SEARCH_EXPORT_LIMIT:
                pages = int(settings.SEARCH_EXPORT_LIMIT // settings.SEARCH_RESULT_LIMIT) - 1
        for page in range(int(pages)):
            results_scrolled = dsl.se.es.scroll(scroll_id=scroll_id, scroll="1m")
            results["hits"]["hits"] += results_scrolled["hits"]["hits"]
    else:
        results = dsl.search(index=RESOURCES_INDEX, id=resourceinstanceid)

    ret = {}
    if results is not None:
        if "hits" not in results:
            if "docs" in results:
[...]

PR #10766, opened 2024-04-13
10644 related resources filter

whatisgalen:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

This PR enables a user to request an additional set of search results comprised of those instances of a chosen Resource Model that are *related* to any of the results from the search query.

Screenshot 2024-04-12 at 6 12 24 PM

Screenshot 2024-04-12 at 6 12 16 PM

* Needs a migration to insert new Search filter into search_component table
* This PR was informally based off a customized Search where all `hits` are returned rather than a page's worth. The all-vs-page_ct approaches should be better accounted for so that the search filter created in this PR knows how to get all the ids from the previous query and also what fraction of the related_results to return

### Issues Solved

#10644

### Checklist

-   [ ] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: 
*   Found by: @whatisgalen 
*   Tested by: @ 
*   Designed by: @whatisgalen

PR #10807, opened 2024-04-19
10804 componentizing core search

whatisgalen:
### Types of changes

### Description of Change

Fundamentally, here's what this PR tries to accomplish:
- *Greater Extensibility:* for search customizations to be more useful to the community, they must be easy-ish to "plug in". Overriding the Search view in a project doesn't really meet that criteria. Creating a component that lives in your project makes a lot more sense. For Arches to be able to switch out its core search logic, however, that core search logic needs to be removed and moved into a component itself
- *Outsource technical debt from the `SearchView` onto `search_component`s:* By stripping down one of the core methods (`search_results`) to just: initialize a SearchEngine, create a response object, and call 3 standard methods from each ordered set of search filters, there's less of a reason to tinker with the search view and development can be channeled into the `search_component`s.

Specifically, here's an overview of the changes:
- a special component called a `SearchLogicComponent` governs both what search components are available to the frontend (and what the frontend logic is) as well as what search components are *required* on the backend and how to handle the search request
- `search.htm` is just a placeholder for whatever template the search-logic component has and `search.js` doesn't do much beyond set up the necessities and send a request to get a SearchView. SearchLogic can be set as default when none is provided, or it can be provided as a named parameter in the url: `search?...&search-logic=custom-search-logic`
- a migration inserts the arches default core logic called `standard-search-logic` along with changes to specific components and adding a config column to all.
- search components extending `BaseFilter` class methods (`append_dsl`, etc.) now take the entire `response_object` and bundled `kwargs` as opposed to specifically a `results` object or `include_provisional` kwargs, respectively.

### Issues Solved

#10804 

### Checklist

#### Ticket Background
*   Found by: @whatisgalen 
*   Tested by: @apeters
*   Designed by: @whatisgalen


apeters (changes requested):
Ok, I think the general idea here is sound, but I do have some changes I'd like to suggest around how things should be structured.

First thing I would recommend is that we remove the `localize-descriptors` filter altogether and put the `post-search-hook` code into the `search-results` method.  I makes the most sense there I think.  It's not really a filter in the traditional sense and something we always want to use and so would never be "disabled" (unlike the "Map Filter" for example which can safely be "disabled").

One observation, both the `core-search` and `localize-descriptors` components were modeled as knockout components as the other search filters are.  The difference here is that these two components have no UI to speak of (reflected in the templates) and might be best modeled as simple view-models.   

Taking what I just said into consideration and recognizing that we now only really have the concept of a `core` search component that you've added, this begs the question.  Maybe the idea of a `core` search component should be expanded to include a UI, essentially all of the UI that currently resides in `views/seach.js`.  `views/search.js` then simply becomes a thin wrapper which loads the `core-search` component.  

This opens up an intriguing pattern.  We could now componentize the entire search experience and if we had an alternate "core" component we could deliver a completely different experience to users.  I could imagine a different search experience for authenticated vs. anonymous users for example.  This would require a bit more work to realize this, but I think you're already down that road.  I think that it would imply that there is a `default` core search component, but if the url included `core=alt` (for example), then a different core component could be rendered.  

Can you explore this enhancement to the pattern?

whatisgalen:
> First thing I would recommend is that we remove the `localize-descriptors` filter altogether and put the `post-search-hook` code into the `search-results` method. It makes the most sense there I think. It's not really a filter in the traditional sense and something we always want to use and so would never be "disabled" (unlike the "Map Filter" for example which can safely be "disabled").

The reason I moved that logic into its own component was because it iterates through the entire set of `results.hits.hits`. If my result set is 50,000 _and_ from cache, it doesn't really make sense to iterate over it (again), which is why I thought it would be better to leave it in component form so that someone could either disable it or edit it there. I could add comments in that component `.py` file to the effect of "Hey this is strongly recommended in some form and the only change we recommend is something more efficient to suit the implementation of your other search component logic."

> This opens up an intriguing pattern. We could now componentize the entire search experience and if we had an alternate "core" component we could deliver a completely different experience to users. I could imagine a different search experience for authenticated vs. anonymous users for example. This would require a bit more work to realize this, but I think you're already down that road. I think that it would imply that there is a `default` core search component, but if the url included `core=alt` (for example), then a different core component could be rendered.
> 
> Can you explore this enhancement to the pattern?

For sure!

arches/app/search/components/core_search.py
```diff
@@ -0,0 +1,83 @@
...
+    "classname": "CoreSearchFilter",
+    "type": "backend",
+    "componentpath": "views/components/search/core-search",
+    "componentname": "core-search",
+    "sortorder": "0",
+    "enabled": True,
+}
+
+class CoreSearchFilter(BaseSearchFilter):
+    def view_data():
```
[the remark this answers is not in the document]
whatisgalen: fix in latest

arches/app/search/components/localize_result_descriptors.py
```diff
@@ -0,0 +1,44 @@
...
+    "type": "backend",
+    "componentpath": "views/components/search/localize-descriptors",
+    "componentname": "localize-descriptors",
+    "sortorder": "0",
+    "enabled": True,
+}
+
+class LocalizeResultDescriptors(BaseSearchFilter):
+    def view_data(self):
+        pass
```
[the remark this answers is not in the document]
whatisgalen: fix in latest

whatisgalen:
@apeters here are the latest changes to branch:
- the template `search.htm` now only instantiates a "core" filter and a "backend" filter. 
- `core-search.htm` now has all the code that previously existed in the search view template (`search.htm`) and instantiations other search components
- localize-descriptors' template was removed
- the `SearchView` in `views/search.js` still does some stuff, but it's still fairly minimal.

whatisgalen:
one additional use case I just found was that this PR would make it easier to enforce a default set of search filters per user now that the `execute_query` method is available; you could create a search_component that arrives last in order of execution and enforces a search policy.

whatisgalen:
Summary of latest changes: individual search components that have "requiredFilters" in other search components now derive this information from a config property on the search_component model instead of having it hard-coded in the viewmodel. A migration for search_componet.config is included.

apeters (changes requested):

arches/app/media/js/views/components/search/localize-descriptors.js
apeters, line 1: I still think that this should be integrated into the search results filter.

whatisgalen: change in latest

arches/app/models/migrations/10804_core_search_filters.py
```diff
@@ -0,0 +1,75 @@
...
+            componentpath,
+            componentname,
+            sortorder,
+            enabled,
+            config
+        ) VALUES (
+            '69695d63-6f03-4536-8da9-841b07116381',
+            'Core Search',
+            '',
+            'core_search.py.py',
```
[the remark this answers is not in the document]
whatisgalen: fix in latest

arches/app/media/js/views/components/search/base-filter.js
```diff
@@ -15,6 +15,7 @@ define([
 
         initialize: function(options) {
             $.extend(this, options);
+            this.requiredFilters = this.getRequiredFilters(this.componentName);
```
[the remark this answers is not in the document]
whatisgalen: fixed in latest

whatisgalen:
Note that in the latest update to the migration, the `core-search` component now has `type:'core'...config:'{"default":true,...}'`

arches/app/models/models.py
```diff
@@ -1183,8 +1183,7 @@ class SearchComponent(models.Model):
     type = models.TextField()
     componentpath = models.TextField(unique=True)
     componentname = models.TextField(unique=True)
```
apeters, line 1185: remove the `unique` conststraint
```suggestion
    componentpath = models.TextField()
```

arches/app/search/components/base.py
```diff
@@ -40,7 +41,15 @@ def view_data(self):
 
         pass
 
-    def post_search_hook(self, search_results_object, results, permitted_nodegroups):
+    def execute_query(self, search_results_object, response_object):
```
apeters, line 44: I think this method should be put into a new `BaseSearch` class.
This class can be implemented by `core_search.py`.

arches/app/search/components/core_search.py
```diff
@@ -0,0 +1,141 @@
...
+        ],
+    },
+}
+
+SEARCH_RESULT_PAGES = (
+    int(settings.SEARCH_EXPORT_LIMIT // settings.SEARCH_RESULT_LIMIT) - 1
+)
+
+
+class CoreSearchFilter(BaseSearchFilter):
```
apeters, line 79: We shouldn't think of this class as being a type of `SearchFilter`.  Instead it could inherit from a new class called `BaseSearch` that could be in base.py.

arches/app/views/search.py
```diff
@@ -348,28 +367,17 @@ def get_dsl_from_search_string(request):
 
 
 def search_results(request, returnDsl=False):
```
apeters, line 369: we should try and push as much of this code into `base.py` or `core_search.py`.  The search view should just be a light weight wrapper that calls out to the core search component to perform much of the core logic.

arches/app/views/search.py
```diff
@@ -81,12 +75,37 @@ def get(self, request):
...
+        ]
+        # TODO: Apply new permission logic after #11005 is merged
+        if user_is_resource_exporter(request.user) and "search-export" in [
+            s["componentname"]
+            for s in core_search_component.config["requiredComponents"]
+        ]:
+            search_components.append(all_search_components_dict["search-export"])
+
+        search_components.append(core_search_component)
+
```
apeters, line 108: I think we should push as much of this off to `base.py` or `core_search.py`

arches/app/media/js/views/components/search/core-search.js
```diff
@@ -0,0 +1,159 @@
...
+            this.query(queryObj);
+        },
+
+        restoreState: function(){
+            if (!this.query()[componentName]) {
+                this.updateQuery();
+            } else {
+                this.sharedStateObject.loading(false);
+            }
+        },
```
apeters, line 150: are these methods even called?

arches/app/media/js/views/components/search/core-search.js
```diff
@@ -0,0 +1,159 @@
...
+    'viewmodels/alert',
+    'views/components/search/base-filter',
+    'templates/views/components/search/core-search.htm',
+], function($, _, ko, arches, AlertViewModel, BaseFilter, coreSearchTemplate) {
+    const componentName = 'core-search';
+    const viewModel = BaseFilter.extend({ 
+        initialize: function(sharedStateObject) {
+            const self = this;
+            sharedStateObject.componentName = componentName;
+            BaseFilter.prototype.initialize.call(this, sharedStateObject);
```
apeters, line 15: I don't think we should think of this as an instance of a `BaseFilter`.  It should rather be it's own thing, or inherit from a new class that mirrors my comments on the `core-search.py` file.

apeters (changes requested):

arches/app/models/migrations/10804_core_search_filters.py
```diff
@@ -0,0 +1,65 @@
...
+            reverse_sql,
+        ),
+        migrations.RemoveField(
+            model_name="searchcomponent",
+            name="enabled",
+        ),
+        migrations.RemoveField(
+            model_name="searchcomponent",
+            name="sortorder",
+        ),
```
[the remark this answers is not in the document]
whatisgalen: fix in latest

arches/app/models/models.py
whatisgalen: fix in latest

arches/app/search/components/base.py
whatisgalen: I've been thinking about this one a lot and I think it still makes sense to have `execute_query` not only be its own method to compartmentalize the logic but also that it originate from the `BaseSearchFilter` class as opposed to the `CoreSearchComponent` class and here's why:  

Search Components that are not Core Search Components should still be able to insert logic at this specific point between `append_dsl` and `post_search_hook`. In the example of [this draft PR](https://github.com/archesproject/arches/pull/10766) where there's a follow-up query to get all those resources related to what's in the result set, it would be ideal for this action to happen *before* `post_search_hook` so that all the logic regarding how to structure or format results applies to this superset of results.

whatisgalen:
I have a couple of things to clarify specific to the frontend approach, but otherwise I've implemented pretty much everything you've made a comment on. Also I left a [comment](https://github.com/archesproject/arches/pull/10807/files#r1661942986) on your review at `execute_query`, happy to discuss that further.

arches/app/media/js/views/search.js
```diff
@@ -46,70 +44,45 @@ define([
...
+        this.getRequiredFilters = function(componentname) {
+            // Map these IDs to their corresponding component names from filtersList
+            const requiredComponentNames = this.requiredFiltersLookup[componentname].map(id => {
+                // Find the component in filtersList that matches the id and return its componentname
+                const component = this.filtersList.find(component => component.searchcomponentid == id);
+                return component ? component.componentname : undefined;
+            }).filter(name => name); // This removes any undefined entries in case some IDs don't match any component
+        
+            return requiredComponentNames;
+        };
```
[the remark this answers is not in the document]
whatisgalen: Not all search filters (`saved-searches` and `search-export` as examples) extend `BaseFilter` so we'd have to enforce that they do (and change those search filters). Changing how those search components work on the frontend feels like it merits its own ticket though so my vote is to just keep this method in `search.py` since it's pretty generic anyway and can be expected to be called on any search component.

whatisgalen:
Here's a summary of the latest changes:
- On the frontend, there's now a generic component called `core-search-component.js` that has all the things you'd need to make your own core-search-component viewmodel (methods that formerly lived in `views/search.js`). `arches-core-search.js` largely extends this but has some overrides
- On the backend, `search/components/base.py` now has the class `CoreSearchComponent`. An instance of this class is used to determine which search components the `SearchView` should send to the frontend, along with a method that entails the overall logic for `search_results` method.
- `SearchFilterFactory` can now return a core component based on a request (either via its `core=` kwarg or by supplying the default core component from the db)

arches/app/media/js/views/components/search/core-search.js
whatisgalen: removed in latest

chiatt (commented):
Could file names not be prefixed with 'arches' (e.g.`arches-core-search`)? We already know it's arches :)

Also, these names are confusing because they are both in the components directory and both "core". Maybe the names could be revised to help distinguish their respective roles?
```
components/search/core-search-component.js
components/search/arches-core-search.js
```

whatisgalen:
> Could file names not be prefixed with 'arches' (e.g.`arches-core-search`)? We already know it's arches :)
> 
> Also, these names are confusing because they are both in the components directory and both "core". Maybe the names could be revised to help distinguish their respective roles?
> 
> ```
> components/search/core-search-component.js
> components/search/arches-core-search.js
> ```

I've refactored `components/search/core-search-component.js` -> `components/search/base-core-search.js` and similarly for the corresponding Python class. For the thing called `"arches-core-search"`, I don't know that it makes sense to call it, say `"default-core-search"` since "default" is a setting on the `search_component.config` in the db. I also don't want to call it simply `"core-search"` because that seems somewhat ambiguous to me paired alongside `"base-core-search"`. We use the "Arches" prefix in some places, like `ArchesTestCase`, `ArchesJSON`, `ArchesStandardPermissionFramework`, `ArchesUserProfileForm`, `ArchesFileReader`, etc., so it doesn't seem to me like an immense departure in pattern. I think `base-core-search` makes the most sense for the former, but did you have particular suggestions for renaming the latter?

whatisgalen:
right now there's a key on the config called `requiredComponents` which essentially sends that list to the frontend to tell it which components to instantiate. However what's required for that core component on the frontend isn't the same as what would be required for that core component on the backend. For example, map-filter may not be required for the arches-core-search component but search-results definitely is.

Based on this thinking, I'm contemplating distinguishing between components that are "associated" or "member" to a core vs components that are a non-negotiable requirement for a core to execute on the backend. So maybe making "availableComponents" be the frontend list and "requiredComponents" be the backend list.

whatisgalen:
todo: fix search-results template

apeters (changes requested):
What I'd like to do is create an exact duplicate of the existing core search but with a different config only.  Everything else is the same except the `componentname` which is used in the url to specify the type of core search to use.
IE: `localhost:8000/search?core=my-cool-new-search`

Screenshot 2024-07-18 at 6 55 05 PM

Right now there is a unique constraint on the `componentpath` column that prevents that.   That should changed if possible.

whatisgalen:
@apeters your request to remove the unique constraint for component path so that search components can be re-used is a reasonable one. However, ko components must be registered with a static, unchanging componentname in order to be called by that componentname in the first place. Thus, even without this constraint, you couldn't assign `"my-new-component"` to the path `views/components/arches-core-search` because the file at that path contains code registering it as `"arches-core-search"`, rendering `"my-new-component"` unfindable, causing a load error. You also can't pass in a componentname as There are two alternatives to achieve what you're after: 
- some frontend approach other than ko components to enable dynamic component declaration
- create a light-weight component for `"my-new-component"` that merely extends `"arches-core-search"`

arches/app/search/components/base.py
```diff
@@ -50,6 +73,92 @@ def post_search_hook(self, search_results_object, results, permitted_nodegroups)
         pass
 
 
+class BaseCoreSearch(BaseSearchFilter):
```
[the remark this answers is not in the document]
whatisgalen: change in latest

arches/app/media/js/views/components/search/base-core-search.js
```diff
@@ -0,0 +1,96 @@
...
+                this.doQuery();
+            }, this);
+        },
+
+        doQuery: function() {
+            let queryObj = JSON.parse(this.queryString());
+            if (self.updateRequest) { self.updateRequest.abort(); }
+            self.updateRequest = $.ajax({
+                type: "GET",
+                url: arches.urls.search_results,
```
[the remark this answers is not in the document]
whatisgalen: fix in latest

arches/app/media/js/views/components/search/arches-core-search.js
```diff
@@ -0,0 +1,114 @@
...
+                    this.selectedPopup('');
+                } else {
+                    this.selectedPopup(componentname);
+                }
+            };
+        },
+
+        doQuery: function() {
+            let queryObj = JSON.parse(this.queryString());
+            queryObj[componentName] = true;
```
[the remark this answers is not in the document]
whatisgalen: change in latest

whatisgalen:
Note that as of the latest changes to this branch, backend search component python modules (e.g. `term_filter.py` and `TermFilter` class) *are* re-usable as they now accept a componentname paramter to tell them which key the relevant search params are. However frontend ko search components are *not* re-usable and new components will have to explicitly extend pre-existing ko search components to take advantage of that code; a new js file with a new name must be created, but the .htm template file from the pre-existing ko component can be re-used, e.g. "custom-term-filter", `custom-term-filter.js`, using `term-filter.htm` template.

chiatt (changes requested):
Search appears to load okay, but I noticed I am getting the following error:  


Also, if I migrate back to the previous commit, switch to the dev/7.6.x branch, and rerun webpack, search fails to load with this error:

arches/app/models/models.py
chiatt: It looks like this has been changed back to unique?

whatisgalen: Yea I reverted the change. I was seeing a weird issue preventing the migration from working, regarding the index of the search_components table (which was generated automatically I think) not being refreshed correctly. That got me thinking more about the change, and then I realized that since componentnames are unique and each componentpath is 1:1 with its component, may as well keep componentpath unique to avoid unintended errors.

apeters (changes requested):

apeters (changes requested):

chiatt (approved):

arches/app/media/js/views/search.js
```diff
@@ -45,71 +43,53 @@ define([
     };
 
     var CommonSearchViewModel = function() {
-        this.filters = {};
-        this.filtersList = _.sortBy(Object.values(SearchComponents), function(filter) {
-            return filter.sortorder;
-        }, this);
+        this.searchComponentVms = {};
```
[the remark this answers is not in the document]
whatisgalen: Formerly there were two distinct vars: `this.filters` and `this.filtersList`. The latter is the list of search component model instances passed from the python view. The former is a lookup of the search component ko components. I'd prefer if the var names were more explicit / semantically named. `searchComponents` could refer to either.

arches/app/media/js/views/components/search/standard-search-logic.js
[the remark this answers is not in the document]
whatisgalen: I talked to @chiatt about this -- given that the setting "default" is in the config, I wondered if having "default" in the filename would confuse developers in thinking that if for whatever reason their custom search didn't work, it would default to the arches search (which wouldn't be the case). Not my hill to die on but that was the reasoning behind the choice of "standard".

arches/app/media/js/views/search.js
```diff
@@ -45,71 +43,53 @@ define([
...
+                return component ? component.componentname : undefined;
+            }).filter(name => name); // This removes any undefined entries in case some IDs don't match any component
+        
+            return requiredComponentNames;
+        };
+        this.getFilterByType = function(type) {
+            const filter = this.filtersList.find(component => component.type == type);
+            if (!filter)
+                return null;
+            return ko.unwrap(this.searchComponentVms[filter.componentname]);
```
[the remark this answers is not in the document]
whatisgalen: `getFilter` hasn't been removed. `getFilterByType` was added for the following reason: Some components need a `term-filter` but for a generic reason, like adding a tag. Other components might need a specific component for a specific reason, which is where `getFilter` would still be used with the component name. `getFilterByType` lets you create a custom version of a component, for ex: `custom-term-filter` without needing to create custom versions of all the other components that call `term-filter`. It's up to each component's logic to determine which methods should be called.

arches/app/media/js/views/components/search/standard-search-logic.js
```diff
@@ -0,0 +1,62 @@
...
+    'views/components/search/base-search-logic',
+    'templates/views/components/search/standard-search-logic.htm',
+], function($, _, ko, arches, AlertViewModel, BaseSearchLogicComponent, standardSearchLogicTemplate) {
+    const componentName = 'standard-search-logic';
+    const viewModel = BaseSearchLogicComponent.extend({ 
+        initialize: function(sharedStateObject) {
+            const self = this;
+            sharedStateObject.componentName = componentName;
+            BaseSearchLogicComponent.prototype.initialize.call(this, sharedStateObject);
+            this.defaultQuery = {"paging-filter": "1", "search-logic":"standard-search-logic", tiles: "true"};
```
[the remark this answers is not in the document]
whatisgalen: Just want to make sure we're 100% sure about this suggested change.

Here's what this change would look like in search.py for example:

```.py
class SearchView(MapBaseManagerView): # django view class
    def get(self, request):
        ...
        search_component_factory = SearchFilterFactory(request)
        search_view_component_instance = search_component_factory.get_search_view_component_instance() # search component of type search-view
        search_components = search_view_component_instance.get_searchview_components() # search components linked to search-view component, unrelated to parent class here
```

Note that there is some confusion between the `search-view` component and its related methods and the django `SearchView` class

arches/app/media/js/views/search.js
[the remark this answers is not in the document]
apeters: Ok, how about this.  We rename `filtersList` to `searchComponentConfigs` and `searchComponentVms` to `searchComponents`

arches/app/media/js/views/search.js
```diff
@@ -45,71 +43,53 @@ define([
...
+        };
         Object.values(SearchComponents).forEach(function(component) {
```
apeters, line 76:
```suggestion
        this.searchComponentConfigs.forEach(function(component) {
```

arches/app/media/js/views/components/search/standard-search-logic.js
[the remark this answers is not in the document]
apeters: Then let's call this `standard-search-view.js`

arches/app/media/js/views/search.js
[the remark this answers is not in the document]
whatisgalen: That seems a little weird given then we'd be potentially accessing `searchComponentConfig.config` to check things like `searchComponentConfig.config.layoutType`

arches/app/templates/views/components/search/standard-search-logic.htm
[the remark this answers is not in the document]
apeters: and this `standard-search-view.htm`

arches/app/media/js/views/components/search/standard-search-logic.js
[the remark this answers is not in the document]
whatisgalen: already in latest commit

apeters:
Can we also remove from the database search_component.config field the `RequiredComponents` and `AvailableComponents`  and replace that with a single array of components called `AppliedFilters`.  You can add a `required` attribute for the required filters.

apeters:
Also, I think we've over used the work `component`.  In reality, everything is a component.  We should use the word `Filter` when contextually it makes most sense.  I also realize that not everything is a filter, but I still think descriptively that it works.

arches/app/search/components/standard_search_view.py
```diff
@@ -0,0 +1,240 @@
...
+        ]
+        if user_is_resource_exporter(self.request.user):
+            search_components.extend(
+                [
+                    available_component
+                    for available_component in self._available_search_components
+                    if available_component.componentname == "search-export"
+                ]
+            )
+
```
apeters, line 186: you should use the propertie (available_search_components) of the class instead of the private variable (_available_search_components)
```suggestion
        search_components = [
            available_component
            for available_component in self.available_search_components
            if available_component.componentname != "search-export"
        ]
        if user_is_resource_exporter(self.request.user):
            search_components.extend(
                [
                    available_component
                    for available_component in self.available_search_components
                    if available_component.componentname == "search-export"
                ]
            )

```

arches/app/search/components/base_search_view.py
```diff
@@ -0,0 +1,163 @@
...
+                item.componentname, float("inf")
+            ),
+        )
+
+    @property
+    def required_search_components(self):
+        return self._required_search_components
+
+    @property
+    def available_search_components(self):
```
apeters, line 117: maybe we should just admit that this is all the filters applied to the search view and call it that
```suggestion
    def search_filters(self):
```

we should also rename the private variables accordingly (eg: _available_search_components --> _search_filters)

arches/app/media/js/views/components/search/advanced-search.js
```diff
@@ -195,11 +195,11 @@ define([
                 queryObj[componentName] = JSON.stringify(advanced);
 
                 if (this.getFilter('term-filter').hasTag(this.tagId) === false) {
```
apeters, line 197:
```suggestion
                if (this.getFilterByType('term-filter').hasTag(this.tagId) === false) {
```

arches/app/media/js/views/components/search/standard-search-view.js
```diff
@@ -0,0 +1,62 @@
...
+    'templates/views/components/search/standard-search-view.htm',
+], function($, _, ko, arches, AlertViewModel, BaseSearchViewComponent, standardSearchViewTemplate) {
+    const componentName = 'standard-search-view';
+    const viewModel = BaseSearchViewComponent.extend({ 
+        initialize: function(sharedStateObject) {
+            const self = this;
+            sharedStateObject.componentName = componentName;
+            BaseSearchViewComponent.prototype.initialize.call(this, sharedStateObject);
+            this.defaultQuery = {"paging-filter": "1", "search-view":"standard-search-view", tiles: "true"};
+
```
apeters, line 17: we could probably remove this line as "required" filters are now defined in the component config
```suggestion
            this.defaultQuery = {"paging-filter": "1", "search-view":"standard-search-view", tiles: "true"};
```

apeters (changes requested):

apeters:
We should remove all values from the "type" column related to layout, now that it is defined in the "config" column.

apeters:
What's the meaning of the for example, the "search-results" filter having defined in it's config a "requiredComponent" of the "map-filter"?

apeters:
Maybe we should come up with a "type" for each filter that is currently blank.  I would also try to avoid using the same value as "componentname".  So for example the type for "TimeFilter" could be "temporal-filter".

whatisgalen:
> Can we also remove from the database search_component.config field the `RequiredComponents` and `AvailableComponents` and replace that with a single array of components called `AppliedFilters`. You can add a `required` attribute for the required filters.

When we discussed this on our call I mentioned that on the frontend, search would be looking for search components that don't have a knockout component, yielding an error, unless we also had a key like "frontend": true or some way to tell search "don't look for a ko component for this search filter".

whatisgalen:
> What's the meaning of the for example, the "search-results" filter having defined in it's config a "requiredComponent" of the "map-filter"?

in the ko search components, some of them had hard-coded values for "requiredComponents". I just moved that into the config. Some frontend components depend on others or at least expect them.

arches/app/search/components/base_search_view.py
whatisgalen: If we're saying that a search component with a frontend ko component is effectively a filter, what would that make search component without one?

apeters: I'm really saying that besides the `search-view` their all filters.  Or we should just generally think about them that way.  Strictly speaking most of them are in fact "filters", but other like search export aren't.  I like "filters".  Components is just too generic.

apeters:
> > Can we also remove from the database search_component.config field the `RequiredComponents` and `AvailableComponents` and replace that with a single array of components called `AppliedFilters`. You can add a `required` attribute for the required filters.
> 
> When we discussed this on our call I mentioned that on the frontend, search would be looking for search components that don't have a knockout component, yielding an error, unless we also had a key like "frontend": true or some way to tell search "do/don't look for a ko component for this search filter". The other thing we get by having two separate arrays here is that you can set a sortorder for a filter on the frontend (used in layout, could potentially be used for logic) that's separate from the sortorder for a filter on the backend (execution order of search filters)

The two objects (`RequiredComponents`, `AvailableComponents`) don't seem to imply front-end or back-end functions.  I think there is only 1 sort during a search operation.  From our discussion the `RequiredComponents` object was just there to enforce filters that need to be applied if none are specified in the url.

whatisgalen:
> > When we discussed this on our call I mentioned that on the frontend, search would be looking for search components that don't have a knockout component, yielding an error, unless we also had a key like "frontend": true or some way to tell search "do/don't look for a ko component for this search filter". The other thing we get by having two separate arrays here is that you can set a sortorder for a filter on the frontend (used in layout, could potentially be used for logic) that's separate from the sortorder for a filter on the backend (execution order of search filters)
> 
> The two objects (`RequiredComponents`, `AvailableComponents`) don't seem to imply front-end or back-end functions. I think there is only 1 sort during a search operation. From our discussion the `RequiredComponents` object was just there to enforce filters that need to be applied if none are specified in the url.

There's a sort of `availableComponents` when the `SearchView` sends search components to the frontend here:
```.py
self._available_search_components = sorted(
            self._available_search_components,
            key=lambda item: available_component_sort_order.get(
                item.componentname, float("inf")
            ),
        )
```

and a sort when the query gets handled via this method called by `search_results`:
```.py
def get_sorted_query_dict(self, query_dict, searchview_component):
        component_sort_order = {
            item["componentname"]: int(item["sortorder"])
            for item in searchview_component.config["requiredComponents"]
        }
        # Sort the query_dict items based on the requiredComponent's sortorder
        sorted_items = sorted(
            query_dict.items(),
            key=lambda item: component_sort_order.get(item[0], float("inf")),
        )
```

whatisgalen:
The latest commits address all requested changes except for finalizing specific values for `search_component.type`

arches/app/search/components/standard_search_view.py
```diff
@@ -0,0 +1,238 @@
...
+                "searchcomponentid": "00673743-8c1c-4cc0-bd85-c073a52e03ec",
+                "sortorder": 13,
+            },
+        ],
+    },
+}
+
+SEARCH_RESULT_PAGES = (
+    int(settings.SEARCH_EXPORT_LIMIT // settings.SEARCH_RESULT_LIMIT) - 1
+)
```
apeters, line 112: this is only used once in this file.  I see no need to pull it out.

arches/app/search/components/standard_search_view.py
```diff
@@ -0,0 +1,238 @@
...
+        resourceinstanceid = self.request.GET.get("id", None)
+        dsl = search_query_object["query"]
+        if for_export or pages:
+            results = dsl.search(index=RESOURCES_INDEX, scroll="1m")
+            scroll_id = results["_scroll_id"]
+            if not pages:
+                if total <= settings.SEARCH_EXPORT_LIMIT:
+                    pages = (total // settings.SEARCH_RESULT_LIMIT) + 1
+                else:
+                    pages = SEARCH_RESULT_PAGES
```
apeters, line 147:
```suggestion
                    pages = int(settings.SEARCH_EXPORT_LIMIT // settings.SEARCH_RESULT_LIMIT) - 1
```

arches/app/search/components/standard_search_view.py
```diff
@@ -0,0 +1,238 @@
...
+        if results is not None:
+            if "hits" not in results:
+                if "docs" in results:
+                    results = {"hits": {"hits": results["docs"]}}
+                else:
+                    results = {"hits": {"hits": [results]}}
+        response_object["results"] = results
+
+    def post_search_hook(self, search_query_object, response_object, **kwargs):
+
```
apeters, line 163: remove this empty line

apeters (changes requested):
Almost there!

arches/app/search/components/standard_search_view.py
whatisgalen: I made this change because further down in the if block on line 144 it's so indented that the black formatting caused its readability to degrade imo:
```.py
pages = int(
        settings.SEARCH_EXPORT_LIMIT
        //
        settings.SEARCH_RESULT_LIMIT
) - 1
```

releases/7.6.0.md
```diff
@@ -15,10 +15,9 @@ Arches 7.6.0 Release Notes
 
 - Adds Github actions to build applications, run tests, and compare coverage between branches for projects.
 
-- Plugins now support the configuration boolean `is_standalone`. Standalone plugins do not appear in the sidebar, and do not display the sidebar or application header.
```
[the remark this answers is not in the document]
whatisgalen: yes, see @jacobtylerwalls [comment](https://github.com/archesproject/arches/pull/10871#discussion_r1711385907)

arches/app/media/js/views/components/search/search-results.js
```diff
@@ -53,23 +52,32 @@ define([
                     this.updateResults();
                 }, this);
 
-                this.searchComponentVms[componentName](this);
+                this.searchFilterVms[componentName](this);
                 this.restoreState();
-                if (this.requiredFiltersLoaded() === false) {
-                    this.requiredFiltersLoaded.subscribe(function () {
-                        this.mapFilter = this.getFilterByType("map-filter");
+                this.mapFilter = this.getFilter("map-filter", false);
```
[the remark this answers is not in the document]
whatisgalen: if you check out line 61: `if (tab === "map-filter") {` the value for tab comes from that search-filter's component name, so even if another filter like `galen-map-filter` was the right type (`"map-filter-type"`), the tab wouldn't be called `"map-filter"`. If that check weren't there, I would agree that it should just be `getFilterByType`.

arches/app/search/components/standard_search_view.py
whatisgalen: change in latest

arches/app/search/components/standard_search_view.py
whatisgalen: change in latest

apeters (approved):
phew!  that took a while but I think it was worth it!

Issue #10814, opened 2024-04-22
Change behavior of QA Type filter in search


ryan86:
Change behavior of QA Type filter from:

Authoritative = Authoritative resources that can include provisional edits

to

Authoritative = Authoritative resources with no provisional edits

Authoritative resources with provisional edits will now be shown in the Provisional QA Type filter.

PR #10815, opened 2024-04-22
Change behavior of Authoritative QA Type filter, re Change behavior of QA Type filter in search #10814

ryan86:


chiatt (approved):
Looks good!

Issue #10812, opened 2024-04-22
Provisional edit flag breaks search result layout 


chiatt:
This occurs when an instance has a geojson node and thus has a 'Map' option in the search result footer. 

<img width="402" alt="image" src="https://github.com/archesproject/arches/assets/1877663/e93c0a2d-787a-4b71-966c-4b0b06919dd8">

PR #10819, opened 2024-04-23
Fix display of search results footer when buttons get crowded, re Provisional edit flag breaks search result layout  #10812

ryan86:
Fix display of search footer when the provisional edit banner and map button are present. re #10812

### DIFF
--- a/arches/app/templates/views/components/search/search-results.htm
+++ b/arches/app/templates/views/components/search/search-results.htm
@@ -43,7 +43,7 @@
             </div>
         </div>
         <div class="search-listing-footer">
-            <div style="flex-grow: 1; padding-bottom: 10px;">
+            <div style="flex-grow: 1;">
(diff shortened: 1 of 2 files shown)


arches/app/templates/views/components/search/search-results.htm
```diff
@@ -43,7 +43,7 @@
             </div>
         </div>
         <div class="search-listing-footer">
-            <div style="flex-grow: 1;">
+            <div style="flex-grow: 1; padding-bottom: 10px;">
```
chiatt, line 46: This looks fine if wrapping is necessary, but if it isn't it causes the footer to be too talk and the text vertically uncentered.

chiatt (changes requested):

chiatt (approved):
Looks good!

PR #10820, opened 2024-04-23
allow non-localized-string widget placeholder to be localized, re #9191

apeters:
forgot to allow for the widget placeholder to be localized, re #9191


chiatt (approved):
Looks good!

Issue #10816, opened 2024-04-22
Feature Idea: Map Feature popup has link to use in Map Filter


whatisgalen:
This feature would be an option or link on the map popup of any map feature (ideally from any source/map layer). Clicking "Search near this point" / "Search within this polygon" would populate the parameter of the Map filter with the feature's geometry, buffering some arbitrary distnace for points and none for polygons (as defaults).

whatisgalen:
Adding a note that the proposed solution in #10827 depends on PR #11056 (fix for #10831) to proceed

whatisgalen:
done in #10827

PR #10827, opened 2024-04-24
Map Filter search by resource layer feature

whatisgalen:
### Types of changes

### Description of Change

https://github.com/archesproject/arches/assets/8366985/3df090fb-5cb2-4472-bf17-c695242561a0

Overview:
- map popup sends afeature to a method in `map-filter.js` called `filterByFeatureGeom()`. 
- the method adds this feature to the map-filter as if a user had drawn a feature.
- the user can alter the feature on the map to adjust the filter as they want.
- clicking a new feature removes the old one from the filter, replacing it with the new

### Issues Solved

#10816 

### Checklist

#### Ticket Background
*   Found by: @whatisgalen 
*   Tested by: @bferguso 
*   Designed by: @whatisgalen 

### Further comments

blocked until PR #11056 gets merged (done)


arches/app/search/components/map_filter.py
```diff
@@ -63,6 +49,41 @@ def append_dsl(self, search_results_object, permitted_nodegroups, include_provis
...
+            feature_geom = geometries[0]["geometry"]
+            spatial_query = create_geoshape_query(feature_geom, spatial_filter)
+            spatial_query.filter(Terms(field="geometries.nodegroup_id", terms=permitted_nodegroups))
+
+            if include_provisional is False:
+                spatial_query.filter(Terms(field="geometries.provisional", terms=["false"]))
+            elif include_provisional == "only provisional":
+                spatial_query.filter(Terms(field="geometries.provisional", terms=["true"]))
+
+            search_query.filter(Nested(path="geometries", query=spatial_query))
```
chiatt, line 86: Just took a quick glance and noticed these lines are essentially the same as lines 43-51 - could be moved to a common function.

whatisgalen: fix in latest

whatisgalen:
My new test is failing, debugging that. In the meantime the feature is definitely testable.

arches/app/media/js/views/components/search/map-filter.js
```diff
@@ -274,6 +275,39 @@ define([
...
+                            "width": this.buffer(),
+                            "unit": this.bufferUnit()
+                        },
+                        "inverted": this.filter.inverted()
+                    });
+                } else {
+                    const tolerance = 0.1; // Degree of Simplification: Lower numbers are less simplified, preserving more detail
+                    const highQuality = true; // Set to true for a slower but higher quality simplification
+                    turf.simplify(feature.geometry, {tolerance: tolerance, highQuality: highQuality, mutate: true});
+                }
```
chiatt, line 295: Simplifying the geometries used in the search filter seems out of scope for the issue associated with this PR ( #10816). It may be something that we should do to shorten the search URL, but changing a geometry without a user's awareness seems like an important enough change that it should be addressed in another issue/PR.

bferguso: @chiatt, @whatisgalen - I'm just trying to finish up another time-critical task and then I'll delve in further, but it seems like it would be a good thing to externalize the the functionality that provides the search geometry into a utility class (kind of like what we did for the map popups) so that custom logic (eg - whether you want to simplify geometries, or getting the geometry by another method, etc, etc) could be overridden by the developer without delving into the core Arches javascript files which are more tightly integrated into the core arches components (cards, views, widgets, etc).

whatisgalen: > @chiatt, @whatisgalen - I'm just trying to finish up another time-critical task and then I'll delve in further, but it seems like it would be a good thing to externalize the the functionality that provides the search geometry into a utility class (kind of like what we did for the map popups) so that custom logic (eg - whether you want to simplify geometries, or getting the geometry by another method, etc, etc) could be overridden by the developer without delving into the core Arches javascript files which are more tightly integrated into the core arches components (cards, views, widgets, etc).

@bferguso thanks for this great feedback. I've incorporated your suggestion into creating a method in the map popup provider to handle the popup feature en route to sending to the mapCard's method. Unless a developer has overridden `viewmodels/map.js`, every popup feature should have direct access to the `mapCard` i.e. Map Filter search filter.

whatisgalen: > Simplifying the geometries used in the search filter seems out of scope for the issue associated with this PR ( #10816). It may be something that we should do to shorten the search URL, but changing a geometry without a user's awareness seems like an important enough change that it should be addressed in another issue/PR.

removed in latest

arches/app/templates/views/components/map-popup.htm
```diff
@@ -9,65 +9,71 @@
...
+                    <!--/ko-->
+                    <!-- ko if: showEditButton -->
+                    <a data-bind="click: function () {
+                        window.open(editURL + ko.unwrap(resourceinstanceid));
+                    }" href="javascript:void(0)">
+                        <i class="ion-ios-refresh-empty"></i>
+                        <span data-bind="text: $root.translations.edit"></span>
+                    </a>
+                    <!--/ko-->
+                    <!-- ko if: ko.unwrap(geometries).length -->
```
bferguso, line 61: I think most of the logic in the map popup will work really well, and allows the logic to be overridden to allow custom geometry retrieval for both resource instances and map overly features. The only catch here is that `geometries` doesn't exist if a geometry from a map overlay is selected so this line throws an exception if it's a map overlay feature and not an Arches resource feature. Can the show/hide logic be implemented in the mapPopupProvider so users that are providing overlay map popups can override this logic?

Also wondering if the visibility of this should also be controlled by an option in the Filter->Select a Filter select box to be consistent with other Map Filter functions.

This maybe be out of scope for this PR, and if so could be followed up by a separate one: Maybe not an issue for geometries that are part of a resource, but filtering by overly-complex geometries in external overlays could potentially cause performance issues (we have a geology layer that would probably be an issue). For overlays layers we could add a configuration variable to the `arches-metadata` map layer object to control whether it's allowed to be used as a source for a map filter geometry.

bferguso (changes requested):

whatisgalen:
> Another question/thought for comment: We've implemented logic that allows multiple geometries to be used as a map filter, applying either a UNION or INTERSECTION across them. Is there any interest in that functionality in Arches Core?

I can envision the UNION case being handled as passing an array of features to the map_filter. The INTERSECTION case would most likely mean a change to the frontend UX. If you could apply a "2nd mapfilter" much like you can apply a 2nd term filter, that seems like the most intuitive way to handle querying by INTERSECTION of spatial features.

bferguso:
> > Another question/thought for comment: We've implemented logic that allows multiple geometries to be used as a map filter, applying either a UNION or INTERSECTION across them. Is there any interest in that functionality in Arches Core?
> 
> I can envision the UNION case being handled as passing an array of features to the map_filter. The INTERSECTION case would most likely mean a change to the frontend UX. If you could apply a "2nd mapfilter" much like you can apply a 2nd term filter, that seems like the most intuitive way to handle querying by INTERSECTION of spatial features.

We've implemented it like this: 
image
The select only shows up if there is > 1 geometry selected.

arches/app/media/js/utils/map-popup-provider.js
```diff
@@ -40,6 +40,19 @@ define(['arches',
...
         },
 
+        /**
+         * This method enables custom logic for how the feature in the popup should be handled and/or mutated en route to the mapFilter.
+         * @param popupFeatureObject - the javascript object of the feature and its associated contexts (e.g. mapCard).
+         * @required @method mapCard.filterByFeatureGeom()
+         * @required send argument: @param feature - a geojson feature object
+         * @optional send argument: @param resourceid
+         */
+        sendFeaturetoMapFilter: function(popupFeatureObject)
```
bferguso, line 50: Should this be `sendFeatureToMapFilter`?

arches/app/media/js/views/components/search/map-filter.js
```diff
@@ -584,6 +618,7 @@ define([
             this.getFilter('term-filter').removeTag('Map Filter Enabled');
             this.draw.deleteAll();
             this.searchGeometries([]);
+            this.setupDraw();
```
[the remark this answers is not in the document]
whatisgalen: fix in latest

arches/app/templates/javascript.htm
```diff
@@ -654,6 +654,7 @@
         map-add-line='{% trans "Add line" as mapAddLine %} "{{ mapAddLine|escapejs }}"'
         map-add-polygon='{% trans "Add polygon" as mapAddPolygon %} "{{ mapAddPolygon|escapejs }}"'
         map-select-drawing='{% trans "Select drawing" as mapSelectDrawing %} "{{ mapSelectDrawing|escapejs }}"'
+        search-near-here='{% trans "Search Near Here" as searchNearHere %} "{{ searchNearHere|escapejs }}"'
```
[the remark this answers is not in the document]
whatisgalen: fix in latest

arches/app/templates/views/components/map-popup.htm
whatisgalen: Thanks for this great suggestion, just moved the visibility logic into the map popup provider

arches/app/media/js/utils/map-popup-provider.js
whatisgalen: fix in latest

arches/app/media/js/views/components/search/map-filter.js
```diff
@@ -274,6 +275,34 @@ define([
...
+                this.draw.set({
+                    "type": "FeatureCollection",
+                    "features": [feature]
+                });
+                if (currentSearchGeoms != null) {
+                    currentSearchGeoms.push(feature);
+                } else {
+                    currentSearchGeoms = [feature];
+                }
+                self.searchGeometries(currentSearchGeoms);
```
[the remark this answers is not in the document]
whatisgalen: the `restoreState` case is a tricky one. I'm wondering if it makes sense to disallow restoring against a map query that lacks the actual feature coordinates and just loads without a map query at all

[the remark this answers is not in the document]
whatisgalen: If the popup feature lacks a resourceid, its geometry does get stored in `this.filter.feature_collection` [here](https://github.com/archesproject/arches/blob/a6de407a2eea7c4d3dee8b6cf47187c3d70e476e/arches/app/media/js/views/components/search/map-filter.js#L473-L497).

arches/app/templates/views/components/map-popup.htm
whatisgalen: > Also wondering if the visibility of this should also be controlled by an option in the Filter->Select a Filter select box to be consistent with other Map Filter functions.

That sounds reasonable; my main hesitation is that the UX feels a lot more intuitive to me if I can click a feature and say "I want to search using *this*" vs...
- first going to the map filter
- then selecting a specific filter functionality ("Filter by Popup")
- then selecting the desired popup feature

whatisgalen:
Just created #10907 regarding multi-feature map-filter queries

arches/app/templates/views/components/map-popup.htm
bferguso: I can buy that argument :).

arches/app/media/js/views/components/search/map-filter.js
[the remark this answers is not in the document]
whatisgalen: I made a change in my latest commits so that `restoreState` will work as expected, drawing a buffered polygon on the map

whatisgalen:
@bferguso I think I got every requested change in here (minus the multi-feature thing, made a new issue for that) and tests are passing. Let me know if you have further feedback upon testing.

arches/app/media/js/utils/map-popup-provider.js
```diff
@@ -1,7 +1,9 @@
-define(['arches',
+define([
+    'arches',
     'knockout',
+    'turf',
```
bferguso, line 4: This is unused so I think it should be removed.

arches/app/media/js/utils/map-popup-provider.js
```diff
@@ -40,6 +42,34 @@ define(['arches',
...
+         * This method enables custom logic for how the feature in the popup should be handled and/or mutated en route to the mapFilter.
+         * @param popupFeatureObject - the javascript object of the feature and its associated contexts (e.g. mapCard).
+         * @required @method mapCard.filterByFeatureGeom()
+         * @required send argument: @param feature - a geojson feature object
+         * @optional send argument: @param resourceid
+         */
+        sendFeatureToMapFilter: function(popupFeatureObject)
+        {
+            let feature = popupFeatureObject.geometries()[0].geom.features[0];
+            // Note that polygons with very high vertex-counts can benefit from simplification.
```
bferguso, line 55: Perhaps this strategy should be included in the documentation rather than in the code base?

arches/app/search/components/map_filter.py
```diff
@@ -28,47 +30,57 @@ def append_dsl(self, search_results_object, permitted_nodegroups, include_provis
...
+                for geom in hit['_source']['geometries']:
+                    if len(geometries) > 0:
+                        break
+                    for feature in geom['geom']['features']:
+                        if len(geometries) > 0:
+                            break
+                        if feature['id'] == spatial_filter["featureid"]:
+                            geometries.append(feature)
+
+            feature_geom = geometries[0]["geometry"]
```
bferguso, line 78: This throws an exception if the resource instance used as the source of the filter is deleted.

arches/app/media/js/utils/map-popup-provider.js
```diff
@@ -40,6 +42,34 @@ define(['arches',
...
+        /**
+         * This method enables custom logic for how the feature in the popup should be handled and/or mutated en route to the mapFilter.
+         * @param popupFeatureObject - the javascript object of the feature and its associated contexts (e.g. mapCard).
+         * @required @method mapCard.filterByFeatureGeom()
+         * @required send argument: @param feature - a geojson feature object
+         * @optional send argument: @param resourceid
+         */
+        sendFeatureToMapFilter: function(popupFeatureObject)
+        {
+            let feature = popupFeatureObject.geometries()[0].geom.features[0];
```
bferguso, line 54: There is an assumption here that there is only one geometry per node. If multiple geometries exist it always selects the first one. If we use the geometry directly, not trying to use a resourceId & nodeId, there should be a way to pull the geometry selected (in processData() I think it can be found using `features.activeFeature.feature.toJSON().geometry.coordinates`).

arches/app/media/js/views/components/search/map-filter.js
```diff
@@ -274,6 +275,35 @@ define([
                 }
             }, this);
 
+            this.filterByFeatureGeom = function(feature, resourceid=null) {
+                if (feature.geometry.type == 'Point') { self.buffer(100); }
```
bferguso, line 279: Do we have access to the buffer value in the filter settings?

bferguso (changes requested):
At a high level I think it's worth a discussion about whether to use resource ID and feature ID or just the geometry itself. I'm also wondering if a geometry should find itself or not, but that's perhaps getting a bit into the weeds.

arches/app/media/js/utils/map-popup-provider.js
whatisgalen: fix in latest

arches/app/media/js/utils/map-popup-provider.js
whatisgalen: fix in latest, created https://github.com/archesproject/arches-docs/issues/417

arches/app/media/js/utils/map-popup-provider.js
whatisgalen: Good catch! I've altered that function to now expect the index of the corresponding feature to instruct it which one to pick off of `popupFeatureObject.geometries()` -- I've also include it as a param for `showFilterByFeature` just in case further inspection on that particular feature is needed to determine whether it should be eligible as a map filter

arches/app/media/js/views/components/search/map-filter.js
whatisgalen: we have access to the buffer in this context given how it's defined [here](https://github.com/archesproject/arches/blob/dev/7.6.x/arches/app/media/js/viewmodels/map-filter.js#L14) in the map-filter viewmodel

whatisgalen:
> At a high level I think it's worth a discussion about whether to use resource ID and feature ID or just the geometry itself. I'm also wondering if a geometry should find itself or not, but that's perhaps getting a bit into the weeds.

I agree on the question of whether a geometry should include itself in the results, but technically speaking if all you have is the geojson coordinates (as opposed to the featureid and resourceid), you would have to exclude based on an exact match of the point or polygon and given how the geometry is converted into GEOS, I can see one tiny change of the 12th decimal causing that match to fail.

The reason I introduced featureid and resourceid at the outset was to both lighten the request payload (compared to geojson strings) and also to build in the option of trying to achieve the above case.

arches/app/search/components/map_filter.py
whatisgalen: Let me know if this is still giving you an error or exception after my latest push to branch

arches/app/media/js/views/components/search/map-filter.js
whatisgalen: for good measure I changed the above line to set a buffer of 25 and only if the existing value for `this.buffer() == 0`

arches/app/media/js/views/components/search/map-filter.js
```diff
@@ -274,6 +275,35 @@ define([
...
+                        "resourceid": resourceid,
+                        "buffer": {
+                            "width": this.buffer(),
+                            "unit": this.bufferUnit()
+                        },
+                        "inverted": this.filter.inverted()
+                    });
+                }
+                let currentSearchGeoms = self.searchGeometries();
+                this.draw.set({
```
bferguso, line 292: This line is giving me a `Uncaught TypeError: cyclic object value for WMS features because the `feature` is also one of the `feature.properties` values. Looking into how to resolve this.

whatisgalen: in your `map-popup-provider.js` file does it work by setting `feature.properties = {};` ?

arches/app/media/js/views/components/search/map-filter.js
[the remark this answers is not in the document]
bferguso: I'm finding that if a feature geometry filter is applied, the search is refreshed, and then the `Map Filter Enabled` search criteria is removed, the filter isn't removed from the URL and the filter remains applied. 

image

Adding another `Filter by Feature` then shows a Request Failed banner, and continues to add geometries to the map filter.

image

arches/app/templates/views/components/map-popup.htm
```diff
@@ -9,65 +9,69 @@
...
+                    <!--/ko-->
+                    <!-- ko if: showEditButton -->
+                    <a data-bind="click: function () {
+                        window.open(editURL + ko.unwrap(resourceinstanceid));
+                    }" href="javascript:void(0)">
+                        <i class="ion-ios-refresh-empty"></i>
+                        <span data-bind="text: $root.translations.edit"></span>
+                    </a>
+                    <!--/ko-->
+                    <a data-bind="click: function() {sendFeatureToMapFilter($data, $index())}, visible: showFilterByFeature($data, $index())" href="#">
```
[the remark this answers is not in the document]
whatisgalen: I'm guessing this has to do with how the `popupFeatureObject` is structured (my assumption in the code is essentially: `feature = geoms[featureGeomIndex].geom.features[0];` but I'm wondering if your test case it's actually: `feature = geoms[0].geom.features[featureGeomIndex];`). Any chance you could log that for me?

whatisgalen:
issues fixed via the latest commits:
- clicking a new feature replaces the old one in the filter with the new, as expected
- the featureid from the clicked feature is used to look up the matching one in the resource descriptor geometries' filters
- the lookup via featureid in the backend map-filter logic has been removed. All payloads now have a geojson feature collection

bferguso (changes requested):
This is looking really good!

I came across one issue - I created a bunch of geometries using the trigger views to do some load testing, and the trigger views don't add the featureid to the `tiledata`. We'll need to populate that value from the trigger views to allow geometries to be added that way.

The featureid is there for geometries created through the system. Wondering if:
1. There should be a migration that adds any missing `featureid`s if they're not present.
2. The trigger views need to be updated to populate `featureid` if populating geometries that way.

I don't use CSV import or bulk importer, but wondering if those processes do/do not populate `featureid` as part of those processes.

I'll manually add the featureid to the tiles that are missing it, but I think the missing values would be problematic for users with geometries not created with the UI.

arches/app/media/js/utils/map-popup-provider.js
```diff
@@ -40,6 +41,45 @@ define(['arches',
...
+
+        /**
+         * Determines whether to show the button for Filter By Feature
+         * @param popupFeatureObject - the javascript object of the feature and its associated contexts (e.g. mapCard).
+         * @returns {boolean} - whether to show "Filter by Feature" on map popup
+         * typically dependent on at least 1 feature with a geometry and/or a featureid/resourceid combo
+         */
+        showFilterByFeature: function(popupFeatureObject) {
+            let foundFeature = null;
+            const strippedFeatureId = popupFeatureObject.feature.properties.featureid.replace(/-/g,"");
```
bferguso, line 72: This throws and exception if the feature doesn't have a featureid. Suggest doing a check for existence first and return false if there is no featureid.

Perhaps refactor to extract common logic that searches for the feature to avoid code duplication? (`findSelectedFeature()`?)

bferguso (changes requested):

chiatt (approved):

chiatt:
@bferguso It looks like @whatisgalen addressed the change you requested. There was a scoping issue that popped up after that fix (which I think you may have observed) but Alexei fixed that in [a90dfaa](https://github.com/archesproject/arches/pull/10827/commits/a90dfaa888bc11f1dfa13008ebd24323e62fec50). I also noticed that the filter button would appear in the map widget within the resource editor. Alexei was able to take care of that as well.

bferguso (approved):
Looks good!

whatisgalen:
Per my [comment](https://github.com/archesproject/arches/pull/10827#issuecomment-2305886850), the featureid was removed from the request payload to the backend (map-filter) because if the user edited the drawn polygon on the map, (for example to cover a larger spatial query area) the feature identified by featureid would differ from the query geometry, and there wasn't a great way to confirm whether the geometry had actually changed (given potential precision differences). This would have yielded search results out of sync with the search query.

However, because the only payload is the geojson feature itself, this means large coordinate sets can exceed the character limit on the GET request header (and url length in the browser). Unfortunately, those requests then fail, hence #11578 and the need for #11582

Issue #10665, opened 2024-03-07
Update ActivityStreams to use an id not url for better error handling in the JSON-LD reader


apeters:
property name should by "id" instead of "url" (found here http://localhost:8000/history)

![Image](https://github.com/archesproject/arches/assets/823773/cb94037c-43ef-4121-b867-67619d51ca5a)

PR #10830, opened 2024-04-24
Use resource id rather than url in ActivityStreams #10665

johnatawnclementawn:
### Types of changes

### Description of Change

Configure ActivityStreams objects to return resource id's rather than resource url.

### Issues Solved

#10665


jacobtylerwalls (approved):
Looks good! Took the liberty of adding a test.

PR #10833, opened 2024-04-25
Abstracting permissions frameworks

philtweir:
### Types of changes

(Ideally, this would not be a breaking change but given it touches quite a few places in code, needs to be confirmed!)

### Description of Change

Following up on ideas and challenges that have come up for us at [Flax & Teal](https://github.com/flaxandteal), where the default permissions approach is not optimal for our use-cases, this aims to make permissions more pluggable.

### Issues Solved

#

### Checklist

_TODO_

-   [x] Unit tests pass locally with my changes (permissions tests do; some unrelated tests not working in my setup, but going to try and confirm in CI)
-   [x] I have added tests that prove my fix is effective or that my feature works (NB: very basic so far, but keen to check direction good)

#### Ticket Background
*   Sponsored by: Required in the course of work by Flax & Teal for Northern Ireland DfC Historic Environment Division (HED) 
*   Found by: @philtweir
*   Tested by: _TODO_ (although a couple of tests have been added)
*   Designed by: @philtweir

### Further comments

[Screencast from 2023-09-23 22-57-41.webm](https://github.com/flaxandteal/arches/assets/1624699/16511af9-527a-4922-8544-80a80456d6f9)

### What is being suggested?

Centralizing (and abstracting) any `django-guardian` calls such that alternative RBAC logic can be swapped in, in a similar way to how storage backends, search components, datatypes, etc. can be currently.

This pulls all the calls into `arches.app.utils.permission_backend`, which holds a singleton `PermissionFramework` object, who must implement all of these methods. Implementations of `PermissionFramework` (an abstract class) can be defined, in `arches.app.permissions` or in Arches Applications, either from scratch or by subclassing another `PermissionFramework`, and set in `settings.py`. Note that `PermissionFramework` is a simple Python class, and is not (but could be, if we would rather) a Django PermissionBackend. 

### Why is this necessary?

At present, users or groups are red-listed rather than green-listed by default, which works well when most data is open, but for instances with substantial amounts of non-public data, it can slow indexing for large numbers of users and falls back to allow if any coding/role-assignment issues occur. Moreover, in these cases, some of the (cascading sequences of) checks make group logic less intuitive, and require hooking extra logic on save to get, for instance, per-group visibility of resources.

---

Screencast shows adding a resource that is visible in search both to the logged-in admin who creates it, and to a public user (Private Browsing). The permissions framework is switched to default-deny in `settings.py` and the server restarted. The Private Browsing window no longer shows any results. (Note, however, that the count does not reflect the filtering - this may be a general issue)

### Outstanding actions

- [ ] `default_deny` needs debugged
- [ ] routines that we have not been using in `arches_standard` may not be tested out
- [ ] tidy the refactor to make sure it still works as expected (our F&T version had accreted some extra changes, that I have pared back)
- [ ] *the implementation here sacrifices clean concise code (esp in `permissions_backend.py`) for least interference*, keeping the same function names everywhere, where possible - this probably is not desirable in the long run, but I would suggest merging this way is easier to amend going forward. That is, for least risk and surprise, we merge something that is a "clearing house" for permissions logic and fits the existing interfaces, instead of a nicer abstraction that would require changing how permissions checks work throughout the codebase (i.e. here, we only need to change `import` statements from `guardian` to `permission_backend` in most existing files).


arches/settings.py
```diff
@@ -371,6 +372,7 @@
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     # "django.middleware.clickjacking.XFrameOptionsMiddleware",
+    #"dauthz.middlewares.request_middleware.RequestMiddleware",
```
philtweir, line 375: These can go, but maybe useful for any reviewers to see what plugging is required, until PR closer to ready.

philtweir:
To avoid doubt - this PR does not include any functionality that I was talking about before re. Casbin, RBAC logic, "principal" users/groups owning resources, Arches ORM, etc. but it does underpin these things in our downstream fork. If any of that fun stuff is desirable for us to upstream to core or build a re-usable Application(s) around, very happy to, but getting this tidied and merged would be Step 1!

whatisgalen:
minor nitpick [here](https://github.com/archesproject/arches/pull/10833/files#diff-d8cc4e918fb19621d4d55d5e0b1751b5ace0062a9fa2542bcb9caa1b571bb928R18) but is there a particular reason to evaluate for a Falsey value (`== False`) vs evaluating for False ('is False')?

arches/app/permissions/arches_default_deny.py
```diff
@@ -0,0 +1,33 @@
...
+    def check_resource_instance_permissions(self, user, resourceid, permission):
+        result = super().check_resource_instance_permissions(user, resourceid, permission)
+
+        if result and result.get("permitted", None) is not None:
+            # This is a safety check - we don't want an unpermissioned user
+            # defaulting to having access (allowing anonymous users is still
+            # possible by assigning appropriate group permissions).
+            if result["permitted"] == "unknown":
+                result["permitted"] = False
+            elif result["permitted"] == False:
```
philtweir, line 18: Great question @whatisgalen :D Should be an `is`

philtweir:
Removed draft as it's at a point for further discussion - appreciate there is more to do on this before it is mergeable!

philtweir:
As it is a performance bottleneck (at least in our setting) that can otherwise be cached, one of the changes I had made was to get `get_nodegroups_by_perm` to return keys, rather than nodegroups -- then the permissions framework does not have to look up the nodegroups if it can map the permissions without loading them all. That means our one returns a list of strings, instead of a list of models, so I have done that here too -- however, if the preference would be to revert to a list of nodegroups, will do so!

whatisgalen:
> Fact I realised may not be obvious - the _default deny_ here (which will need further discussion) is not for nodegroups, but for instances. That is, it does not default-deny access to parts of the graph, but to resource instances that there is no explicit access to. If the former is preferable, that should be a fairly simple switch/combination.
> 
> We found a naive default-deny framework creates two challenges:
> 

This change makes a lot of intuitive sense to me. If I create a new resource model intended for a minority subset of users in Arches 7.6 right now, I need to exclude all other users. Because the default is currently that all users have access to a new Resource Model, one only thinks of group permissions in the case of denial or restriction. I think there would be more engagement with how we define group and role-based permissions if one were prompted to specify permissions out the outset.

..also agree this would warrant quite a few changes in the codebase, but absolutely warranted imo.

chiatt:
> Fact I realised may not be obvious - the _default deny_ here (which will need further discussion) is not for nodegroups, but for instances. That is, it does not default-deny access to parts of the graph, but to resource instances that there is no explicit access to. If the former is preferable, that should be a fairly simple switch/combination.
> 
> We found a naive default-deny framework creates two challenges:
> 
> * lots of instances viewable by lots of users means lots of `view_resourceinstance` permissions need added
> * users need to be given permissions post-sign up (automatically or manually) to see anything at all
> 
> so, most likely, an implementer would want to subclass `ArchesDefaultDenyPermissionFramework` to add programmatic rules appropriate to their deployment, perhaps with Django groups.
> 
> Moreover, it currently does not cover ES searches - I can add that (as we have done something along these lines in our own project), but it involves changes to `arches.app.search.*` so keen to get an OK on direction of travel for the core component here first. I have added `get_sets_for_user` to the PermissionFramework as a placeholder (_set_ being a named grouping of instances/permissions that a user can access, with implementation details left up to the framework), but in this PR it does not action anything (yet).

Thanks for your comments here Phil - they're very helpful. I think ultimately we may want to use a consistent default deny strategy for nodegroups and instances, but it makes sense to me to start with just instances for now. The need for applying default deny to resource instances is much more urgent. In regards to the two challenges you mention, I think lots of users + lots of instances can be handled by relying on groups. In cases when a new user is added, I think it's ok that they can't see anything. In fact, that's very desirable.

arches/app/views/base.py
```diff
@@ -52,7 +52,9 @@ def get_context_data(self, **kwargs):
             if self.request.user.has_perm("view_plugin", plugin):
                 context["plugins"].append(plugin)
 
-        createable = get_createable_resource_types(self.request.user)
+        createable = list(
+            models.GraphModel.objects.filter(pk__in=list(get_createable_resource_types(self.request.user))).all()
```
chiatt, line 56: I get the following error in the report view when I reach this line:
`django.core.exceptions.ValidationError: ['“<model name>” is not a valid UUID.']`

The `get_createable_resource_types` method already returns a list of graphs that a user has access to. 
```
(Pdb) get_createable_resource_types(self.request.user)                            
[<GraphModel: Report>, <GraphModel: Test Concept Load>, <GraphModel: Creature>, <GraphModel: Resource>, <GraphModel: Test>]
```
If I revert this line back to ```createable = get_createable_resource_types(self.request.user)``` the report view loads as expected.

chiatt (changes requested):
When I opened the search page I got and error that `SYSTEM_SETTINGS_RESOURCE_MODEL_ID = "ff623370-fa12-11e6-b98b-6c4008b05c4c"` was not available in the settings. Interestingly, that setting is defined only in `system_settings.py` and not in `settings.py`. It seems like it should be in both places, but @apeters might have more insight there. 

In any case, to get the search page to load I had to add `SYSTEM_SETTINGS_RESOURCE_MODEL_ID` to my settings.py file. I'm not sure why the behavior is different on this branch, but maybe it doesn't matter. To address this, `SYSTEM_SETTINGS_RESOURCE_MODEL_ID = "ff623370-fa12-11e6-b98b-6c4008b05c4c"` should just be added to `settings.py`

arches/app/views/base.py
philtweir: It looks like this was a mistake in the rebasing - Thanks @chiatt ! In our fork, I had changed it to list IDs - the main advantage is:

* most uses of `get_createable_resource_types` or `get_nodegroups_by_perm` do not use the full graph object, only a count or IDs
* as they are used in permissions, they get hit every request and we saw a decent speed-up if we could skip a full Graph/NodeGroup load

By making these return strings, the permissions framework can pull back less from the DB in (e.g.) API requests or, if it can cache (as our Casbin implementation) then it can fulfill these without needing to hit the DB at all.

If that logic makes sense then it might be tidier than what I have here, if I move this Django model call into a `get_createable_resource_models` wrapper, and pull in the `get_createable_resource_types` (to return `list[str]`) change from our fork?

chiatt: Thanks Phil - Yeah, that sounds like a really nice improvement.

chiatt:
Hey @philtweir - I created a new branch in the arches repo off of the dev/7.6.x branch called `permissions_core`. I've changed the base of this PR to that branch so that we can get your work moved into the arches repo a little sooner than if we were merging it directly into dev/7.6.x. This should make it a little easier for others to help contribute to this work. Specifically I think we can help update `permissions_mananger.js` to fully support default deny.

arches/app/views/base.py
philtweir: Added in.

PR #10834, opened 2024-04-25
Return blank tiles from get_tile_data() #10829

### LINKED ISSUE #10829: Tiledata value where all jsonb values are null causes datatype.get_display_value to fail (7.5.x)
Resources that have tiles where all values are None are causing get_display_value to fail for nodes in them.

image

image

This appears to be caused by a previous bug fix for `BaseDataType.get_tile_data` that now causes the function to return None if only None values are found.

Commit: https://github.com/archesproject/arches/pull/10115/commits/e4a519e531aae9e31b3671a116fdfb72a6d8e62e

A number of the `datatype.get_display_value` functions do not handle being returned:

- concept
- concept-list
- date
- edtf
- file-list
- domain
- resource-intance
- node (handled but incorrect error message).

I am unsure if there are rules stopping these type of value empty tiles being created, but this issue has been made obvious having migrating an existing dataset

jacobtylerwalls:
### Types of changes

### Description of Change
Return blank tiles (holding only None values) from `get_tile_data()` rather than returning an implicit None. For example, return `{node_id: None}` rather than `None`.

Datatypes often drill into the result of `get_type_data()` expecting a dict, and do not check for `None`. (Perhaps they should, since there is still a code path that could return None after logging an exception. Or, we could elevate that logged exception to a runtime exception, but that's outside the scope here.)

Tiles might hold only None if a tile is added for a resource and then edited to remove all values. Or they might have been created in a data migration (#9704, 7.5.0).

Refs 4b02d35

### History
https://github.com/archesproject/arches/commit/4b02d35d097230f273d34e8aac86204efe5d2057 changed the if/else logic so that tiles with only falsy values log an exception and return None implicitly (7.5.0)
https://github.com/archesproject/arches/commit/e4a519e531aae9e31b3671a116fdfb72a6d8e62e tightened that up to "only None values" to exempt falsy values like [] in the file-list datatype. (7.5.0)
be7a130284f5aa955bbfa73ad10051eb1333d1e2 removed the logged exception for blank tiles but kept the implicit None return (7.5.2)

Targeting dev/7.5.x as this is a regression in 7.5.0.

### Issues Solved

Closes #10829

### Checklist

#### Ticket Background
*   Sponsored by: Farallon
*   Found by: @aj-he

### DIFF
--- a/arches/app/datatypes/base.py
+++ b/arches/app/datatypes/base.py
@@ -202,6 +202,8 @@ class BaseDataType(object):
             return provisionaledits[userid]["value"]
         elif not data:
             logger.exception(_("Tile has no authoritative or provisional data"))
+        else:
+            return data
 
 
     def get_display_value(self, tile, node, **kwargs):
--- a/releases/7.5.3.md
+++ b/releases/7.5.3.md
@@ -1,0 +1,52 @@
+Arches 7.5.3 Release Notes
+--------------------------
+
+### Bug Fixes and Enhancements
+
+- Prevent get_tile_data() from returning None for tiles holding only None, #10829
+
+
+### Upgrading Arches
+
+1. Upgrade to version 7.5.0 before proceeding. If upgrading from an earlier version, refer to the upgrade process in the [Version 7.5.0 release notes](https://github.com/archesproject/arches/blob/dev/7.5.x/releases/7.5.0.md)
+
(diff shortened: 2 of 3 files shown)


aj-he:
Seems to have fixed the initial error. I am running some load package tests.

aj-he (approved):
this fixes the issue

Issue #9708, opened 2023-06-30
Importing datetime values require a timezone, but should be optional


azaroth42:
xsd:dateTime and ISO8601 do not require timezones to be present, however when importing a dateTime value through JSON-LD, the node will note match without a timezone explicitly in the value.

For example, sending

```
{
  "begin_of_the_begin": "1978-04-18T00:00:00Z",
  "end_of_the_end": "1978-04-18T23:59:59"
}
```

deals with the earlier date correctly and fails for the end date with the following error:

```
ValueError: Data does not match any actual node, despite prop/class combination http://www.cidoc-crm.org/cidoc-crm/P82b_end_of_the_end http://www.w3.org/2001/XMLSchema#dateTime:
{'@type': 'http://www.w3.org/2001/XMLSchema#dateTime', '@value': '1978-04-18T23:59:59'}
```

Propose that all datetimes should have Z automatically appended if sent without a timezone.

azaroth42:
Or that the `DATE_FORMATS` setting have the timezone-less options by default (having re-found that particular setting)

ryan86:
It appears the Z timezone has meaning: 

```If the time is in UTC, add a Z directly after the time without a space. Z is the zone designator for the zero UTC offset. "09:30 UTC" is therefore represented as "09:30Z" or "T0930Z". "14:45:15 UTC" would be "14:45:15Z" or "T144515Z".```

If time zone is required, maybe Z is as good as any. Would it be better to prompt the user or to add a flag to the command to add a timezone value with a default being 'Z'?

apeters:
I think the consensus is that we should add a flag on json-ld import to allow users to specify the default timezone if none is given in the data.

jacobtylerwalls:
Fixed in #10836

PR #10836, opened 2024-04-25
Add a default timezone arg to jsonld import, re yale#9708

ryan86:
Add timezone flag to jsonld import to indicate timezone for data that may not include a timezone by default, re #9708


arches/app/utils/data_management/resources/formats/rdffile.py
```diff
@@ -650,6 +650,13 @@ def find_matching_branch(self, k, v, tree_node, result, tile=None, indent=0):
                         else:
                             self.printline(f"Could not validate {values} as a {o['datatype']}", indent + 1)
                     else:
+                        if o["datatype"].datatype_name == 'date':
+                            try:
```
apeters, line 654:
```suggestion
                        if self.default_timezone and o["datatype"].datatype_name == 'date':
```

ryan86: Won't this fail if the value can't be parsed on this line

`datetime.datetime.strptime(value, "%Y-%m-%d %H:%M:%S")`

apeters: sorry meant my suggestion to be this:
```suggestion
                        if self.default_timezone and o["datatype"].datatype_name == 'date':
                            try:
```

apeters (approved):
looks good

Issue #10843, opened 2024-04-29
Check geometries for maximum Lucene byte size 


chiatt:
The geojson-feature-collection datatype prevents geometries from being saved with greater than 1500 vertices. This limit is in place because Lucene restricts strings to 32766 bytes. Without this limit a geom that exceeds 32766 bytes would fail to index. Fifteen hundred vertices was a safe estimate for the number of coordinates that would exceed the Lucene cap. 

Before this limit was imposed, we were using a string length limit, however, that number didn't have as much meaning to users as a vertex count and it wasn't very reliable because a unicode string can exceed 32766 bytes with fewer characters.

The problem with the vertex approach is that a valid geometry (with fewer than 32766 bytes) may be rejected if exceeds 1500 vertices. So, Instead of relying on string length or vertices, we should get the byte count, allowing as many geometries to pass as possible.

If the limit is exceeded, the user should be notified in the UI exactly why their resource failed to validate.

From the ES docs:
```The value for ignore_above is the character count, but Lucene counts bytes. If you use UTF-8 text with many non-ASCII characters, you may want to set the limit to 32766 / 4 = 8191 since UTF-8 characters may occupy at most 4 bytes.```
https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html

ryan86:
Superseded by #10855

PR #10845, opened 2024-04-30
Change vertice count to byte count when validating geoms re Check geometries for maximum Lucene byte size  #10843

ryan86:
Change vertices count to byte count to conform with Lucene documentation on large geoms. re #10843


ryan86:
Superseded by #10855

Issue #10798, opened 2024-04-18
Yale - Import data from external source to arches via UI


azaroth42:
A tracker issue to help Rob / Yale better import data from one system into Arches via a UI.

In particular:
  * A page should allow an authenticated user to enter a URL and LUX would then retrieve that record as JSON-LD and import it, with the existing UUID if possible
  * The import process should also import the records that the initial one referenced by retrieving them individually
  * It should allow the user to request a zip file of records from LUX to avoid the above repeated calls
  * It should report errors to the user as to which records did not import, hopefully with an instructive error message as to why they didn't import
  * The multi-import steps could be a workflow if that would be easier/better?

Bad example code:
[example_code.zip](https://github.com/archesproject/arches/files/15028987/example_code.zip)

whatisgalen:
@azaroth42 are you spec'ing this out as an import module?

apeters:
> * It should allow the user to request a zip file of records from LUX to avoid the above repeated calls

@azaroth42 At the moment we'll be focusing on just allowing users to upload a zip file of resources formatted with the same directory structure as the current json-ld loader uses from the command line.

apeters:
@azaroth42 Also, we'll assume that this operation either overwrites, or adds a new resource, like the current json-ld loader command does.

azaroth42:
Yes, agreed re overwrite/add. Building the zip file we can work on. Doesn't seem too hard for the small number of records per interaction to build it only in memory.

Issue #10860, opened 2024-05-02
Flaky test: `test_save_provisional_from_athoritative`


jacobtylerwalls:
Failure occasionally seen on CI looks like:

```py
ERROR: test_save_provisional_from_athoritative (tests.models.tile_model_tests.TileTests.test_save_provisional_from_athoritative)
Test that a provisional edit is created when a user that is not a
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/arches/arches/tests/models/tile_model_tests.py", line [26](https://github.com/archesproject/arches/actions/runs/8925493498/job/24514272746#step:11:27)7, in test_save_provisional_from_athoritative
    self.assertEqual(provisional_tile.data["72048cb3-adbc-11e6-9ccf-14109fd34195"]["en"]["value"], "AUTHORITATIVE")
                     ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: '72048cb3-adbc-11e6-9ccf-14109fd34195'

----------------------------------------------------------------------
Ran 332 tests in [29](https://github.com/archesproject/arches/actions/runs/8925493498/job/24514272746#step:11:30)8.753s

FAILED (errors=1)
```

PR #10863, opened 2024-05-02
Add lenient file-type checking mode #10862

### LINKED ISSUE #10862: Add "lenient" file-checking mode
`validate_file_type()` runs some [safety checks](https://github.com/archesproject/arches/blob/0c89e7d117f4c639f1ce706f3dfaa82462fb1ca4/arches/app/utils/file_validator.py#L51) on uploaded files:
- determinable file type
- safe zip file
- well-formatted xlsx or csv files
- permitted file type

... but only if `FILE_TYPE_CHECKING = True`.

Proposal:
- ~change new project template to set FILE_TYPE_CHECKING = True in 7.6~
- ~Emit DeprecationWarning if it's set False~
- ~Remove the setting in 8.0~
- See #10863

jacobtylerwalls:
### Types of changes

### Description of Change
There are several useful safety checks guarded behind the FILE_TYPE_CHECKING setting, not all of which have to do with the file type, strictly speaking.

The main reason this was off by default was the lack of a "lenient" option to allow through files whose extensions cannot be determined.

Add this lenient mode, default new projects to it, and deprecate the old true/false values.

### Issues Solved
Closes #10862

### Checklist

#### Ticket Background
*   Sponsored by: Yale (Found while dev'ing #10798) & Farallon

### Demo
Run any management command (runserver, check, test, etc.) with `python -Wall::DeprecationWarning manage.py ...`
```py
/web_root/arches/arches/apps.py:13: DeprecationWarning: Providing boolean values to FILE_TYPE_CHECKING is deprecated. Starting with Arches 8.0, the only allowed options will be None, 'lenient', and 'strict'.
```

### DIFF
--- a/arches/app/etl_modules/jsonld_importer.py
+++ b/arches/app/etl_modules/jsonld_importer.py
@@ -99,6 +99,27 @@ class JSONLDImporter(BaseImportModule):
                 ),
             }
 
+        try:
+            self.read_zip_file(content, result)
+        except zipfile.BadZipFile as e:
+            return {
+                "success": False,
+                "data": FileValidationError(
+                    message=e.args[0],
+                    code=400,
+                ),
+            }
+
+        if not result["summary"]["files"]:
+            title = _("Invalid Uploaded File")
+            message = _(
+                "This file has missing information or invalid formatting. Make sure the file is complete and in the expected format."
+            )
+            return {"success": False, "data": {"title": title, "message": message}}
+
+        return {"success": True, "data": result}
+
+    def read_zip_file(self, content, result):
         with zipfile.ZipFile(content, "r") as zip_ref:
             files = zip_ref.infolist()
             for file in files:
@@ -126,15 +147,6 @@ class JSONLDImporter(BaseImportModule):
                         destination = destination / part
 
                     default_storage.save(destination, f)
-
-        if not result["summary"]["files"]:
-            title = _("Invalid Uploaded File")
-            message = _(
-                "This file has missing information or invalid formatting. Make sure the file is complete and in the expected format."
-            )
-            return {"success": False, "data": {"title": title, "message": message}}
-
-        return {"success": True, "data": result}
 
     def validate_uploaded_file(self, file):
         path = Path(file.name)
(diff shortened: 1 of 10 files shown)


aarongundel (commented):
I like where this is heading, but I do have some concerns here.  

One of the problems with file type checking (and the primary reason that it's optional) is because the filetype guesser is not exhaustive.  There are lots of file types that it will not (and cannot) validate.  A simple example is a text file with a txt extension.  So if file type checking is permanently true, even if you add "txt" to FILE_TYPES - it will be returned as "unknown" and validation will fail.  See test_unknown_filetypes in file_validator.py.  We do some manual checking of xlsx and csv in there for this reason (they cannot be validated any other way - and this is admittedly not the best way to validate them - that is, actually reading potentially malicious content into the csv reader to verify that the file is safe!).  I don't know of a good way around this issue.  We could potentially allow for custom validation of files.  That is to say, if you want a txt file, you can write a custom validator for said txt files and use your own logic to verify that it's acceptable.  Maybe this is not an issue that needs to be solved now, since the option is still around until 8.  But if we are telling people to turn it on to test it - a subset of users are almost certain to do so in production, in which case we're going to get questions on why files are not uploading after enabling the setting we warned them to enable.

arches/app/etl_modules/jsonld_importer.py
```diff
@@ -99,6 +99,27 @@ def read(self, request):
...
+        try:
+            self.read_zip_file(content, result)
+        except zipfile.BadZipFile as e:
+            return {
+                "success": False,
+                "data": FileValidationError(
+                    message=e.args[0],
+                    code=400,
+                ),
+            }
```
jacobtylerwalls, line 111: Noticed that in #10885 I forgot to surface this error in the UI where FILE_TYPE_CHECKING is false and we go straight to trying read a file as a zip file.

tests/models/tile_model_tests.py
```diff
@@ -312,6 +312,11 @@ def test_save_provisional_from_athoritative(self):
 
         provisionaledits = provisional_tile.provisionaledits
         self.assertEqual(tiles.count(), 2)
+        self.assertIn(
```
jacobtylerwalls, line 315: Unrelated, but not really worth a whole new PR for it: follow-up to #10860 where I wanted to add some more debug output, but was stymied by the KeyError being raised much earlier.

aarongundel (approved):
This is a a nice improvement!

jacobtylerwalls:
Thanks for the review!

Issue #10868, opened 2024-05-03
importing concept preflabels in transform_value_for_tile should also call `.strip()` to remove whitespace


whatisgalen:
older data exports of Arches might have leading or trailing whitespace because of whatever string formatting was being used. Rather than asking an importer to manually remove those, the datatype should just strip it the same way it does for validating concept uuid values.

PR #10869, opened 2024-05-03
strips leading, trailing whitespace from concept labels in transform_…

whatisgalen:
…value_for_tile, re #10868

### Types of changes

### Issues Solved

#10868

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/datatypes/concept_types.py
+++ b/arches/app/datatypes/concept_types.py
@@ -209,6 +209,7 @@ class ConceptDataType(BaseConceptDataType):
             if value == "":
                 value = None
             else:
+                value = value.strip()
                 try:
                     value = self.lookup_labelid_from_label(value, kwargs)
                 except:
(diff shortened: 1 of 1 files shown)


arches/app/datatypes/concept_types.py
```diff
@@ -209,6 +209,7 @@ def transform_value_for_tile(self, value, **kwargs):
             if value == "":
                 value = None
             else:
+                value = value.strip()
```
chiatt, line 212: I'm a little concerned that there could be existing systems with trailing spaces in their labels and that this change would break their import. 

I think the way that we could address this is rather than stripping here, we could strip in `lookup_label` doing something like:

```if concept[1] in (label, label.strip())```

That also has the advantage of striping in only once rather than in both `transform_value_for_tile` methods.

chiatt (changes requested):

arches/app/datatypes/concept_types.py
whatisgalen: excellent suggestion, fix in latest

PR #10871, opened 2024-05-03
10090 intuitive term search

whatisgalen:
### Types of changes

### Description of Change

- When a Term Match is selected from the search_terms dropdown in search, that query will now contain both a term match query *and* a nodegroup match query. This means it will query only tiles of the context label's nodegroup and no other strings on the document body.
- When typing in a string in the term search, it will now be labeled "Contains Term: [string]" to make explicit what it will search for
- Resources whose displaynames match a case-sensitive term will be boosted higher in the search term results
- Search Term filter tests now include a nodegroupid in the hard-coded term filter object

### TL;DR
Prior to this PR, Term search UX in arches was:
- enter a search term
- see a list of matching terms: [node name] - [value matching term]
- click a term
- returns results where *any* string *anywhere* on that resource matches your [value matching term]

Proposed UX:
- enter a search term
- ...
- returns results where *only* strings for the *term-match nodegroup* on a resource match [value matching term]

### Issues Solved

#10090

### Checklist

#### Ticket Background
*   Found by: @whatisgalen 
*   Tested by: @ 
*   Designed by: @whatisgalen @apeters


releases/7.6.0.md
```diff
@@ -15,8 +15,6 @@ Arches 7.6.0 Release Notes
 
 - Adds Github actions to build applications, run tests, and compare coverage between branches for projects.
 
-- Plugins now support the configuration boolean `is_standalone`. Standalone plugins do not appear in the sidebar, and do not display the sidebar or application header.
```
chiatt: It looks like this must have been accidentally removed

chiatt (changes requested):

releases/7.6.0.md
jacobtylerwalls: Actually this is correct -- this was already removed in dev/7.6.x, so this looks like a preparatory change for a merge.

chiatt (approved):
Looks good!

PR #10874, opened 2024-05-03
Fix "sortorder" values are not being preserved during a JSON import

jkemper-pca:
When a json file is used for the `import_business_data` command, the "sortorder" values are not being used and all tiles are resetting to all be "sortorder: 0".

### Example

1. Initial Manual Data Entry
*Note the order and the resource name at the top*
Fixed_Report
[Initial_Export.json which is used for all `import_business_data` commands](https://github.com/archesproject/arches/files/15204092/Initial_Export.json)
1. After using `export_business_data` followed by `import_business_data`
*Note the same areas as last step are now in an incorrect order (this is not a consistent order but is random each time)*
Reimported_Report
[Reimported_Export.json](https://github.com/archesproject/arches/files/15204094/Reimported_Export.json)
1. After adding the fix and running `import_business_data` on the same data file as the previous step
*This has the same order as the original data*
Fixed_Report
[Fixed_Export.json](https://github.com/archesproject/arches/files/15204095/Fixed_Export.json)

### Types of changes

### Description of Change
This change adds one line to ingest code for a JSON import and gets the supplied "sortorder" value for a tile similar to how it gets values for "nodegroup_id", "parenttile_id", etc.

### Issues Solved
Now, JSON data that is imported will maintain its order.
#

### Checklist

#### Ticket Background
*   Found by: @jkemper-pca 
*   Tested by: @jkemper-pca 
*   Designed by: @jkemper-pca

### DIFF
--- a/arches/app/utils/data_management/resources/formats/archesfile.py
+++ b/arches/app/utils/data_management/resources/formats/archesfile.py
@@ -176,6 +176,7 @@ class ArchesFileReader(Reader):
                                 "resourceinstance": resourceinstance,
                                 "parenttile_id": str(src_tile["parenttile_id"]) if src_tile["parenttile_id"] else None,
                                 "nodegroup_id": str(src_tile["nodegroup_id"]) if src_tile["nodegroup_id"] else None,
+                                "sortorder": str(src_tile["sortorder"]) if src_tile["sortorder"] else 0,
                                 "data": src_tile["data"],
                             }
                             new_values = {"tileid": uuid.UUID(str(src_tile["tileid"]))}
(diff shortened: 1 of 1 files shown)


whatisgalen (changes requested):
I suspect it isn't breaking the tile save, but I imagine we want sortorder to be cast to `int()` instead of `str()`

whatisgalen (approved):
looks good!

whatisgalen:
@jkemper-pca thanks for submitting this PR! In the future, you can first create an "Issue" and then in a Pull Request specify "Issues Solved: #[number of the issue]" that way others can comment if it's something they might already be working on or if there's more to the problem than meets the eye.

mradamcox:
Just encountered this issue on an old 6.2.0 branch. Thanks for finding and fixing it!

Issue #10875, opened 2024-05-05
Append ResourcInstance datatype ontology metadata to ES document


whatisgalen:
The Resource Relations Index is deprecated. Currently, when ResourceInstance tile data gets indexed as tile data, it appends to the document nearly everything *except the ontology data*. While adding these properties to the document would have implications for string-based searches, there also isn't a way to get this data except via the database. Including the ontology and inverse-ontology properties (either solely as their ids or ids + localized strings) would build a foundation for resource-relationship-based searches that currently cannot exist.

PR #10876, opened 2024-05-05
10875 Includes ontology property in search strings

whatisgalen:
### Types of changes

### Description of Change

- (quality of life improvement) consolidates redundant logic into base datatype method `get_nodevalue_as_list()`, implemented in resource-instance and concept datatypes
- includes `ontologyProperty` and `inverseOntologyProperty` (and the tile nodegroupid, whether provisional) in the document `"strings"` so that if a search term matches the text value of a relationship, it will show up in the suggested terms and be usable as a term query

### Why does this matter?
Resource-instance datatype tiledata isn't searchable in the term search except via the resource displayname of a related instance. The ontology property (relationship) is arguably just as important if not more than the displayname of related instances. Given that there is *zero exposure of any resource's relationship to search*, adding it in to term search as a searchable string seemed a logical start.
### Issues Solved

#10875 

### Checklist

#### Ticket Background
*   Found by: @whatisgalen 
*   Tested by: @ 
*   Designed by: @whatisgalen 

### Further comments

**Note** that #10898 should be merged first as this PR has cherry-picked commits from that PR


arches/app/datatypes/base.py
```diff
@@ -377,6 +377,13 @@ def collects_multiple_values(self):
         """
 
         return False
+    
+    def get_nodevalue_as_list(self, nodevalue):
```
chiatt, line 381: Couldn't this method name just be `get_nodevalues`? I think it's generally good to avoid using types in variable names. This method demonstrates why given that it can return either a list or a tuple.

chiatt (changes requested):
Since concept relationships were introduced to describe relationships, most Arches instances will likely rely on concept labels rather than ontology properties because concepts can be more descriptive and are human readable. This is true even for Arches for Science which has semantic models and once had ontology relationships, but has since migrated to using concept relationships.

I mention this because this PR only supports ontology properties, but probably should support concept relationships as well.

whatisgalen:
> Since concept relationships were introduced to describe relationships, most Arches instances will likely rely on concept labels rather than ontology properties because concepts can be more descriptive and are human readable. This is true even for Arches for Science which has semantic models and once had ontology relationships, but has since migrated to using concept relationships.
> 
> I mention this because this PR only supports ontology properties, but probably should support concept relationships as well.

I had forgotten that defining relationships out of collections was a thing. I tried to test it out but it didn't work. I think we should absolutely include the values in the document strings, though.

arches/app/datatypes/base.py
whatisgalen: fix in latest

whatisgalen:
> Since concept relationships were introduced to describe relationships, most Arches instances will likely rely on concept labels rather than ontology properties because concepts can be more descriptive and are human readable. This is true even for Arches for Science which has semantic models and once had ontology relationships, but has since migrated to using concept relationships.
> 
> I mention this because this PR only supports ontology properties, but probably should support concept relationships as well.

One approach to accomplish the lookup needed in order to index concept value labels as strings (vs their ids) is to leverage the ConceptDataType's lookup_value method. To access that from the `ResourceInstanceDataType` class, PR #10898 would need to be merged.

whatisgalen:
I realized that the concept *datatype* actually just calls a method from the concept *model* [here](https://github.com/archesproject/arches/blob/dev/7.6.x/arches/app/models/concept.py#L1480-L1483). While we could have the `ResourceInstanceDataType` do the same thing, I think it's better to go through the concept datatype as an intermediary anyway.

philtweir:
This looks good to me - this might be a bit of a daft question, but would you have an example of a motivating use-case -- I wasn't sure if I was missing something, since the ontology property is defined on the model, but would this be, for example, if I wanted to find all monuments with a child monument present?

arches/app/datatypes/datatypes.py
```diff
@@ -2413,13 +2408,18 @@ def get_display_value(self, tile, node, **kwargs):
             except:
                 logger.info(f'Resource with id "{resourceid}" not in the system.')
         return ", ".join(items)
+    
+    def get_relationship_display_value(self, relationship_valueid):
+        concept_datatype_instance = self.datatype_factory.get_instance('concept')
+        concept_preflabel = concept_datatype_instance.get_pref_label(relationship_valueid)
```
chiatt, line 2414: This will always return the preflabel in `en-US` because that's the default. You'll need to get the apps active language using something like this:

```
from django.utils.translation import get_language
lang = get_language()
```

All the datatype is doing in `get_pref_label` is calling a concept method. You could just call that same method and then you don't need the datatype:
```
from arches.app.models.concept import get_preflabel_from_valueid
get_preflabel_from_valueid(valid, lang)['value']
```

chiatt (changes requested):

arches/app/datatypes/datatypes.py
whatisgalen: Good catch on the language.
> All the datatype is doing in get_pref_label is calling a concept method. You could just call that same method and then you don't need the datatype

Yea I was reflecting on that for a little while. To be honest it still feels ambiguous to me whether the concept model's method should be called, or whether the concept datatype's method (which I recognize calls the former) should be called. Something about calling the model directly from a different datatype _feels_ like an end-run around the concept-datatype, like _it_ ought to be the intermediary to get a preflabel from a valueid, but it's hard to justify that case given that it simply calls the model method.

whatisgalen:
> This looks good to me - this might be a bit of a daft question, but would you have an example of a motivating use-case -- I wasn't sure if I was missing something, since the ontology property is defined on the model, but would this be, for example, if I wanted to find all monuments with a child monument present?

So if the relationship type (present in resource-instance tile data) matters to your search, this PR lets you search for instances of, say "Historic Resource" resource model, and lets you include "is district contributor" as a search term. If your resource-instance tile data had such a relationship, you're now filtering based on relationships straight from term search as opposed to in the advanced search. Combined with my other PR #10871 this would actually filter the results not just to a string match of "is district contributor", but a nodegroup-based field match (node values for "Related Entities" node must contain a relationship: "is district contributor").

arches/app/datatypes/datatypes.py
chiatt: I think it's better just to use the concept method than the concept datatype. If the resource instance datatype needed a really complex method from the concept datatype, then it might be different because it could require repeating a lot of code. However, in this case the Concept model has exactly what you need without writing anything new. Because of that, using the datatype isn't any DRYer than just using the Concept method. Instead, it  adds an extra layer of code between the resource-instance datatype and the Concept method.

When we move over to using controlled lists in v8, things should be pretty similar. The ControlledList model could just have a method that returns the prefLabel. The datatype doesn't actually need a 'get_pref_label' method because the tile actually stores the label in its data. So when we start using controlled list values to describe relationships then could also use the ControlledList model like you do the Concept model.

whatisgalen: @chiatt changes pushed to latest

whatisgalen:
@apeters any chance I could get a quick review on this? TL;DR is that resource-instance datatype relationship data gets included in the search strings

apeters (changes requested):
I also noticed that when we append a new string to the "strings" section of the resource index document, that we also append a term(terms) to the "terms" index as well so that those "strings" show up in the dropdown.

arches/app/datatypes/datatypes.py
```diff
@@ -2232,29 +2235,73 @@ def to_json(self, tile, node):
...
+                        uuid.UUID(ontology_property_item)
+                        relationship = (
+                            self.get_relationship_display_value(ontology_property_item)
+                            or ontology_property_item
+                        )
+                        terms.append(SearchTerm(value=relationship, lang=lang))
+                    except ValueError:
+                        terms.append(
+                            SearchTerm(value=ontology_property_item, lang=lang)
+                        )
```
apeters, line 2301: I think we should just leave the lang= kwarg blank "" as we really don't know the language of the values

apeters: IE 
```
SearchTerm(value=..., lang="")
```

apeters (changes requested):

apeters (approved):
look good!

Issue #10877, opened 2024-05-06
Assigning string representation of UUID to model instances causes undefined ORM behavior


jacobtylerwalls:
**Problem**
Assigning a string representation of a UUID to a model instance causes undefined behavior in the ORM.

For example, we shouldn't be doing this:
```py
PHYSICAL_THING_GRAPH_ID = "9519cb4f-b25b-11e9-8c7b-a4d18cec433a"
new_resource = Resource(graph_id=PHYSICAL_THING_GRAPH_ID)
```
Someone reported an [issue](https://code.djangoproject.com/ticket/35434) against Django for the same symptom I found in #10743 where an `ObjectDoesNotExist` escaped from `prefetch_related_objects()`.

The reply from the ORM maintainers was: _don't assign stringified UUID's to your model instances; you'll see subtle breakages._

The rationale is that although Postgres will implicitly convert strings to UUID when you hit the database, you shouldn't expect the ORM to constantly perform type coercion everywhere -- when instantiating an instance, you should take care to use the correct types.

Quick demo of sneaky breakages:

```py
In [1]: from models import GraphModel

In [2]: original = GraphModel.objects.first()

In [3]: unsafe = GraphModel(str(original.pk))

In [4]: all_graphs = GraphModel.objects.all()

In [5]: all_graphs.contains(unsafe)
Out[5]: True

In [6]: len(all_graphs)
Out[6]: 1

In [7]: all_graphs.contains(unsafe)
Out[7]: False
```

***
I think we need to audit Arches and fix these. I hacked out a linter plugin to look for affected locations, and I found 162 locations. Not as bad as I thought!

<details>
<summary>Locations</summary>

```
python3 -m pylint --load-plugins arches.check_uuidfield_assignment --disable=all --enable=stringified-uuid-assignment arches tests --recursive=y
************* Module arches.app.utils.v3utils
arches/app/utils/v3utils.py:11:58: E2400: Direct assignment of a stringified UUID to a UUIDField produces undefined behavior. (stringified-uuid-assignment)
************* Module arches.app.utils.data_management.resources.formats.rdffile
arches/app/utils/data_management/resources/formats/rdffile.py:840:28: E2400: Direct assignment of a stringified UUID to a UUIDField produces undefined behavior. (stringified-uuid-assignment)
arches/app/utils/data_management/resources/formats/rdffile.py:842:28: E2400: Direct assignment of a stringified UUID to a UUIDField produces undefined behavior. (stringified-uuid-assignment)
[...]

PR #10881, opened 2024-05-06
Remove workaround in optimize_resource_iteration() that caused N+1 queries

jacobtylerwalls:
Follow-up to https://github.com/archesproject/arches/commit/f2861f22a34e7bb096bbc95bfcf9188d70611ccf.

### Types of changes

### Description of Change
@bferguso loosened the type annotation of `index_resources_using_singleprocessing` to take an iterable of resources, so in f2861f22a34e7bb096bbc95bfcf9188d70611ccf I made sure to handle that. But in doing so I hit the footgun described in #10877. I found a workaround that gave up on n+1 queries at the graph level, but the better fix is to just clean the fields to ensure graph ids are UUIDs.

For more information, see #10877.

### DIFF
--- a/arches/app/utils/index_database.py
+++ b/arches/app/utils/index_database.py
@@ -218,8 +218,7 @@ def optimize_resource_iteration(resources: Iterable[Resource], chunk_size: int):
         )
     else:  # public API that arches itself does not currently use
         for r in resources:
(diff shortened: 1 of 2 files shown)


whatisgalen (approved):
fix works as expected!

PR #10882, opened 2024-05-07
#10848 Support controlled lists in arches packages

### LINKED ISSUE #10848: [controlled-lists] Support controlled lists in arches packages
Add import & export commands to support controlled lists

johnatawnclementawn:
### Types of changes

### Description of Change

Adds the ability to export and import controlled lists from the CLI

** To be merged after #10569 **

### Issues Solved

Resolves #10848

### Checklist


arches/management/commands/packages.py
```diff
@@ -1114,6 +1124,128 @@ def import_reference_data(self, data_source, overwrite="ignore", stage="stage",
...
+                for sheet in wb.sheetnames:
+                    if sheet == "ControlledList":
+                        self.import_sheet_to_model(wb[sheet], models.ControlledList)
+                    elif sheet == "ControlledListItem":
+                        self.import_sheet_to_model(wb[sheet], models.ControlledListItem)
+                    elif sheet == "ControlledListItemValue":
+                        self.import_sheet_to_model(wb[sheet], models.ControlledListItemValue)
+                # validate all data
+                for model in [models.ControlledList, models.ControlledListItem, models.ControlledListItemValue]:
+                    for instance in model.objects.all():
```
jacobtylerwalls, line 1141: Need to limit to validating only the objects created in this command.

arches/management/commands/packages.py
```diff
@@ -1114,6 +1124,128 @@ def import_reference_data(self, data_source, overwrite="ignore", stage="stage",
...
+                        instance.full_clean()
+                self.stdout.write('Data imported successfully from {0}'.format(source))
+        else:
+            self.stdout.write('The source file does not exist. Please rerun this command with a valid source file.')
+            sys.exit()
```
jacobtylerwalls, line 1146:
```suggestion
```
Suggest letting django handle exiting, otherwise AFAICT you can't call this command from other code with `call_command()`.

arches/management/commands/packages.py
```diff
@@ -1114,6 +1124,128 @@ def import_reference_data(self, data_source, overwrite="ignore", stage="stage",
...
+        # Process row data and create instances of the model
+        instances = []
+        list_items_with_parent = {}
+        for row in import_table:
+            instance = model()
+            for field in fields:
+                is_fk = field["is_fk"]
+                field_name = field["name"]
+                value = (row[field_name] if row[field_name] else None) # might be ''
+                if value and is_fk and model == models.ControlledListItem:
```
jacobtylerwalls, line 1169: Comparing with a class I'd suggest `is` instead of equality.

arches/management/commands/packages.py
```diff
@@ -1114,6 +1124,128 @@ def import_reference_data(self, data_source, overwrite="ignore", stage="stage",
...
+                    elif field_name == "parent":
+                        # stash list items with parent relationships to create relationships after all list items have been created
+                        # stashed object in the form of {child_list_item_instance : parent_list_item_pk, ...}
+                        list_items_with_parent[instance] = value
+                elif value and is_fk and model == models.ControlledListItemValue:
+                    if field_name == "valuetype":
+                        valuetype = models.DValueType.objects.get(valuetype = value)
+                        setattr(instance, field_name, valuetype)
+                    elif field_name == "language":
+                        related_language = models.Language.objects.get(code=value)
```
jacobtylerwalls, line 1182: Any of these .get()'s could fail, but at least the error would be pretty self-explanatory. This one, though, I could see catching a user depending on whether or not they have concepts/languages in their new project. Do we want to add a hint around this with a follow-up action?

arches/management/commands/packages.py
```diff
@@ -1114,6 +1124,128 @@ def import_reference_data(self, data_source, overwrite="ignore", stage="stage",
...
+        self.export_model_to_sheet(wb, models.ControlledListItemValue)
+        
+        # if data_dest == ".":
+        #     data_dest = os.path.dirname(settings.SYSTEM_SETTINGS_LOCAL_PATH)
+        if data_dest != "" and data_dest != ".":
+            wb.save(os.path.join(data_dest, "controlled_lists.xlsx"))
+            self.stdout.write("Data exported successfully to controlled_lists.xlsx")
+        else:
+            utils.print_message("No destination directory specified. Please rerun this command with the '-d' parameter populated.")
+            sys.exit()
```
jacobtylerwalls, line 1219: as above

arches/management/commands/packages.py
```diff
@@ -1114,6 +1124,128 @@ def import_reference_data(self, data_source, overwrite="ignore", stage="stage",
...
+        self.export_model_to_sheet(wb, models.ControlledListItem)
+        self.export_model_to_sheet(wb, models.ControlledListItemValue)
+        
+        # if data_dest == ".":
+        #     data_dest = os.path.dirname(settings.SYSTEM_SETTINGS_LOCAL_PATH)
+        if data_dest != "" and data_dest != ".":
+            wb.save(os.path.join(data_dest, "controlled_lists.xlsx"))
+            self.stdout.write("Data exported successfully to controlled_lists.xlsx")
+        else:
+            utils.print_message("No destination directory specified. Please rerun this command with the '-d' parameter populated.")
```
jacobtylerwalls, line 1218: Ideally we'd find a way to do this with `self.stdout.write()`, but if you want to keep the "*" * 80 I guess you could still factor that in.

arches/management/commands/packages.py
```diff
@@ -1114,6 +1124,128 @@ def import_reference_data(self, data_source, overwrite="ignore", stage="stage",
...
+            ws = wb
+        else:
+            ws = wb.create_sheet(title=model.__name__)
+        fields = [{"name": field.name, "datatype": field.get_internal_type()} for field in model._meta.fields]
+        ws.append(field["name"] for field in fields)
+        for instance in model.objects.all():
+            row_data = []
+            for field in fields:
+                value = getattr(instance, field["name"])
+                if (isinstance(value, models.ControlledList) or isinstance(value, models.ControlledListItem) or isinstance(value, models.ControlledListItemValue)):
```
jacobtylerwalls, line 1234: You can pass an iterable of types to isinstance
```suggestion
                if isinstance(value, (models.ControlledList, models.ControlledListItem, models.ControlledListItemValue)):
```

tests/commands/controlled_lists_import_export_tests.py
```diff
@@ -0,0 +1,29 @@
...
+class ControlledListsImportExportTests(ArchesTestCase):
+
+    @classmethod
+    def setUpTestData(cls):
+        from tests.views.controlled_lists_tests import ControlledListTests
+        return ControlledListTests.setUpTestData()
+
+    
+    def test_export_controlled_list(self):
+        management.call_command("packages", operation='export_controlled_lists', dest_dir=settings.TEST_ROOT)
```
jacobtylerwalls, line 29: To clean up this file after the test runs, you can do this before the call_command:
```py
        dest = Path(TEST_ROOT) / "path_to_the_sheet.xlsx"  # needs `from pathlib import Path` at top of file
        self.addCleanup(os.remove, dest)
```

That way, whether call_command passes or fails, the os.remove() will be attempted and will delete your sheet. It's a lot like wrapping the meat of your test in a try/finally. The callback that you would ordinarily put in the finally is what you pass to addCleanup.

Tangent: If call_command fails so early that it doesn't even write a sheet, your cleanup might raise FileNotFoundError, but your test is failing anyway in that scenario, so two errors in that scenario isn't a developer experience terrible enough to spend more time making the cleanup function more robust.

jacobtylerwalls, line 29: If you pass the `stdout` keyword arg to this command with some sort of StringIO() object, it will capture the output so that the test is run quietly. (more "neighborly" to keep your tests quiet in the happy path)

[example](https://github.com/archesproject/arches/pull/10915)

arches/management/commands/packages.py
johnatawnclementawn: I've added a check for missing languages - I'm not sure if we should automatically create a new language if it is missing from the system during this ETL. I seem to recall a conversation about `sys.exit()` being used after that check, but it seems like we're using that elsewhere in `packages.py` in similar situations.

jacobtylerwalls (approved):

jacobtylerwalls:
Just noting that I think (by design) we're omitting images for now.

PR #10885, opened 2024-05-07
Add JSON-LD bulk import module #10798

### LINKED ISSUE #10798: Yale - Import data from external source to arches via UI

jacobtylerwalls:
### Types of changes

### Description of Change
Add a bulk import module that calls the CLI loader for JSON-LD.

Add some hooks to BaseImportModule to make it more extensible for things like:
- shelling out to a CLI command
- wrapping the tile save in a transaction to support overwriting (if resources must be deleted before excess tile checks, that shouldn't become permanent if the tile save later fails)

### Issues Solved

Closes #10798

### Checklist

#### Ticket Background
*   Sponsored by: Yale LUX
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @ 

### Further comments
Follow-up tickets I found as part of this work:
- #10887
- #10889

### Testing instructions
I tested with the sample models in the unit tests. I opened a django shell, ran most of `arches.tests.importer.jsonld_import_tests.JsonLDImportTests.setUpClass()`, hacking bits in/out as needed, and then manually created some resources of the various models, viewed the JSON-LD at the /resources route, and zipped that up according to the instructions in [#10798.](https://github.com/archesproject/arches/issues/10798#issuecomment-2075563323)


arches/app/etl_modules/base_import_module.py
```diff
@@ -107,13 +116,21 @@ def get_graph_tree(self, graphid):
...
             node_lookup = {str(row[1]): {"depth": int(row[5]), "cardinality": row[7]} for row in rows}
-            nodes = Node.objects.filter(graph_id=graphid)
+            nodes = Node.objects.filter(graph_id=graphid).select_related("nodegroup")
             for node in nodes:
                 nodeid = str(node.nodeid)
                 if nodeid in node_lookup:
                     node_lookup[nodeid]["alias"] = node.alias
                     node_lookup[nodeid]["datatype"] = node.datatype
                     node_lookup[nodeid]["config"] = node.config
+                elif not node.istopnode:
```
jacobtylerwalls, line 126: This is here to work around #10889

arches/app/etl_modules/base_import_module.py
```diff
@@ -158,40 +175,66 @@ def get_node_lookup(self, nodes):
         return lookup
 
     def run_load_task(self, userid, files, summary, result, temp_dir, loadid):
-        with connection.cursor() as cursor:
```
jacobtylerwalls, line 161: I factored out stage_files() and check_tile_cardinality(), and then wrapped what remained in a try/finally so that temp files were always cleaned up.

arches/app/etl_modules/base_import_module.py
```diff
@@ -158,40 +175,66 @@ def get_node_lookup(self, nodes):
...
+            self.delete_from_default_storage(self.temp_dir)
+        except FileNotFoundError:
+            pass
+
+    def validate_uploaded_file(self, file):
         pass
 
+    def stage_files(self, files, summary, cursor):
+        for file in files:
+            self.stage_excel_file(file, summary, cursor)
```
jacobtylerwalls, line 218: In future work, we could move this to the excel importers and let this base method just be abstract.

arches/app/etl_modules/base_import_module.py
```diff
@@ -207,38 +250,36 @@ def validate(self, loadid):
...
-        try:
-            self.delete_from_default_storage(self.temp_dir)
-        except (FileNotFoundError):
-            pass
+
         result = {"summary": {"name": content.name, "size": self.filesize_format(content.size), "files": {}}}
         validator = FileValidator()
-        if len(validator.validate_file_type(content)) > 0:
+        extension = content.name.split(".")[-1] or None
+        if len(validator.validate_file_type(content, extension=extension)) > 0:
```
jacobtylerwalls, line 260: Needed to get the extension before the first call to validate_file_type for the body of that method to work correctly.

You may want to test with FILE_TYPE_CHECKING = True.

arches/app/etl_modules/save.py
```diff
@@ -11,115 +11,131 @@
...
 
+def save_to_tiles(userid, loadid):
+    with connection.cursor() as cursor:
+        disable_tile_triggers(cursor, loadid)
+        error_saving_tiles = _save_to_tiles(cursor, loadid)
+        reenable_tile_triggers(cursor, loadid)
+        if error_saving_tiles:
+            return error_saving_tiles
+
+        _post_save_edit_log(cursor, userid, loadid)
```
jacobtylerwalls, line 22: This is the only change to this file: decomposing save_to_tiles() into functions that can be reordered as the JSON-LD importer needs to. Changing the indent level caused an ugly diff, but that's all that's going on below.

arches/app/models/models.py
```diff
@@ -1797,8 +1797,8 @@ class Meta:
 
 class LoadErrors(models.Model):
     load_event = models.ForeignKey(LoadEvent, db_column="loadid", on_delete=models.CASCADE)
-    nodegroup = models.ForeignKey("NodeGroup", db_column="nodegroupid", null=True, on_delete=models.CASCADE)
-    node = models.ForeignKey("Node", db_column="nodeid", null=True, on_delete=models.CASCADE)
+    nodegroup = models.ForeignKey("NodeGroup", db_column="nodegroupid", null=True, blank=True, on_delete=models.CASCADE)
+    node = models.ForeignKey("Node", db_column="nodeid", null=True, blank=True, on_delete=models.CASCADE)
```
jacobtylerwalls, line 1801: Allows running `full_clean()` or `clean_fields()` for validation purposes.

arches/app/utils/file_validator.py
```diff
@@ -14,15 +15,17 @@ def __init__(self):
 
     def test_unknown_filetypes(self, file, extension=None):
         errors = []
-        if extension in settings.FILE_TYPES:
-            if extension == "xlsx":
+        match extension:
+            case "DS_Store":
+                self.logger.log(logging.WARN, "DS_Store file encountered, proceeding with caution.")
```
jacobtylerwalls, line 20: These get skipped later in `validate_file_type()`, but adding some future safety here in case that ever changes, or there are other callers to this method.

tests/bulkdata/single_csv_tests.py
```diff
@@ -45,13 +44,17 @@ def setUpClass(cls):
...
+            admin = User(pk=1, username="admin", is_staff=True, is_superuser=True)
+            # Use bulk_create to avoid headaches with post_save signal and content types
+            User.objects.bulk_create([admin])
+
         graph.publish(user=admin)
-    
 
-    @override_settings(
-        MEDIA_ROOT = os.path.join(settings.ROOT_DIR, "..", "tests/fixtures/data")
-    )
```
jacobtylerwalls, line 54: Took Cyrus's idea here to override MEDIA_ROOT and just put that in the test_settings file.

tests/views/command_line_tests.py
```diff
@@ -43,6 +46,9 @@ def setUpClass(cls):
             with captured_stdout():
                 management.call_command("packages", operation="load_package", source=test_pkg_path, yes=True, verbosity=0)
 
+        path_to_cheesy_image = Path(settings.MEDIA_ROOT) / "uploadedfiles" / "test.png"
+        cls.addClassCleanup(os.unlink, path_to_cheesy_image)
```
jacobtylerwalls, line 50: That ended up exposing this file under a place tracked by git, so I cleaned it up.

jacobtylerwalls:
@apeters occurred to me while AfK: I wonder if the very last commit broke something, you could try reverting it.

apeters (changes requested):
I'm seeing any error details when the load fails even though I can see more info in the db.
In the UI:
Screenshot 2024-05-13 at 6 50 30 PM
From the load_errors table:
Screenshot 2024-05-13 at 6 52 38 PM

It would be nice if the error message showed the actual name of the .json file that triggered the error instead of just the "block".

Also, it would be nice if users could opt to pass the --ignore-errors flag so that all resources that pass are loaded to the db and the ones that don't pass log errors in the ui.

jacobtylerwalls:
Thanks, good notes. The current error UX looks like this:

error-ui

Clicking "Full error report" gives:

full-error-report

> Additionally I notice that the errors shown in the UI specify Validation Errors, but the errors I'm seeing are really load errors. Maybe we need both, but for load errors it may be enough to show the json file that errored and it's corresponding error message.

I understood load errors and validation errors to be synonyms in this context. What was the distinction you had in mind?

> Also, it would be nice if users could opt to pass the --ignore-errors flag so that all resources that pass are loaded to the db and the ones that don't pass log errors in the ui.

This is an interesting idea, but I think the whole interface of the bulk data manager assumes that you're not going to perform a load unless everything passes. In the scenario you describe, you'd have a partial success / partial failure state. I don't think the bulk data manager will support that.

jacobtylerwalls:
PS -- I could copy over the `message` you see in that JSON dump with detailed information into the `error` field so that it appears in the UI without clicking for more details. That's an easy change, but could clutter the table UI (and other node datatype validation errors that _didn't_ happen during "parsing a block" could no longer be distinguished).

apeters (changes requested):
Looking at the Validation Errors table I would change `Column Name` to `File Name`.  Also the `Count` column appears to be incorrect (0 index issue?).  Are the `Node Alias` and `Details` columns used?

Screenshot 2024-06-03 at 2 22 54 PM

jacobtylerwalls:
Solved the magic string and the column names by just overwriting the template block 👍 

> Looking at the Validation Errors table I would change Column Name to File Name. Also the Count column appears to be incorrect (0 index issue?). Are the Node Alias and Details columns used?

- Node alias should be populated if a tile failed validation as part of staging the tiles. (Most of the failures you and I have been seeing during testing are the "early failures" in the CLI. As we discussed, I'm using a dummy node for those, so showing that dummy node alias here would not be helpful.) I could add an N/A if you think it's best.
- Details: fixed -- query wasn't quite right
- Count: same

jacobtylerwalls:
The validation error table looks like this now (with the first "Details" link expanded):

validation-error-table

apeters (approved):
This looks great @jacobtylerwalls .  Thanks for all those last minute UI tweaks.

Issue #10891, opened 2024-05-08
Makemessages does not pick up custom translatable tags


apeters:
We have 2 custom translatable tags
`quoted_trans` and `jsescaped_trans` that the django `makemessages` command does not pickup.  
That means that strings are not registered in the .po files and will subsequently not be translated.

PR #10893, opened 2024-05-08
10891 fix makemessages and custom tags

apeters:
### Types of changes

### Issues Solved

#10891


apeters:
To test this run the command before and after the fix.  You should notice that before the fix, strings marked with either the `quoted_trans` or `jsescaped_trans` tag don't get entries in the .po file (make sure to delete any .po files ahead of time).
```
django-admin makemessages --all  --extension=htm,py
```

chiatt (approved):
Looks good to me!

Issue #10897, opened 2024-05-09
Individual datatype instances from DataTypeFactory cannot access each other


whatisgalen:
Generally, there are cases where you're indexing a resource and you might want one datatype to be able to talk to another datatype, for example if you wanted to leverage String datatype methods to handle strings in Concept or FileList, or if you wanted EDTF to leverage Number datatype methods, etc.

In a use-case example of `ResourceInstanceDataType`, the tiledata relationships can either be ontology properties or concepts. The concept datatype is a de facto authority on how to lookup a label by value, but the `ResourceInstanceDataType` instance cannot access the `ConceptDataType` instance originating from the same `DataTypeFactory`.

This ticket was created out of #10876 to accommodate concept-based relationships also being included in `document["strings"]`

PR #10898, opened 2024-05-09
10897 datatype factory cross-instance access

whatisgalen:
### Types of changes

### Description of Change

passes a reference to a "parent" `DataTypeFactory` to the "child" datatype instances instantiated

### Issues Solved

#10897 

### Checklist

#### Ticket Background
*   Found by: @whatisgalen 
*   Tested by: @ 
*   Designed by: @whatisgalen


chiatt (commented):
I spoke with @apeters and @chrabyrd about this change and we agree that it isn't actually necessary. The one case where a datatype could use another datatype's method is [here](https://github.com/archesproject/arches/pull/10876#discussion_r1612184210), but even in this case it would probably be better not to rely on another datatype at all.

PR #10776, opened 2024-04-15
Adds Vue component testing #10774

chrabyrd:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

Adds frontend testing to arches-core and projects.

MUST UPDATE `package.json` AND REBUILD `yarn.lock` ONCE https://github.com/archesproject/arches-dev-dependencies/pull/17 HAS BEEN MERGED

### Issues Solved

#10774

PR #10922, opened 2024-05-14
10855_change_precision_of_incoming_geoms

ryan86:
- Make precision reduction function recursive
- Separate bbox and precision/byte size validation
- Add tests


jacobtylerwalls:
@ryan86 The webpack failure should be resolved in a few minutes, once I merge #10776

tests/utils/datatype_tests.py
```diff
@@ -55,6 +56,50 @@ def test_tile_transform(self):
...
+    def test_validate_reduce_byte_size(self):
+        f = open("tests/fixtures/problematic_excessive_vertices.geojson")
+        geom = json.load(f)
+        geom_datatype = DataTypeFactory().get_instance("geojson-feature-collection")
+        errors = geom_datatype.validate(geom)
+        self.assertEqual(len(errors), 0)        
+
+
+    def test_validate_bbox(self):
+        settings.DATA_VALIDATION_BBOX = [(
```
jacobtylerwalls, line 70: This has the potential to leave side effects on the next test. The doc'd way to do this in django is with `@override_settings`, but b/c arches subclasses the django settings, you also have to wrap the action your test takes with `@sync_overridden_test_settings_to_arches`, or possibly use it as a second decorator, although I haven't explored that.

There's an example in `test_arches_application_extension_implicit`.

tests/utils/datatype_tests.py
```diff
@@ -55,6 +56,50 @@ def test_tile_transform(self):
         with self.assertRaises(ValueError):
             boolean.transform_value_for_tile(None)
 
+class GeoJsonDataTypeTest(ArchesTestCase):
+
+    def test_validate_reduce_byte_size(self):
+        f = open("tests/fixtures/problematic_excessive_vertices.geojson")
```
jacobtylerwalls, line 62: If you open with `with open(...` you won't leave the file handle open.

arches/app/datatypes/datatypes.py
```diff
@@ -1041,7 +1042,26 @@ def update(self, tile, data, nodeid=None, action=None):
         updated_data = tile.data[nodeid]
         return updated_data
 
+    def find_num(self, current_item):
+            if len(current_item) and type(current_item[0]) == float:
```
jacobtylerwalls, line 1046: If you isinstance() instead of check equality or identity, you'll be flexible enough in case somebody uses a numpy float.

Asserting against an exact type is usually a problem.

tests/utils/datatype_tests.py
```diff
@@ -55,6 +56,50 @@ def test_tile_transform(self):
...
+              52.666192057898854
+            ),
+            (
+              12.948801570473677,
+              52.666192057898854
+            )]
+
+        geom_datatype = DataTypeFactory().get_instance("geojson-feature-collection")
+
+        with self.subTest():
```
jacobtylerwalls, line 93: If you give the subTest a label with a kwarg, you can tell the two failures apart.

jacobtylerwalls (commented):
Just swung by to see if webpack was happier, noticed a couple things esp. in the tests.

tests/utils/datatype_tests.py
```diff
@@ -55,6 +57,52 @@ def test_tile_transform(self):
...
+class GeoJsonDataTypeTest(ArchesTestCase):
+
+    def test_validate_reduce_byte_size(self):
+        with open("tests/fixtures/problematic_excessive_vertices.geojson") as f:
+            geom = json.load(f)
+        geom_datatype = DataTypeFactory().get_instance("geojson-feature-collection")
+        errors = geom_datatype.validate(geom)
+        self.assertEqual(len(errors), 0)        
+
+
```
chiatt, line 69: Just one line of space between methods in a class

chiatt (changes requested):

chiatt (approved):

PR #10926, opened 2024-05-15
Unsilence type declarations from typed third-party modules #10925

### LINKED ISSUE #10925: [typescript] Unsilence types from third-party modules
I figured out why I've been [unable to import types from Vue](https://github.com/archesproject/arches/commit/87a303912ab5d9cb80d77e6911a1d43187f0a449). It's because we declared it was untyped by adding it to declarations.d.ts:

https://github.com/archesproject/arches/blob/760d4826201e5ad6d87f6fd3350a86d10a318578/arches/app/src/declarations.d.ts#L67

Rolling that back allows me to import types and gives better ts linting:

```ts
arches/app/src/components/ControlledListManager/ItemCharacteristic.vue:11:7 - error TS2322: Type '{ readonly label?: any; readonly field?: any; }' is not assignable to type '{ field: "uri"; label: string; }'.
  Property 'field' is optional in type '{ readonly label?: any; readonly field?: any; }' but required in type '{ field: "uri"; label: string; }'.

11 const props: { field: "uri", label: string } = defineProps(["field", "label"]);
```

Suggesting to roll those declarations back for third-party modules. We can leave arches in there, as well as any older js libraries that aren't on DefinitelyTyped.

typescript docs on "silence warnings about a particular module without writing a declaration file":
https://www.typescriptlang.org/docs/handbook/2/type-declarations.html#your-own-definitions

jacobtylerwalls:
### Types of changes

### Description of Change
Types from third-party modules like `vue` and `js-cookie` weren't importable because we were silencing them in this file. Instead, we should only silence with `declare module ...` modules that aren't typed or lack types on DefinitelyTyped.

In this PR, I just checked npm and removed all packages that had either a "TS" logo for typescript or a "DT" logo for DefinitelyTyped.

[ts reference](https://www.typescriptlang.org/docs/handbook/2/type-declarations.html#your-own-definitions)

With this change, we can now use "type-based props declaration" with interfaces in the way [suggested](https://vuejs.org/guide/typescript/composition-api.html#using-script-setup) in the vue docs, see portion beginning "However, it is usually more straightforward...".

This solves items 3 and 4 described [here](https://github.com/archesproject/arches/pull/10569#issuecomment-1924787017).

### Issues Solved
Closes #10925

### Checklist

#### Ticket Background
*   Tested by: @ 
*   Designed by: @ 

### Further comments
To see this in action, [see](https://github.com/archesproject/arches/actions/runs/9115406228) the controlled-list-item-editor branch to see work-in-progress at cleaning up typescript usage now that we're getting types from Vue.


chrabyrd (changes requested):
Good catch. I think line 3 needs to be changed, and a note added into the project `declarations.d.ts` template explaining when declarations should be manually added.

arches/install/arches-templates/project_name/src/declarations.d.ts
```diff
@@ -1,7 +1,8 @@
 // import declarations from other projects or Arches core
 import('../../node_modules/arches/arches/app/src/declarations.d.ts');
 
-// declare modules that have been added to your project (should mirror `package.json`)
+// declare untyped modules that have been added to your project in `package.json`
+// Module homepage on npmjs.com uses logos "TS" or "DT" to indicate if typed
```
jacobtylerwalls, line 5: I didn't dive into the difference here, but if someone uses a moudle that's on DefinitelyTyped, the ts compiler provides a nice hint to install the types as a dev dependency a la https://github.com/archesproject/arches-dev-dependencies/pull/21, so I figured this is good enough.

chrabyrd (approved):
lgtm!

Issue #10934, opened 2024-05-16
Boolean datatype validation broken


whatisgalen:
implementation of `str_to_bool` in `BooleanDatatype.validate` fails on case difference for values like `"True"` vs `"true"`.
We should either call `.lower()` on the string passed to `str_to_bool` or bypass it altogether if we already know `isinstance(value, bool)` is `True`.

replicable in `dev/7.6.x`

jacobtylerwalls:
Hey @whatisgalen 👋 .

I remember spotting that when refactoring it in 7.6, and I added a test to be sure, notice "True" is a rejected value:

https://github.com/archesproject/arches/blob/e4145e4cd546210c1bde18a04634fb36ec8d548b/tests/utils/datatype_tests.py#L38

I think it makes sense to change it, but are there any backwards compatibility implications?

jacobtylerwalls:
Just so you think I'm not crazy, this is what we were using before. Agree that it's probably too strict if you can assure me there aren't bc implications :-)
https://docs.python.org/3.11/distutils/apiref.html#distutils.util.strtobool

whatisgalen:
@jacobtylerwalls Yea I'm seeing those tests fail right now. I'm looking at the 3 instances where `str_to_bool` gets used and I can't see any harm in including `'False'` and `'True'` as acceptable values, or lowercasing incoming values.

whatisgalen:
Also the fact that this bug got through means that we should most likely test the datatypes the same way they get called/used by an actual client when saving tiles. So maybe a "tile_post_datatypes_test" or something (with a better name).

whatisgalen:
A little more background:

As of 7.6, when the nodevalue of a boolean node gets validated in `Tile.validate()`, it's validated as a python bool `True` or `False`. Calling `str(True)` becomes `"True"` which due to the upper case T fails the `str_to_bool` logic. My guess is that either [this older way](https://github.com/archesproject/arches/blob/dev/6.2.x/arches/app/datatypes/datatypes.py#L256-L257) somehow handled that without issue or something about how the frontend tile request payload gets loaded into the python tile view may have changed.

PR #10937, opened 2024-05-16
Fix for boolean datatype validation

whatisgalen:
…0934

### Types of changes

### Issues Solved

#10934

### Checklist

#### Ticket Background
*   Found by: @whatisgalen 
*   Tested by: @ 
*   Designed by: @whatisgalen

### DIFF
--- a/arches/app/utils/string_utils.py
+++ b/arches/app/utils/string_utils.py
@@ -1,5 +1,5 @@
 def str_to_bool(value):
-    match value:
+    match value.lower():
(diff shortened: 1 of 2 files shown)


jacobtylerwalls (approved):
Thanks, Galen. I just checked and 7.5 isn't affected, so we don't need a changelog.

I missed this in #10462, apparently the implementation I was porting from did handle this:
https://github.com/python/cpython/blob/35c799d79177b962ddace2fa068101465570a29a/Lib/distutils/util.py#L315

Issue #10931, opened 2024-05-15
Allow users to define min and max zoom levels of a resource layer


chiatt:
Users should be able to define the min and max zoom level of a resource layer in the map layer manager. This will allow users to control the zoom levels at which mvt tiles are requested (in conjunction with #10929).

Issue #10929, opened 2024-05-15
Allow min and max zoom to be applied to resource map sources


chiatt:
Map sources for resource layers are created in the datatype `get_map_source` method, so we can't persist changes to the source. It would be helpful to add `minzoom` and `maxzoom` properties to a resource source so that requests to the mvt endpoint could be restricted to a range of zoom levels. These properties could be read from the node config. This would prevent the database from getting overwhelmed with requests to calculate mvt tiles in cases where there is a large number of geometries (200k +)

PR #10943, opened 2024-05-16
Allow min & maxzoom to be applied to a resource map source

chiatt:
### Types of changes

### Description of Change
Allows minzoom and maxzoom properties to be added to the map source of a resource so that geometry requests can be limited to a specific range. 

Currently zoom properties must be added using sql:
```
update nodes set config = config || '{"minzoom": 10}' where nodeid = 'some geojson node id';
```

The ability to modify these properties in the UI will be addressed here: #10931

### Issues Solved

#10929

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/datatypes/base.py
+++ b/arches/app/datatypes/base.py
@@ -109,6 +109,11 @@ class BaseDataType(object):
         if node is None:
             return None
         source_config = {"type": "vector", "tiles": [tileserver_url]}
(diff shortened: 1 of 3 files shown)


njkim (approved):
Works as expected

Issue #10719, opened 2024-03-27
Ensure all package loads in setUpClass() take place inside a transaction


jacobtylerwalls:
#10711 works toward isolating the package loads necessary for test cases, but there were a handful of package loads that couldn't fully be moved after `super().setUpClass()` initiates a transaction (so that they roll back cleanly when the class is torn down).

https://github.com/archesproject/arches/blob/350e4a7cd2cca3b17176cf31bbcbf5b5282ff695/tests/views/api_tests.py#L41-L44

This can cause annoying red herring test failures if the package load fails for unrelated reasons as experienced in https://github.com/archesproject/arches/pull/9613#discussion_r1540413026.

We should identify the root cause and fix the setup (or adjust the failing tests).

jacobtylerwalls:
- [ ] permission_tests
- [ ] api_tests
- [ ] command_line_tests
- [ ] resource_tests

PR #10948, opened 2024-05-20
7.6 -> 8.0

jacobtylerwalls:


jacobtylerwalls:
> I think fixing https://github.com/archesproject/arches/issues/10719 is the sanest way to fix the error shown in the unit tests. 

Whoops, that's actually unrelated: the underlying deal is graph slugs no longer being strictly unique in v8, will put up a PR that adjusts the test setup, then this branch should turn green.

jacobtylerwalls:
Will reattempt with a fresh branch to solve the merge conflict

jacobtylerwalls:
Actually, there's an interesting merge conflict here re: future graphs, so I would be more comfortable with @chrabyrd's eyes on it 🙏

Issue #10839, opened 2024-04-26
Cloning RM results in source/original RM and its instances becoming inaccessible


tris-ots:
So far I've only been able to reproduce this issue in Arches 7.5.1 and with a [specific ontology](https://github.com/coherit/cidoc4arches), and only with published RMs. I was unable to reproduce the error using 7.5.0, or with the more standard CIDOC CRM 6.2 [ontology package](https://github.com/archesproject/cidoc-crm-ontology).  

Steps to reproduce:
- Clone a published RM in the Arches Designer.
- Delete the cloned RM.
- Attempt to view the original RM in the Arches Designer. 

Expected behavior: 
- The original RM and its resource instances, if any, are unaffected.

Observed behavior: 
- Clicking on the original RM in the Arches Designer returns a Django error page. 
- Resource Instances of the original RM are no longer present or searchable on the Search page.

The error: 
```
  DoesNotExist at /graph_designer/e63cea5e-01a1-11ef-ab4f-5254004626ce

  NodeGroup matching query does not exist.

  Request Method: 	GET
  Request URL: 	http://localhost:8080/graph_designer/e63cea5e-01a1-11ef-ab4f-5254004626ce
  Django Version: 	4.2.10
  Exception Type: 	DoesNotExist
  Exception Value: 	

  NodeGroup matching query does not exist.

  Exception Location: 	/opt/arches/.venv/lib/python3.11/site-packages/django/db/models/query.py, line 637, in get
  Raised during: 	arches.app.views.graph.GraphDesignerView
  Python Executable: 	/opt/arches/.venv/bin/uwsgi
  Python Version: 	3.11.2
  Python Path: 	

  ['/opt/arches/arches_project',
  '.',
  '',
  '/usr/lib/python311.zip',
  '/usr/lib/python3.11',
  '/usr/lib/python3.11/lib-dynload',
  '/opt/arches/.venv/lib/python3.11/site-packages',
  '/opt/arches/arches_project/arches_project']

  Server time: 	Tue, 23 Apr 2024 17:11:34 -0500

  Traceback (most recent call last):
    File "/opt/arches/.venv/lib/python3.11/site-packages/django/db/models/fields/related_descriptors.py", line 218, in __get__
      rel_obj = self.field.get_cached_value(instance)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #10952, opened 2024-05-20
Fix clone published graph bug #10839

chrabyrd:
### Types of changes

### Issues Solved

#10839

### DIFF
--- a/arches/app/views/graph.py
+++ b/arches/app/views/graph.py
@@ -421,14 +421,26 @@ class GraphDataView(View):
                 elif self.action == "export_branch":
                     clone_data = graph.copy(root=data)
                     clone_data["copy"].slug = None
+                    clone_data["copy"].publication = None
+
                     clone_data["copy"].save()
+
+                    if bool(graph.publication_id):
+                        clone_data["copy"].publish(user=request.user)
+
                     ret = {"success": True, "graphid": clone_data["copy"].pk}
 
                 elif self.action == "clone_graph":
                     clone_data = graph.copy()
                     ret = clone_data["copy"]
                     ret.slug = None
+                    ret.publication = None
+
                     ret.save()
+
+                    if bool(graph.publication_id):
+                        ret.publish(user=request.user)
+
                     ret.copy_functions(graph, [clone_data["nodes"], clone_data["nodegroups"]])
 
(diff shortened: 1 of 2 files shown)


arches/app/views/graph.py
```diff
@@ -422,13 +422,20 @@ def post(self, request, graphid=None):
...
                 elif self.action == "clone_graph":
                     clone_data = graph.copy()
                     ret = clone_data["copy"]
                     ret.slug = None
+                    ret.publication = None
+
                     ret.save()
+
+                    if bool(graph.publication_id):
+                        ret.publish(user=request.user)
```
jacobtylerwalls, line 437: Noticed no tests fail when I revert this. Do you have an idea why that might be?

kfogel: @jacobtylerwalls Just confirming: the new ` test_graph_clone_on_unpublished_graph()` test doesn't fail even when the above change to `arches/app/views/graph.py` is absent?

jacobtylerwalls: Right. This may be the right changeset after all, but it's just easier for me to verify that as a reviewer if we have a test that fails without the changeset in the application code.

chrabyrd: @jacobtylerwalls this should be g2g now. It's an interesting interaction with how we're hydrating graphs -- the return of `clone_graph` is correct but instantiating a cloned graph elsewhere in the application will error without the changes.

jacobtylerwalls: Thanks, that's very helpful context.

jacobtylerwalls, line 437: 👍 

I just had a quick look at save() and publish() and it doesn't look like we're in a scenario where need to refresh_from_db() before doing this. (Whereas if this were an add_node() operation that created node aliases in postgres on save, I could see that maybe being a thing to be careful about.)

jacobtylerwalls (approved):
+1 tested. Thanks for the updates!

(I can approve the cherry-pick for 7.5 when you put it up, too.)

jacobtylerwalls:
(A changelog included with the 7.5 PR would be 💯 )

kfogel:
Thanks to you both, @chrabyrd and @jacobtylerwalls!

PR #10968, opened 2024-05-23
Refinements to reference datatype

chiatt:
### Description of Change
Cleanup in response to @chrabyrd comments on #10569


arches/app/media/js/viewmodels/reference-select.js
```diff
@@ -110,8 +110,7 @@ define([
                 if (item.uri) {
                     let text = self.getPrefLabel(item.values) || arches.translations.searching + '...';
                     NAME_LOOKUP[item.uri] = {"prefLabel": text, "labels": item.values.filter(val => this.isLabel(val)), "listid": item.controlled_list_id};
-                    text = indentation + text;
-                    return text;
+                    return text = indentation + text;
```
chrabyrd, line 113: nit: `return indentation + text`

chrabyrd (changes requested):
overall looks great! Just 1 nit and it should be g2g 👍

chrabyrd (approved):
lgtm!

Issue #9282, opened 2023-02-28
Yarn 1.x reaching end of life


chiatt:
In 2020 yarn 1.22, the version we currently use in Arches, [entered maintenance mode](https://classic.yarnpkg.com/en/docs/install#mac-stable). Moving to yarn 3.x will require some [effort](https://yarnpkg.com/getting-started/migration#why-should-you-migrate) and likely impact Arches project maintainers. It may be necessary in the coming months to either upgrade to the latest version of yarn or to replace yarn with npm (or possibly even [pnpm](https://www.atatus.com/blog/npm-vs-yarn-vs-pnpm/)). Although yarn does provide some performance benefits during installation, moving to npm would mean managing one less dependency.

chiatt:
That's a good point @bferguso. We should aim to do that.

whatisgalen:
following

chrabyrd:
resolved in #10980

PR #10980, opened 2024-05-28
Convert from yarn to npm #9282

chrabyrd:
### Types of changes

### Issues Solved

#9282


docker/Readme.md
```diff
@@ -315,10 +315,10 @@ The volume section at **point 4** also mounts a `settings_local.py` into the con
...
 
-##### Yarn components
+##### npm components
 Because your volume from `step 4` is laid over the files in your container, files generated during image `build` are hidden.  
 In order to install newly added static file packages, run the following command:  
-    `docker-compose -f .\docker-compose-local.yml run arches install_yarn_components`
+    `docker-compose -f .\docker-compose-local.yml run arches install_npm_components`
 
 ##### docker-compose-local.yml
 Here is a link to a working [docker-compose-local.yml](https://gist.github.com/jmunowitch/c63fa39be4651b9bf2f0b1abc69f7479) file
```
jacobtylerwalls, line 324: Let's delete this, it doesn't even resolve to anything

package.json
```diff
@@ -102,15 +97,14 @@
...
-        "select2": "node_modules/select2/select2",
         "select-woo": "node_modules/select-woo/dist/js/selectWoo.full",
         "select-woo-src": "node_modules/select-woo/src/js/select2",
         "bootstrap": "node_modules/bootstrap/dist/js/bootstrap.min",
         "jquery-ui": "node_modules/jqueryui/jquery-ui.min",
         "backbone": "node_modules/backbone/backbone-min",
         "underscore": "node_modules/underscore/underscore-min",
         "jquery-validate": "node_modules/jquery-validation/dist/jquery.validate.min",
-        "d3": "node_modules/d3/dist/d3.min",
+        "d3": "node_modules/d3/src/index.js",
```
jacobtylerwalls, line 107: I'm unable to build the frontend unless I revert this change. Do you know why?

jacobtylerwalls (changes requested):
d3 caused a problem with my build, do you see the same?

docker/Readme.md
chrabyrd: could you clarify? `install_npm_components` matches to a function declared in `entrypoint.sh`. Admittedly I don't run docker so my context is weak

package.json
chrabyrd: hmm just checked this and it's working for me -- and breaks the other way. Did you update the project's `package.json` to point to this branch?

```
    "dependencies": {
        "arches": "archesproject/arches#cbyrd-9282-convert-from-yarn-to-npm"
    },
```

docker/Readme.md
jacobtylerwalls: no worries -- I meant the comment right below your change, github couldn't let me comment directly on an unchanged line. the gist.github.com/... link goes nowhere

package.json
jacobtylerwalls: whoops, I definitely missed that--lgtm!

jacobtylerwalls (approved):
LGTM, sending over to Aaron to have a look.

aarongundel (approved):
This is really good to see.  A couple of notes from my testing (that you already saw) - relative pathing in the package.json file will break with npm because of the hoisting differences between yarn and npm.  Our workaround is to avoid using relative pathing in the package.json.  I am not sure if this is worth documenting or not, since it's not really a solution.  

I did notice performance impacts when doing a full build of arches vgm.  Not too surprising, but noteworthy.  

You may want to note that if someone is using a container based setup, they'll need to update any node references therein.  For instance, my entrypoint script for vgm was calling yarn install directly.  Some other projects will likely be impacted by that change.  It could end up being the case if that if you're not diligent about removing yarn from your container setup, you'd have dueling package managers - and that'll probably be a receipe for a forum post.  

Other than that, looks good!  Your call on documentation updates.

PR #11005, opened 2024-06-05
Permissions frameworks core merge into to 7.6.x

aarongundel:


tests/permissions/base_permissions_framework_test.py
```diff
@@ -0,0 +1,109 @@
...
+from arches.app.utils.betterJSONSerializer import JSONSerializer, JSONDeserializer
+from django.contrib.auth.models import User
+from django.contrib.auth.models import Group
+from arches.app.models.models import GraphModel, ResourceInstance, Node
+from arches.app.models.resource import Resource
+from arches.app.permissions.arches_default_deny import (
+    ArchesDefaultDenyPermissionFramework,
+)
+
+
```
jacobtylerwalls, line 34: Elsewhere we've been adding the boilerplate comment here to copy/paste to the command line

tests/permissions/base_permissions_framework_test.py
```diff
@@ -0,0 +1,109 @@
...
+        self.framework = self.FRAMEWORK()
+        resource = Resource.objects.get(pk=self.resource_instance_id)
+        resource.graph_id = self.data_type_graphid
+        resource.remove_resource_instance_permissions()
+
+    def tearDown(self):
+        ResourceInstance.objects.filter(graph_id=self.data_type_graphid).delete()
+
+    @classmethod
+    def add_users(cls):
```
jacobtylerwalls, line 52: Noticed this is just riffing on the resource_tests.py util. Can we DRY it out by moving it to a test utils file?

tests/permissions/base_permissions_framework_test.py
```diff
@@ -0,0 +1,109 @@
...
+        self.resource_instance_id = "f562c2fa-48d3-4798-a723-10209806c068"
+        self.user = User.objects.get(username="ben")
+        self.group = Group.objects.get(pk=2)
+        self.framework = self.FRAMEWORK()
+        resource = Resource.objects.get(pk=self.resource_instance_id)
+        resource.graph_id = self.data_type_graphid
+        resource.remove_resource_instance_permissions()
+
+    def tearDown(self):
+        ResourceInstance.objects.filter(graph_id=self.data_type_graphid).delete()
```
jacobtylerwalls, line 49: This shouldn't be necessary, TestCase subclasses roll back after each test to the savepoint at the end of setUpTestData().

tests/permissions/base_permissions_framework_test.py
```diff
@@ -0,0 +1,109 @@
...
+from django.contrib.auth.models import Group
+from arches.app.models.models import GraphModel, ResourceInstance, Node
+from arches.app.models.resource import Resource
+from arches.app.permissions.arches_default_deny import (
+    ArchesDefaultDenyPermissionFramework,
+)
+
+
+class ArchesPermissionFrameworkTestCase(ArchesTestCase):
+    def setUp(self):
```
jacobtylerwalls, line 36: Nothing fails if I do this to allow the setup to run 1x per class instead of per test...
```suggestion
    @classmethod
    def setUpTestData(cls):
        cls.add_users()
        # change self to cls below
```
...

tests/permissions/base_permissions_framework_test.py
```diff
@@ -0,0 +1,109 @@
...
+        self.expected_resource_count = 2
+        self.client = Client()
```
jacobtylerwalls, line 38: client already exists on self

tests/permissions/permissions_arches_default_deny_tests.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+        self.framework.assign_perm("view_resourceinstance", self.group, resource)
+        can_access_with_view_permission = self.framework.user_can_read_resource(
+            self.user, self.resource_instance_id
+        )
+        self.assertTrue(
+            implicit_permission is False
+            and can_access_without_view_permission is False
+            and can_access_with_view_permission is True
+        )
```
jacobtylerwalls, line 65: Debugging will be easier if this is separated into three assertions

tests/permissions/permissions_arches_default_deny_tests.py
```diff
@@ -0,0 +1,118 @@
...
+        nodes = Node.objects.filter(graph_id=resource.graph_id)
+        for node in nodes:
+            if node.nodegroup:
+                self.framework.assign_perm(
+                    "no_access_to_nodegroup", self.group, node.nodegroup
+                )
+        hasperms = self.framework.user_has_resource_model_permissions(
+            self.user, ["models.read_nodegroup"], resource
+        )
+        self.assertTrue(hasperms is False)
```
jacobtylerwalls, line 85: I usually see assertIs(..., False) if the exact value is important.

tests/permissions/permissions_arches_default_deny_tests.py
```diff
@@ -0,0 +1,118 @@
...
+            ben.id in restrictions["cannot_write"],
+            sam.id in restrictions["cannot_delete"],
+            sam.id in restrictions["no_access"],
+            admin.id not in restrictions["cannot_read"],
+            admin.id not in restrictions["cannot_write"],
+            admin.id not in restrictions["cannot_delete"],
+            admin.id not in restrictions["no_access"],
+        ]
+
+        self.assertTrue(all(results) is True)
```
jacobtylerwalls, line 118: Probably better as `with self.subTest(...`

tests/permissions/permissions_arches_standard_tests.py
```diff
@@ -0,0 +1,139 @@
...
+            admin.id not in restrictions["cannot_write"],
+            admin.id not in restrictions["cannot_delete"],
+            admin.id not in restrictions["no_access"],
+        ]
+
+        self.assertTrue(all(results) is True)
+
+    def test_inclusion_array(self):
+        inclusions = self.framework.get_permission_inclusions()
+        self.assertTrue(len(inclusions) == 4)
```
jacobtylerwalls, line 125: I pretty much always avoid assertTrue unless I'm really trying to check truthiness of something of indeterminate length, you just get way better debug output.

tests/permissions/permissions_arches_standard_tests.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @mock.patch("django.contrib.auth.models.User")
+    def test_permission_search_filter(self, mock_User):
+        mock_User.id = 12
+        filter = self.framework.get_permission_search_filter(mock_User)
+        filter_text = str(filter.dsl)
+        assert filter_text.find("permissions.users_with_no_access") != -1
+        assert filter_text.find(str(mock_User.id)) != -1
```
jacobtylerwalls, line 133: Would suggest to avoid pytest asserts

jacobtylerwalls (changes requested):
Very cool! Thanks for letting me have a look while this was still in draft. I have a few suggestions.

releases/7.6.0.md
```diff
@@ -1,52 +1,51 @@
...
-- Adds frontend testing and Python testing for projects
+-   Adds frontend testing and Python testing for projects
 
-- Adds Github actions to build applications, run tests, and compare coverage between branches for projects.
+-   Adds Github actions to build applications, run tests, and compare coverage between branches for projects.
 
-- Plugins now support the configuration boolean `is_standalone`. Standalone plugins do not appear in the sidebar, and do not display the sidebar or application header.
+-   Plugins now support the configuration boolean `is_standalone`. Standalone plugins do not appear in the sidebar, and do not display the sidebar or application header.
 
-- 10862 JSON-LD bulk import module
```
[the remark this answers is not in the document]
aarongundel: Aggressive prettier auto-formatting made these things easy to miss.  I disabled that and reloaded this file from base then redid the change.  Thanks for the catch.

tests/permissions/permissions_arches_default_deny_tests.py
aarongundel: That's a neat trick.

arches/app/media/js/views/components/search/search-results.js
```diff
@@ -250,8 +250,8 @@ function($, _, BaseFilter, bootstrap, arches, select2, ko, koMapping, GraphModel
                             selected: ko.computed(function() {
                                 return result._source.resourceinstanceid === ko.unwrap(self.selectedResourceId);
                             }),
-                            canRead: result._source.permissions && result._source.permissions.users_without_read_perm.indexOf(this.userid) < 0 && self.userCanReadResources,
-                            canEdit: result._source.permissions && result._source.permissions.users_without_edit_perm.indexOf(this.userid) < 0 && self.userCanEditResources,
+                                canRead: result["can_read"],
+                                canEdit: result["can_edit"],
```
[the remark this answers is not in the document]
aarongundel: This is actually due to a change to how the get_resource_types_by_perm method works.  It only returns UUIDs now.  So I ended up needing to fix this at the resource_type_filter view level

arches/app/models/resource.py
```diff
@@ -75,6 +78,27 @@ def __init__(self, *args, **kwargs):
...
         self.node_datatypes = None
 
+    def get_instance_creator(self, user=None):
+        creatorid = None
+        can_edit = None
+        if (
+            models.EditLog.objects.filter(resourceinstanceid=self.resourceinstanceid)
+            .filter(edittype="create")
+            .exists()
+        ):
```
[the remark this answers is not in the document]
aarongundel: This whole method is likely to change with an upcoming merge.

arches/app/views/graph.py
```diff
@@ -234,9 +234,8 @@ def get_ontology_namespaces(self):
         return ontology_namespaces
 
     def get(self, request, graphid):
-
         if graphid == settings.SYSTEM_SETTINGS_RESOURCE_MODEL_ID:
-            if not request.user.groups.filter(name="System Administrator").exists():
+            if not group_required("System Administrator", raise_exception=True):
                 raise PermissionDenied
```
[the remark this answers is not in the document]
aarongundel: I am not sure I follow.  If the requested graphid is something other than the Systems Settings Resource Model ID, then the group need not be System Administrator - so I am not sure how this would be used in a decorator.  I see how the raise is duplicative, though.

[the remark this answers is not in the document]
jacobtylerwalls: Oh you're quite right, I missed that when squinting at this in the files tab.

arches/app/permissions/arches_standard.py
```diff
@@ -0,0 +1,926 @@
...
+                > 0
+            )
+        return False
+
+    def get_resource_types_by_perm(
+        self, user: User, perms: str | Iterable[str]
+    ) -> list[str]:
+        graphs = set()
+        nodegroups = self.get_nodegroups_by_perm(user, perms)
+        for node in Node.objects.filter(nodegroup__in=nodegroups).prefetch_related(
```
jacobtylerwalls, line 515: select_related when going from many -> one, otherwise an unnecessary query is made.
```suggestion
        for node in Node.objects.filter(nodegroup__in=nodegroups).select_related(
```

arches/app/permissions/arches_standard.py
```diff
@@ -0,0 +1,926 @@
...
+        )
+
+    def user_can_read_concepts(self, user: User) -> bool:
+        """
+        Requires that a user is a part of the RDM Administrator group
+
+        """
+
+        if user.is_authenticated:
+            return bool(user.groups.filter(name="RDM Administrator").exists())
```
jacobtylerwalls, line 679: suggest removing the bool, but I see elsewhere the extra bool cast seems to be intentional, so I leave it to you :-)

arches/app/permissions/arches_standard.py
```diff
@@ -0,0 +1,926 @@
...
+    def has_perm(self, user_obj: User, perm: str, obj: Model | None = None) -> bool:
+        # check if user_obj and object are supported (pulled directly from guardian)
+        support, user_obj = check_support(user_obj, obj)
+        if not support:
+            return False
+
+        if "." in perm:
+            app_label, perm = perm.split(".")
+            if obj is None:
+                raise RuntimeError("Passed perm has app label of '%s' and obj is None")
```
jacobtylerwalls, line 797: ValueError seems more idiomatic to me.

arches/app/permissions/arches_standard.py
```diff
@@ -0,0 +1,926 @@
...
+        user_permission_cache = caches["user_permission"]
+        current_user_cached_permissions = user_permission_cache.get(str(user.pk), {})
+
+        if current_user_cached_permissions.get("user_permissions"):
+            user_permissions = current_user_cached_permissions.get("user_permissions")
+        else:
+            user_permissions = set()
+
+            for group in user.groups.all():
+                for group_permission in group.permissions.all():
```
jacobtylerwalls, line 834: Pre-existing issue, but seems like we should have prefetch_related("permissions") before this.

arches/settings.py
```diff
@@ -489,8 +488,10 @@
 # RECAPTCHA_USE_SSL = False
 NOCAPTCHA = True
 # RECAPTCHA_PROXY = 'http://127.0.0.1:8000'
+SILENCED_SYSTEM_CHECKS = ["guardian.W001"]
+
 if DEBUG is True:
-    SILENCED_SYSTEM_CHECKS = ["captcha.recaptcha_test_key_error"]
+    SILENCED_SYSTEM_CHECKS += ["captcha.recaptcha_test_key_error"]
```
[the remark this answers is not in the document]
aarongundel: Makes me wonder if we should add it to the upgrade notes.  Mentioned this to @chiatt  yesterday - not sure if it makes sense to override these silenced checks by default at the project level.  Makes more sense to me to let the user override them if there's a need.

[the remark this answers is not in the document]
aarongundel: removed the lines from settings.py-tpl

arches/app/permissions/arches_standard.py
```diff
@@ -468,6 +467,10 @@ def user_has_resource_model_permissions(
 
         if resource:
             graph_id = resource.graph_id
+            if graph_id is None:
+                raise ValueError(
+                    "graph_id must not be None to check resource permissions"
+                )
```
jacobtylerwalls, line 473: What if we deindent this so that we also catch when both params are None?

tests/permissions/base_permissions_framework_test.py
```diff
@@ -0,0 +1,68 @@
...
+    @classmethod
+    def setUpClass(cls):
+        cls.data_type_graphid = "330802c5-95bd-11e8-b7ac-acde48001122"
+        if not GraphModel.objects.filter(pk=cls.data_type_graphid).exists():
+            # TODO: Fix this to run inside transaction, i.e. after super().setUpClass()
+            # https://github.com/archesproject/arches/issues/10719
+            test_pkg_path = os.path.join(
+                test_settings.TEST_ROOT, "fixtures", "testing_prj", "testing_prj", "pkg"
+            )
+            path_to_cheesy_image = test_pkg_path / "uploadedfiles" / "test.png"
```
jacobtylerwalls, line 57: Sorry, I should have clarified that `/` only works on `pathlib.Path`, not strings, so take your pick of casting to Path or os.path.join, which I see we're doing in the vicinity

arches/app/permissions/arches_standard.py
```diff
@@ -0,0 +1,926 @@
...
+            else:
+                formatted_perms.append(perm)
+
+    permitted_nodegroups = {}
+    checker: ObjectPermissionChecker = CachedObjectPermissionChecker(
+        user_or_group,
+        NodeGroup,
+    )
+
+    for nodegroup in NodeGroup.objects.all():
```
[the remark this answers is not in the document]
aarongundel: Can you give me some more info here?  I'm not sure where nodegroups are cached in the checker.

arches/settings.py
```diff
@@ -660,6 +659,9 @@
     # Don't forget to use absolute paths, not relative paths.
 )
 
+PERMISSION_LOCATIONS = [
+    "arches.app.permissions",
+]
```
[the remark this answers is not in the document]
aarongundel: This setting isn't actually used anywhere at this point.  It can probably be removed.

tests/permissions/permissions_arches_standard_tests.py
```diff
@@ -0,0 +1,127 @@
...
+            sam.id in restrictions["cannot_delete"],
+            sam.id in restrictions["no_access"],
+            admin.id not in restrictions["cannot_read"],
+            admin.id not in restrictions["cannot_write"],
+            admin.id not in restrictions["cannot_delete"],
+            admin.id not in restrictions["no_access"],
+        ]
+
+        for result in results:
+            with self.subTest(result=result):
```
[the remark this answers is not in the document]
aarongundel: it is indeed much easier to figure out.  I've corrected this.

arches/app/views/api.py
```diff
@@ -2105,8 +2095,8 @@ def get(self, request):
             )
             result = cursor.fetchall()
             permitted_nodegroups = [
```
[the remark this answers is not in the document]
aarongundel: I changed this signature to return UUIDs.  This works in most cases, only minor tweaks were required from there.  So hopefully it lessens the overall impact of the method changes.

arches/wsgi.py
```diff
@@ -35,5 +35,6 @@
 application = get_wsgi_application()
 
 from arches.app.models.system_settings import settings
+from arches.app.models import permissions
```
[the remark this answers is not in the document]
aarongundel: I yanked this and saw no ill effects.

arches/app/views/api.py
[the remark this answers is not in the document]
jacobtylerwalls: Thanks, I think that's much improved.

arches/app/permissions/arches_standard.py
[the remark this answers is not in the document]
jacobtylerwalls: Sure thing. If you glance at `checker.get_perms`, you see that it just returns from `self._obj_perms_cache`. We could iterate that directly (assuming it's not empty, but I'm pretty certain it's never empty):
```py
(Pdb) len(connection.queries)
15
(Pdb) uuid_strings = [key[1] for key in checker._obj_perms_cache]
(Pdb) uuid_strings
['20000000-0000-0000-0000-100000000000', 'a34b72d0-2f49-11ef-b803-1ea1ac521290', 'a34b7424-2f49-11ef-b803-1ea1ac521290', 'a34b65b0-2f49-11ef-b803-1ea1ac521290', 'a34b677c-2f49-11ef-b803-1ea1ac521290', 'a34b6aa6-2f49-11ef-b803-1ea1ac521290', 'a34b68e4-2f49-11ef-b803-1ea1ac521290', 'a34b62b8-2f49-11ef-b803-1ea1ac521290', 'a34b6bfa-2f49-11ef-b803-1ea1ac521290', 'a34b6d44-2f49-11ef-b803-1ea1ac521290', 'a34b6ede-2f49-11ef-b803-1ea1ac521290', 'a34b7050-2f49-11ef-b803-1ea1ac521290', 'a34b7190-2f49-11ef-b803-1ea1ac521290', 'feb95991-fa14-11e6-974f-6c4008b05c4c', '32f2d663-fae4-11e6-83ad-6c4008b05c4c', 'c5a2a914-fadd-11e6-8f8f-6c4008b05c4c', '0e8fe87a-4148-11e7-aa8b-c4b301baab9f', '0e8ff675-4148-11e7-9c88-c4b301baab9f', '0e8fe457-4148-11e7-9c10-c4b301baab9f', 'c5a2a2e8-fadd-11e6-9849-6c4008b05c4c', 'c5a2b1dc-fadd-11e6-a726-6c4008b05c4c', 'c5a2b530-fadd-11e6-b5f6-6c4008b05c4c', 'd0987880-fad8-11e6-8cce-6c4008b05c4c', '0e8fdef0-4148-11e7-8330-c4b301baab9f', '0e8ff31c-4148-11e7-998a-c4b301baab9f']
(Pdb) ng = NodeGroup(UUID(uuid_strings[0]))
(Pdb) checker.get_perms(ng)
['delete_nodegroup', 'no_access_to_nodegroup', 'read_nodegroup', 'write_nodegroup']
(Pdb) len(connection.queries)
15
```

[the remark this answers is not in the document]
jacobtylerwalls: Feel free to push back if you think this is too tenuous, just wanted to make sure it was considered.

arches/app/permissions/arches_standard.py
```diff
@@ -0,0 +1,925 @@
...
+        user_permission_cache = caches["user_permission"]
+        current_user_cached_permissions = user_permission_cache.get(str(user.pk), {})
+
+        if current_user_cached_permissions.get("user_permissions"):
+            user_permissions = current_user_cached_permissions.get("user_permissions")
+        else:
+            user_permissions = set()
+            user.groups.prefetch_related("permissions")
+
+            for group in user.groups.all():
```
jacobtylerwalls, line 832: I was thinking like:
```suggestion
            for group in user.groups.prefetch_related("permissions").all():
```

arches/app/models/migrations/10999_add_principaluserid_to_resources.py
```diff
@@ -0,0 +1,37 @@
...
+
+    dependencies = [
+        ("models", "10798_jsonld_importer"),
+    ]
+
+    @transaction.atomic
+    def forward_func(state, editor):
+        for resource in Resource.objects.all():
+            resource.principaluser_id = resource.get_instance_creator()
+            resource.save()
```
jacobtylerwalls, line 20: This is a one-time only thing, so it's probably okay to have an N+1 queries problem here with one editlog query per resource. But we could look into a solution for it, especially if that's a use case we anticipate having in the application: are there cases where we iterate a bunch of resources and should be "joining" the EditLog table instead of querying for every row?

A couple other questions:
- do we need iterator() here in case there are deployments with a very high count of resources?
- should this be atomic? Django [cautions](https://docs.djangoproject.com/en/5.0/howto/writing-migrations/#non-atomic-migrations) that large data migrations can be slow when performed inside transactions

aarongundel: Yeah, this is not great.  A ton of resources would probably end up making for a long upgrade time.

arches/settings.py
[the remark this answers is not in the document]
jacobtylerwalls: We just looked at this together, and I think we have good test coverage without necessarily needing to add tests for this, but I leave it to you.

jacobtylerwalls (approved):
Great, thanks for the updates!

arches/app/permissions/arches_standard.py
chrabyrd, line 1: nit: more descriptive filename here. What is `arches_standard` to a new user?

arches/app/search/components/resource_type_filter.py
```diff
@@ -56,5 +57,11 @@ def append_dsl(
 
     def view_data(self):
         return {
-            "resources": get_resource_types_by_perm(self.request.user, "read_nodegroup")
+            "resources": list(
+                Graph.objects.filter(
```
chrabyrd, line 61: `get_resource_types_by_perm` returns `GraphModel`s . We could save a little overhead here if it's also possible to return those here instead of `Graph`s

chrabyrd (changes requested):
overall looks/works well! 👍 

Had some minor nits but yeah this is a good change

arches/app/search/components/resource_type_filter.py
chiatt: @aarongundel I wouldn't be surprised if the use of the proxy model here is the source of the performance issues that you've noticed.

arches/app/permissions/arches_standard.py
aarongundel: This is difficult to change.  My gut, at first, was to say "arches_default_allow" - but it's inherited into arches_default_deny - and that just feels wrong.  arches_default_permission_framework is "better" but the default will (in theory) be changing in 8.x - so this class name would have to change, and that also feels wrong.  I hate anything called "legacy" in code.  I am not sure what to change this to - my inclination is to leave it alone because I can't think up a better name.

arches/app/const.py
```diff
@@ -21,3 +21,4 @@ class ExtensionType(Enum):
     ETL_MODULES = "etl_modules"
     FUNCTIONS = "functions"
     SEARCH_COMPONENTS = "search_components"
+    PERMISSIONS_FRAMEWORKS = "permissions_frameworks"
```
jacobtylerwalls: As you already discovered, changing the right hand side will cause `module_importer.get_directories` to expect to find permissions modules under my_arches_app/permissions_framework.py instead of permissions.py where you have it now. So either renaming both to PERMISSIONS or adding special case in module_importer I think is needed if we want to change this.

Makes me think we might want a quick test for this after all next to the other ones for that function.

arches/app/permissions/arches_standard.py
```diff
@@ -0,0 +1,931 @@
...
+            return (
+                len(self.get_resource_types_by_perm(user, ["models.read_nodegroup"]))
+                > 0
+            )
+        return False
+
+    def get_resource_types_by_perm(
+        self, user: User, perms: str | Iterable[str]
+    ) -> list[str]:
+        graphs = list()
```
jacobtylerwalls, line 514: unused (overwritten)

arches/app/permissions/arches_standard.py
```diff
@@ -0,0 +1,931 @@
...
+
+        nodegroups = self.get_nodegroups_by_perm(user, perms)
+        graphs = (
+            Node.objects.values("graph_id")
+            .annotate(graph_count=Count("graph_id"))
```
jacobtylerwalls, line 519: looks unused?

arches/app/views/resource.py
```diff
@@ -467,7 +447,7 @@ class ResourcePermissionDataView(View):
 
     def get(self, request):
         resourceid = request.GET.get("instanceid", None)
-        resource_instance = models.ResourceInstance.objects.get(pk=resourceid)
+        resource_instance = Resource.objects.get(pk=resourceid)
```
chrabyrd, line 450: Can this stay as the `ResourceInstance` model?

arches/app/views/resource.py
```diff
@@ -484,7 +464,7 @@ def post(self, request):
             data = JSONDeserializer().deserialize(request.body)
             self.apply_permissions(data, request.user)
             if "instanceid" in data:
-                resource = models.ResourceInstance.objects.get(pk=data["instanceid"])
+                resource = Resource.objects.get(pk=data["instanceid"])
```
chrabyrd, line 467: Can this stay as the `ResourceInstance` model?

arches/app/views/resource.py
```diff
@@ -553,7 +533,7 @@ def get_instance_permissions(self, resource_instance):
             len(get_users_with_perms(resource_instance))
             + len(get_groups_with_perms(resource_instance))
         ) > 1
-        instance_creator = get_instance_creator(resource_instance)
+        instance_creator = resource_instance.get_instance_creator_and_edit_permissions()
```
chrabyrd, line 536: I get this is why `ResourceInstance` is being replaced with `Resource` in this file. However loading that proxy model may also cause some additional overhead and it might be worth rethinking the pattern

arches/app/views/resource.py
```diff
@@ -582,7 +562,7 @@ def make_instance_public(self, resourceinstanceid, graphid=None):
     def apply_permissions(self, data, user, revert=False):
         with transaction.atomic():
             for instance in data["selectedInstances"]:
-                resource_instance = models.ResourceInstance.objects.get(
+                resource_instance = Resource.objects.get(
```
chrabyrd, line 565: Can this stay as the `ResourceInstance` model?

arches/app/permissions/arches_standard.py
```diff
@@ -0,0 +1,931 @@
...
+            .filter(
+                Q(nodegroup__in=nodegroups)
+                & ~Q(graph_id=settings.SYSTEM_SETTINGS_RESOURCE_MODEL_ID)
+                & Q(graph__isresource=True)
+            )
+            .values_list("graph_id")
```
jacobtylerwalls, line 525: flat=True avoids having to .[0] later.

arches/app/permissions/arches_standard.py
chrabyrd: bleh it may be overkill but abstract a `permissions_base` and have `arches_default_allow` and `arches_default_deny` built over top?

Again this is a naming nit and if it proves to be too much trouble no worries

chrabyrd (changes requested):

arches/app/views/resource.py
aarongundel: I ended up moving these creator methods back into the base class, then I reverted those instances.

arches/app/const.py
aarongundel: I think I'll punt on that change for now in the interest of getting this in.

arches/app/permissions/arches_standard.py
aarongundel: That is probably the _right_ way to do it, but I'm going to punt on it for now in the interest of getting this PR in.

chrabyrd: yeah that's fair

arches/app/models/models.py
```diff
@@ -1162,6 +1162,29 @@ class ResourceInstance(models.Model):
...
+
+    def get_instance_creator(self):
+        create_record = EditLog.objects.filter(
+            resourceinstanceid=self.resourceinstanceid, edittype="create"
+        ).first()
+
+        if create_record:
+            creatorid = create_record.userid
+
+        if creatorid is None or creatorid == "":
```
jacobtylerwalls, line 1183: Wondering if you might convert to int right here, rather than later in get_instance_creator_and_edit_permissions(), and go ahead and catch any exception (to handle other strings besides `''` that will throw if you call int). Similar to #11104.

jacobtylerwalls (approved):
👍

chiatt (approved):
While there are some refinements that can probably still be made, overall this looks really great and tested very well with Arches for Science.

chrabyrd (approved):

PR #11016, opened 2024-06-07
Declare arches applications in AppConfig #11009

### LINKED ISSUE #11009: Introspect AppConfig instead of requiring users to declare ARCHES_APPLICATIONS
In future work, what do you think about removing the ARCHES_APPLICATIONS setting altogether? It's occurred to me before that it's the job of the arches application, not the installer, to know whether the app is arches-related.

If the AppConfig in an arches application had `is_arches_application = True`, then anywhere in core that reads from that setting could just interrogate it like this (probably as a property or something):
```py
In [1]: from django.apps import apps

In [2]: apps.get_app_config('arches_templating').is_arches_application
Out[2]: True
```
...and then anybody installing it doesn't have to know or care about installing it as an ARCHES_APPLICATION, it just goes into INSTALLED_APPS like every other django app and call it a day.

_Originally posted by @jacobtylerwalls in https://github.com/archesproject/arches/pull/11002#discussion_r1629797485_

jacobtylerwalls:
### Types of changes

### Description of Change
If we shift the burden from declaring arches applications from the installer to the publisher, then the installer can just put it in `INSTALLED_APPS` like any django app. We can then presumably extend on this pattern to let the publisher of the app declare whether their app has graphs, business_data, etc., without any further friction for the installer.

The publisher would do this in apps.py:
```py
from django.apps import AppConfig


class MyArchesApp(AppConfig):
    name = 'my_arches_app'
    verbose_name = 'My Arches App'
    is_arches_application = True
```
### Issues Solved

Closes #11009

### Checklist

#### Ticket Background
*   Tested by: @ 
*   Designed by: @


arches/app/utils/thumbnail_factory.py
```diff
@@ -39,4 +39,11 @@ def create(self):
...
 
-ThumbnailGeneratorInstance = ThumbnailFactory().create()
+
+# system_settings does queries (in update_from_db), but queries are
+# not allowed at the module level, so guard under apps.ready.
+# https://github.com/django/django/blob/adae619426b6f50046b3daaa744db52989c9d6db/docs/topics/testing/overview.txt#L201
+if apps.ready:
+    ThumbnailGeneratorInstance = ThumbnailFactory().create()
+else:
+    ThumbnailGeneratorInstance = object()
```
jacobtylerwalls, line 49: This needs more testing to make sure I didn't break the functionality, but as linked in the docs, there could be subtle breakages from having module-level queries anyway, so now's a good time to do something about it.

releases/7.6.0.md
```diff
@@ -272,6 +272,58 @@ Minor incompatibilities:
         ```
         LOCALE_PATHS.insert(0, os.path.join(APP_ROOT, 'locale'))
         ```
+    4. Remove the `ARCHES_APPLICATIONS` setting. If your application is an Arches Application (designed for redistribution), you should create an AppConfig in apps.py with the class attribute `is_arches_application = True`.
+    5. Replace:
```
jacobtylerwalls, line 276: We might be able to add some backward compatibility here by keeping the arches_applications argument as a synonym for the new installed_apps argument. I like the idea of doing that to remove a little friction.

arches/settings_utils.py
```diff
@@ -104,27 +144,45 @@ def build_templates_config(root_dir, debug, app_root=None, arches_applications=N
     ]
 
 
-def transmit_webpack_django_config(
-    root_dir, app_root, static_url, public_server_address, webpack_development_server_port, arches_applications=None
-):
-    arches_applications_paths = {}
+def transmit_webpack_django_config(**kwargs):
```
jacobtylerwalls, line 147: Only by the time I ran into the issue with an ambiguous `import celery` statement (see note below -- checked with @aarongundel and he's run into this on deployments, too) -- did I start needing to feed additional settings. Had the idea to just feed them all in, see `**our_settings` below.

We lose some explicitness, but we hopefully gain some concision in the settings.py templates and avoid churn if we have to keep iterating on this in future versions.

arches/app/utils/thumbnail_factory.py
jacobtylerwalls: I ended up debugging this because django.setup() imports arches in a particular way.

jacobtylerwalls: Never mind, I just worked around this by hiding the circular import of this module.

I think the "module-time import" is not an issue because `THUMBNAIL_GENERATOR` is defined in core arches and so we never go to the database to get the setting. Resolving.

arches/install/arches-templates/project_name/settings.py-tpl
```diff
@@ -411,12 +409,4 @@ except ImportError as e:
...
-    transmit_webpack_django_config(
-        root_dir=ROOT_DIR,
-        app_root=APP_ROOT,
-        arches_applications=ARCHES_APPLICATIONS,
-        public_server_address=PUBLIC_SERVER_ADDRESS,
-        static_url=STATIC_URL,
-        webpack_development_server_port=WEBPACK_DEVELOPMENT_SERVER_PORT,
-    )
-
+    transmit_webpack_django_config(**locals())
```
chrabyrd, line 412: ❤️

arches/app/utils/thumbnail_factory.py
```diff
@@ -19,6 +19,7 @@
 import logging
 import importlib
 
+from django.apps import apps
```
chrabyrd, line 22: cruft?

arches/settings.py
```diff
@@ -21,14 +21,7 @@
...
 
-
-try:
-    from settings_utils import *
-except ModuleNotFoundError:
-    try:
-        from .settings_utils import *
-    except ModuleNotFoundError:
-        pass
+from arches.settings_utils import *
```
chrabyrd, line 24: 👍

arches/settings_utils.py
```diff
@@ -1,24 +1,67 @@
...
+        yield
+    finally:
+        sys.path = _orig_sys_path
+
+
+def arches_applications():
+    from arches import settings as core_settings
+
+    return [
+        mod.name for mod in arches_applications_modules(core_settings.INSTALLED_APPS)
```
chrabyrd, line 29: nit: `module`

arches/settings_utils.py
```diff
@@ -1,24 +1,67 @@
...
+        if path in sys.path:
+            sys.path.remove(path)
+            sys.path.append(path)
+    try:
+        yield
+    finally:
+        sys.path = _orig_sys_path
+
+
+def arches_applications():
```
chrabyrd, line 25: nit: `list_arches_applications` or similar name

arches/settings_utils.py
```diff
@@ -1,24 +1,67 @@
...
+
+
+def arches_applications_modules(installed_apps):
+    from arches import settings as core_settings
+
+    arches_applications_modules = []
+
+    for app in installed_apps or []:
+        if app in core_settings.INSTALLED_APPS:
+            continue
```
chrabyrd, line 40: I'm unsure if this logic block is good for future-proofing. I believe in the future Arches core will used Arches applications

arches/settings_utils.py
```diff
@@ -1,24 +1,67 @@
...
+        try:
+            config = AppConfig.create(app)
+        except ImportError:
+            # Something not pip-installed (the project) might not be importable yet.
+            continue
+        if getattr(config, "is_arches_application", False):
+            arches_applications_modules.append(config.module)
+            break
```
chrabyrd, line 48: why break here? will `arches_applications_modules` return a list of multiple arches apps?

arches/settings_utils.py
```diff
@@ -1,24 +1,67 @@
...
+from contextlib import contextmanager
 
-from pathlib import Path
+import django
+from django.apps import apps, AppConfig
+from django.conf import global_settings, settings
+
+
+@contextmanager
+def move_to_end_of_sys_path(*paths):
```
chrabyrd, line 13: nit: private method yeah? `_move_to_end_of_sys_path`

arches/settings_utils.py
```diff
@@ -1,24 +1,67 @@
...
+
+def arches_applications():
+    from arches import settings as core_settings
+
+    return [
+        mod.name for mod in arches_applications_modules(core_settings.INSTALLED_APPS)
+    ]
+
+
+def arches_applications_modules(installed_apps):
```
chrabyrd, line 33: nit: there's a lot of overlap for having both this and `arches_applications` as public methods. Since `arches_applications` just calls this, could this be moved to a private method and the callsites adjust to instead call `arches_applications`?

arches/settings_utils.py
```diff
@@ -1,24 +1,67 @@
...
 def build_staticfiles_dirs(
-    root_dir, app_root=None, arches_applications=None, additional_directories=None
+    root_dir, app_root=None, installed_apps=None, additional_directories=None
 ):
     """
     Builds the STATICFILES_DIRS tuple with respect to ordering projects,
     packages, additional directories.
 
-    Arguements
+    Arguments
```
chrabyrd, line 60: 😅

arches/settings_utils.py
chrabyrd: Agreed, in this case this is a better pattern 👍

arches/settings_utils.py
```diff
@@ -121,38 +150,44 @@ def build_templates_config(
...
-    root_dir,
-    app_root,
-    static_url,
-    public_server_address,
-    webpack_development_server_port,
-    arches_applications=None,
-):
-    arches_applications_paths = {}
+def transmit_webpack_django_config(**kwargs):
+    is_core = kwargs["APP_NAME"] == "Arches"
```
chrabyrd, line 154: nit: `is_arches_core`

arches/settings_utils.py
```diff
@@ -121,38 +150,44 @@ def build_templates_config(
...
+    our_settings = {k: v for k, v in kwargs.items() if k.isupper()}
```
chrabyrd, line 155: nit: what is `our_settings`? Can this be renamed to something more descriptive?

releases/7.6.0.md
chrabyrd: IMO I'd rather bite the bullet and deal with the friction rather than have the stray setting && small amount of tech debt it'll take to remove it later. Not a super strong opinion though

releases/7.6.0.md
```diff
@@ -301,12 +355,11 @@ Minor incompatibilities:
 
     # Only handle i18n routing in active project. This will still handle the routes provided by Arches core and Arches applications,
     # but handling i18n routes in multiple places causes application errors.
-    if settings.APP_NAME != "Arches" and settings.APP_NAME not in settings.ARCHES_APPLICATIONS:
+    if settings.APP_NAME != "Arches" and settings.APP_NAME not in arches_applications():
```
chrabyrd, line 358: This logic will need reworking I think -- `settings.APP_NAME` is in `INSTALLED_APPS`, and I'm assuming the default for all Arches projects will be that `is_arches_application` is true.

releases/7.6.0.md
```diff
@@ -285,6 +285,58 @@ Minor incompatibilities:
         ```
         LOCALE_PATHS.insert(0, os.path.join(APP_ROOT, 'locale'))
         ```
+    4. Remove the `ARCHES_APPLICATIONS` setting. If your application is an Arches Application (designed for redistribution), you should create an AppConfig in apps.py with the class attribute `is_arches_application = True`.
```
chrabyrd, line 288: I'd argue that this should be included in `arches-templates` and `updateproject.py` , so that the assumption/paradigm continues that all Arches projects are apps, and the only difference is if they're declared in `INSTALLED_APPS` or not.

chrabyrd (commented):
I know this is still a draft but I saw you requested a review. Overall looks good, haven't given it a spin yet but I can grok what's going on.

arches/settings_utils.py
jacobtylerwalls: Can do in this case. Is it the case you prefer multi-line blocks over slightly abbreviated names?

arches/settings_utils.py
jacobtylerwalls: What do you think about starting to use type annotations, at least for function return types -- rather than squeezing this into the name?

arches/settings_utils.py
jacobtylerwalls: thanks, cruft from first attempt

arches/settings_utils.py
jacobtylerwalls: Probably can make it private. The callsites grab the path, not just the name, so we can expose another tiny method like `arches_applications_paths` if you think that's a good direction.

releases/7.6.0.md
jacobtylerwalls: Yeah, this is (one reason) why I wasn't assuming all projects are "arches applications" -- we'll need to find out if there are any blockers to collapsing that distinction.

arches/settings_utils.py
chrabyrd: Yep, every time. I'd rather have an 86-char variable name that tells you exactly what the thing is if necessary, rather an abbreviate it.

Although we're not enforcing it, I like [Google's python styleguide](https://google.github.io/styleguide/pyguide.html#316-naming)

And always [PEP20](https://peps.python.org/pep-0020/#the-zen-of-python) -- which could be argued that this falls under the 2nd aphorism 😀

arches/settings_utils.py
chrabyrd: for me it's not about typing, but naming functions as actions, though typing is never a bad idea 😄

arches/settings_utils.py
jacobtylerwalls: Thanks for the link, there's some great stuff in there.

Just to add to where I'm coming from: in the google style guide, under the single character section, it says "Generally speaking, descriptiveness should be proportional to the name’s scope of visibility." I think that describes why I was okay with `mod` here. It deemphasizes the comprehension variable that falls out of scope immediately. Making it take up more space gives it more importance, whereas abbreviating it suggests to the developer it’s less important than the variable being iterated over and the variable holding the result.

If I encountered a new code base where every list comprehension variable was given as much importance as the other variables, taking up multi-line blocks etc., I would find it a little harder to get up to speed.

This isn’t a list comprehension example, but earlier in that style guide, this signature is given as a blessed example, abbreviating `imag` for `imaginary`:
```py
Yes: def complex(real, imag=0.0): return Magic(r=real, i=imag)
```

...but it's a little hard to tell if that was intentional, because that example was focused on the spaces around `=`, not comprehensions.

In the comprehension section, I see:
```py
result = [mapping_expr for value in iterable if filter_expr]
```
So `mapping_expr` instead of `mapping_expression`, same with `filter_expr` etc.

***
I won't ask to change module back to mod, but I do think we need to use common sense and not rely on a one-size-fits-all rule. I also fully support the idea of erring on the side of longer names. It was a bit forbidding to work on the rdffile.py file recently, which uses a lot of one-and-two-char names like `vi`.

chrabyrd: No that's fair, and honestly a nitpicky on my part. Appreciate it and I'll be mindful in the future about letting some short-naming slide 👍

releases/7.6.0.md
jacobtylerwalls: They will have to change settings.py either way, the question is just whether to allow changing it in two fewer places, but after sitting with it for a bit it does have the potential to be confusing, so after sitting with it for a bit, it's probably better to not leave in the stray argument.

releases/7.6.0.md
jacobtylerwalls: Now I just check if the project is last in INSTALLED_APPS. Open to ideas.

arches/settings_utils.py
```diff
@@ -121,38 +129,63 @@ def build_templates_config(
...
-def transmit_webpack_django_config(
-    root_dir,
-    app_root,
-    static_url,
-    public_server_address,
-    webpack_development_server_port,
-    arches_applications=None,
-):
-    arches_applications_paths = {}
+def inject_arches_applications_directories():
```
chrabyrd, line 132: The logic in this function doesn't hold up to all the conditional cases in the other `settings_utils` functions whose logic it's replacing. Can they not exist unaltered?

chrabyrd (changes requested):
Overall getting there but there's still some stuff to work out IMO

chrabyrd:
@jacobtylerwalls 


I appreciate all the hard work you've put into this PR. As it’s taking shape, I’m starting to feel that we might be adding more complexity than necessary, and it may be overall better to just stick with the settings declaration. I’d love to hear your thoughts on this and I'm for sure open to discussion.

arches/app/models/models.py
```diff
@@ -412,6 +411,8 @@ def save(self, *args, **kwargs):
         super(File, self).save(*args, **kwargs)
 
     def make_thumbnail(self, kwargs_from_save_call, force=False):
+        from arches.app.utils.thumbnail_factory import ThumbnailGeneratorInstance
```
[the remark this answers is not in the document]
jacobtylerwalls: Yep, if you don't hide the circular import like this, you get this when transmitting settings. py to nodejs:
```py
  File "/Users/jwalls/prj/arches/arches/app/utils/thumbnail_factory.py", line 48, in <module>
    ThumbnailGeneratorInstance = ThumbnailFactory().create()
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jwalls/prj/arches/arches/app/utils/thumbnail_factory.py", line 29, in create
    if settings.THUMBNAIL_GENERATOR:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jwalls/prj/arches/arches/app/models/system_settings.py", line 69, in __getattr__
    self.update_from_db()
  File "/Users/jwalls/prj/arches/arches/app/models/system_settings.py", line 90, in update_from_db
    from arches.app.datatypes.datatypes import DataTypeFactory
  File "/Users/jwalls/prj/arches/arches/app/datatypes/datatypes.py", line 29, in <module>
    from arches.app.utils.permission_backend import user_is_resource_reviewer
  File "/Users/jwalls/prj/arches/arches/app/utils/permission_backend.py", line 19, in <module>
    from arches.app.models.models import ResourceInstance, MapLayer
ImportError: cannot import name 'ResourceInstance' from partially initialized module 'arches.app.models.models' (most likely due to a circular import) (/Users/jwalls/prj/arches/arches/app/models/models.py)
```

arches/install/arches-templates/project_name/settings.py-tpl
```diff
@@ -130,11 +130,9 @@ INSTALLED_APPS = (
     "oauth2_provider",
     "django_celery_results",
     # "silk",
-    "{{ project_name }}",
+    "{{ project_name }}",  # must be last
```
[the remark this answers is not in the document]
jacobtylerwalls: Right, I don't love this either. To my knowledge, the only consumer that needs to know whether some arches project is the last arches project in the chain is urls.py, so it knows whether to append the i18n routes or not.

More on that below.

arches/install/arches-templates/project_name/apps.py-tpl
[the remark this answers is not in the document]
jacobtylerwalls: We have system checks defined in the apps.py in arches core. In terms of not having an AppConfig there, it's just because we're taking the default one from Django. If we decide core arches is an "arches application" you are correct we will need an explicit AppConfig, but I didn't think we were doing that.

releases/7.6.0.md
```diff
@@ -312,17 +364,34 @@ Minor incompatibilities:
...
+    if settings.APP_NAME == settings.INSTALLED_APPS[-1]:
         if settings.SHOW_LANGUAGE_SWITCH is True:
             urlpatterns = i18n_patterns(*urlpatterns)
 
         urlpatterns.append(path("i18n/", include("django.conf.urls.i18n")))
-        
     ```
 
-11. Run `python manage.py updateproject`
+11. Add an `apps.py` in the project root:
```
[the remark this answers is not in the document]
jacobtylerwalls: Sure, no problem. The only reason I waited was because I wasn't sure what `verbose_name` to derive for an arbitrary name, but we can just do whatever Django does by default.

arches/settings_utils.py
jacobtylerwalls: Happy to walk through my thought process, this part is still pretty request-for-comment.

There are three places in settings.py where we need information about arches applications. Two of them are in the settings themselves, in other words, in the middle of creating a constant in settings.py we call a settings util to inject stuff. The third place is at the bottom of settings.py where we transmit certain settings to nodejs to build the frontend.

The basic idea in this PR is to shift the burden of declaring arches applications to the arches application, using the standard Django AppConfig and app registry. That's not available until you run django.setup(), which requires you to know what your settings are, which was basically the first thing you and I both found when hacking on this -- we're caught in a bind where we want to dynamically alter settings based on the value of other settings/apps.

So the ideas I've been toying with here are non-invasive ways to call django.setup(). django.setup() assumes you have settings you can feed to it. So that's a perfect fit for that third settings.py location where we're feeding settings. It's not a great fit for the first two cases where we're basically injecting stuff into the assignment of a settings constant. A few options for those cases:

- forget the app registry/documented django interface, just directly import the module with importlib (pros: works, cons: hacky, doesn't account for the way django lets you define multiple configs, and it's figured out for you)
- manually create AppConfigs, kind of like the registry does (pros: kinda works, cons: might bork later callers of django.setup()?)
- defer the injecting of arches application templates/staticfiles information to the apps.py ready() method (pros: more idiomatic for django, cons: ...?)

This latest iteration of this PR is taking this last approach.

> The logic in this function doesn't hold up to all the conditional cases in the other settings_utils functions whose logic it's replacing.

This latest approach wasn't supposed to miss any logic, so if you can help point me to what I'm missing, I'd very appreciative 🙏

jacobtylerwalls:
> I’m starting to feel that we might be adding more complexity than necessary, and it may be overall better to just stick with the settings declaration. I’d love to hear your thoughts on this and I'm for sure open to discussion.

I hear that. Let me take one last go at removing the requirement at a specific INSTALLED_APPS ordering and see how it looks after that.

arches/install/arches-templates/project_name/settings.py-tpl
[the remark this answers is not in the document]
jacobtylerwalls: Found a way to do this without relying on the INSTALLED_APPS order, see [aa108b3](https://github.com/archesproject/arches/pull/11016/commits/aa108b3c12dd9f656daf8702db8fcb6c9ee1fbf3).

arches/install/arches-templates/project_name/urls.py-tpl
```diff
@@ -9,7 +9,7 @@ urlpatterns = [
 
 # Only handle i18n routing in active project. This will still handle the routes provided by Arches core and Arches applications,
 # but handling i18n routes in multiple places causes application errors.
-if settings.APP_NAME != "Arches" and settings.APP_NAME not in settings.ARCHES_APPLICATIONS:
+if settings.ROOT_URLCONF == __name__:
```
chrabyrd, line 12: clever! 🤓

arches/management/commands/updateproject.py
```diff
@@ -173,6 +174,28 @@ def update_to_v7_6(self):
                             os.path.join(dirpath, filename[:-7] + ".py"),
                         )
 
+        if os.path.isfile(os.path.join(settings.APP_ROOT, "apps.py")):
```
chrabyrd, line 177: this will need to be hardened, as `updateproject` calls all  methods in this file. So conceivably it could get called in a future iteration where `is_arches_application=True` has already been added

arches/management/commands/updateproject.py
```diff
@@ -262,7 +286,10 @@ def update_to_v7_6(self):
                 file.close()
 
                 updated_file_data = (
-                    file_data.replace("{{ project_name }}", settings.APP_NAME)
+                    file_data.replace(
+                        "{{ project_name_title_case }}", settings.APP_NAME.title()
```
chrabyrd, line 290: why is this / verbose_name in `apps.py` needed?

arches/settings_utils.py
chrabyrd: ah fair, on second readthrough this makes sense

chrabyrd, line 132: An issue I have with this is that it's obfuscating how `STATICFILES_DIRS` and `TEMPLATES` gets built. Imagine I'm a new Arches user and I want to know how my Arches-application templates are getting rendered. The function called in `settings.py` has no reference to it and without combing the rest of `settings_utils` I'm left not knowing how the thing is made.

This could be a signposting issue. But preferably I'd like the thing to get built in all the same place.

arches/install/arches-templates/project_name/apps.py-tpl
chrabyrd, line 1: IMO Arches-core should have this same pattern. There's some app-breaking logic I worked around in an abandoned ticket in `importer.py` and `index_database.py` -- so if, say, `setup_db` gives you grief after the change I know a way around it.

chrabyrd (changes requested):
Ok this is looking pretty solid.  There's a few things I've brought up and I'll kick the tires once more before approving if we choose to stay on this path 👍

arches/management/commands/updateproject.py
jacobtylerwalls: I'm doing this string replacement in two places:
- one, the class name for the AppConfig class, e.g. DiscoConfig. If the project name is "disco", we don't want a class called "discoConfig", we want "DiscoConfig".
- two, for the `verbose_name`: it's shown in all the [basic doc'd examples](https://docs.djangoproject.com/en/5.0/ref/applications/). We could decide to remove it, since it defaults to label.title() anyway, but I liked the idea of leaving it in to "leave it in the user's face" so they might consider using a better one if it looks wrong, e.g. "Arches for Science" instead of "Arches_For_Science" (default)

arches/app/models/models.py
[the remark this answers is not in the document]
jacobtylerwalls: If that change is acceptable, happy to do it 👍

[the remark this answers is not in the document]
jacobtylerwalls: I have a related change I'd like to make to increase safety around this, so I'll pull this out into a new PR

arches/management/commands/updateproject.py
jacobtylerwalls: Sorry, I'm not quite following. If apps.py exists in a future iteration, or even today (in the case of arches-templating), this just logs a UserWarning telling the user what to do. Do you mean change the wording of the UserWarning?

chrabyrd: ah sorry, I mean read the file && check if `is_arches_application=True` already exists. Someone could feasibly make this change to move from 7.5 -> 7.6, but if we release a 7.6.1 that requires running `updateproject` again they'd still get a warning.

arches/settings_utils.py
chrabyrd: Yeah if this able to be removed entirely in favor of that setting change, I'd say it's worth being folded into this ticket.

arches/management/commands/updateproject.py
chiatt: I think I see @chrabyrd's point, but at the same time it's very unlikely that we would ask a user to run updateproject on a patch release.

chrabyrd: True and fair @chiatt , but then how about an 8.0 release? Presumably `updateproject` will be run, and it if we follow the current pattern it will run all functions declared here.

Also open to a discussion around changing the pattern of course.

chiatt: Good point, but I should point out that this method is named `update_to_v7_6`. Version 8.0 is still a long way off and generally the updateproject file changes significantly between feature releases, let alone a major one.

chiatt: We expect people to be upgraded to the last feature release before an upgrade. So we can remove past upgrade methods and probably should.

chrabyrd: Just be clear, you're saying that you expect this pattern to change? https://github.com/archesproject/arches/blob/dev/7.6.x/arches/management/commands/updateproject.py#L26

and that we _don't_ just tack `self.update_to_v8_0()` to the end of it? If so, cool and I'm on board 👍

chrabyrd: ah -- once again you beat me to the comment 🙂

releases/7.6.0.md
```diff
@@ -302,7 +303,9 @@ Minor incompatibilities:
     3. If it exists, remove `${project_name}/webpack/webpack-user-config.js`
 
 10. In settings.py:
-    1. Remove `"compressor"` from `INSTALLED_APPS`
+    1. In `INSTALLED_APPS`:
+        - remove `"compressor"`
+        - add `"arches.apps"`
```
jacobtylerwalls, line 308: There's a fanciness judgment call here ✨:

if we don't like having to add "arches.apps" to be able to let django find "arches.apps.templates" by default, we can create a custom loader that does something fancy like: look for every app except core arches under /templates, but look for core arches under app/templates. I just paused for input before adding such an invasive workaround.

jacobtylerwalls: Another option is to just keep the approach before we discussed on Tuesday at core standup and just admit a little indirection.

releases/7.6.0.md
```diff
@@ -302,7 +303,10 @@ Minor incompatibilities:
     3. If it exists, remove `${project_name}/webpack/webpack-user-config.js`
 
 10. In settings.py:
-    1. Remove `"compressor"` from `INSTALLED_APPS`
+    1. In `INSTALLED_APPS`:
+        - remove `"compressor"`
+        - add `"arches.app"`
+        - Reorder anything also present in `ARCHES_APPLICATIONS` to a position before `"arches.app"` to preserve their higher preference with respect to templates and static files.
```
chrabyrd, line 309: We're spawning projects' `settings.py` with `arches.app` somewhere in the middle. And I can forsee a huge headache caused by needing to ensure `arches.app` goes at the end of installed apps. I can't imagine a scenario where a user would want arches-core templates to overwrite all arches_applications templates (+ static files).

I like this pattern better than `inject()`-ing inside `apps.py`. IMO it would be worth adding a small bit of logic in `settings.py` to append this to `INSTALLED_APPS`

chrabyrd (changes requested):
overall looking good just a few last things

releases/7.6.0.md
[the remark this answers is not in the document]
jacobtylerwalls: Oh, the change in thumbnail_factory.py? I guess I'm not sure we can prescribe whether the user truly wants the system-settings-graph-settings or not. We should probably just fix the circular import in system_settings.py. I'll open a ticket.

releases/7.6.0.md
jacobtylerwalls: Nice idea!

releases/7.6.0.md
```diff
@@ -302,7 +303,12 @@ Minor incompatibilities:
     3. If it exists, remove `${project_name}/webpack/webpack-user-config.js`
 
 10. In settings.py:
-    1. Remove `"compressor"` from `INSTALLED_APPS`
+    1. In `INSTALLED_APPS`:
+        - remove `"compressor"`
+        - Add this line afterward to ensure "arches.app" (provider of core arches templates) always follows any Arches Applications in `INSTALLED_APPS`:
```
chrabyrd, line 308: nit: this should be a separate step

releases/7.6.0.md
```diff
@@ -316,6 +322,67 @@ Minor incompatibilities:
...
+        STATICFILES_DIRS =  build_staticfiles_dirs(
+            root_dir=ROOT_DIR,
+            app_root=APP_ROOT,
+            arches_applications=ARCHES_APPLICATIONS,
+        )
+        ```
+        with:
+        ```
+        STATICFILES_DIRS =  build_staticfiles_dirs(
+            root_dir=ROOT_DIR,
```
chrabyrd, line 337: I think this line is not needed?

chrabyrd (changes requested):
small nits in release doc, otherwise I'm happy with how this turned out from a design perspective.

One major bug I found is there seems to be an issue with static files.

Repro steps:
- Make a project ( `foo` )
- Make a second project ( `bar` )
- Install `bar` in your virtual ENV
- Add `bar` to INSTALLED_APPS inside `foo`
- In `foo`'s `media/reports/default.js` add a console log
- In `bar`'s `media/reports/default.js` add a different console log
- For kicks, in arches core `media/reports/default.js` add yet another console log
- in `foo`, build the static bundle, create a dummy graph/resource instance and navigate to the report. It should show the console log in `foo`'s  default.js
- delete `foo/media/js/reports/default.js`
- re-build the static bundle

The expectation is that I would then see `bar`'s console log. However I see arches-core console log.

jacobtylerwalls:
> The expectation is that I would then see bar's console log. However I see arches-core console log.

Thanks for the repro steps -- I followed it and got the same result.

Then I retested on dev/7.6.x, and I found that in the sanity-check step before deleting any files, I see my `foo` log statement, when we would expect `bar`. And then removing and rebuilding obviously falls back to core.

So, my impression is that we're improving the situation slightly with this PR. Do you see the same?

jacobtylerwalls:
(Hm, if project files are supposed to overwrite arches applications files, perhaps the new behavior isn't really an improvement.)

chrabyrd:
Ah -- yeah sanity checks for the win. Once https://github.com/archesproject/arches/pull/11147 is merged into `dev/7.5.x` and  `dev/7.5.x` is merged into `dev/7.6.x` and then those changes merged/tested there, we should be g2g 👍

jacobtylerwalls:
Retested the scenario from above, looking clear now 😎

chrabyrd (changes requested):
Hmm the regression goes away but now the project is coming across as an arches_application

```
Data imported from settings.py: {
  APP_ROOT: '/Users/cbyrd/Projects/foo/foo',
  ARCHES_APPLICATIONS: [ 'foo', 'bar' ],
  ARCHES_APPLICATIONS_PATHS: {
    foo: '/Users/cbyrd/Projects/foo/foo',
    bar: '/Users/cbyrd/Projects/bar/bar'
  },
  SITE_PACKAGES_DIRECTORY: '/Users/cbyrd/Projects/ENV/lib/python3.10/site-packages',
  PUBLIC_SERVER_ADDRESS: 'http://localhost:8000/',
  ROOT_DIR: '/Users/cbyrd/Projects/arches/arches',
  STATIC_URL: '/static/',
  WEBPACK_DEVELOPMENT_SERVER_PORT: 9000
}
```

which I _almost don't hate_. There could be conversation there where we really don't care about `APP_ROOT` and instead just treat any project like an Arches app. But IMO that could require some intense changes and is likely out-of-scope for this

jacobtylerwalls:
Ah I thought we had consensus on collapsing the app/project distinction, but no problem to cleave that off here and propose it separately to vet it throughly 👍 

I'll just flip the bit in the apps.py template to set `is_arches_application = False` and just document that when you're ready to graduate to a reusable app you need to flip the bit.

Hopefully flipping the bit will become unnecessary!

chrabyrd:
> Ah I thought we had consensus on collapsing the app/project distinction, but no problem to cleave that off here and propose it separately to vet it throughly 👍
> 
> I'll just flip the bit in the apps.py template to set `is_arches_application = False` and just document that when you're ready to graduate to a reusable app you need to flip the bit.
> 
> Hopefully flipping the bit will become unnecessary!

Ah I may have missed that. Ideally I'd rather have the settings_util updated to filter our the project its called from, but if that's not kosher I'd much rather have it as it currently is instead of needing the user to flip a bit to make a project an Arches app

jacobtylerwalls:
Oh perfect, I'll do that.

Then I would want to clarify the settings.py <--> webpack interface to use a key like  "arches applications other than this one/app root" (overly descriptive ? 😛 ) so we can keep using `is_arches_application` in apps.py to represent "anything built with arches", no bit flipping necessary.

Then we can decide what we mean by "arches application" -- do we mean "anything you build with arches"? (my hope)

chrabyrd:
Oh fair! Yeah I'm down to update the key to something like `EXTERNAL_ARCHES_APPLICATIONS`.


But taking a step back, this all may be unnecessary. I'd say don't a change a thing today, I'd like some time to think if what we currently have will break anything. Because the more I think on it, the more _sorta like_ having the project also be counted as an Arches application 👍

jacobtylerwalls:
No worries. My expectation is that it will give implementers a little more control over _what overrides what_ via just reading from the order of INSTALLED_APPS. We can provide sensible defaults, but ultimately the implementer can do whatever they want. (Should feel natural to them if coming from a Django background, which was my takeaway from the convo after core standup the other day.)

arches/install/arches-templates/project_name/settings.py-tpl
```diff
@@ -145,10 +145,12 @@ INSTALLED_APPS = (
     "oauth2_provider",
     "django_celery_results",
     # "silk",
-    "{{ project_name }}",
+    "{{ project_name }}",  # Ensure the project is listed before any other arches applications
```
chrabyrd, line 148: Added this instruction, though I'm wondering if there's a better way to ensure ordering?

chrabyrd (commented):
@jacobtylerwalls this looks good and I'm on board. Rolled a few changes into this having to do with the ordering of Project -> Arches Apps -> Arches core, and some polish around always having at least 1 Arches app now. Once the merge conflict has been resolved and you can confirm that this is still functioning how you'd like, this should be good to approve 👍

arches/install/arches-templates/project_name/settings.py-tpl
jacobtylerwalls: Nice. Let's see how this goes in practice: we can keep tinkering with the docs without it being a breaking change. I'm thinking it's the responsibility of the readme in the arches application to make this as clear as possible for their users.

arches/management/commands/updateproject.py
johnatawnclementawn: > We expect people to be upgraded to the last feature release before an upgrade. So we can remove past upgrade methods and probably should.

Resolved in #11176

jacobtylerwalls:
Thanks for the review @chrabyrd!

chrabyrd (approved):
lgtm! 👍

PR #11023, opened 2024-06-10
7.6.x -> 8.0.x

jacobtylerwalls:
There are a number of ways to test this, but one possibility is to focus on the merge commit and ensure the correct 7.6 changes trickled up.

If you want to check the black commit, you can run black yourself (after cherry-picking the prep commit) and diff the output.


chrabyrd (approved):
lgtm! 👍

PR #11047, opened 2024-06-13
Fine-tune project template tooling

jacobtylerwalls:
Noticed in arches-lingo:
- these constants in base_test.py are only used in core.
- without specifying tab width, prettier defaulted to two spaces. we missed this in core because we use an .editorconfig
- separate formatting steps

### DIFF
--- a/arches/install/arches-templates/.github/workflows/main.yml
+++ b/arches/install/arches-templates/.github/workflows/main.yml
@@ -61,9 +61,12 @@ jobs:
             echo "package.json not found, skipping npm commands."
           fi
 
-      - name: Check formatting
+      - name: Check frontend formatting with prettier
         run: |
           npm run prettier:check
+
+      - name: Check backend formatting with black
+        run: |
           black . --check --exclude=node_modules
 
       - name: Run frontend tests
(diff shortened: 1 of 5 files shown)


arches/install/arches-templates/.github/workflows/main.yml
```diff
@@ -61,9 +61,12 @@ jobs:
...
           fi
 
-      - name: Check formatting
+      - name: Check frontend formatting with prettier
         run: |
           npm run prettier:check
+
+      - name: Check backend formatting with black
+        run: |
           black . --check --exclude=node_modules
```
jacobtylerwalls, line 70: Separated these as we got multiple reports of users finding the concat'ed logging of prettier and black confusing.

The other approach here would be to just run the pre-commit hooks in CI, but we have some overlap with the webpack build step there. I think this works for now, but we can reevaluate if we add other hooks.

jacobtylerwalls:
Hey @chrabyrd, let me know if you're available for a review. Thanks.

chrabyrd (approved):
lgtm! 👍

PR #11053, opened 2024-06-18
Ensure arches applications work with github actions #11049

### LINKED ISSUE #11049: Ensure Github actions work with projects that use Arches applications

chrabyrd:
### Types of changes

### Description of Change

Allows to GH test runners to work with ARCHES_APPLICATIONS

### Issues Solved

Closes #11049 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chrabyrd:
Still waiting to update 7.6.0 and maybe arches-docs, but wanted to get agreement on method beforehand 👍

aarongundel:
After talking with you @chrabyrd, it makes sense what you're trying to do here.  It is nice to have a system that just works when you deploy it, and this will go a long way towards doing that.  At the same time I keep wondering if we're being too prescriptive in the project templates.  Perhaps not, but assuming that user will make use of github actions in their project may or may not be true.  That makes me wonder if this should be in core (or perhaps broken out into some other repository? obviously that would add complexity and perhaps make this moot).  At the same time, I fully understand and appreciate the reasoning behind this PR.

chrabyrd:
> After talking with you @chrabyrd, it makes sense what you're trying to do here. It is nice to have a system that just works when you deploy it, and this will go a long way towards doing that. At the same time I keep wondering if we're being too prescriptive in the project templates. Perhaps not, but assuming that user will make use of github actions in their project may or may not be true. That makes me wonder if this should be in core (or perhaps broken out into some other repository? obviously that would add complexity and perhaps make this moot). At the same time, I fully understand and appreciate the reasoning behind this PR. To be clear - I have no real objections to this method.

:wave: @aarongundel !

I appreciate the insight. Yeah I had thought of making this into an Arches application, and therefore having it be an easy drop-in for projects that desire it. However it seems that Github actions are one of the few things that won't work with an  arches application -- it requires the `.github` folder and associated files to be physically in the repo.  And you're right, it's an assumption that the project will use github to host their Arches project, and that they'll want to use actions. However I'm of the mind that automatic linting/testing/coverage check will want to largely be used and we should make it as easy as possible to implement. I guess if the argument boils down to opt-in or opt-out, I believe that it makes more sense for a project to be opt-out.

.github/actions/install-arches-applications/action.yml
```diff
@@ -0,0 +1,29 @@
+name: 'Install Arches Applications'
```
jacobtylerwalls, line 1: What do you think about renaming this to something like Install Private Arches Applications? The description right below reads very clearly, explaining you probably don't need this if you've declared a (public) python dependency in the usual way.

.github/actions/build-and-test-branch/action.yml
```diff
@@ -0,0 +1,115 @@
...
+    - name: Set up Elasticsearch
+      uses: ankane/setup-elasticsearch@v1
+      with:
+        elasticsearch-version: 8
+
+    - name: Install Python packages
+      run: |
+        python -m pip install --upgrade pip
+        pip install .
+        pip install -r ${{ inputs.project-name }}/install/requirements.txt || true
```
jacobtylerwalls, line 31: Can I ask why this line is necessary? I thought we wanted to encourage putting python dependencies in pyproject.toml in the usual way now that we updated the project templates. Then all you need is `pip install .` and you get everything. (AFAICT, we really only need the install directory now to pass development dependencies to pip.)

Perhaps the README's on arches applications like AfS need to clarify this now that the project templates are updated.

jacobtylerwalls (commented):
I understand better what's going on here. We want to allow people to use private arches applications with the regular project templates. What do you think about updating the PR title and issue to reflect that?

.github/actions/build-and-test-branch/action.yml
jacobtylerwalls: btw, AfS already declares python dependencies on other django apps and arches apps (like arches_templating) in this way, and doesn't have an install dir because it doesn't have its own dev dependencies.

.github/actions/install-arches-applications/action.yml
chrabyrd: Eh I'm unsure, because it's not always private. Let's say you have an application that you've hosted on github but not pypi. You could feasibly use this pattern instead of linking in `requirements.txt`. I agree there needs to be better signposting though

.github/actions/build-and-test-branch/action.yml
jacobtylerwalls: I already started on #11057 :-)

chrabyrd: cool, I'm happy to couch this until #11057  is merged in. However I'd like like to get it in #1041 is merged, if possible I'd like test coverage to get folded into that

chrabyrd:
Un-couching this now 👍

arches/install/arches-templates/.github/actions/install-arches-applications/action.yml
```diff
@@ -0,0 +1,29 @@
+name: 'Install Arches Applications'
+description: 'Manually edit this file to install all Arches Applications declared in settings.py, but not declared in `requirements.txt`'
```
jacobtylerwalls, line 2: Should update this reference and a few lines below

arches/install/arches_admin.py
```diff
@@ -135,6 +135,7 @@ def handle(self, options):
             ".coveragerc",
             "pyproject.toml",
             ".pre-commit-config.yaml",
+            ".github/workflows/main.yml",
```
jacobtylerwalls, line 138: I think this will create mixed slashes on Windows, but I see we have this problem in several places already, so I'll open a new issue for it.

PR #11055, opened 2024-06-18
Fix SyntaxWarning

jacobtylerwalls:
See recent GitHub actions logs with this warning emitted:
```py
/home/runner/work/arches/arches/arches/app/search/components/sort_results.py:25: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if sort_param is not None and sort_param is not "":
```

Could possibly be simplified even further to `if sort_param:`, but would rather keep this scoped smaller / easier to review.


johnatawnclementawn (approved):
lgtm!

PR #11056, opened 2024-06-18
10831 link geojson featureid

### LINKED ISSUE #10831: include database link between geoms in geojson_geometries and individual features in a tile.data feature collection
This issue is essentially about the fact that each geospatial feature exists independent of the tile and resource it's associated with because there can be N features per tile. Basically if you're looking for a particular geom via its `featureid`, you can't find it via the database, only ES. If a given tile has N geoms, the db doesn't know one from the other except via that integer primary key. I would characterize a solution to this as "Nice to Have" as opposed to "necessary" because the `"id"` property on the feature is part of the ES `terms` index, so each individual feature is searchable if you know that id. [This change](https://github.com/archesproject/arches/pull/10827/commits/40959570cb3f87a330ecc461405809f64ab1b3b2) 
 included in PR #10827 also ensures that when we create new tiles with known feature ids they don't get overwritten.

whatisgalen:
### Types of changes

### Description of Change

- creates a new field on the model `GeoJSONGeometry` called `featureid` that points to the `"id"` property on geojson features in tiledata
- creates a migration to add this field and update both the `refresh_geojson_geometries()` and `refresh_tile_geojson_geometries()` postgres functions to save the `"id"` property from tile data features into `featureid` for corresponding rows in `geojson_geometries`
- includes `featureid` as property on MVT called `tile_featureid`

### Issues Solved

Closes #10831 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Found by: @whatisgalen 
*   Tested by: @robgaston 
*   Designed by: @whatisgalen


arches/app/models/migrations/10799_geojsongeometry_featureid.py
robgaston, line 1: @whatisgalen I notice the where clauses have changed in the functions you edited here, but I don't understand why that should be necessary for this change.  It seems to me like we should only be adding the one new column (`featureid`) and not doing anything else.  Am I missing something?

arches/app/models/migrations/10799_geojsongeometry_featureid.py
```diff
@@ -0,0 +1,195 @@
...
+                                )
+                            )).geom,
+                            4326
+                        ),
+                        3857
+                    )
+                ) AS geom,
+                (json_array_elements(
+                    t.tiledata::json->n.nodeid::text->'features'
+                )::json->>'id')::uuid AS featureid
```
robgaston, line 46: the cast to `json` here is unnecessary, I think, as `json_array_elements` returns `json`

arches/app/views/api.py
```diff
@@ -455,8 +455,10 @@ def get(self, request, nodeid, zoom, x, y):
                         cursor.execute(
                             """SELECT ST_AsMVT(tile, %s, 4096, 'geom', 'id') FROM (SELECT tileid,
                                 id,
+                                featureid AS tile_featureid,
                                 resourceinstanceid,
                                 nodeid,
+                                featureid::text AS tile_featureid,
```
robgaston, line 461: you seem to have the column duplicated in this select statement.  Also, is there a reason to use the name `tile_featureid` over the simpler and more consistent `featureid`?

robgaston (changes requested):

arches/app/models/migrations/10799_geojsongeometry_featureid.py
whatisgalen: I can't remember why I added the GROUP BY clause but I've removed it and reformatted the lines so the where clauses are unchanged

arches/app/models/migrations/10799_geojsongeometry_featureid.py
whatisgalen: change in latest

arches/app/views/api.py
whatisgalen: change in latest

robgaston (approved):

PR #11067, opened 2024-06-19
Add principal user

philtweir:
### Types of changes

### Description of Change

Adds the concept of a principal user who can bypass normal permissions logic.

### Issues Solved

Closes #

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/models/resource.py
```diff
@@ -544,7 +553,8 @@ def get_documents_to_index(
 
         document["tiles"] = tiles
         document["permissions"] = {
-            "creator": [self.get_instance_creator()["creatorid"]]
+            "creator": [self.get_instance_creator()["creatorid"]],
```
philtweir, line 556: Based on the call last week, this does seem different to the principal_user case, so I can definitely see the value in having both - but one question, does searching the edit-log table add overhead to get_documents_to_index? (in fact, that is one of the main reasons I had not gone the "userid-on-table" direction, as we are also trying to trim reindexing time)

arches/app/models/migrations/10999_add_principaluserid_to_resources.py
```diff
@@ -0,0 +1,16 @@
...
+
+from django.conf import settings
+from django.db import migrations, models
+from django.db.models import deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("models", "9979_update_stage_for_bulk_edit"),
```
philtweir, line 11: I have to admit, I have not had time to test this as an isolated change, so this seemed like the correct dependency, but would need confirmed.

arches/app/search/components/search_results.py
```diff
@@ -72,6 +73,15 @@ def append_dsl(
             GeoBoundsAgg(field="points.point", name="bounds")
         )
         nested_agg.add_aggregation(nested_agg_filter)
+
+        if self.user and self.user.id:
+            search_query = Bool()
+            subsearch_query = Bool()
+            # TODO: call to permissions framework with subsearch_query
```
philtweir, line 80: This is a placeholder, but presumably the next step would be to add a method to the permissions frameworks to do something/nothing with the user.

(As a side-note, on our fork, we have tried to get a little more separation between the request and the business logic - ideally we would not hand the full request object to the permissions framework, but the user and perhaps a parameter dictionary; does that sound reasonable?)

arches/app/models/resource.py
```diff
@@ -461,6 +468,8 @@ def index(self, context=None):
                         )
                         es_index.index_document(document=doc, id=doc_id)
 
+            resource_indexed.send(sender=self.__class__, instance=self)
```
philtweir, line 471: This is maybe out of scope, so happy to remove (I had not intentionally added it to this commit when reordering commits), but we use it to react to a potential change of permissions data (for example, to refresh a permissions cache entry), and that seems a more likely use-case with these changes.

arches/app/media/js/views/components/search/search-results.js
```diff
@@ -226,6 +226,9 @@ function($, _, BaseFilter, bootstrap, arches, select2, ko, koMapping, GraphModel
                             resourceinstanceid: result._source.resourceinstanceid,
                             displaydescription: result._source.displaydescription,
                             alternativelanguage: result._source.displayname_language != arches.activeLanguage,
+                            principaluser: ko.computed(function () {
+                                return result._source.permissions.principal_user && result._source.permissions.principal_user.includes(ko.unwrap(self.userid));
```
philtweir, line 230: This may or may not be desirable.

philtweir: (but definitely helps for permissions debugging)

arches/app/search/components/search_results.py
philtweir: It may also be prudent to do the principal user "subquery" logic in the permissions framework too, so it just gets the search_query back, as there are definitely cases where the principal_user should not have access to some or all of a resource. (e.g. an application form)

aarongundel (approved):
Thanks for this!

arches/app/models/resource.py
chiatt: @philtweir - We're actually all in on the principal user idea. Our intention was to remove the creator in favor of the principal_user just so that we don't have to do that edit_log query anymore. A small performance win - but I'm sure helpful when processing large volumes of resources. Now that Aaron has merged this PR, he's going to go ahead and remove the creator. It looks like a pretty minor change.

PR #11068, opened 2024-06-20
Keeps original tile-data around so that tile changes can be picked up

philtweir:
### Types of changes

### Description of Change

Keeps the original tile data in the `data` dict that will initialize the "edited" tile, so that (e.g.) tile-save receivers can see from the new tile's `tile._original_data` member, what was in there (and, consequently, what has changed).

### Issues Solved

Closes #

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/views/tile.py
+++ b/arches/app/views/tile.py
@@ -156,6 +156,7 @@ class TileData(View):
                 if tile_id is not None and tile_id != "":
                     try:
                         old_tile = Tile.objects.get(pk=tile_id)
+                        data["_original_data"] = old_tile.data
                     except ObjectDoesNotExist as e:
                         return self.handle_save_error(
                             e,


philtweir:
Can add a test here as well, but keen to get an OK in principle first!

arches/app/views/tile.py
```diff
@@ -156,6 +156,7 @@ def post(self, request):
                 if tile_id is not None and tile_id != "":
                     try:
                         old_tile = Tile.objects.get(pk=tile_id)
+                        data["_original_data"] = old_tile.data
```
chiatt, line 159: My concern with this change is that it isn't used anywhere within core. We avoid changes to core that support external applications, but aren't clearly part of the API because they can make Arches difficult to maintain. Anyone who wants to modify this post method could easily change or remove this line because it would appear to be cruft. On the other hand, if the outside application stops using it, that change wouldn't be apparent within arches, and it would indeed be cruft at that point. We don't want a situation where developers are afraid to make changes to Arches because they could unintentionally break applications that may or may not be using it.

chiatt (commented):
I looked to see if the 'original_data' value is available in the post response, but I don't see it - probably because original_value isn't a property on the tile model, so it doesn't get serialized. I assumed the intent was to send the data back in the response, but If not, I'm not sure how you might be using this change.

Currently in the UI we just store the last value and compare it with the post response. Is that not possible in your use case? If not, maybe there's another way we can approach this. 



h

arches/app/views/tile.py
jacobtylerwalls: My recollection from Slack/standup is that the use case is subclassing the view. Ultimately decomposing this view might make it easier to subclass, but until then, this seems like a noninvasive change to me.

We could just document it with a single test case and then we'd know not to break it (or we could choose to break it if/only if decomposing the view to encourage easier subclassing.)

apeters: I have to agree with @chiatt that adding this, even if it's just one line, sets a dangerous precedent.  I know that there are other developers out there that would point to this and say "well since you allowed this here, can you allow my edit?"  I don't want to be super dogmatic, but this does feel like a slippery slope.  Even if we had a test and documented it, I would still think it's not a great idea.  

Now onto an alternate solution.
If all we want to do is check the previous revision of a tile on save in a signal, you could easily query the edit_log for that information.
```
SELECT *
FROM public.edit_log
WHERE tileinstanceid = 'c85dbd19-9697-4cb7-8997-2c72f0ff3132'
ORDER BY timestamp DESC
LIMIT 2
```
This would give you the current edit and the previous edit.

I do realize that this is an additional db query which is not usually preferred, but I was able to test this against a very large instance of arches that contains 40M edit_log entries.  After adding an index to the `tileinstanceid` column (which you have to have), it took milliseconds to execute that query.

chiatt: > I have to agree with @chiatt that adding this, even if it's just one line, sets a dangerous precedent. I know that there are other developers out there that would point to this and say "well since you allowed this here, can you allow my edit?" I don't want to be super dogmatic, but this does feel like a slippery slope. Even if we had a test and documented it, I would still think it's not a great idea.
> 
> Now onto an alternate solution. If all we want to do is check the previous revision of a tile on save in a signal, you could easily query the edit_log for that information.
> 
> ```
> SELECT *
> FROM public.edit_log
> WHERE tileinstanceid = 'c85dbd19-9697-4cb7-8997-2c72f0ff3132'
> ORDER BY timestamp DESC
> LIMIT 2
> ```
> 
> This would give you the current edit and the previous edit.
> 
> I do realize that this is an additional db query which is not usually preferred, but I was able to test this against a very large instance of arches that contains 40M edit_log entries. After adding an index to the `tileinstanceid` column (which you have to have), it took milliseconds to execute that query.

I'm going to close this PR given @apeters comments and proposed solution.

PR #11072, opened 2024-06-20
Isolate controlled lists into separate django app

jacobtylerwalls:
### Description of Change
These models and views were siloed enough that it raises the question of whether they should be isolated in the django app registry as well to facilitate independent migrations, etc.

### TODO:
- [x] Remove reference datatype from core (probably a bug here with arches applications)
- [x] Rename models/tables to remove redundant word "Controlled" now that the app name contains that word
- [x] Move/update controlled list package command
- [x] Edit plugin config in initial migration: the path is probably wrong
- [x] Functional tests (have only run the unit tests so far, haven't built webpack)
- [x] Nuke run_tests.py, or spin up a new ticket for reorganizing the test runner entrypoint
- [x] Assess packaging impact


tests/run_tests.py
```diff
@@ -1,38 +0,0 @@
...
-        argv = ["tests"]
-
-    TestRunner = get_runner(test_settings)
-    test_runner = TestRunner([], interactive=True)
-    failures = test_runner.run_tests(argv)
-    sys.exit(failures)
-
-
-if __name__ == "__main__":
-    run_all(sys.argv)
```
jacobtylerwalls, line 38: This didn't seem to work as is, and couldn't verify it worked with the new controlledlists tests. We're moving to pyproject.toml, so we may be removing this as soon as 7.6 anyway.

jacobtylerwalls:
@johnatawnclementawn Heads up, the table names changed in the latest, so the `__arches_migrate_collections_to_clm` function needs an update when you have a chance. Thanks.

chrabyrd (changes requested):
Haven't gone too deeply yet but there's a few off-the-cuff things to start:

- can `controlledlists` be delineated?

- can `controlledlists` have its own `src` dir? My thinking is that it should be as seperate from arches as possible. Happy to jump in and hammer out the webpack update for this branch when ready 👍

chrabyrd:
This should be superseded by the work in https://github.com/archesproject/arches/issues/11201

jacobtylerwalls:
Some notes @johnatawnclementawn:
- I pushed a few more commits but you can't see them from the closed PR view, so make sure you use the [branch comparison](https://github.com/archesproject/arches/compare/dev/8.0.x...jtw/isolate-controlled-lists)
- After creating the project, I would (if it were me) work commit by commit, doing a `git checkout <hash>` in arches and copying files over, adjusting names as needed, etc (this will help me as a reviewer)
- Some can be ignored, e.g. the first commit on this branch, as well as the very recent ones adding aliases to the core arches webpack config
- looks like I hadn't yet moved the translations out of javascript.htm, so I would look at the javascript.htm changes in the [reversion PR](https://github.com/archesproject/arches/pull/11209/files#diff-076035587e211aef326ed1561044587af7b83c04a532da48cc0acef16b464528) and make sure those port over

PR #11086, opened 2024-06-25
update to convert multipart geometries to single part

### LINKED ISSUE #10972: Transform multi-part geoms to single-part feature collections when saving tile data
The convert_multipart_to_singlepart should work for this. It can be called in transform_value_for_tile. It probably needs to also be in pre_tile_save. There's some overlap in the roles of these functions that might need review.

aarongundel:
Fixes #10972


arches/app/datatypes/core/geojson_feature_collection.py
```diff
@@ -0,0 +1,430 @@
...
+    def check_geojson_value(self, value):
+        if type(value) is str:
+            geojson = json.loads(value)
+        else:
+            geojson = value
+
+        features = []
+        if geojson["type"] == "FeatureCollection":
+            for feature in geojson["features"]:
+                if feature["geometry"]["type"].find("Multi") != -1:
```
chiatt, line 112: With Python strings you can check for string patterns directly: ` if "Multi" in feature["geometry"]["type"]:`

arches/app/datatypes/core/geojson_feature_collection.py
```diff
@@ -0,0 +1,430 @@
...
+            return geojson
+        else:
+            raise TypeError
+
+    def transform_value_for_tile(self, value, **kwargs):
+        if "format" in kwargs and kwargs["format"] == "esrijson":
+            arches_geojson = self.geo_utils.arcgisjson_to_geojson(value)
+        else:
+            try:
+                arches_geojson = self.check_geojson_value(value)
```
chiatt, line 137: This check also needs to apply to the arches_json on line 134 in case multiparts are coming as `arcgisjson` (as would be the case from the esri addin).

chiatt (changes requested):
This is working great, but still needs to be applied to data coming in as archesjson.

arches/app/datatypes/core/geojson_feature_collection.py
aarongundel: old habits...

arches/app/datatypes/core/geojson_feature_collection.py
aarongundel: a bit circuitous with the dict conversion, but this should work.

arches/app/datatypes/core/geojson_feature_collection.py
```diff
@@ -0,0 +1,434 @@
...
+
+    def validate_from_rdf(self, value):
+        if type(value) == str:
+            # first deserialize it from a string
+            value = json.loads(value)
+        return self.validate(value)
+
+    def pre_tile_save(self, tile, nodeid):
+        if tile.data[nodeid] == "":
+            tile.data[nodeid] = None
```
chiatt, line 432: This will allow `check_geojson_value` to run against `None` if the node value is already `None`.  So I think this check should simply be: 
```
if tile.data[nodeid]:
    tile.data[nodeid] = self.check_geojson_value(tile.data[nodeid])
```

Changing the value from and empty string to None should probably go in the datatype's `clean` method which gets called before this one on save.

The `clean` method gets called before this one, and this logic should actually be there.

chiatt (changes requested):

chiatt (approved):

Issue #11041, opened 2024-06-12
Upgrade `PrimeVue` to 4.0.0-rc.2


chrabyrd:
New release just dropped ( https://www.npmjs.com/package/primevue/v/4.0.0-beta.5 ), and [the issue I raised has been addressed.](https://github.com/primefaces/primevue/issues/5811)

Hopefully we can get this working now 👍

chrabyrd:
It looks like they've moved [themes to it's own package](https://github.com/primefaces/primevue/blob/4.0.0-beta.5/README.md)

but nothing has been updated yet, and even though the [packaged exists on npm](https://www.npmjs.com/package/@primevue/themes) , it appears to not be able to be found when adding it to Arches. Rolling off for now and waiting for a documentation update from the team.

jacobtylerwalls:
Fiexed in #11093

Issue #11043, opened 2024-06-12
Exact-pin all frontend dependencies 


chrabyrd:
This ticket is meant to start a discussion: 

I believe we should exact-pin our frontend dependencies. I believe the stability and reduced chance for application drift is worth the additional overhead of needing to manually update dependencies on a set schedule ( say, every minor+ release ? )

aj-he:
@chrabyrd this would be great. We've had a number of occasions when we've scheduled a release that breaks because overnight a dependency releases something that then breaks deployment pipeline!

jacobtylerwalls:
> EDIT: actually, this tends to happen more with pypi packages than front end so could this be extended to pip installs too?

@aj-he Most python dep's are pinned exactly, looks like only a handful aren't. The Django pin is loosened only for security updates on the assumption those are always stable and to not require fast new patch releases of arches. Do you recall what gave you trouble before?

jacobtylerwalls:
Re: python deps, just noticed today that setuptools advises against (for libraries, like arches--projects being another matter):

https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata

PR #11093, opened 2024-06-26
Update primevue and primevue themes

ryan86:
Update primevue and prime vue themes re #11041


package.json
```diff
@@ -73,7 +74,7 @@
         "moment": "^2.29.4",
         "nouislider": "11.0.3",
         "numeral": "^2.0.6",
-        "primevue": "4.0.0-beta.3",
+        "primevue": "^4.0.0-rc.2",
```
jacobtylerwalls, line 77: Looks like this was an exact pin before, so I'd keep it that way for now, even though this whole pinning strategy is primed (ho ho ho) for a revamp in #11043.

jacobtylerwalls (approved):
A little hello world with a primevue button:

hey-world

***
Thanks!

PR #11095, opened 2024-06-26
Persist dirty state of string widget #11092

### LINKED ISSUE #11092: Unsaved edits are not persisting in text widget
While the widget indicates that there is an unsaved edit when a user switches cards, returning to the widget clears the unsaved value as well as the unsaved indicator in the card tree. 

This was previously resolved by #10027, but seems to have regressed. Also, one of the changes made in the previous fix [here](https://github.com/archesproject/arches/blob/686760ff0203a5ae9ab57e71c0082cac7285c8a9/arches/app/media/js/views/components/widgets/text.js#L183-L185), breaks the text widget in AFS workflows causing it to replace an existing value with an empty string right after instantiation. 

The rich-text-widget doesn't have these issues, so it may be helpful to take a look at how it's implemented in resolving this bug.

chiatt:
### Types of changes

### Description of Change
Persists unsaved value in widget if a user switches between cards in the card tree.

### Issues Solved

Closes #11092

### Checklist

### DIFF
--- a/arches/app/media/js/viewmodels/widget.js
+++ b/arches/app/media/js/viewmodels/widget.js
@@ -96,7 +96,7 @@ define([
 
         if (ko.isObservable(this.value) && ko.isObservable(this.defaultValue)) {
             var defaultValue = this.defaultValue();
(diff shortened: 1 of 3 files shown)


aarongundel (approved):
confirmed issue and fix.

PR #11104, opened 2024-06-28
Catch error when creatorid is ''

ryan86:
Fix issue where provisional editor id is null

### DIFF
--- a/arches/app/models/resource.py
+++ b/arches/app/models/resource.py
@@ -648,11 +648,13 @@ class Resource(models.ResourceInstance):
                     else False
                 )
                 if resource_is_provisional is True:
-                    creator_id = int(
-                        EditLog.objects.filter(
-                            resourceinstanceid=self.pk, edittype="create"
-                        ).values_list("userid")[0][0]
-                    )
+                    creator_id = EditLog.objects.filter(
+                        resourceinstanceid=self.pk, edittype="create"
+                    ).values_list("userid")[0][0]
+                    try:
+                        creator_id = int(creator_id)
+                    except ValueError:
+                        creator_id = ""
                     if creator_id == user.id:
                         permit_deletion = True
             else:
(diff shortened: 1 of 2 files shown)


jacobtylerwalls (approved):
Verified this fixes the AfS issue. Thanks!

arches/app/models/resource.py
```diff
@@ -648,11 +648,13 @@ def delete(self, user={}, index=True, transaction_id=None):
...
-                            resourceinstanceid=self.pk, edittype="create"
-                        ).values_list("userid")[0][0]
-                    )
+                    creator_id = EditLog.objects.filter(
+                        resourceinstanceid=self.pk, edittype="create"
+                    ).values_list("userid")[0][0]
+                    try:
+                        creator_id = int(creator_id)
+                    except ValueError:
+                        creator_id = ""
```
jacobtylerwalls, line 657: This could just `pass` so that if this block grows later we don't wonder why the `""` was used.

jacobtylerwalls (changes requested):
Hey, now that github actions are back up the test coverage is failing. Can you add a test for this scenario?

tests/models/resource_test.py
```diff
@@ -373,6 +373,17 @@ def test_provisional_user_can_delete_own_resource(self):
             result = test_resource.delete(user=user)
             self.assertTrue(result)
 
+        with self.subTest(user="can't delete"):
```
jacobtylerwalls, line 376: Next time, I wouldn't reuse a subTest label, but it's not the worst thing since unlike assertions in a for loop, you at least have separate line numbers here.

The flip side is that using subTest outside a loop seems like not as strong a choice as a test class with shared setup and then separate test methods for the separate assertions, but it's fine to blend in for now 👍

jacobtylerwalls (approved):
Thanks for adding the test!

PR #11119, opened 2024-07-02
Reraise ImproperlyConfigured in SystemSettings #11118

### LINKED ISSUE #11118: Wide catch in `SystemSettings.__getattr__` might cause unexpected database fetch
While developing #11016, I discovered an edge case where SystemSettings, which is supposed to read from settings.py first if can, falls back unexpectedly to database settings.

The scenario is:
- during the Django startup phase (which #11016 manually invokes when transmitting django settings to nodejs/webpack)
- the arches models are imported
- one of those modules has a module-level constant that calls ThumbnailFactory.create()
- that calls settings.THUMBNAIL_GENERATOR
- but settings aren't ready yet, because we're in the middle of the Django startup phase, so Django throws ImproperlyConfigured
- that's caught by the wide catch, and we go to the database unexpectedly
- then, that happened to throw a circular import error, which is the only reason @chrabyrd and I noticed it

Suggest to catch and reraise ImproperlyConfigured instead of proceeding onward. Will demonstrate in a PR.

jacobtylerwalls:
### Types of changes

### Description of Change
Protects against silent failure shown in issue

### Issues Solved
Closes #11118

### Checklist

#### Ticket Background

### Testing considerations
A 'real life' scenario for testing:

Edit the `if __name__ ...` block of settings.py in core arches to include this (along the lines of #11009):
```py
    import django
    from django.conf import global_settings, settings

    settings_defined_here = {k: v for k, v in locals().items() if k.isupper()}
    settings.configure(default_settings=global_settings, **settings_defined_here)
    django.setup()
```

Then cd to the directory containing settings.py and attempt to `python settings.py`. With this change you should get a more appropriate failure instead of a red herring circular import error (emanating because we went to fetch from the database, which doesn't seem correct, and which won't necessarily always fail in future similar situations).

(Then, in #11016, we're going to address the underlying issue with the thumbnail factory by using django.conf settings, not ours)

### DIFF
--- a/arches/app/models/system_settings.py
+++ b/arches/app/models/system_settings.py
@@ -17,8 +17,8 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 """
 
 from django.conf import LazySettings
-from django.db import ProgrammingError, OperationalError
-from django.utils.functional import empty
+from django.core.exceptions import ImproperlyConfigured
+
 from arches.app.models import models
 
 
@@ -42,7 +42,6 @@ class SystemSettings(LazySettings):
 
     def __init__(self, *args, **kwargs):
         super(SystemSettings, self).__init__(*args, **kwargs)
-        # print self
 
     def __str__(self):
         ret = []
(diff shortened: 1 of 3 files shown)


johnatawnclementawn (approved):
Looks good to me 👍🏼

PR #11122, opened 2024-07-02
11080 handle lines

chiatt:
Supports both line and polygon geometries and changes `split_geom` to return feature collections rather than geometries.

A file with a large polygon and line for testing: [biggeojson_w_line.json](https://github.com/user-attachments/files/16072755/biggeojson_w_line.json)

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -1364,7 +1364,11 @@ class GeojsonFeatureCollectionDataType(BaseDataType):
                 }
             )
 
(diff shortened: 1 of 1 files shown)


bferguso:
This looks good and s working as expected - thanks for resolving the issues with this. As an aside / other issue, I'm seeing this: 

image

I'm guessing this is coming from mapbox-gl. I wonder if we should be considering this too. Everything seemed to render and work/search fine so I'm not sure if the error has any substantial effect.

bferguso (approved):
Looks good to me!

Issue #11139, opened 2024-07-05
GeoJSON export issue - card but no geometry in some resources


csmith-he:
Scenario: 

- At least one resource within the GeoJSON export has a geospatial coordinates card containing geometry.  

![image](https://github.com/archesproject/arches/assets/38533555/e87a9357-1a4e-4078-be33-202a49f66310)

- At least one resource within the GeoJSON export has a geospatial coordinates card which doesn't contain geometry

![image](https://github.com/archesproject/arches/assets/38533555/fe4f4d09-8e07-48f3-a484-cc98b8c948c8)

When the search export is run, the to_geojson code handles each resource, whether or not is has geometry, as though is has geometry which means it fails on line 402 in arches/arches/app/search/search_export.py 

![image](https://github.com/archesproject/arches/assets/38533555/386a703e-b3c9-4873-a73d-4beccee34f7f)

![image](https://github.com/archesproject/arches/assets/38533555/2dddb225-2bae-4450-aba4-7ca5d0eb92c3)

There needs to be some validation at this point to ignore resources with no geometry

PR #11140, opened 2024-07-05
11139 Check value is geometry before proceeding with export to geojson

csmith-he:
### Types of changes

### Description of Change

Puts an isinstance check on the geometry field value to make sure the following code is only run on values that whole valid geometry

### Issues Solved
#11139 

### Checklist

    - [X] dev/7.5.x (main support): regressions, crashing bugs, security issues, major bugs in new features
-   [X] I added a changelog in arches/releases

#### Ticket Background
*   Sponsored by: HistoricEngland
*   Found by: @csmith-he 
*   Tested by: @csmith-he

### Further comments

Required for Arches for HERs

### DIFF
--- a/arches/app/search/search_export.py
+++ b/arches/app/search/search_export.py
@@ -398,13 +398,14 @@ class SearchResultsExporter(object):
                             properties[header] = None
                 for key, value in instance.items():
                     if key == geometry_field:
-                        geometry = GEOSGeometry(value, srid=4326)
-                        for geom in geometry:
-                            feature = {}
-                            feature["geometry"] = JSONDeserializer().deserialize(GEOSGeometry(geom, srid=4326).json)
-                            feature["type"] = "Feature"
-                            feature["properties"] = properties
-                            features.append(feature)
+                        if isinstance(value, GeometryCollection):
+                            geometry = GEOSGeometry(value, srid=4326)
+                            for geom in geometry:
+                                feature = {}
+                                feature["geometry"] = JSONDeserializer().deserialize(GEOSGeometry(geom, srid=4326).json)
+                                feature["type"] = "Feature"
+                                feature["properties"] = properties
+                                features.append(feature)
 
         feature_collection = {"type": "FeatureCollection", "features": features}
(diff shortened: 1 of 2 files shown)


releases/7.5.3.md
```diff
@@ -14,6 +14,7 @@ Arches 7.5.3 Release Notes
 - Fixes print from resource manager #11035
 - Displays configured maximum file size and accepted image file formats in photo gallery #11022
 - Alphabetically sort graphs by name in search Resource Type filter #10646
+- Adds check to GeoJSON export that value is a valid geometry # 11139
```
chiatt, line 17: If you move this up a few lines, that will resolve the merge conflict. Also, there is a space between the # and the issue number that will prevent github from creating a link.

arches/app/search/search_export.py
```diff
@@ -398,13 +398,14 @@ def to_geojson(self, instances, headers, name):  # a part of the code exists in
...
                 for key, value in instance.items():
                     if key == geometry_field:
-                        geometry = GEOSGeometry(value, srid=4326)
-                        for geom in geometry:
-                            feature = {}
-                            feature["geometry"] = JSONDeserializer().deserialize(GEOSGeometry(geom, srid=4326).json)
-                            feature["type"] = "Feature"
-                            feature["properties"] = properties
-                            features.append(feature)
+                        if isinstance(value, GeometryCollection):
```
chiatt, line 401: It might be better just to check if there is a value:
```
if value:
```
That should handle empty strings and an `None`. If value something other than one of those, then we probably want it to throw and error because that would help point to a different (and maybe more serious issue).

chiatt (changes requested):

arches/app/search/search_export.py
csmith-he: @chiatt Part of the issue that that a string of 'None' is being passed as the value, which then makes the code fall over:

image

That's why I went with the method of checking the value was a geometry

chiatt (approved):

PR #11156, opened 2024-07-08
Remove celery check from search to improve page load performance

### LINKED ISSUE #11155: Remove celery check to improve search page performance
Checking for the availability of a celery worker is an expensive call that can take between 1 and 6 seconds. We call this on every load of the search page to confirm that celery is available for large search exports. If it's not available and a user is trying to export more instances than is allowed without celery, we notify the user that they must reduce their search because celery isn't running. This isn't very helpful to most users who aren't going to know what vegetables have to do with their search export. 

We should just remove this check entirely and simply inform users that their search exceeds the export limit and that they have to refine it.

chiatt:
### Types of changes

-   [x] Performance improvement 

### Description of Change
Remove celery check from search to improve page load performance

### Issues Solved

Closes #11155

### Checklist


arches/app/templates/views/components/search/search-export.htm
```diff
@@ -216,25 +216,16 @@
         <!-- /ko -->
     </div>
 
-    <!-- Celery Warning -->
-    <!-- ko let: {show_celery_warning: ((total() > {{app_settings.SEARCH_EXPORT_IMMEDIATE_DOWNLOAD_THRESHOLD}}) && (celeryRunning() !== 'True') || (format() === 'html' && total() > {{app_settings.SEARCH_EXPORT_IMMEDIATE_DOWNLOAD_THRESHOLD_HTML_FORMAT}} && celeryRunning() !== 'True'))} -->
-    <div 
-        id="export-warning"
```
jacobtylerwalls, line 222: Seems like we should also remove the `aria-describedby` that's using this (and maybe the `ko let` comment?)

arches/app/templates/views/components/search/search-export.htm
```diff
@@ -216,25 +216,16 @@
...
-    ></div>
-
     <!-- Download -->
+    <!-- ko let: {show_celery_warning: ((total() > {{app_settings.SEARCH_EXPORT_IMMEDIATE_DOWNLOAD_THRESHOLD}}) || (format() === 'html' && total() > {{app_settings.SEARCH_EXPORT_IMMEDIATE_DOWNLOAD_THRESHOLD_HTML_FORMAT}}))} -->
     <div class="search-export download">
         <button class="btn btn-lg btn-primary btn-active-primary"
             type="button"
             data-bind="
-                css: {disabled: (total() === 0) || (total() > {{app_settings.SEARCH_EXPORT_IMMEDIATE_DOWNLOAD_THRESHOLD}} && celeryRunning() !== 'True') || (format() === 'html' && total() > {{app_settings.SEARCH_EXPORT_IMMEDIATE_DOWNLOAD_THRESHOLD_HTML_FORMAT}} && celeryRunning() !== 'True') },
+                css: {disabled: (total() === 0) || (total() > {{app_settings.SEARCH_EXPORT_IMMEDIATE_DOWNLOAD_THRESHOLD}}) || (format() === 'html' && total() > {{app_settings.SEARCH_EXPORT_IMMEDIATE_DOWNLOAD_THRESHOLD_HTML_FORMAT}}) },
```
jacobtylerwalls, line 225: Pre existing issue, but while here would be nice to make the button disabled in fact instead of only styled as disabled.

jacobtylerwalls: ...unless clicking the button is the only way to get feedback, as I don't see any other message around the button if Celery is not running.

chiatt:
Instead of removing this, we can just do a Broker check by setting the `CELERY_CHECK_ONLY_INSPECT_BROKER` setting to True (If we are relatively confident that a worker is running).

PR #11157, opened 2024-07-09
Raise lower bound on Django to 4.2.14

jacobtylerwalls:
7.5.x


chiatt (approved):
Thanks for catching this @jacobtylerwalls!

PR #11158, opened 2024-07-09
Fix i18n text/json field serialization in Django admin #10597

### LINKED ISSUE #10597: Plugins can't be edited in the Django admin
After `Plugin.name` was localized in #10517, you can no longer use the Django admin to edit a plugin. The non-localized name will load in the form, and then saving the form throws "Enter a valid JSON" when saving the field.

Workaround is to hand-edit the name to look like localized JSON again.

Could block people from using the Django admin to set `{"show": True}` in the config field.

Version: 7.6 dev

jacobtylerwalls:
Follow-up to 9334ae1.

### Types of changes

### Description of Change
**Before**
Attempting to edit an internationalized field like `Plugin.name` (string) or `DDataType.defaultconfig` (json) would throw errors in the Django admin ("Enter a valid JSON"). Also, it's possible that if a user worked around this by hand-escaping the JSON, there was a possibility of data loss since not all saved language values (for strings) were populated into the form.

**After**
Now, these issues are fixed. Some testing notes below.

### Issues Solved
Closes #10597

### Checklist

#### Ticket Background
*   Tested by: @ 
*   Designed by: @ 

### Further comments
The `encode()` method was added in 9334ae1d609338462a1743ca3c082a3b73c4b061, because it's what's called when  `json.dumps()` is [called in the Django forms layer](https://github.com/django/django/blob/e095c7612d49dbe371e9c7edd76ba99b6bc4f9f6/django/forms/fields.py#L1401) with this encoder as the `cls` keyword.

### Testing considerations
- [ ] Start with multiple saved values for internationalized string and json fields. The Plugin model has one of each (name, config).
- [ ] Ensure the form loads with all saved values, and can be saved back with or without edits
- [ ] Ensure when Plugin.icon is nulled out, and you attempt to save, then re-enter the missing icon and save back, that there is no "Enter a valid JSON" again on the string/json fields.

### DIFF
--- a/arches/app/utils/betterJSONSerializer.py
+++ b/arches/app/utils/betterJSONSerializer.py
@@ -53,7 +53,11 @@ class JSONSerializer(object):
         self.utf_encode = False
 
     def encode(self, obj):
-        return self.serializeToPython(obj, **self._options)
+        if isinstance(obj, (I18n_JSON, I18n_String)):
+            # Ensure raw JSON strings are re-encoded ("'" -> "\"")
+            # and non-active language values are preserved.
+            return DjangoJSONEncoder().encode(obj.raw_value)
+        return self.serialize(obj, **self._options)
 
     def serializeToPython(self, obj, **options):
         # allow users to override any kwargs passed into the __init__ method
@@ -71,7 +75,7 @@ class JSONSerializer(object):
 
     def serialize(self, obj, **options):
         obj = self.serializeToPython(obj, **options)
-        # prevent raw strings from begin re-encoded
+        # prevent raw strings from being re-encoded
         # this is especially important when doing bulk operations in elasticsearch
         if isinstance(obj, str):
             return obj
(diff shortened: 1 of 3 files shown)


arches/app/utils/betterJSONSerializer.py
```diff
@@ -53,7 +53,11 @@ def __init__(self, **options):
         self.utf_encode = False
 
     def encode(self, obj):
-        return self.serializeToPython(obj, **self._options)
+        if isinstance(obj, (I18n_JSON, I18n_String)):
+            # Ensure raw JSON strings are re-encoded ("'" -> "\"")
+            # and non-active language values are preserved.
+            return DjangoJSONEncoder().encode(obj.raw_value)
+        return self.serialize(obj, **self._options)
```
jacobtylerwalls, line 60: `serializeToPython()` doesn't seem to return strings in every case, so I backed out to `serialize()`, which does. That clarified why I can't just `serialize()` for the I18n fields -- `serialize()` refuses to re-encode, but we _need_ to re-encode in this case to preserve double quotes.

chiatt (approved):
Looks good!

PR #11189, opened 2024-07-15
Move compatibility check from Django template to system check #11114

### LINKED ISSUE #11114: Project's core arches dependency range defined in two places
Defining a project-level dependency in core arches can happen in two places:
- pyproject.toml, as of 7.6
- settings.py: `MIN_ARCHES_VERSION`, and `MAX_ARCHES_VERSION`

Instead of having this in two places, which can cause drift, we could move the check in compatibility.py out of template rendering and into a Django system check. This probably still has value as a check, since not every developer `pip install`'s their own projects, especially if they're not redistributing their app as a reusable app or doing custom development, or using GitHub etc.

If the check is defined at the project level, we can put it in the `apps.py` that will exist as of #11016.

jacobtylerwalls:
### Types of changes

### Description of Change
Avoid maintaining a project's range of compatible arches versions in two places. Also, allows the error to surface even if an application is not rendering Django templates (e.g. vue app only).

### Issues Solved
Closes #11114

### Checklist

#### Ticket Background

### Testing considerations
python manage.py check (or showmigrations, etc)
- test project (without pip installation)
- test pip-installed arches application
- test missing arches requirement
- test invalid arches requirement (not semantic)
- test requirement too high, too low, exact pin to wrong version


arches/apps.py
```diff
@@ -51,3 +58,62 @@ def check_cache_backend(app_configs, **kwargs):
...
+            else:
+                raise ValueError
+        except ValueError:
+            errors.append(
+                Error(
+                    f"Invalid or missing arches requirement",
+                    obj=config.name,
+                    hint=project_requirements,
+                    id="arches.E002",
+                )
```
jacobtylerwalls, line 106: Happy to demote this to a warning if that's better. As is, a new project spun up off a version under dev (e.g. 7.6.0.a1) will fail this check (and thus fail to runserver, since any error-level system check will cause manage.py to quit) because a new project is a created with a range of >=7.6.0,<7.7.0 and 7.6.0-a1 is lower than 7.6.0.

We can also adjust the template during the alpha phase as long as we remember to set it to final before releasing final.

chiatt: What about making it a warning if DEBUG==True and an error otherwise?

releases/7.6.0.md
```diff
@@ -170,7 +172,7 @@ Minor incompatibilities:
     is now a more attractive target for overriding than `run_load_task()`.
 -   `unzip_file()` moved from `arches.setup` to `arches.app.utils.zip`
 -   Version-related utils moved from `arches.setup` to `arches.version`
-
+-   `arches.app.utils.compatibility` was removed in favor of a Django system check.
```
chiatt, line 175: Maybe we should also mention that the settings MIN_ARCHES_VERSION and MAX_ARCHES_VERSION are no longer in use.

PR #11192, opened 2024-07-16
Adds mvt query scoping

aarongundel:


robgaston:
@aarongundel @chiatt do you think it would be worth reworking the return of MVT geoms and related columns as a CTE as shown in the [example](https://postgis.net/docs/ST_AsMVT.html)?  It seems like it is likely a way to improve performance as well by reducing query volume.  ie something like:
```
"""WITH mvtgeom AS (SELECT tileid,
    id,
    resourceinstanceid,
    nodeid,
    ST_AsMVTGeom(
        geom,
        TileBBox(%s, %s, %s, 3857)
    ) AS geom,
    1 AS total
FROM geojson_geometries
WHERE nodeid = %s and resourceinstanceid not in %s and (geom && ST_TileEnvelope(%s, %s, %s)))
SELECT ST_AsMVT(mvtgeom.*,%s)
FROM mvtgeom"""
[zoom, x, y, nodeid, resource_ids, zoom, x, y, nodeid]
```

chiatt (approved):
Looks good - and definitely supercharges the mvt api!

PR #11200, opened 2024-07-17
Remove python2 profiler

jacobtylerwalls:
Discovered while debugging the ArchesTestRunner that python unittest discovery imports modules, and might throw when importing these modules because `hotshot` no longer exists.


aarongundel (approved):
LGTM -  I know of no dependencies here.

Issue #11201, opened 2024-07-17
Convert Controlled List Manager into Arches Application


chrabyrd:

jacobtylerwalls:
@ekansa I'm curious if you have a first impression about to how to handle documentation for arches applications for which we anticipate wide adoption (e.g >95%). Should we maintain the docs for them at archesproject/arches-docs also?

PR #11209, opened 2024-07-18
Remove controlled list manager and reference datatype

jacobtylerwalls:
### Types of changes

### Description of Change
8.0 is 9 months out. In the meantime, as part of #11201, we'll be exploring the idea of publishing the Controlled List Manager & reference datatype as an arches application maintained outside of the core arches repository.

Thus, the work merged in #10569 needs to be reverted.

A separate commit maintains some useful changes that were mixed in with that work and can be merged down into 7.6. Some (but not all) of those changes would be necessary if we decide that the controlled list manager will support 7.6.

Follow-up PR to discuss that: #11210

### Issues Solved
Refs #11201


chrabyrd (approved):
lgtm!

PR #11210, opened 2024-07-18
Pick out useful changes from controlled list work

jacobtylerwalls:
As described in https://github.com/archesproject/arches/pull/11209#issue-2417288140, there were a few useful changes mixed in with the controlled list work that we should preserve in core, ranging from trivial typo fixes to a new hook for datatype validation.

Adding this to 7.6 preserves the ability for the (separately maintained) Controlled List Manager & Reference Datatype to be compatible with 7.6 if it chooses. See #11201.


chiatt (approved):
Looks good!

PR #11235, opened 2024-07-23
Consume core arches types in projects #11234

### LINKED ISSUE #11234: Project typescript linter doesn't find types exported from core arches
#11210 added the following type to core arches, as a baby step to prove out shipping types that projects/arches applications such as the controlled list manager will find useful:

https://github.com/archesproject/arches/blob/2dc5abe88102c3bf80342e4a83a039cdc0810a3c/arches/app/src/arches/types.ts#L1-L8

However, this import, which worked in #10569 when part of arches core:

```ts
import type { Language } from "@/arches/types";
```

Fails (at the ts linter only, not the bundle) in an arches application (archesproject/arches-references@9fa562a) with:

```ts
arches_references/src/arches-references/utils.ts:4:31 - error TS2307: Cannot find module '@/arches/types' or its corresponding type declarations.

4 import type { Language } from "@/arches/types";
                                ~~~~~~~~~~~~~~~~

[6:02:58 PM] Found 12 errors. Watching for file changes.
```

jacobtylerwalls:
### Types of changes

### Description of Change
Addresses unreleased issue in new 7.6 feature.

As described in #11234, the ts linter at the project level cannot resolve a core-arches import aliased with `@` without these changes. See below for an alternative I ruled out.

[Docs](https://www.typescriptlang.org/tsconfig/#baseUrl) for the `baseUrl` field advise removing it anyway:
> This feature was designed for use in conjunction with AMD module loaders in the browser, and is not recommended in any other context. As of TypeScript 4.1, baseUrl is no longer required to be set when using [paths](https://www.typescriptlang.org/tsconfig/#paths).

@chrabyrd I'm totally open to other solutions here; this was just what got things working for me and I wanted to volley it over to you to get your opinion.

@johnatawnclementawn let me know if making the adjustment suggested here to tsconfig.json at the project level resolves it for you also 🙏 

### Issues Solved
Closes #11234

### Checklist

#### Ticket Background
*   Sponsored by:  Getty Conservation Institute

### Further comments
An alternative I considered was changing the imports to end in `.ts`. That moved the error from the import to the usage, showing errors at the usage sites like:
```ts
arches_references/src/arches-references/utils.ts:76:23 - error TS2709: Cannot use namespace 'Language' as a type.

76     selectedLanguage: Language,
                         ~~~~~~~~

[6:11:18 PM] Found 16 errors. Watching for file changes.
```

Fine, even though Language _is_ a type, you can surrender to the linter and pretend it's not a type by taking `typeof`. So let's say you change every use of `Language` to `typeof Language`. Now you have 0 warnings, but now you have no useful type-checking. Try to access `.garbage` on a `Language` object, and notice typescript doesn't complain. Whereas with this change you retain useful type-checking.

### DIFF
--- a/arches/install/arches-templates/tsconfig.json
+++ b/arches/install/arches-templates/tsconfig.json
@@ -14,11 +14,14 @@
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "skipLibCheck": true,
-    "baseUrl": "{{ project_name }}/src/",
     "paths": {
+      "@/arches/*": [
+        "./node_modules/arches/arches/app/src/arches/*.ts",
+        "./node_modules/arches/arches/app/src/arches/*.vue"
+      ],
       "@/*": [
-        "*.ts",
-        "*.vue"
+        "./{{ project_name }}/src/*.ts",
+        "./{{ project_name }}/src/*.vue"
       ]
     }
   },


chrabyrd (changes requested):
Couple of thoughts:

- cool with `baseUrl` going away, good doc dive there
- I'd prefer if importing is the same pattern across the board. Either everything in `src` has its extension appended or nothing does. So if this also works for vue the 👍 
- Is there any other way to do this? To me this isn't extendable. If a project wanted to pull a type from a different project, they'd need to update `tsconfig.json` manually. Can `declarations.d.ts` be leveraged to deal with this instead?

jacobtylerwalls:
> I'd prefer if importing is the same pattern across the board. Either everything in src has its extension appended or nothing does. So if this also works for vue the 👍

I think the reason it doesn't work for vue is that you wouldn't expose types from the script tag of a vue SFC and import them elsewhere with `import type`. For example, a pared down version of import statements in the controlled list manager in its current state looks like:

```ts
import { LIGHT_GRAY } from "@/arches-references/theme.ts";

import { dataIsList } from "@/arches-references/utils.ts";
import ListTree from "@/arches-references/components/tree/ListTree.vue";

import type { Ref } from "vue";
import type { Language } from "@/arches/types";
import type { Selectable } from "@/arches-references/types";
```

The first two blocks are functional imports, and they both use the *.ts and *.vue extensions. So IMO they are consistent with each other. Striking the *.vue extension causes webpack to not build, as you already found before.

What I'm talking about now is the type-only import in the third block. Changing that to add `.ts` causes the problems described in the issue. And it's consistent with third-party type imports like `import type { Ref } from "vue";`.

> Is there any other way to do this? To me this isn't extendable. If a project wanted to pull a type from a different project, they'd need to update tsconfig.json manually.

I agree with you, and I think it's worth proving that out if someone has a cycle to put to it. ☕  But we might want to say "consuming arches application types with the `@` alias in another project" is an 8.0 feature to prove out, whereas consuming a core arches type with the `@` alias in `anything else` is a 7.6 bugfix. WDYT?

johnatawnclementawn (commented):
This works for the use case of arches-references. Is `{{ project_name }}` supposed to be interpolated during `arches-project create` or does that need to be manually changed per project?

chrabyrd:
> I agree with you, and I think it's worth proving that out if someone has a cycle to put to it. ☕  But we might want to say "consuming arches application types with the `@` alias in another project" is an 8.0 feature to prove out, whereas consuming a core arches type with the `@` alias in _anything else_ is a 7.6 bugfix. WDYT?

I can put a cycle to this. If it's not addressed here then it will be addressed almost immediately in the Lingo work, so might as well do it now 👍

jacobtylerwalls:
>  Is {{ project_name }} supposed to be interpolated during arches-project create or does that need to be manually changed per project?

It's interpolated during create project or manage.py updateproject (if you haven't run it yet), so AFAICT the expectation is that projects developing against 7.6 alphas will need to rerun updateproject when 7.6 is finalized.

jacobtylerwalls:
@chrabyrd an alternative I just realized to changing these aliases is to just not use an alias for core arches (or a third-party app) when developing a project:

For example, by changing:
```ts
import type { Language } from "@/arches/types";
```
to:

```ts
import type { Language } from "arches/arches/app/src/arches/types";
```

everything is peachy with the old config.

jacobtylerwalls:
(Oh, missed your prior message. Thanks for giving it a look!)

chrabyrd:
> (Oh, missed your prior message. Thanks for giving it a look!)

No worries! Also this raises a good point about the `@` symbol and if we should even use it 🤔

jacobtylerwalls:
Superseded by #11254

PR #11265, opened 2024-08-01
Sync app template treatment of ignored system checks setting

jacobtylerwalls:
Follow-up to a1ecddc.

This setting already exists when imported from arches, so prefer extending to overwriting.

Was getting this check at the project level:
```sh
?: (guardian.W001) Guardian authentication backend is not hooked. You can add this in settings as eg: `AUTHENTICATION_BACKENDS = ('django.contrib.auth.backends.ModelBackend', 'guardian.backends.ObjectPermissionBackend')`.
```


jacobtylerwalls:
Oh whoops, we discussed this [before](https://github.com/archesproject/arches/pull/11005#discussion_r1647874617). Setting to draft until we can talk about it.

aarongundel (approved):
LGTM

Issue #11180, opened 2024-07-12
Provide the ability for users to find resources that do not have geojson tiles


chiatt:
Users need to be able to identify resources that are either missing tiles with geometries (either no tile or a tile with no value on the geometry node).

Similar to #11179

Issue #11179, opened 2024-07-12
 Provide the ability for users to search for resources without associated files


chiatt:
Users should be able to search on the absence of file data. This may be a change to advanced search in core Arches

PR #11270, opened 2024-08-02
11179, 11180 search for resource w/o files or geometry

apeters:
### Types of changes

### Description of Change
Adds the ability to search for resources that have/don't have files or have/don't have geometry as well as the ability to search on file size and geometry type.

re #11179, re #11180

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


whatisgalen:
this looks like an awesome change. Any chance for fileList advanced search there could be an option for a search criteria of number of files?

chiatt:
The geospatial features work great as does file search when no file is present. However, min and max file size isn't working as expected. In every case I get all documents as if no filter was applied

https://github.com/user-attachments/assets/b40ea107-c121-47b0-9082-ffbf802c9e8e

chiatt (changes requested):
Geospatial and file null search is working as expected. However, file size searches are not. I think that the units are incorrect in the UI, or they need to be converted from mb to kb before the query is performed.

I created two resources - one with a 5.8mb file and one with a 122kb file. I confirmed in the tile data that their sizes were recorded properly. You can see in the video that at 1MB with a greater than search, we should only get the 5.8mb file, but we get both files until the search is increased to 1000mb. 

With a less than search, we get only 1 file at 1000mb, when we should get both if the search was performed in MB. If the search is performed in KB, then the result would make sense.

https://github.com/user-attachments/assets/dde56a4d-4df7-4d09-a7ba-6ebc606b43b4

chiatt (approved):

chiatt:
> this looks like an awesome change. Any chance for fileList advanced search there could be an option for a search criteria of number of files?
> 
> I think you could do it with these changes in `datatypes.py` at line 1308:
> 
> ```python
> def append_to_document(self, document, nodevalue, nodeid, tile, provisional=False):
>         try:
>             for f in tile.data[str(nodeid)]:
>                 val = {
>                     "string": f["name"],
>                     "nodegroup_id": tile.nodegroup_id,
>                     "provisional": provisional,
>                 }
>                 document["strings"].append(val)
>             file_ct_val = {
>                "number": float(len(tile.data[str(nodeid)])),
>                 "nodegroup_id": tile.nodegroup_id,
>                 "provisional": provisional,
>             }
>             document["numbers"].append(file_ct_val)
> ```
> 
> And at line 1692:
> 
> ```python
>  def get_es_mapping(self, nodeid):
>         ret = {
>             "properties": {
>                 "tiles": {
>                     "type": "nested",
>                     "properties": {
>                         "data": {"properties": {str(nodeid): self.default_es_mapping(), "file_count": {"type": "double"}}},
>                     },
>                 }
>             }
>         }
>         return ret
> ```
> 
> And at line 1304:
> 
> ```python
> if value["op"] == "gte" or value["op"] == "lte":
>     operators = {"gte": None, "lte": None, "lt": None, "gt": None}
>     operators[value["op"]] = float(value["val"]) * 1000
>     search_query = Range(
>         field="tiles.data.%s.size" % (str(node.pk)), **operators
>     )
>  elif value["op"] == "eq":
>     field_path = "tiles.data.file_count"
>     search_query = Terms(**{field_path: float(value["val"])})
> ```
> 
> And then on the frontend:
> 
> in `file-list.htm` (change at line 13):
> 
> ```
> select2Config:{
>                 data: [
>                     { text: $root.translations.sizeGreaterThan, id: 'gte' },
>                     { text: $root.translations.sizeLessThan, id: 'lte' },
>                     { text: $root.translations.hasNoValue, id: 'null' },
>                     { text: $root.translations.hasAnyValue, id: 'not_null' },
>                     { text: $root.translations.hasNumberOfFiles, id: 'eq' },
>                 ],
>                 value: op,
> ```
> 
> and in `javascript.htm` (change at line 336):
> 
> ```
> has-number-of-files='{% trans "Has number of files" as hasNumberOfFiles %} "{{ hasNumberOfFiles|escapejs }}"'
> ```

I think this is a good suggestion, but it should probably be an issue for v8.

PR #11276, opened 2024-08-02
Fix updateproject after declarations.d.ts template moved

jacobtylerwalls:
Follow-up to 37d12ca.


apeters (approved):
funnily enough I just submitted a fix for this too!  You beat me to it....

PR #11282, opened 2024-08-05
Loosen line endings check to ignore ontology (.xml) files

jacobtylerwalls:
### Types of changes

### Description of Change
A unix line endings check was added in 7.6 to improve collaboration and prevent checking in code files with mixed line endings. It might be a bit overkill to enforce this for data files like ontology files, however, which we noticed when upgrading Arches for Science and discussed [here](https://github.com/archesproject/arches-for-science/pull/1576#issuecomment-2266277859).

### DIFF
--- a/.github/actions/build-and-test-branch/action.yml
+++ b/.github/actions/build-and-test-branch/action.yml
@@ -61,9 +61,9 @@ runs:
         black . --check --exclude=node_modules
       shell: bash
 
-    - name: Check line endings
+    - name: Check line endings on all but ontology (.xml) files
       run: |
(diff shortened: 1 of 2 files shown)


chiatt (approved):

chiatt:
Not sure why coverage check is indicating a decrease. Nothing in this PR should have lowered it.

jacobtylerwalls:
Probably coverage was lowered slightly on the target after latest merges to dev/7.6.x but those coverage-lowering commits were missing on this branch. I could have rebased, but not worth it 👍

PR #11284, opened 2024-08-05
Add session-based auth REST API endpoints for login, logout #11261

### LINKED ISSUE #11261: Implement API-based user authentication
This is related to: https://github.com/archesproject/arches-lingo/issues/13

Base goal is to have User login/logout functional. Stretch goal is to also include a registration API. MFA/SSO are not scoped in this ticket.

jacobtylerwalls:
### Types of changes

### Description of Change
To support arches applications that do not use the Arches login page interface, create REST API endpoints for login and logout. See extended discussion below.

### Issues Solved
Closes #11261

### Checklist

    - [x] dev/8.0.x

#### Ticket Background

### Further comments
- Third-party packages?
  - Django apps with REST APIs often use third-party packages like Django Rest Framework or django-ninja to implement auth.
  - There is value in not reimplementing the auth "wheel" if possible.
  - However, this is most useful if the project has already committed to using those frameworks for their overall REST APIs. AFAICT that decision is not in scope for this ticket.
  - Django Rest Framework encourages you to [write a LoginView for your login page anyway](https://www.django-rest-framework.org/api-guide/authentication/#sessionauthentication) and inherit from Django's LoginView, as I've done here, so I'm not convinced the views would look any different if we adopted DRF here: we would just be adding plumbing to avoid boilerplate when _subsequent DRF views_ need authentication classes. But we don't know that we're writing any further DRF views. A project is free to do that if they wish and grab these views as part of that.
- Session-based or token-based auth?
  - [According](https://www.django-rest-framework.org/api-guide/authentication/#sessionauthentication) to DRF:
    > Session authentication is appropriate for AJAX clients that are running in the same session context as your website.
  - django-ninja also [defaults to session auth
](https://django-ninja.dev/guides/authentication/)
  - We already have an OAuth based token login endpoint, I'd prefer not to unnecessarily add a duplicate.
  - Suggest opting for the simplest implementation until we confront harder use cases (social auth etc).

### Testing instructions
- See [lingo PR](https://github.com/archesproject/arches-lingo/pull/39)


chrabyrd (commented):
This looks good, just 2 things to talk about:

- Is there a chance to DRY up `views/auth.py`? Or is there enough difference between the interaction that keeping the difference would be prudent?

- I imagine this is a larger conversation, but is Arches eventually going to use DRF? If so this may be a good opportunity to get our foot in the door.

jacobtylerwalls:
> Or is there enough difference between the interaction that keeping the difference would be prudent?

Yeah, almost every line of that other view deals with redirects and template renders. Even the ratelimit decorator can't be shared, b/c we're doing some gymnastics to double the limit to account for the fact that our auth form requires two clicks to submit.

chrabyrd (approved):
lgtm! 👍

PR #11307, opened 2024-08-08
Soften compatibility check to warning in debug mode re #11114

jacobtylerwalls:
Improve the compatibility check from #11114 by softening it to a warning in debug mode. This allows people to continue to develop projects while their arches compatibility is in flux (perhaps with a URL reference to a branch).

Also clarify/standardize the message prefixes with application names, so that they appear like:

```
arches-references: (arches.W001) Cache backend does not support rate-limiting
```
instead of:
```
?: (arches.W001) Cache backend does not support rate-limiting
```


chiatt (approved):
Looks good!

PR #11318, opened 2024-08-09
Ensure longer model names can be used #11317

### LINKED ISSUE #11317: Initial migration can fail if you have a long `db_table` in Model.meta
See failing run on arches-references:
```py
  File "/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django.db.utils.DataError: value too long for type character varying(50)
```

Cause seems to be the call in 0001_initial.py to `create_permissions()` without ensuring that the migrations are ordered in a way that you get the benefit of the [migration](https://github.com/django/django/pull/2857) in the django admin app that boosted the column length on the model names.

Somewhat spooky that this only appeared in an arches app and not while this feature was under dev in arches core, but it's probably something to do with the contenttypes app and how it makes apps visible to other apps.

jacobtylerwalls:
### Types of changes

### Description of Change
Before, longer model names might fail (due to indeterminate inter-app ordering of migrations) if the initial migration for Arches ran before the migration from the Django admin app that lifts the length of the permission name field.

### Issues Solved

Closes #11317

### Checklist

    - [x] dev/8.0.x

### DIFF
--- a/arches/app/models/migrations/0001_initial.py
+++ b/arches/app/models/migrations/0001_initial.py
@@ -112,6 +112,7 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ("contenttypes", "0002_remove_content_type_name"),
+        ("auth", "0002_alter_permission_name_max_length"),
     ]
 
     initial = True
--- a/releases/8.0.0.md
+++ b/releases/8.0.0.md
@@ -7,7 +7,8 @@ Arches 8.0.0 Release Notes
 
 ### Additional highlights
 
-- Add session-based REST APIs for login, logout [#11261]
+- Add session-based REST APIs for login, logout [#11261](https://github.com/archesproject/arches/issues/11261)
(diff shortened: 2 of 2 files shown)


arches/app/models/migrations/0001_initial.py
```diff
@@ -112,6 +112,7 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ("contenttypes", "0002_remove_content_type_name"),
+        ("auth", "0002_alter_permission_name_max_length"),
```
jacobtylerwalls, line 115: This dependency is necessary because this migration itself calls create_permission()

chrabyrd (approved):
❤️

PR #11319, opened 2024-08-09
Add token-based CSS theming #11262

### LINKED ISSUE #11262: Create Arches PrimeVue4 token-based theming pattern

jacobtylerwalls:
### Types of changes

### Description of Change
Begin building a theme preset for arches that can be used/extended in arches applications also.

[PrimeVue docs](https://primevue.org/theming/styled/)

### Issues Solved
Closes #11262

### Checklist

    - [x] dev/8.0.x

#### Accessibility Checklist

#### Ticket Background
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/media/js/utils/create-vue-application.js
```diff
@@ -7,16 +7,60 @@ import StyleClass from 'primevue/styleclass';
...
-import Aura from '@primevue/themes/aura';
-
+import { definePreset } from '@primevue/themes';
 import { createApp } from 'vue';
 import { createGettext } from "vue3-gettext";
+import Aura from '@primevue/themes/aura';
 
 import arches from 'arches';
 
+export const ArchesPreset = definePreset(Aura, {
```
jacobtylerwalls, line 17: I'm opening to moving this to its own file, of course.

chrabyrd: Yep! Please move this to its own file when you get a chance. IMO it should live in `src/arches` but happy to have a discussion around it.

arches/app/media/js/utils/create-vue-application.js
```diff
@@ -7,16 +7,60 @@ import StyleClass from 'primevue/styleclass';
...
+    primitive: {
+        sky: {
```
chrabyrd, line 19: why the `sky` naming?

chrabyrd (changes requested):
Nice! Exciting to start getting theming buckled down

arches/app/media/js/utils/create-vue-application.js
jacobtylerwalls: I'm extending Aura, which has a color named Sky in the palette. I replaced the darkest value sky-950 with our existing sidebar color b/c it was very close. Didn't want to create 10 other new shades of arches blue ;)

arches/app/media/js/utils/create-vue-application.js
jacobtylerwalls: We could use cyan instead. [See here](https://primevue.org/theming/styled/)

arches/app/media/js/utils/create-vue-application.js
chrabyrd: Ah -- I think it's worth having a short dev discussion around color naming just to be sure a good foundation is laid for this. I know that PrimeVue is set up to have many colors for many situations but I believe the Arches palette will/(should?) have like 6-8 colors 👍

arches/app/src/arches/theme.ts
```diff
@@ -0,0 +1,71 @@
...
+    },
+});
+
+export const DEFAULT_THEME = {
+    theme: {
+        preset: ArchesPreset,
+        options: {
+            prefix: "p",
+            darkModeSelector: "system",
+            cssLayer: true,
```
jacobtylerwalls, line 68: I think we can get away without backporting the cssLayer false -> true bit flip, since we can say it's a "new feature" to be able to isolate the arches CSS a la https://github.com/archesproject/arches-references/pull/16#discussion_r1714303546.

arches/app/media/js/utils/create-vue-application.js
jacobtylerwalls: I added a few primitive tokens namespaced under arches and then linked to them from some new semantic tokens. I think this is the idea, LMK how it looks to you. I left in comments that are kind of duplicative for now but show where I extracted the values from in the legacy UI.

arches/app/src/arches/theme.ts
chrabyrd, line 1: nit: How do you feel about `src/arches/themes/default.ts`? I don't feel strongly either way really, but worth a quick discussion IMO

jacobtylerwalls: I like it, let's do it.

chrabyrd:
My assumption is that basic theme-switching will be stubbed out here, but lemme know if that assumption is wrong 👍

jacobtylerwalls:
I think there's value in doing that in follow-up work in another PR, as theme switching isn't really blocking further arches-references development, but having this in is.

arches/app/src/arches/themes/default.ts
```diff
@@ -0,0 +1,63 @@
+import { definePreset } from "@primevue/themes";
+import Aura from "@primevue/themes/aura";
+
+export const ArchesPreset = definePreset(Aura, {
```
chrabyrd, line 4: I'd like this first iteration to be intentional, more of a blueprint for what's getting fleshed out in further iterations. 

I like the idea of `arches` primitives, but I'm unsure of supporting legacy coloring, or at very least without a `-legacy` tag. Ideally moving forward there will be a small subset of colors that define an out-of-the-box Arches application. On that, the color gradation of `arches.blue` seems to buck the logic of what PrimeVue's doing with their other colors: moving up/down the gradient scale. 

Although I like `arches` primitives, maybe it's _not_ the right solution? Maybe our solution is just all semantic tokens: `arches.navigation.blue`, `arches.background`, `arches.button.red` etc etc. 

Should we have a `primary` semantic color that's just iterating the `sky` gradient? IMO I'd say pull that out.

The component overrides, do you think these are generic enough to be universal? Would we ever want a datatable or button label without the added fontWeight, for instance?

`darkModeSelector` should be related to a function that takes in user interaction with the theming selector. Doesn't need to be done in this PR but it's a case to handle around

jacobtylerwalls: Agree on the legacy namespace.

> On that, the color gradation of arches.blue seems to buck the logic of what PrimeVue's doing with their other colors: moving up/down the gradient scale.

Right, my thought also. When we do have that 6-8 color subset, hopefully we can draw from one of those gradients.

jacobtylerwalls: > Should we have a primary semantic color that's just iterating the sky gradient? IMO I'd say pull that out.

Aura uses emerald (green) as its primary color. I'm happy to remove it, I just thought that arches had mostly blue as its primary color, so I wanted to repoint to a blue shade.

jacobtylerwalls: > The component overrides, do you think these are generic enough to be universal? Would we ever want a datatable or button label without the added fontWeight, for instance?

Let's include folks who care about design details in this specific conversation, but yes, the only reason these are here are to enforce it sitewide (obviously subject to further undoing at the component level case by case).

The splitter handle override is definitely universal, if I had the time I'd report it to primevue as a bug. I bet someone fixes it eventually and we can back it out.

arches/app/src/arches/themes/default.ts
```diff
@@ -0,0 +1,63 @@
...
+    },
+});
+
+export const DEFAULT_THEME = {
+    theme: {
+        preset: ArchesPreset,
+        options: {
+            prefix: "p",
+            darkModeSelector: "system",
+            cssLayer: true,
```
jacobtylerwalls, line 60: Do you want to keep this `true` even though we're about to back out the `revert-layer` value?

jacobtylerwalls: Ugh, actually, it causes some problems if we back out the revert-layer stuff. Never mind.

chrabyrd: eh I'd say keep it for now. It should make overwriting the theme easier in the future 👍

arches/app/src/arches/themes/default.ts
jacobtylerwalls: > Maybe our solution is just all semantic tokens: arches.navigation.blue, arches.background, arches.button.red etc etc.

I like that. And if we go that route, then having the legacy tokens as primitives makes sense, because their context is not relevant now, they're not swappable, they're frozen in time.

jacobtylerwalls: With the caveat that I'm kind of hoping arches.navigation.blue resolves to sky.800 if possible rather than a raw value of our own :D

chrabyrd: > Let's include folks who care about design details in this specific conversation, but yes, the only reason these are here are to enforce it sitewide (obviously subject to further undoing at the component level case by case).
> 
> The splitter handle override is definitely universal, if I had the time I'd report it to primevue as a bug. I bet someone fixes it eventually and we can back it out.

Agreed that we should include other people in this conversation 👍 . That said, my kneejerk reaction here is to keep universal overrides at a minimum.

jacobtylerwalls: I'll pull it out -- I'm wondering if this is because something else was going awry with font sizes in the controlled list manager (w/r/t arches chrome and such).

The extra weight looks a little silly in lingo.

arches/app/src/arches/themes/default.ts
jacobtylerwalls: I tried, but the font-awesome icons are all borked. I think we should punt for now, unfortunately.

chrabyrd: no worries! Don't feel super strongly about it either way at the moment 👍

arches/app/src/arches/themes/default.ts
```diff
@@ -0,0 +1,81 @@
...
+import Aura from "@primevue/themes/aura";
+
+const archesColors = Object.freeze({
+    blue: "#579ddb",
+    green: "#3acaa1",
+    red: "#f75d3f",
+});
+
+export const ArchesPreset = definePreset(Aura, {
+    primitive: {
```
chrabyrd, line 11: should a `blue` palette be added as well?

arches/app/src/arches/themes/default.ts
```diff
@@ -0,0 +1,81 @@
...
+        navigation: {
+            list: {
+                padding: "0",
+            },
+            item: {
+                padding: "1rem",
+            },
+            color: "{arches.legacy.sidebar}",
+        },
+        iconSize: "small",
```
chrabyrd: I would recommend remove this, I don't believe we should have our "default" icon size be small ( or that we should offer a "default" icon size)

arches/app/src/arches/themes/default.ts
```diff
@@ -0,0 +1,68 @@
...
+        // Additional tokens
+        content: {
+            gap: "1rem",
+        },
```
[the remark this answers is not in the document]
chrabyrd: yes please, back it out. I feel `content` isn't very descriptive

arches/app/src/arches/themes/default.ts
```diff
@@ -0,0 +1,81 @@
...
+            legacy: {
+                sidebar: "#2d3c4b",
+            },
+        },
+        green: palette(archesColors.green),
+        red: palette(archesColors.red),
+    },
+    semantic: {
+        primary: palette(archesColors.blue),
+        navigation: {
```
chrabyrd, line 26: IMO this needs to either be removed or have more descriptive naming. `arches-navigation-list` could refer to many things. `arches-navigation-list-padding` is better. same with color IMO but I feel less strongly there

arches/app/src/arches/themes/default.ts
```diff
@@ -0,0 +1,81 @@
...
+            color: "{arches.legacy.sidebar}",
+        },
+        iconSize: "small",
+        // custom tokens
+        content: {
+            gap: "1rem",
+        },
+        fontFamily: "system-ui, sans-serif",
+    },
+    components: {
```
chrabyrd, line 39: IMO we should use overrides with a _very_ light touch. I get the splitter handle && agree. The other stuff not so much

chrabyrd (changes requested):
Overall looks good and I think there's just a few large-stroke things to hash out

arches/app/src/arches/themes/default.ts
jacobtylerwalls: PrimeVue themes come with an iconSize whether we like it or not (1rem), so I have a feeling we will want to override it with _something_. But we can definitely punt.

arches/app/src/arches/themes/default.ts
jacobtylerwalls: This is another one that comes from primevue theming. (Sorry, tried to clarify this by putting "custom" tokens beneath a comment, but as of now we won't have any custom tokens.) This is what we get if we do nothing:

```js
navigation
  item
    {padding: '0.5rem 0.75rem', borderRadius: '{border.radius.sm}', gap: '0.5rem'}
  list 
    {padding: '0.25rem 0.25rem', gap: '2px'}
  submenuIcon
    {size: '0.875rem'}
  submenuLabel
    {padding: '0.5rem 0.75rem', fontWeight: '600'}
```

I think leaving this in is _less_ opinionated by taking some of these paddings down to 0 or at least equal X/Y

***
On second look, the navigation-color one is custom, so I'll find a way to rename/clarify.

arches/app/src/arches/themes/default.ts
jacobtylerwalls: Going with a lighter touch, so it's gone again

jacobtylerwalls:
@chrabyrd No doubt we'll want to evaluate those navigation tokens again with the benefit of some usage experience, but I think this is ready for another look!

chrabyrd (approved):
🚀 🚀 🚀

PR #11347, opened 2024-08-16
Update nonlocalized string search to correct the null search performance issue

aarongundel:
There's an issue with the null/empty search performance for non-localized string nodes in very large datasets.  Performance is dramatically slower because of the base painless script.  This bypasses that problem only for this datatype. 

Also extracting nonlocalized string (similar to geojson feature collection datatype).

### DIFF
--- a/arches/app/datatypes/core/__init__.py
+++ b/arches/app/datatypes/core/__init__.py
@@ -1,1 +1,2 @@
 from .geojson_feature_collection import *
+from .non_localized_string import *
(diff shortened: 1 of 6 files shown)


chiatt (approved):

apeters:
I was interested in the fix, but after looking at it I realized that this will fail the `Has no value` search when a single resource instance with a cardinality "n" card has one tile with a null "non-localized-string" node and another tile with a value in the "non-localized-string" node.  This was the core reason for using the "painless" script from what I can recall.  

Not sure how to resolve this @aarongundel and @chiatt

PR #11352, opened 2024-08-19
Omit `SESSION_COOKIE_SECURE` in development

jacobtylerwalls:
Refs archesproject/arches-lingo#59
Follow-up to f8076a3.

This should be set True in production, but the development server doesn't run over HTTPS, so this is too aggressive for our default dev settings.

[Reference](https://docs.djangoproject.com/en/5.0/ref/settings/#session-cookie-secure)


chrabyrd (approved):
noticed the documentation update 🙏 . lgtm!

PR #11398, opened 2024-08-25
do not allow modification of permissions if set by default perms

aarongundel:
Does not allow modification of permissions for a user/group if they are set in default permissions.  

Also fixes issue with "first" record in permissions manager - disables superuser permission modifications.  

Also fixes issue with cls being passed to get_perms_for_model


tests/utils/datatypes/datatype_tests.py
```diff
@@ -0,0 +1,248 @@
+"""
```
jacobtylerwalls, line 1: Can you remove datatype_tests.py so we don't end up with two copies?

aarongundel:
@chiatt checked with @jacobtylerwalls on this, coverage is probably going down because I cleaned up the imports in a couple places.  Otherwise, everything in here is tested and good to go.

chiatt (approved):

PR #11412, opened 2024-08-29
also catch type error for userid being None

aarongundel:
userid can be of NoneType, which will return a TypeError.  pass on this to invoke the DEFAULT_RESOURCE_IMPORT_USER setting.


chiatt (approved):
Looks good!

PR #11415, opened 2024-08-29
Handle error arising from a failed search

whatisgalen:
### Types of changes

-   [x] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

This PR primarily fixes a bug where, if an error occurred in a search request, only one arg got returned when a tuple of arguments are expected. Beyond that, it **replaces the flow of**: 
- `search_results` passes mutable args (`response_object`, `search_query_object`) to a search-view which then returns them back
- **...with**: `search_results` does not pass these args to search-view and expects brand new ones (d58ff4ecd628919695cb4178ded376108d536e01 and db4ac33dde38d04353eeaa2666172564ef775784)
...thus reducing the complexity of state management by wholly deferring to the search-view.

It also allows `response_object` to be a `str` for the case of search results which were already encoded in a json string (d46c1633c8dbed230a35ab84ecaab18606240eb1)

### Issues Solved

Closes #11414 

### Checklist

-   I targeted one of these branches:
    - [x] dev/7.6.x (under development): features, bugfixes not covered below
    - [ ] dev/7.5.x (main support): regressions, crashing bugs, security issues, major bugs in new features
    - [ ] dev/6.2.x (extended support): major security issues, data loss issues
-   [ ] I added a changelog in arches/releases
-   [ ] I submitted a PR to arches-docs (if appropriate)
-   [ ] Unit tests pass locally with my changes
-   [ ] I added tests that prove my fix is effective or that my feature works
-   [ ] My test fails on the target branch

#### Accessibility Checklist

[Developer Guide](https://arches.readthedocs.io/en/stable/developing/advanced/accessibility/)

| Topic            | Changed | Retested |
| ---------------- | ------- | -------- |
| Color contrast   |         |          |
| Form fields      |         |          |
| Headings         |         |          |
| Links            |         |          |
| Keyboard         |         |          |
| Responsive Design|         |          |
| HTML validation  |         |          |
| Screen reader    |         |          |

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

PR #11417, opened 2024-08-30
Fix noisy test

jacobtylerwalls:
Fix noisy output from recently merged test:

```py
.......Not Acceptable: /search/resources..........
```

While re-running tests, noticed a relatively slow test (unrelated) with a straightforward fix.


whatisgalen (approved):
Looks good! Also the search_tests weirdness gets fixed via #11415

jacobtylerwalls:
Thanks for the review!

PR #11438, opened 2024-09-06
Speed up test suite by 15% and fix blocking issue #11418 and issue #10719

### LINKED ISSUE #10719: Ensure all package loads in setUpClass() take place inside a transaction

### LINKED ISSUE #11418: Recently merged test hangs on `input()` call
On CI, we're [getting](https://github.com/archesproject/arches/actions/runs/10626277878/job/29457567371) an `EOFError` emanating from an `input()` call. Locally the situation is worse, it just hangs.

To reproduce:
```
python manage.py test tests.search.search_export_tests --settings="tests.test_settings"
```

Result: hangs:

Stacktrace from CTRL-C'ing:
```py
  File "/Users/jwalls/prj/arches/tests/search/search_export_tests.py", line 66, in setUpTestData
    cls.ensure_test_resource_models_are_loaded()
  File "/Users/jwalls/prj/arches/tests/base_test.py", line 110, in ensure_test_resource_models_are_loaded
    errs, importer = ResourceGraphImporter(
                     ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jwalls/prj/arches/arches/app/utils/data_management/resource_graphs/importer.py", line 152, in import_graph
    overwrite_input = input(
```

***
cc/ @whatisgalen @apeters

jacobtylerwalls:
### Types of changes

### Description of Change
A couple recent developments made it worth refactoring some things in the test suite.

1 -- #11270 changed how test packages are loaded, now every test graph is loaded for every test case whether test case needs it or not, which added overhead
2 -- #11418 describes a hanging test that was preventing running the whole suite, it's likely due to the above

All of this was faster to solve by just biting the bullet and fixing #10719. The only blocker was refactoring `refresh_geojson_geometries()` in the packages command from a nested function to a staticmethod that could be more easily patched.

### Benchmark
**PR**
```
Ran 401 tests in 213.807s
```

**dev/7.6.x** if you hack out the failing input() call
```
Ran 401 tests in 252.148s
```

### Issues Solved
Fixes #11418
Fixes #10719

### Checklist

#### Ticket Background
*   Sponsored by: Farallon

### Further comments
To load graph data now in a test, I basically cribbed the interface from Django, which lets you declare a `fixtures` class attribute on your test. But instead of using that, since we (for now) want to continue dogfooding our own graph importers, I called it `graph_fixtures` and hooked it into the same interface @apeters added in #11270.

So in the future to use sample graphs, you would:
- subclass ArchesTestCase
- add graph_fixtures attribute
- not depend on it until after `super().setUpTestData()`

### DIFF
--- a/arches/management/commands/packages.py
+++ b/arches/management/commands/packages.py
@@ -663,6 +663,11 @@ class Command(BaseCommand):
                 print("Could not save system settings")
             self.export_package_settings(dest_dir, "true")
 
+    @staticmethod
+    def update_resource_geojson_geometries():
+        with connection.cursor() as cursor:
+            cursor.execute("SELECT * FROM refresh_geojson_geometries();")
+
     def load_package(
         self,
         source,
@@ -1177,10 +1182,6 @@ class Command(BaseCommand):
         def load_etl_modules(package_dir):
             load_extensions(package_dir, "etl_modules", "etl_module")
 
-        def update_resource_geojson_geometries():
-            with connection.cursor() as cursor:
-                cursor.execute("SELECT * FROM refresh_geojson_geometries();")
-
         def load_apps(package_dir):
             package_apps = glob.glob(os.path.join(package_dir, "apps", "*"))
             for app in package_apps:
@@ -1299,7 +1300,7 @@ class Command(BaseCommand):
             for css_file in css_files:
                 shutil.copy(css_file, css_dest)
         print("Refreshing the resource view")
-        update_resource_geojson_geometries()
+        self.update_resource_geojson_geometries()
         print("loading post sql")
         load_sql(package_location, "post_sql")
         print("loading templates")
--- a/tests/base_test.py
+++ b/tests/base_test.py
@@ -115,6 +115,8 @@ class ArchesTestCase(TestCase):
     def setUpTestData(cls):
         LanguageSynchronizer.synchronize_settings_with_db(update_published_graphs=False)
         cls.loadOntology()
+        if not cls.graph_fixtures:
+            return
         for path in test_settings.RESOURCE_GRAPH_LOCATIONS:
             file_paths = [
                 file_path
--- a/tests/base_test.py
+++ b/tests/base_test.py
@@ -17,6 +17,8 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 """
 
 import os
+from unittest import mock
+from pathlib import Path
 
 from django.db import connection
 from django.core import management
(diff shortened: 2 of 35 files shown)


arches/management/commands/packages.py
```diff
@@ -663,6 +663,11 @@ def create_package(self, dest_dir):
                 print("Could not save system settings")
             self.export_package_settings(dest_dir, "true")
 
+    @staticmethod
+    def update_resource_geojson_geometries():
+        with connection.cursor() as cursor:
+            cursor.execute("SELECT * FROM refresh_geojson_geometries();")
```
jacobtylerwalls, line 669: This pg function truncates tables, which you can't do in a transaction, which all our standard Django test cases are wrapped in. De-indenting this allows me to more easily patch it out when we don't need it.

If you've ever seen a wall of text with extra red herring errors after one test failed, it's likely that this wasn't patched out.

tests/base_test.py
```diff
@@ -52,6 +54,12 @@
 
 
 class ArchesTestCase(TestCase):
+    graph_fixtures = []
+    """
+    Similar to TestCase.fixtures, but uses ResourceGraphImporter to avoid flushing.
+    Uses the name of the .json file (case-sensitive), not graph name.
```
jacobtylerwalls, line 60: We can adjust this so you have to provide the .json suffix if desired.

tests/base_test.py
```diff
@@ -135,16 +107,77 @@ def setUpClass(cls):
...
-        cursor = connection.cursor()
         sql = "DELETE FROM public.oauth2_provider_application WHERE id = 44;"
-        cursor.execute(sql)
+        with connection.cursor() as cursor:
+            cursor.execute(sql)
         super().tearDownClass()
-
-    @classmethod
-    def deleteGraph(cls, root):
-        graph = Graph.objects.get(graphid=str(root))
```
jacobtylerwalls, line 149: Now that we're being more careful about syncing ES during tests that care about ES, there's less of a need to run proxy model deletes to ensure side effects are performed. We can just rely on the transaction rolling back to leave PG in the correct state.

tests/test_settings.py
```diff
@@ -20,22 +20,31 @@
 
 from arches.settings import *
 
-try:
-    from django.utils.translation import gettext_lazy as _
-except ImportError:  # unable to import prior to installing requirements
-    pass
+from django.utils.translation import gettext_lazy as _
```
jacobtylerwalls, line 23: From the git blame, the try/except was here for when we had setup.py, but we've migrated to pyproject.toml

chiatt (approved):
Looks good!

PR #11439, opened 2024-09-06
Speed up indexing with cursor_tuple_fraction pg option #11382

### LINKED ISSUE #11382: Tune performance of iterator() calls with cursor_tuple_fraction
We use `QuerySet.iterator()` in a couple places for indexing and deletion, and we tend to fully exhaust those iterators, so we would probably benefit from setting the [cursor_tuple_fraction option](https://docs.djangoproject.com/en/5.1/ref/databases/#server-side-cursors):

> By default, PostgreSQL assumes that only the first 10% of the results of cursor queries will be fetched. The query planner spends less time planning the query and starts returning results faster, but this could diminish performance if more than 10% of the results are retrieved.

We may wish to set this to 0.9.

jacobtylerwalls:
### Types of changes

### Description of Change
See #11382

### Issues Solved

Closes #11382

### Checklist

#### Ticket Background
*   Sponsored by: Farallon

### DIFF
--- a/arches/install/arches-templates/project_name/settings.py-tpl
+++ b/arches/install/arches-templates/project_name/settings.py-tpl
@@ -106,7 +106,9 @@ DATABASES = {
         "ENGINE": "django.contrib.gis.db.backends.postgis",
         "HOST": "localhost",
         "NAME": "{{ project_name }}",
-        "OPTIONS": {},
+        "OPTIONS": {
+            "options": "-c cursor_tuple_fraction=1",
+        },
         "PASSWORD": "postgis",
         "PORT": "5432",
         "POSTGIS_TEMPLATE": "template_postgis",
--- a/arches/install/arches-templates/tests/test_settings.py-tpl
+++ b/arches/install/arches-templates/tests/test_settings.py-tpl
@@ -39,7 +39,9 @@ DATABASES = {
         "ENGINE": "django.contrib.gis.db.backends.postgis",
         "HOST": "localhost",
         "NAME": "{{ project_name }}",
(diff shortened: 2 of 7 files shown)


aarongundel:
I don't understand what this is doing - I don't see this option listed https://docs.djangoproject.com/en/5.1/ref/databases/#postgresql-connection-settings in here.  Can you expand on this or point me to some documentation?

jacobtylerwalls:
Sorry, the link was buried in the issue:

> We use QuerySet.iterator() in a couple places for indexing and deletion, and we tend to fully exhaust those iterators, so we would probably benefit from setting the [cursor_tuple_fraction option](https://docs.djangoproject.com/en/5.1/ref/databases/#server-side-cursors):

> By default, PostgreSQL assumes that only the first 10% of the results of cursor queries will be fetched. The query planner spends less time planning the query and starts returning results faster, but this could diminish performance if more than 10% of the results are retrieved.

[Django link](https://docs.djangoproject.com/en/5.1/ref/databases/#server-side-cursors)
[PG link](https://www.postgresql.org/docs/16/runtime-config-query.html#GUC-CURSOR-TUPLE-FRACTION)


This a little bit _getting out in front of the issue_ since I don't have an example in hand where we're being bitten by this (e.g. where we're suffering with a slower query plan), but it seems good to go ahead and bias this toward fetching the whole result, as every iterator() we've used is always fully exhausted.

PS -- I'm already using this in the lingo settings, where the concept tree is built up using an iterator().

aarongundel (approved):
LGTM

PR #11446, opened 2024-09-10
Fix term-filter.clear method #11443

### LINKED ISSUE #11443: term-filter.clear() clears selectWoo dom elements of tags but not terms unless new ajax query
When a search-filter gets cleared, the UI should also reflect that. In a search-view that doesn't make a new ajax request when the query changes back to the default (no search parameters), the selectWoo elements part of the term-filter persist even though the underlying `tags` and `terms` have already been removed thus (from `views/components/search/term-filter.js`):
```.js
clear: function() {
            this.filter.terms.removeAll();
            this.filter.tags.removeAll();
        }
```

See video below for example:

https://github.com/user-attachments/assets/5dc07ace-0385-4689-846d-9e415430aa53

whatisgalen:
### Types of changes

### Description of Change

When the `term-filter` is used to include search terms, the ko.observableArray `terms` gets a new `term` object but the ko.observableArray `tags` does not. The logic in the term-search binding only updates the selectWoo UI when `tags` gets updated, not `terms` (unless a user individually clicks "x" on the term, or the ensuing query updates the term-filter state). This PR adds a subscription to `terms` to update (empty) the selectWoo if there are neither tags nor terms.

### Issues Solved

Closes #11443 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/bindings/term-search.js
+++ b/arches/app/media/js/bindings/term-search.js
@@ -27,6 +27,9 @@ define([
                     searchbox.tags.push(option);
                 }); 
                 searchbox.trigger('change');
+                if (ko.unwrap(terms).length == 0 && searchbox.tags.length == 0){
+                    searchbox.empty();
+                }
             });
             terms.subscribe(function(terms) {
                 if (terms.length == 0 && searchbox.tags.length == 0)
--- a/arches/app/media/js/bindings/term-search.js
+++ b/arches/app/media/js/bindings/term-search.js
@@ -28,6 +28,10 @@ define([
                 }); 
                 searchbox.trigger('change');
             });
+            terms.subscribe(function(terms) {
+                if (terms.length == 0 && searchbox.tags.length == 0)
+                    searchbox.empty();
+            });
             var self = this;
             this.stripMarkup = function(m){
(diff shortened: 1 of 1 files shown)


arches/app/media/js/bindings/term-search.js
```diff
@@ -28,6 +28,10 @@ define([
                 }); 
                 searchbox.trigger('change');
             });
```
apeters, line 30: If you add both a resource-type filter along with a term-filter and hit the clear button you'll notice that you'll need this logic here too:

```suggestion
                if (terms.length == 0 && searchbox.tags.length == 0){
                    searchbox.empty();
                }
```

apeters (changes requested):

arches/app/media/js/bindings/term-search.js
whatisgalen: I'm assuming this committed change was limited to line 29 as otherwise the deleted `});` breaks things

apeters (approved):
looks good

PR #11451, opened 2024-09-11
Remove types.ts

jacobtylerwalls:
Helpful typescript interfaces for vue development will be fleshed out further in archesproject/arches-vue-utils rather than core.


robgaston (approved):
lgtm

PR #11465, opened 2024-09-17
performance improvements for indexing with default deny

aarongundel:
There are some issues with indexing performance using default deny for large datasets.  These seem to be tied to the number of resources, but in all cases there is a 10x difference between default allow and default deny performance.  This change attempts to improve that discrepancy, which is largely a result of querying for object permissions for users and groups - a set of four groups of queries that must occur for each object.  Default Allow has a much faster implementation here.  This PR (instead of starting with the user and getting their permissions on a given object) gets all of the permissions that exist for a particular object and uses that (complete) set of permissions.


aarongundel:
pending testing on a full dataset with 700k+ records.

Issue #10873, opened 2024-05-03
Card/widget dirty state is not resolvable after discarding card/widget changes in Graph Designer


chrabyrd:
This results in un-dismissable buttons that only appear when the Graph Designer component has a dirty state. This appears to be related to i18n changes.

Steps to reproduce:

1. Create a new Graph
2. Add a node to the graph ( eg. `string` )
3. Navigate to the cards tab
4. Make a change to the Widget label or Card Title
5. Attempt to discard the changes via the `Discard Card Edits` button
6. Notice the changes are discarded, but the dirty state is not resolved

Diff of card and `._card`:

card:
```
{
    "cards": [],
    "widgets": [
        {
            "card_id": "04626822-08ca-11ef-8d26-acde48001122",
            "config": {
                "en": "",
                "width": "100%",
                "maxLength": null,
                "uneditable": false,
                "placeholder": "Enter text",
                "defaultValue": "",
                "i18n_properties": [
                    "placeholder"
                ],
                "label": ""
            },
            "id": "046269bc-08ca-11ef-8d26-acde48001122",
            "label": "",
            "node_id": "04626660-08ca-11ef-8d26-acde48001122",
            "sortorder": null,
            "source_identifier_id": "8924711c-3dae-44d2-a5cd-2513d9f78159",
            "visible": true,
            "widget_id": "10000000-0000-0000-0000-000000000001",
            "disabled": false
        }
    ],
    "tiles": [],
    "cardid": "04626822-08ca-11ef-8d26-acde48001122",
    "nodegroup_id": "04626660-08ca-11ef-8d26-acde48001122",
    "cardinality": "1",
    "name": "string",
    "instructions": "",
    "cssclass": null,
    "helptext": "",
    "helpenabled": false,
    "helptitle": "",
    "visible": true,
    "active": true,
    "sortorder": null,
    "component_id": "f05e4d3a-53c1-11e8-b0ea-784f435179ea",
    "config": null,
    "constraints": [
        {
            "uniquetoallinstances": false,
            "nodes": [],
            "cardid": "04626822-08ca-11ef-8d26-acde48001122",
            "constraintid": "92e1e7c5-90f2-4cab-ae0a-f01e5181199a"
        }
    ],
    "id": "04626822-08ca-11ef-8d26-acde48001122",
    "description": null,
    "graph_id": "0462653e-08ca-11ef-8d26-acde48001122",
[...]

PR #11480, opened 2024-09-20
Remove placeholder from i18n props for non-localized-string widget re #10873

jacobtylerwalls:
### Types of changes

### Description of Change
Found while dev'ing #10873: I would not expect the placeholder on a non-localized string widget to be localized.

This is not the fix for #10873, I just noticed this before reducing the scope of my fix to the string widget only.

### Issues Solved
Refs #10873

#### Ticket Background
*   Sponsored by: Farallon


jacobtylerwalls:
Looks like this was explicitly added in #10820, closing

Issue #8022, opened 2021-12-14
Enhance Arches security with two-factor authentication


robgaston:
To improve the user security of Arches implementations, two-factor authentication should be optionally implemented as part of user authentication.  To support this we will need to identify the best way to integrate two-factor authentication into Django, and then provide a way for users to optionally implement this at the project-level.

> Version 2 of the RDM will be able to support subscription services for thesauri objects such as concepts and collections.  To further improve user permissions and enhance Arches’ security... will implement two-factor authentication

chrabyrd:
resolved in #8079

PR #11491, opened 2024-09-24
Remove (hypothetical) side effects from GET calls #10664

### LINKED ISSUE #10664: `GeoJSON.get()` and `MVT.get()` views have side effects
Was reading up on CSRF today and recalled that HTTP GET requests shouldn't have side-effects (and for that reason aren't protected against CSRF).

We have a couple GET views that save `UserProfile` rows:
https://github.com/archesproject/arches/blob/3e5493154ffde7de680d3e0a0518b14b8c1aa554/arches/app/views/api.py#L179-L180
https://github.com/archesproject/arches/blob/3e5493154ffde7de680d3e0a0518b14b8c1aa554/arches/app/views/api.py#L275-L276

I can't see any attack vector here, so this is just a cleanup opportunity. We seem to be doing this merely for the sake of calling the `UserProfile.viewable_nodegroups` property. Creating a db row just for the sake of calling a function is something we can avoid: one, to be more RESTful, and two, to shave off a db query.

jacobtylerwalls:
### Types of changes

### Description of Change
UserProfiles always exist since #8022.
1ca5c32 removed some cruft guards for missing profiles. This removes the rest, which makes it more clear that these GETs do not have side-effects (and thus CSRF attack vectors).

### Issues Solved

Closes #10664

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -1263,8 +1263,6 @@ class FileListDataType(BaseDataType):
             # this does not get called when saving data from the mobile app
             previously_saved_tile = models.TileModel.objects.filter(pk=tile.tileid)
             user = request.user
-            if hasattr(request.user, "userprofile") is not True:
-                models.UserProfile.objects.create(user=request.user)
             user_is_reviewer = user_is_resource_reviewer(request.user)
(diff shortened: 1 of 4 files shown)


johnatawnclementawn (approved):
LGTM 👍🏼  - nice catch

Issue #9858, opened 2023-08-01
Saving new node of type string raises 500


jacobtylerwalls:
EDIT: see[ simpler reproducer](https://github.com/archesproject/arches/issues/9858#issuecomment-1673729350)

#### Reproducer ####
1. Start a new graph, choose an ontology
2. Create a node under the root
3. Click save edits: Graph Validation Error appears as expected (child node doesn't define ontology)
4. Discard edits.
5. Edit the data type of the child node to something other than semantic.
6. Save edits.
7. Create new node under root.
8. Publish graph.
9. Navigate to Add Resource Instances
10. Click browser back button.

#### Result ####
```py
Traceback (most recent call last):
  File "/Users/jwalls/prj/venv/lib/python3.10/site-packages/django/core/handlers/exception.py", line 56, in inner
    response = get_response(request)
  File "/Users/jwalls/prj/venv/lib/python3.10/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/jwalls/prj/venv/lib/python3.10/site-packages/django/views/generic/base.py", line 84, in view
    return self.dispatch(request, *args, **kwargs)
  File "/Users/jwalls/prj/venv/lib/python3.10/site-packages/django/utils/decorators.py", line 46, in _wrapper
    return bound_method(*args, **kwargs)
  File "/Users/jwalls/prj/venv/lib/python3.10/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/Users/jwalls/prj/venv/lib/python3.10/site-packages/django/views/generic/base.py", line 119, in dispatch
    return handler(request, *args, **kwargs)
  File "/Users/jwalls/prj/arches/arches/app/views/graph.py", line 518, in post
    card = Card(data)
  File "/Users/jwalls/prj/arches/arches/app/models/card.py", line 165, in __init__
    widget_model.save()
  File "/Users/jwalls/prj/venv/lib/python3.10/site-packages/django/db/models/base.py", line 806, in save
    self.save_base(
  File "/Users/jwalls/prj/venv/lib/python3.10/site-packages/django/db/models/base.py", line 857, in save_base
    updated = self._save_table(
  File "/Users/jwalls/prj/venv/lib/python3.10/site-packages/django/db/models/base.py", line 1000, in _save_table
    results = self._do_insert(
[...]

PR #11498, opened 2024-09-26
Fix error preventing widget changes

### LINKED ISSUE #10899: Cannot change widget template (changes don't persist)
Tested in stable/7.5.2 and dev/7.5.x

Changing the template of a card does not seem to work after saving and refreshing the graph.

Tested here with the rich text template:

https://github.com/archesproject/arches/assets/119508494/0b34d877-b21e-450c-a5cd-75eb371c3496

Steps to reproduce:
1. Create a node of either string or concept datatype (easiest to test)
2. On the card, change the template to either rich-text or radio buttons
3. Save the card edits
4. Refresh the graph page
5. Navigate back to the card and see the template hasn't changed (i.e it will still display the text template not rich-text)

Then trying to change it to rich-text again produces an integrity error like the following:
```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "cards_x_nodes_x_widgets_nodeid_cardid_widgetid_976dc542_uniq"
DETAIL:  Key (nodeid, cardid, widgetid)=(eaf17eaa-0d56-11ef-b8bc-00155d34a3f9, 3880e6b4-1b5a-450e-97e6-1c0eea178210, 10000000-0000-0000-0000-000000000005) already exists.
```

chiatt:
### Types of changes

### Description of Change

Replaced `get_or_create` because it was taking widget_id as an argument. This worked well for newly created records, but if a record already existed and the widget was being changed to a different widget, this would create a second cardsxnodesxwidget record for the node.

This change reintroduces #9858, however, that error was due in part to making two immediate calls to the CardView endpoint to update the card and permissions trees [here](https://github.com/archesproject/arches/blob/96051bb2d3fc1edd45b24071c53e2004a6313070/arches/app/media/js/views/graph-designer.js#L310-L311). Throttling the second call by 50ms prevents the issue, however, a better long term solution would be to avoid the second call and find a better way of ensuring the trees stay in sync.

### Issues Solved

Closes #10899

### Checklist

### DIFF
--- a/arches/app/models/card.py
+++ b/arches/app/models/card.py
@@ -154,28 +154,18 @@ class Card(models.CardModel):
 
                 if "widgets" in args[0]:
                     for widget in args[0]["widgets"]:
-                        cardxnodexwidgetid = widget.get("id", None)
                         node_id = widget.get("node_id", None)
                         card_id = widget.get("card_id", None)
                         widget_id = widget.get("widget_id", None)
-                        if cardxnodexwidgetid is None and (
-                            node_id is not None
-                            and card_id is not None
-                            and widget_id is not None
-                        ):
-                            widget_model, _ = (
-                                models.CardXNodeXWidget.objects.get_or_create(
-                                    node_id=node_id,
-                                    card_id=card_id,
-                                    widget_id=widget_id,
-                                )
-                            )
-                        else:
+ 
+                        try:
+                            widget_model = models.CardXNodeXWidget.objects.get(node_id=node_id, card_id=card_id)
+                        except models.CardXNodeXWidget.DoesNotExist:
                             widget_model = models.CardXNodeXWidget()
-                            widget_model.pk = cardxnodexwidgetid
                             widget_model.node_id = node_id
                             widget_model.card_id = card_id
-                            widget_model.widget_id = widget_id
+
+                        widget_model.widget_id = widget_id
                         widget_model.config = widget.get("config", {})
                         widget_model.label = widget.get("label", "")
                         widget_model.visible = widget.get("visible", None)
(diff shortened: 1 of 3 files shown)


arches/app/models/card.py
```diff
@@ -154,35 +154,24 @@ def __init__(self, *args, **kwargs):
 
                 if "widgets" in args[0]:
                     for widget in args[0]["widgets"]:
-                        cardxnodexwidgetid = widget.get("id", None)
                         node_id = widget.get("node_id", None)
                         card_id = widget.get("card_id", None)
                         widget_id = widget.get("widget_id", None)
-                        if cardxnodexwidgetid is None and (
```
jacobtylerwalls, line 161: Ah, thanks for finding the root cause. This fixes it for me without a throttle by deferring all the work to the db instead of using a retrieve-update-save pattern, which is where the potential for race conditions creeps in:

```diff
diff --git a/arches/app/models/card.py b/arches/app/models/card.py
index 57eb529dbe..44746350ca 100644
--- a/arches/app/models/card.py
+++ b/arches/app/models/card.py
@@ -163,10 +163,10 @@ class Card(models.CardModel):
                             and widget_id is not None
                         ):
                             widget_model, _ = (
-                                models.CardXNodeXWidget.objects.get_or_create(
+                                models.CardXNodeXWidget.objects.update_or_create(
                                     node_id=node_id,
                                     card_id=card_id,
-                                    widget_id=widget_id,
+                                    defaults={"widget_id": uuid.UUID(widget_id)},
                                 )
                             )
```

chiatt: Ah - you're right. That fixed the issue without the the throttle. Thanks!

jacobtylerwalls (approved):
+1 tested

PR #11506, opened 2024-09-27
Indicate if user has no etl module permissions

### LINKED ISSUE #11450: Notify users if they have no ETL module access
Due to changes in permissions, resource reviewers no longer have access to bulk data manager modules. This is actually the expected behavior. However, this requires an admin to give groups/users object permissions to each module. To address this change in behavior we should:
- Inform users that they do not have access to etl modules if no etl modules are permitted
- Provide a means of granting and revoking permissions in bulk for a given group

chiatt:
### Types of changes

### Description of Change
Indicates to user that they have no permissions to any bulk modules if they do not find any etl modules on the Bulk Data Manager page. This is because permissions were not enforced in 7.5, however, that issue is fixed in 7.6, so users accustomed to seeing modules may now find that that they can't see any in 7.6. Additionally, this PR adds a management command to allow an admin to grant or revoke permissions to etl modules in bulk.

### Issues Solved

Closes #11450

### Checklist

### DIFF
--- a/arches/app/templates/views/components/plugins/etl-manager.htm
+++ b/arches/app/templates/views/components/plugins/etl-manager.htm
@@ -27,7 +27,7 @@
         <div class="workflow-select-plugin">
             <div class="workflow-select-card-container">
                 <!--ko if: etlModules.length === 0 -->
-                <div style="display: flex; justify-content: center; padding: 25px; width: 100%"><div data-bind="text: $parent.translations.bulkDataPermissionNotification" style="font-size: 1.9rem">TEST</div></div>
+                <div style="display: flex; justify-content: center; padding: 25px; width: 100%"><div data-bind="text: $parent.translations.bulkDataPermissionNotification" style="font-size: 1.9rem"></div></div>
                 <!--/ko-->
                 <!-- ko foreach: {data: etlModules, as: 'etlmodule', noChildContext: true} -->
                     <!-- ko if: activeModules() == 'import' && (!moduleSearchString() || !!moduleSearchString() && etlmodule.name.toLowerCase().includes(moduleSearchString().toLowerCase()) ) -->
--- a/arches/app/templates/views/components/plugins/etl-manager.htm
+++ b/arches/app/templates/views/components/plugins/etl-manager.htm
@@ -26,6 +26,9 @@
         <!-- ko ifnot: loading() -->
         <div class="workflow-select-plugin">
             <div class="workflow-select-card-container">
+                <!--ko if: etlModules.length === 0 -->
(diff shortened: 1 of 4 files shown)


arches/management/commands/permissions.py
```diff
@@ -0,0 +1,97 @@
...
+
+        if action == "grant":
+            permission_action = getattr(permission_backend, "assign_perm")
+        elif action == "revoke":
+            permission_action = getattr(permission_backend, "remove_perm")
+
+        try:
+            permission_group = Group.objects.get(name=group)
+        except Group.DoesNotExist:
+            print(f"{group} does not exist")
```
jacobtylerwalls, line 92: Sorry for the drive by, but do you mind using self.stdout.write instead of print in commands, for ease of testing and redirecting?

jacobtylerwalls: Or self.stderr perhaps better here

tests/commands/test_permissions.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+from django.core.management import call_command
+from tests.base_test import ArchesTestCase
+from django.contrib.auth.models import Group, User
+from arches.app.utils.permission_backend import PermissionBackend
+from arches.app.models import models
+
+
+class PermissionsCommandTet(ArchesTestCase):
```
aarongundel, line 12: typo with test here

arches/management/commands/permissions.py
```diff
@@ -0,0 +1,97 @@
...
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+"""
+
+from django.core.management.base import BaseCommand
+from django.contrib.auth.models import Group
+from arches.app.models import models
+import arches.app.utils.permission_backend as permission_backend
+
+
+class Command(BaseCommand):
```
aarongundel, line 25: Might be useful to provide a full example command for documentation?

arches/app/templates/views/components/plugins/etl-manager.htm
```diff
@@ -26,6 +26,9 @@
         <!-- ko ifnot: loading() -->
         <div class="workflow-select-plugin">
             <div class="workflow-select-card-container">
+                <!--ko if: etlModules.length === 0 -->
+                <div style="display: flex; justify-content: center; padding: 25px; width: 100%"><div data-bind="text: $parent.translations.bulkDataPermissionNotification" style="font-size: 1.9rem">TEST</div></div>
```
aarongundel, line 30: Should never be visible, but the word TEST should probably be removed.

aarongundel (changes requested):
looks good, just a couple minor items here

arches/management/commands/permissions.py
chiatt: Good idea - done

tests/commands/test_permissions.py
chiatt: Nice catch

aarongundel (approved):
LGTM

PR #11508, opened 2024-09-27
adds resource geoutils #11507

### LINKED ISSUE #11507: Add geoutils for buffering and finding resources within a feature collection

chrabyrd:
### Types of changes

### Description of Change

I DO NOT EXPECT THIS TO BE MERGED INTO THE 7.6 release. This is a sister ticket to https://github.com/fargeo/ca-shpo-online/pull/159

### Issues Solved

Closes #11507


arches/app/src/arches/declarations.d.ts
```diff
@@ -4,6 +4,7 @@ declare module "arches";
 // declare untyped modules that have been added to your project in `package.json`
 declare module "@babel/runtime";
 declare module "@mapbox/geojsonhint";
+declare module "@mapbox/mapbox-gl-draw";
```
jacobtylerwalls, line 7: There are definitions for this on definitely typed, so let's install them as a dev dependency instead of type-ignoring the module.

arches/app/utils/geo_utils.py
```diff
@@ -94,3 +99,142 @@ def convert_geos_geom_collection_to_feature_collection(self, geometry):
...
+
+        return {
+            "type": "FeatureCollection",
+            "features": buffered_features,
+        }
+
+    def get_intersection_and_difference_of_feature_collections(
+        self, feature_collection_1, feature_collection_2
+    ):
+        def _build_feature_collection(geometries):
```
robgaston, line 174: a couple thoughts here:
- `geometries` appears to be a single GEOS geometry, so I think we should rename it to `geometry`
- is there a reason for decomposing multi-part geometries before creating features?  those are valid geojson geometry types, so there doesn't seem to be a strict requirement for that in my mind.

arches/app/utils/geo_utils.py
```diff
@@ -94,3 +99,142 @@ def convert_geos_geom_collection_to_feature_collection(self, geometry):
...
+
+        feature_collection_2_geometries = [
+            GEOSGeometry(json.dumps(feature["geometry"]))
+            for feature in feature_collection_2["features"]
+        ]
+        feature_collection_2_geometries_union = GeometryCollection(
+            *feature_collection_2_geometries
+        ).unary_union
+
+        return {
```
robgaston, line 224: i'm thinking it might be better to just return a single feature collection here and use properties on the features rather than keys at the top level to distinguish "types" of features.  this would then return actual geojson rather than geojson nested inside an object which seems more usable

robgaston (changes requested):
these functions look generally good (a few comments on `get_intersection_and_difference_of_feature_collections`) and I think they should go into core as built here.  I'd say go ahead and write those tests so we can merge this in. thanks!

chrabyrd:
closing in favor of moving geoutils to SHPO. These can/will be re-introduced to core on an as-need basis

PR #11513, opened 2024-10-01
Reduce queries for published graphs when indexing

jacobtylerwalls:
### Types of changes

### Description of Change
Before, when indexing resources there were 2 queries per graph to get the published graph in the active language. One for the GraphXPublishedGraph row, one for the PublishedGraph row. This was then cached in memory, persisting either for a short period (on the CLI) or potentially forever for server-side operations, which could lead to retrieving stale data.

Now, these are prefetched during `optimize_resource_iteration`, so that it takes 2 queries total per chunk, no matter how many graphs are involved. A typical chunk size is 500 (2000 // 8), so if two adjacent chunks each contain 500 resources of the same graph, then this might _slightly_ increase queries, because we are no longer perpetually caching, but with a benefit of less memory overhead and better correctness for server-side operations--_but in all other cases this should reduce queries by avoiding repetitive published graph queries._

### Checklist
    - [x] dev/8

#### Ticket Background
*   Sponsored by: Farallon
*   Found by: @chiatt

### DIFF
--- a/arches/app/utils/index_database.py
+++ b/arches/app/utils/index_database.py
@@ -12,6 +12,8 @@ django.setup()
 from datetime import datetime
 from django.db import connection, connections
 from django.db.models import prefetch_related_objects, Prefetch, Q, QuerySet
+from django.utils.translation import get_language
+
 from arches.app.models import models
 from arches.app.models.resource import Resource
 from arches.app.models.system_settings import settings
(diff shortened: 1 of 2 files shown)


arches/app/utils/index_database.py
```diff
@@ -263,10 +263,19 @@ def index_resources_using_singleprocessing(
...
+                try:
+                    resource.serialized_graph = (
+                        resource.graph.publication.published_graph_active_lang[
+                            0
+                        ].serialized_graph
+                    )
+                except IndexError:
+                    resource.serialized_graph = None
+
                 resource.set_node_datatypes(node_datatypes)
```
jacobtylerwalls, line 278: This might be a pretty good candidate to attack next?

jacobtylerwalls: never mind, it's gathered at the top of this function.

jacobtylerwalls:
After pairing with @apeters on this, I think we're going to go in the direction of:
- pulling this query for published graphs out of the resource batch iteration and do it earlier, passing the result as an arg to `index_resources_using_singleprocessing`

jacobtylerwalls:
Coverage check failure due to falling denominator.

jacobtylerwalls:
@whatisgalen perhaps you'd like to review?

chiatt (approved):
Looks good to me!

PR #11530, opened 2024-10-07
Bump primevue to 4.1.0

jacobtylerwalls:
Allows using primefaces/primevue#6367 (so we can back out the current set of workarounds in arches-references/arches-lingo).

Nothing to test at the moment, just unblocking that work to remove workarounds.


johnatawnclementawn (approved):
LGTM 👍🏼

PR #11533, opened 2024-10-08
Return 404 for nonexistent search components #11532

### LINKED ISSUE #11532: search_home route raises AttributeError for missing search components
- Visit /search?search-view=nonexistent

We could respond with 404 instead of letting AttributeError escape.

Will send a fix that I cooked up while following along with the walk through from the dev meeting, which I very much enjoyed :-)

```py
  File "/Users/jwalls/prj/arches/arches/app/views/search.py", line 84, in get
    search_components = searchview_instance.get_searchview_filters()
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Exception Type: AttributeError at /search
Exception Value: 'NoneType' object has no attribute 'get_searchview_filters'
```

jacobtylerwalls:
### Types of changes

### Description of Change
Respond with a 404 view instead of an unhandled 500 exception for requests to the search_home route with a nonexistent search-view param.

### Issues Solved

Closes #11532

### Checklist

#### Ticket Background
*   Sponsored by: Farallon

### Further comments
We can target 8 instead if this doesn't qualify as a "major" bug in a new feature, but would be nice since the LTS was recent to avoid issuing 500s in people's server logs if we can.

### DIFF
--- a/arches/app/views/search.py
+++ b/arches/app/views/search.py
@@ -18,13 +18,13 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 import logging
 import os
-from django.contrib.auth import authenticate
+
 from django.contrib.gis.geos import GEOSGeometry
 from django.core.cache import cache
 from django.db import connection
+from django.http import Http404
 from django.shortcuts import render
 from django.utils.translation import gettext as _
-from django.utils.decorators import method_decorator
 from arches.app.models.models import (
     MapMarker,
     GraphModel,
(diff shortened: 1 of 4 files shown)


whatisgalen (changes requested):
@jacobtylerwalls this seems like a good change on its face especially from the point of view of the backend. However with being more concise about the type of error code (404 vs 500) the http404 seems to bypass the error handling in `search_results` and the end user is no longer in Arches search but is now sort of stranded on whatever the 404 page is. I don't actually think that should be the UX because there's nothing to tell you what went wrong in a prod deployment. I also think that maybe the 404 should be reserved for something more like a non-existent url path as opposed to a query parameter that leads to a non-existent thing in the database. @apeters I think you likely have thoughts on this as well.

apeters:
@whatisgalen to me this is really like submitting an incorrect route to Arches.  You're asking for a base page layout that doesn't exist.  I think throwing an `http404` makes sense in that case.

apeters (approved):
I'll approve this once the merge conflicts are fixed

PR #11542, opened 2024-10-14
Avoid refetching resource instances inside permissions framework

jacobtylerwalls:
### Types of changes

### Description of Change
The methods in the permissions framework mostly expect to be provided with a resource id. Each method then fetches an instance from the database. If the caller already has a model instance, they can't provide it to avoid the refetch. This PR now allows the caller to do that.

### Checklist

#### Ticket Background
*   Sponsored by: Farallon

### DIFF
--- a/tests/views/api/test_permissions.py
+++ b/tests/views/api/test_permissions.py
@@ -1,0 +1,35 @@
+from django.db import connection
+from django.test import TestCase
+from django.test.utils import CaptureQueriesContext
+from django.urls import reverse
+
+from arches.app.models.graph import Graph
+from arches.app.models.models import ResourceInstance
+
+# these tests can be run from the command line via
+# python manage.py test tests.views.api.test_permissions --settings="tests.test_settings"
+
+
+class InstancePermissionsAPITest(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.graph = Graph.new(
+            name="INSTANCE_PERMISSIONS_TEST_GRAPH",
+            is_resource=True,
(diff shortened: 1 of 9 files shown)


tests/views/api/test_permissions.py
```diff
@@ -0,0 +1,35 @@
...
+        resource = ResourceInstance.objects.create(graph=self.graph)
+        with CaptureQueriesContext(connection) as queries:
+            response = self.client.get(
+                reverse("api_instance_permissions"),
+                QUERY_STRING=f"resourceinstanceid={resource.pk}",
+            )
+        resource_selects = [
+            q for q in queries if q["sql"].startswith('SELECT "resource_instances"')
+        ]
+        self.assertEqual(len(resource_selects), 1, list(queries))
```
jacobtylerwalls, line 32: On the target branch, this fails like so:

```
FAIL: test_get (tests.views.api.test_permissions.InstancePermissionsAPITest.test_get)
  File "/Users/jwalls/prj/arches/tests/views/api/test_permissions.py", line 33, in test_get
    self.assertEqual(len(resource_selects), 1, queries)
AssertionError: 3 != 1 : <django.test.utils.CaptureQueriesContext object at 0x1183095e0>

Ran 1 test in 0.070s
```

Not the most high traffic endpoint, but an example of what this PR allows you to save in other cases (e.g. resource API)

aarongundel (approved):
This looks great to me.  I'm a bit sad we can't target 7.6, but this'll help make a good case for an upgrade to 8.

PR #11547, opened 2024-10-15
Check for searchview default types (#11546)

### LINKED ISSUE #11546: distinguish `"default":true` search-view from `"searchViewDefault":true` so one search-view doesn't need to do everything
When you set a `search-view` to `default: true` so that it gets used by default when a user goes to the Search page, that search-view is now used by default every single time the `search_results` url is hit. This means that if you have a search-view that doesn't use, say, the vanilla `"term-filter"`, any query parameters nested under `"term-filter"` will be discarded and those searches will not work as expected (for example the resource-instance select widget in the resource-instance datatype). 

One way to fix this would be to incorporate new keys on the search-view config json to distinguish a search-view as a "globalDefault" vs one that is just the "searchViewDefault". That way a developer doesn't need to worry about side effects from their custom searchview when normal Arches search requests get sent. 

A workaround is to include all default Arches search filters in the `linkedSearchFilters` array in the search-view's database config. However even so, the main search-view logic would need to anticipate however core arches (`standard-search-view`) interacts with those search filters.

whatisgalen:
### Types of changes

### Issues Solved

Closes #11546

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/search/components/base.py
+++ b/arches/app/search/components/base.py
@@ -112,10 +112,8 @@ class SearchFilterFactory(object):
         return searchview_component_name
 
     def get_searchview_instance(self, **kwargs):
-        default_property = kwargs.get("default_property", None)
-        searchview_component_name = self.get_searchview_name(
-            default_property=default_property
-        )
+        default_type = kwargs.get("default_type", None)
+        searchview_component_name = self.get_searchview_name(default_type=default_type)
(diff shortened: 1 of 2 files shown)


chiatt:
Closing as stale

PR #11549, opened 2024-10-16
Fix for issue preventing validation of url with only whitespace

### LINKED ISSUE #10796: URL datatype - cleaning empty string to None can prevent next data entry
URL datatype
v7.5.2

- Enter a value for URL, save
- Delete and type a space, then backspace, save
- Enter another value

Result: cannot save

https://github.com/archesproject/arches/assets/38668450/66a17458-6566-4ad6-ade5-16dedd0b578b

_Originally posted by @jacobtylerwalls in https://github.com/archesproject/arches/issues/10642#issuecomment-2064521593_

CWDamm-Kint:
### Types of changes

### Description of Change

Previously, the clean function was assigning `tile.data[nodeid]` as None if the url provided was entirely whitespace, which prevented it from being validated. This then prevented further data entry after saving. 

The clean function now separates out some of the logic, so that the url_label is replaced with None if blank or whitespace, but the full `tile.data[nodeid]` is only assigned as None if both values are falsey. This means that blank but non-empty url strings still go through to validation. 

### Issues Solved

Closes #10796

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Tested by: @CWDamm-Kint
*   Designed by: @CWDamm-Kint

### DIFF
--- a/arches/app/datatypes/url.py
+++ b/arches/app/datatypes/url.py
@@ -307,8 +307,12 @@ class URLDataType(BaseDataType):
     def clean(self, tile, nodeid):
         if data := tile.data[nodeid]:
             try:
-                if not any([val.strip() for val in data.values()]):
-                    tile.data[nodeid] = None
+                if not any([data.values()]): 
+                    tile.data[nodeid] = None 
+
+                if not data["url_label"].strip(): 
+                    tile.data[nodeid]["url_label"] = None 
+
             except:
                 pass  # Let self.validate handle malformed data
 


samuelscandrett:
This has been targeted at 7.6.x (rather than 8.0.x) because it was recently raised from an AfHER context, causing a semi-data loss state, as the entire tile needs to be deleted in order to update a URL.  If the nodegroup held a small set of nodes then it wouldn't be as major, but due to the number of nodes in some of the AfHER nodegroups (e.g. External Cross References, or some Monument/Area Phases) it means that a lot of data must be deleted and re-inputted.

Since AfHER will be on 7.6.x soon (and for a long period of time), it'd be great to get semi-data loss scenarios like this into 7.6.
Interested in others thoughts!

samuelscandrett:
I've noticed that this fix doesn't set the node value to null if there is no data, instead it looks as such:

```
{
  "5ed66e9e-b71a-11ef-a84a-00155dbda2ca": {
    "url": "",
    "url_label": null
  }
}
```

We need to the data to appear as null in the backend, but still editable in the front end.  

This issue is related to this recent string issue https://github.com/archesproject/arches/issues/11658 as they both concern the back-end having a null value but the front-end either not reflecting this.

chiatt:
I just tried to reproduce the bug that this ticket addresses #10796, but I couldn't reproduce the issue given the instructions in the ticket. It seems like it addresses a different problem. Maybe a different issue is needed?

samuelscandrett:
> I just tried to reproduce the bug that this ticket addresses #10796, but I couldn't reproduce the issue given the instructions in the ticket. It seems like it addresses a different problem. Maybe a different issue is needed?

I can reproduce it in 7.6, but only with a newly created URL.  Once I refresh the page and try to edit it further times the issue doesn't occur (shown below).

https://github.com/user-attachments/assets/15ba7c98-d957-472a-a71a-b7226ef095db

The second input that doesn't trigger a dirty state will have it's value persist in the node preview, while being constantly removed from the input box (this is after going to and from different nodegroups)
image

It's odd how this issue occurs after the inital saving of the value when it is null, but after refreshing the page data can be added and reverted to null with no problems.

samuelscandrett:
PR closed and replaced by https://github.com/archesproject/arches/pull/11710

PR #11562, opened 2024-10-22
#11561 Load all components from arches apps before looking in core

### LINKED ISSUE #11561: Not all Arches App components are loaded in load-component-dependencies
1. Using Arches-Lingo, `pip install .` to install dependencies (arches, arches-references, arches-vue-utils)
2. `python manage.py runserver` and `npm run start`,
3. Visit controlled list manager (http://localhost:8000/plugins/controlled-list-manager)
4. You should observe an error in the console like this: `Error: Cannot find module './views/components/plugins/controlled-list-manager'`

In `load-component-dependencies.js`, the long try/catch block only iterates over one Arches app before giving up and looking for the component in Core.

johnatawnclementawn:
### Types of changes

### Issues Solved

Closes #11561

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @johnatawnclementawn  
*   Tested by: @ 
*   Designed by: @


arches/app/media/js/utils/load-component-dependencies.js
```diff
@@ -60,7 +60,11 @@ define([], function() {
                                                             require(`${LINKED_APPLICATION_PATH_8}/media/js/${componentPath}`);
                                                             break;
                                                         }
-                                                        catch { throw new Error(); } // if all attempts fail within the loop, throw error for outer try/catch
+                                                        catch {
+                                                            if (i === ARCHES_APPLICATIONS.length - 1) { 
+                                                                throw new Error();
```
jacobtylerwalls: keeping the comment would be nice

jacobtylerwalls (approved):
+1 tested, will give Christopher a chance to take a look, too.

arches/app/media/js/utils/load-component-dependencies.js
```diff
@@ -60,7 +60,11 @@ define([], function() {
                                                             require(`${LINKED_APPLICATION_PATH_8}/media/js/${componentPath}`);
                                                             break;
                                                         }
-                                                        catch { throw new Error(); } // if all attempts fail within the loop, throw error for outer try/catch
+                                                        catch {
+                                                            if (i === ARCHES_APPLICATIONS.length - 1) { 
+                                                                throw new Error(); 
+                                                            }
```
jacobtylerwalls: Will need a changelog

jacobtylerwalls:
So this definitely improves the situation, and I'm happy if we merge it, but I just noticed that this is still going to return wrong results in terms of overrides by application ordering if something that shouldn't override first gets to override first simply because it's installed with `-e` and meets one of the try/catch conditions instead of progressing to the next loop iteration over the site-packages paths. We probably want to refactor this so that each of the 9 try/catch blocks tries a pair of `require()` calls, once in site-packages, once in the editable path.

chrabyrd:
👀

chrabyrd:
@johnatawnclementawn @jacobtylerwalls 

yeah I can't wait to get rid of this file. Pretty sure it's just there to support KOJS stuff. Anywho I gave this another going-over to avoid logic against comparing the length of ARCHES_APPLICATIONS. This lgtm but should be looked at again by you two 👍 

@jacobtylerwalls could you give a reproducable example of you're talking about? Trying to grok it.

chrabyrd (approved):

jacobtylerwalls:
> @jacobtylerwalls could you give a reproducable example of you're talking about? Trying to grok it.

Sure, the idea is that:
- your ordered arches applications are FOO, BAR, BAZ
- BAR and BAZ contain the component
- FOO and BAR are installed in site-pacakges, but BAZ is installed with -e, so it gets a linked_application_path

Expect: bar component is found
Actual: baz component is found, because we iterate like this:
```
check FOO in site packages
     check all editable paths for all applications
```

so we find baz before continuing the loop to fin bar in site-packages. Granted I haven't set this up with to actually test, just seemed this way from a code read.

releases/7.6.2.md
```diff
@@ -11,6 +11,7 @@
 -   Fail with 404 rather than 500 (AttributeError) for missing search components in search_home route #[11532](https://github.com/archesproject/arches/issues/11532)
 -   Adds an alert message before running the bulk deletion #[11405](https://github.com/archesproject/arches/issues/11405)
 -   Fixes a bug where getting the json-ld of a resource with many related resources could crash the system #[11572](https://github.com/archesproject/arches/issues/11572)
+-   Fixes bug that prevented some Arches Application front-end components from being loaded #[11561](https://github.com/archesproject/arches/issues/11561)
```
jacobtylerwalls: We missed 7.6.2 with this, so let's bootstrap a new file

jacobtylerwalls (approved):

chrabyrd:
> > @jacobtylerwalls could you give a reproducable example of you're talking about? Trying to grok it.
> 
> Sure, the idea is that:
> 
> * your ordered arches applications are FOO, BAR, BAZ
> * BAR and BAZ contain the component
> * FOO and BAR are installed in site-pacakges, but BAZ is installed with -e, so it gets a linked_application_path
> 
> Expect: bar component is found Actual: baz component is found, because we iterate like this:
> 
> ```
> check FOO in site packages
>      check all editable paths for all applications
> ```
> 
> so we find baz before continuing the loop to fin bar in site-packages. Granted I haven't set this up with to actually test, just seemed this way from a code read.

ugh. yeah. I think this will require a change to webpack as there's no guarantee LINKED_APPLICATION_PATH_${num} matches the ARCHES_APPLICATION iteration value, so a double require block won't work. 

However, like you I'm happy to have this merged in and take care of the hardening in a followup, as anything that requires user action doesn't belong in a patch release if we can help it 😅

PR #11569, opened 2024-10-23
Fix failure to reverse migrate #11568

### LINKED ISSUE #11568: 10799_geojsongeometry_featureid can't be reversed
From a fully migrated 7.6/8.0.dev db:

```sh
python manage.py migrate models 10999_update_principaluser
```
```
django.db.utils.OperationalError: cannot ALTER TABLE "geojson_geometries" because it has pending trigger events
```
I've found that calling `refresh_geojson_geometries()` inside a RunSQL operation is pretty fragile because the function truncates a table and disturbs the per-migration transaction. Things usually work if you isolate the single migration operation involving that function into its own file.

That said, 7.6 has already shipped, so we can't easily insert a new migration into the middle of the history. We could cheat a little bit and just move this AddField operation into the previous migration and a comment explaining what/why:

https://github.com/archesproject/arches/blob/8ed87f361c63b1e432d048c6f3b3b7c7f72d8a57/arches/app/models/migrations/10799_geojsongeometry_featureid.py#L188-L192

jacobtylerwalls:
### Types of changes

### Description of Change
Fix failure to reverse migrate over 7.6. features.

### Issues Solved

Closes #11568

### Checklist

### Further comments
The only backward incompatibility I can see is if someone has only partially migrated their database on 7.6.0 to the migration changed in this file, then upgrades to 7.6.2, and attempts to migrate forward, but then they're not following the upgrade instructions: to fully migrate to 7.6.0 before applying a patch release, so all bets are off.


whatisgalen (approved):
Looks good -- thanks for fixing this thorny issue!

Issue #11447, opened 2024-09-11
Files are being opened without encoding arg


jacobtylerwalls:
Linters like [pylint](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/unspecified-encoding.html) and [ruff](https://docs.astral.sh/ruff/rules/unspecified-encoding/) both discourage opening files without an `encoding` arg, which can cause problems when files written on one platform are opened on another. Because arches is a general library used cross-platform, this affects us.

One instance that came up was #11337 and was fixed, but there are dozens of other issues you can see if you turn on `EncodingWarning` (see [PEP 597](https://peps.python.org/pep-0597/)), which you don't need a linter for:
```py
/Users/jwalls/prj/arches/arches/management/commands/load_ontology.py:84: EncodingWarning: 'encoding' argument not specified
  with open(config_file, "r") as f:
/Users/jwalls/prj/arches/tests/base_test.py:129: EncodingWarning: 'encoding' argument not specified
```

We have a few choices:

- add utf-8 encoding everywhere
- or, document that windows users should run python in utf-8 mode
- or, just want until Python 3.15 when utf-8 mode defaults on, but python has deferred things like this before, so there's no guarantee it will happen on schedule

Suggest adding default encoding and turning EncodingWarning on CI and elevating it to an error. This will function like a linter without having to decide on adopting a linter like pylint or ruff.

(Finally, the django_storages backend allows setting default encoding apparently, but that's only a potential partial fix for that one backend and for files going through the storages interface.)

chiatt:
I like the idea of adding explicit encoding with a CI check.

PR #11571, opened 2024-10-23
Generate static urls

### LINKED ISSUE #11567: Generate static urls for frontend consumption

chrabyrd:
### Types of changes

### Description of Change

This removes dependency on `arches_urls.htm` for all Vue frontend work. 

### Issues Solved

Closes #11567 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/templates/base-root.htm
```diff
@@ -22,7 +22,7 @@
 <!DOCTYPE html>
 <!--[if IE 8]> <html lang="en" class="ie8"> <![endif]-->
 <!--[if IE 9]> <html lang="en" class="ie9"> <![endif]-->
-<!--[if !IE]><!--> <html lang="en"> <!--<![endif]-->
+<!--[if !IE]><!--> <html lang="{{ app_settings.ACTIVE_LANGUAGE }}"> <!--<![endif]-->
```
jacobtylerwalls: Thinking about this more since yesterday, this might be alright after all, given the current pattern of using the i18n pattern routes from Django. On language switch, reload the page, see the updated /lang/... route.  Hash out a different pattern if/when we need it.

arches/app/templates/base.htm
```diff
@@ -17,16 +17,9 @@
...
 
-{% load static %}
 {% load i18n %}
-{% load webpack_static from webpack_loader %}
 {% load render_bundle from webpack_loader %}
 
-<!DOCTYPE html>
-<!--[if IE 8]> <html lang="en" class="ie8"> <![endif]-->
-<!--[if IE 9]> <html lang="en" class="ie9"> <![endif]-->
-<!--[if !IE]><!--> <html lang="en"> <!--<![endif]-->
```
jacobtylerwalls, line 28: 💸

arches/app/utils/frontend_configuration_utils.py
```diff
@@ -0,0 +1,198 @@
...
+def _generate_frontend_configuration_directory():
+    destination_dir = os.path.realpath(
+        os.path.join(_get_base_path(), "..", "frontend_configuration")
+    )
+
+    os.makedirs(destination_dir, exist_ok=True)
+
+
+def _generate_urls_json():
+    def generate_human_readable_urls(patterns, prefix="", namespace="", result={}):
```
jacobtylerwalls: I try to avoid mutable default arguments like `[]` and `{}`, although I do see you've wrapped this all in a closure. Just wondering if you'd prefer to make this a little more explicit so someone else isn't surprised that `result` persists between calls.

releases/8.0.0.md
```diff
@@ -69,9 +72,9 @@ JavaScript:
 
 1. In settings.py, add the following key to `DATABASES` to [improve indexing performance](https://github.com/archesproject/arches/issues/11382):
     ```
-        "OPTIONS": {
-            "options": "-c cursor_tuple_fraction=1",
-        },
+    "OPTIONS": {
+        "options": "-c cursor_tuple_fraction=1",
+    },
```
jacobtylerwalls: I had this intended like that for easier copy-pasting, since it's two levels deep. Maybe it's better to keep the indenting and clarify in the verbiage that this is all underneath "default:".

releases/8.0.0.md
```diff
@@ -80,21 +83,36 @@ JavaScript:
...
     ```
 
-2. Within your project, with your Python 3 virtual environment activated:
+1. Update `.gitignore`:
+    1. Remove `.tsconfig-paths.json` and `.frontend-configuration-settings.json`
+    2. Add `frontend_configuration`
+
+1. Replace the `generate_frontend_configuration` import statement in `apps.py`:
+    ```
+    from arches.settings_utils import generate_frontend_configuration
```
jacobtylerwalls: I toyed with importing this into the old location so that projects wouldn't need to update, but things broke, so 👍

releases/8.0.0.md
```diff
@@ -57,6 +58,8 @@ JavaScript:
 
 - `ensure_userprofile_exists()` was removed from the `Tile` model.
 
+- The `arches` object is no longer available in frontend components that exist outside of defined Arches patterns. If your project contains Vue components that are mounted on templates that inherit directly from `base-root.htm`, they will need to be updated to use `generateArchesURL` and the settings API instead.
```
jacobtylerwalls: Suggest something a little more specific than "defined patterns", which makes sense to people who read on about `base-root.htm` and know what that is, but others might need help understanding if this applies to them.

arches/app/utils/frontend_configuration_utils.py
```diff
@@ -0,0 +1,198 @@
...
+        def join_paths(*args):
+            return "/".join(filter(None, (arg.strip("/") for arg in args)))
```
jacobtylerwalls: We seem to be losing the trailing slash on some routes. Leaving aside whether it's a good idea to have trailing slashes required in your routes, if projects _do_ have that, I think we should leave it to them (and whatever third-party packages even outside of arches they depend on).

For instance, with status quo, we have on javascript.htm:
```
reorder_cards="/reorder_cards/"
```
but now in urls.json:
```
    "reorder_cards": "/reorder_cards",
```
giving:
```js
> _arches_utils_generate_arches_url_ts__WEBPACK_IMPORTED_MODULE_11__["default"]('reorder_cards')
'/reorder_cards'
```

Which will cause a redirect (or not) depending on the value of `APPEND_SLASH` and yada yada.

Can we avoid a behavior change here?

jacobtylerwalls (changes requested):
Principle makes sense. Noticed one behavior change I think we should look into. Will kick the tires again after.

releases/8.0.0.md
chrabyrd: oh no fair enough, I just ran the doc through a formatter and that popped out. Happy to revert

arches/app/templates/base-root.htm
jacobtylerwalls: It's early days for sure, but this also matches how I'm treating the system language on the vue side as basically read-only. When we implement the language switcher, we can still leave it read-only based on the updated response from the server.

arches/app/utils/frontend_configuration_utils.py
chrabyrd: nice catch, took a little bit of thinking but made that adjustment 👍

jacobtylerwalls (changes requested):
I cross-checked this against the new `listurls` [feature](https://github.com/django/django/pull/18769) likely to land in Django 5.2 and noticed that by keying the dict on `name`, we're overwriting and losing additional routes that lack a name, e.g. `tileserver/<path>`.

https://github.com/archesproject/arches/blob/dbd103a7ddaebbedc5f2c87c44d43898cf7c45d7/arches/urls.py#L692

Maybe we can have an additional data structure to hold unnamed routes?

releases/8.0.0.md
```diff
@@ -77,37 +80,63 @@ JavaScript:
...
-3. Create editable_future_graphs for your Resource Models using the command `python manage.py graph create_editable_future_graphs`. This will publish new versions of each Graph.
+1. Within your project, with your Python 3 virtual environment activated:
+    1.  Migrate your database:
+        ```
+        python manage.py migrate
+        ```
+
+    1. Then update your project:
+        ```
+        python manage.py updateproject
```
chiatt: I think it would be good to let people know that they can delete their `arches_urls.htm` file (I'd suggest using update project, but that might be a little intrusive)

chiatt (changes requested):
Can `arches_urls.htm` be removed from the project template?

jacobtylerwalls (commented):
Hey, just circling back here. I think the issue I mentioned above about unnamed routes is not so serious, given that we weren't using them on the front end before.

But there's a similar issue with routes we do use on the frontend like `plugins`.

There are four routes named "plugins". This is fine, since `reverse()` [matches on the number and names of args](https://docs.djangoproject.com/en/5.1/topics/http/urls/#naming-url-patterns):

https://github.com/archesproject/arches/blob/f3390eff2ff5fe8eca899d2b3f199868fcea1575/arches/urls.py#L652-L655

But since we're keying on name, we can hold at most one. The one that wins out is:
```
    "plugins": "/plugins/{slug}/{path}",
```

That's the form we happen to be using in the controlled list manager, e.g. /plugins/controlled-list-manager/item/pk

However, the bulk data manager uses
/plugins/bulk-data-manager,

And I don't think a client could generate the URL for that, if this test case is right:
```diff
diff --git a/arches/app/src/arches/utils/generate-arches-url.test.ts b/arches/app/src/arches/utils/generate-arches-url.test.ts
index 68fab4bf88..d7f752450f 100644
--- a/arches/app/src/arches/utils/generate-arches-url.test.ts
+++ b/arches/app/src/arches/utils/generate-arches-url.test.ts
@@ -7,9 +7,14 @@ global.ARCHES_URLS = {
     another_url: "/admin/another/{id}",
     multi_interpolation_url:
         "/{language_code}/resource/{resource_id}/edit/{field_id}/version/{version_id}",
+    plugins: "/plugins/{slug}/{path}",
 };
 
 describe("generateArchesURL", () => {
+    it("should return simple plugin url", () => {
+        const result = generateArchesURL("plugins", { slug: "bulk-data-manager" });
+        expect(result).toBe("/plugins/bulk-data-manager");
+    });
     it("should return a valid URL with specified language code and parameters", () => {
         const result = generateArchesURL("example_url", { id: "123" }, "fr");
         expect(result).toBe("/fr/admin/example/123");
```

```
 FAIL  arches/app/src/arches/utils/generate-arches-url.test.ts > generateArchesURL > should return simple plugin url
AssertionError: expected '/plugins/bulk-data-manager/{path}' to be '/plugins/bulk-data-manager' // Object.is equality

- Expected
+ Received

- /plugins/bulk-data-manager
+ /plugins/bulk-data-manager/{path}
```

Do you see the same?

releases/8.0.0.md
chrabyrd: ya agreed that'd be a little intrusive. Happy to update the doc and then in the next major version of Arches remove support for it entirely

chrabyrd:
> I cross-checked this against the new `listurls` [feature](https://github.com/django/django/pull/18769) likely to land in Django 5.2 and noticed that by keying the dict on `name`, we're overwriting and losing additional routes that lack a name, e.g. `tileserver/<path>`.
> 
> https://github.com/archesproject/arches/blob/dbd103a7ddaebbedc5f2c87c44d43898cf7c45d7/arches/urls.py#L692
> 
> Maybe we can have an additional data structure to hold unnamed routes?

I opted for the logic to, when a route is unnamed, default to it's first path. eg

```
    re_path(r"^iiifserver/(?P<path>.*)$", IIIFServerProxyView.as_view()),
```

```
    "iiifserver": [
        {
            "url": "/iiifserver/{path}",
            "params": [
                "path"
            ]
        }
    ],
```

chrabyrd:
> Hey, just circling back here. I think the issue I mentioned above about unnamed routes is not so serious, given that we weren't using them on the front end before.
> 
> But there's a similar issue with routes we do use on the frontend like `plugins`.
> 
> There are four routes named "plugins". This is fine, since `reverse()` [matches on the number and names of args](https://docs.djangoproject.com/en/5.1/topics/http/urls/#naming-url-patterns):
> 
> https://github.com/archesproject/arches/blob/f3390eff2ff5fe8eca899d2b3f199868fcea1575/arches/urls.py#L652-L655
> 
> But since we're keying on name, we can hold at most one. The one that wins out is:
> 
> ```
>     "plugins": "/plugins/{slug}/{path}",
> ```
> 
> That's the form we happen to be using in the controlled list manager, e.g. /plugins/controlled-list-manager/item/pk
> 
> However, the bulk data manager uses /plugins/bulk-data-manager,
> 
> And I don't think a client could generate the URL for that, if this test case is right:
> 
> ```diff
> diff --git a/arches/app/src/arches/utils/generate-arches-url.test.ts b/arches/app/src/arches/utils/generate-arches-url.test.ts
> index 68fab4bf88..d7f752450f 100644
> --- a/arches/app/src/arches/utils/generate-arches-url.test.ts
> +++ b/arches/app/src/arches/utils/generate-arches-url.test.ts
> @@ -7,9 +7,14 @@ global.ARCHES_URLS = {
>      another_url: "/admin/another/{id}",
>      multi_interpolation_url:
>          "/{language_code}/resource/{resource_id}/edit/{field_id}/version/{version_id}",
> +    plugins: "/plugins/{slug}/{path}",
>  };
>  
>  describe("generateArchesURL", () => {
> +    it("should return simple plugin url", () => {
> +        const result = generateArchesURL("plugins", { slug: "bulk-data-manager" });
> +        expect(result).toBe("/plugins/bulk-data-manager");
> +    });
>      it("should return a valid URL with specified language code and parameters", () => {
>          const result = generateArchesURL("example_url", { id: "123" }, "fr");
>          expect(result).toBe("/fr/admin/example/123");
> ```
> 
> ```
>  FAIL  arches/app/src/arches/utils/generate-arches-url.test.ts > generateArchesURL > should return simple plugin url
> AssertionError: expected '/plugins/bulk-data-manager/{path}' to be '/plugins/bulk-data-manager' // Object.is equality
> 
> - Expected
> + Received
> 
> - /plugins/bulk-data-manager
> + /plugins/bulk-data-manager/{path}
> ```
> 
> Do you see the same?

This is a good catch. I've updated the format of the JSON ( && related JS function ) to handle for this case:

```
    "plugins": [
        {
            "url": "/plugins/{pluginid}",
            "params": [
                "pluginid"
            ]
        },
        {
            "url": "/plugins/{pluginid}/{path}",
            "params": [
                "pluginid",
                "path"
            ]
        },
        {
            "url": "/plugins/{slug}",
            "params": [
                "slug"
            ]
        },
        {
            "url": "/plugins/{slug}/{path}",
            "params": [
                "slug",
                "path"
            ]
        }
    ],
```

chrabyrd:
@chiatt @jacobtylerwalls  This should hopefully be g2g now 👍

.github/actions/build-and-test-branch/action.yml
```diff
@@ -33,7 +33,7 @@ runs:
 
     - name: Ensure frontend configuration files exist
       run: |
-          python manage.py check
+          python manage.py check --tag=compatibility
```
jacobtylerwalls: This change was removed in b9dc94f when removing the failed experiment with writing a system check to check languages & published graphs. Better not to add this complexity back, I think.

.github/actions/build-and-test-branch/action.yml
```diff
@@ -74,7 +74,7 @@ runs:
 
     - name: Check for missing migrations
       run: |
-        python manage.py makemigrations --check
+        python manage.py makemigrations --check --skip-checks
```
jacobtylerwalls: Same

arches/install/arches-templates/.github/actions/build-and-test-branch/action.yml
```diff
@@ -33,7 +33,7 @@ runs:
     
     - name: Ensure frontend configuration files exist
       run: |
-          python manage.py check
+          python manage.py check --tag=compatibility
```
jacobtylerwalls: and here and below

releases/8.0.0.md
```diff
@@ -160,7 +163,20 @@ JavaScript:
     - Add `"Framework :: Django :: 5.2"`
     - Remove any classifier beginning with "License :: "
 
-1. Remove "3.10" from the `python-version` matrix in `.github/workflows/main.yml`.
+1. Update `.gitignore`:
```
jacobtylerwalls, line 169: Needs consolidation with the .gitignore instruction on L184

releases/8.0.0.md
jacobtylerwalls: Actually it works without a breaking change if we do the import at the bottom of the file, but since this is 8.0 and some breaking changes are to be expected, I think it's fine to require this to be updated.

releases/8.0.0.md
```diff
@@ -427,16 +451,40 @@ JavaScript:
...
+        <!-- updated format -->
 
-    <img id="map-buffer-search-help-gif" src="{% webpack_static 'img/help/map-buffer-search.gif' %}" aria-hidden="true"></img>
-    ```
+        <img id="map-buffer-search-help-gif" src="{% webpack_static 'img/help/map-buffer-search.gif' %}" aria-hidden="true"></img>
+        ```
+
+1. (optional) Remove `templates/arches_urls.htm`:
+    - This change will be enforced in later Arches versions.
+    - Instead of accessing urls via `import arches from "arches";`, you can instead use the `generateArchesUrl` utility.
```
jacobtylerwalls: Url -> URL

releases/8.0.0.md
```diff
@@ -427,16 +451,40 @@ JavaScript:
...
+        ```html
+        <!-- updated format -->
 
-    <img id="map-buffer-search-help-gif" src="{% webpack_static 'img/help/map-buffer-search.gif' %}" aria-hidden="true"></img>
-    ```
+        <img id="map-buffer-search-help-gif" src="{% webpack_static 'img/help/map-buffer-search.gif' %}" aria-hidden="true"></img>
+        ```
+
+1. (optional) Remove `templates/arches_urls.htm`:
+    - This change will be enforced in later Arches versions.
```
jacobtylerwalls: Do you want to reiterate here for clarity that it's required now if you extend from base-root?

releases/8.0.0.md
```diff
@@ -427,16 +451,40 @@ JavaScript:
...
+    - Instead of accessing urls via `import arches from "arches";`, you can instead use the `generateArchesUrl` utility.
+
+        ```js
+        // defunct format
+
+        import arches from "arches";
+
+        const url = `${arches.urls.my_awesome_url}?my_awesome_query_arg='foo'`
```
jacobtylerwalls: I don't think this example is equivalent to the below, as `generateArchesURL` doesn't seem to build query params, just path components.

webpack/webpack.config.dev.js
```diff
@@ -39,7 +39,16 @@ module.exports = () => {
                     }),
                     new StylelintPlugin({
                         files: Path.join('src', '**/*.s?(a|c)ss'),
-                    })
+                    }),
+                    {
+                        apply: (compiler) => {
+                            compiler.hooks.afterCompile.tap('WatchArchesUrlsPlugin', (compilation) => {
+                                compilation.fileDependencies.add(
+                                    Path.resolve(__dirname, APP_ROOT, '..', 'frontend_configuration', 'urls.json')
```
jacobtylerwalls, line 47: The 7.6 upgrade notes had to call out making this location writable, so I figure we need to do that in 8.0 also.

arches/app/utils/frontend_configuration_utils.py
```diff
@@ -0,0 +1,221 @@
...
+    )
+
+    with open(destination_path, "w") as file:
+        json.dump(tsconfig_paths_data, file, indent=4)
+
+
+def _get_base_path():
+    return (
+        os.path.realpath(settings.ROOT_DIR)
+        if settings.APP_NAME == "Arches"
```
jacobtylerwalls: This setting has given us a lot of indigestion. We found after 7.6 that projects were bootstrapped with tiles that overrode this setting, and behavior difference between devserver and wsgi.py, see #11702 and #11700. Can we find something else?

arches/app/media/js/utils/create-vue-application.js
```diff
@@ -10,8 +10,9 @@ import Tooltip from 'primevue/tooltip';
 import { createApp } from 'vue';
 import { createGettext } from "vue3-gettext";
 
-import arches from 'arches';
 import { DEFAULT_THEME } from "@/arches/themes/default.ts";
+import generateArchesURL from '@/arches/utils/generate-arches-url.ts';
```
jacobtylerwalls: Getting a webpack warning:

```
WARNING in ../arches/arches/app/media/js/utils/create-vue-application.js 28:49-66
export 'default' (imported as 'generateArchesURL') was not found in '@/arches/utils/generate-arches-url.ts' (possible exports: generateArchesURL)
```

releases/8.0.0.md
```diff
@@ -427,16 +451,40 @@ JavaScript:
...
+        ```js
+        // updated format
+
+        import { generateArchesURL } from "@/arches/utils/generate-arches-url.ts";
+
+        const url = generateArchesURL("my_awesome_url", { my_awesome_query_arg: "foo" });
+        fetch(url).then(resp => {...});
+        ```
+
+    - If you update your project to remove any reference to `arches.urls`, you can then safely delete `templates/arches_urls.htm`
```
jacobtylerwalls, line 523: I tested lingo and it works great. I tried to access the CLM from lingo, with or without these changes, and it breaks:
```
Uncaught Error: Unknown component 'controlled-list-manager'
```

Did I miss a step (can you reproduce on your end?)

arches/app/utils/frontend_configuration_utils.py
```diff
@@ -0,0 +1,221 @@
...
+
+    sorted_urls = {
+        name: urls_grouped_by_name[name] for name in sorted(urls_grouped_by_name)
+    }
+
+    destination_path = os.path.realpath(
+        os.path.join(_get_base_path(), "..", "frontend_configuration", "urls.json")
+    )
+
+    with open(destination_path, "w") as file:
```
jacobtylerwalls: I know this is pre-existing, but let's use this opportunity to pass "utf-8" since we depend on that later in the node code, see #11447. Couple other places in this file.

arches/app/utils/frontend_configuration_utils.py
```diff
@@ -0,0 +1,221 @@
...
+            return regex
+        return ""
+
+    def build_urls(patterns, prefix="", namespace="", collected_urls=None):
+        if collected_urls is None:
+            collected_urls = {}
+
+        for pattern in patterns:
+            if isinstance(pattern, URLPattern):
+                full_name = f"{namespace}{pattern.name}" if pattern.name else None
```
jacobtylerwalls, line 62: I noticed when dumping the urls that the django admin app correctly [namespaces its urls](https://docs.djangoproject.com/en/5.1/topics/http/urls/#url-namespaces). I'll open a follow-up issue for starting to do that in arches (and lingo & controlled-lists etc) as a good practice.

This is also mentioned in Django [tutorial 3](https://docs.djangoproject.com/en/5.1/intro/tutorial03/#namespacing-url-names).

Then, we might want `generateArchesURL` to take an optional namespace arg, but that can be a follow-up.

jacobtylerwalls (commented):
Thanks for updating those edge cases, this is looking really great. My main question is just whether you can still load the controlled list manager? The rest of the feedback is minor.

jacobtylerwalls:
@chiatt still has a good shout [here](https://github.com/archesproject/arches/pull/11571#pullrequestreview-2431040090)

jacobtylerwalls:
Ah, maybe it's already gone? I don't see where it was removed in the PR but I'm sure I'm just missing it.

arches/app/utils/frontend_configuration_utils.py
chrabyrd: Yeah namespacing URLs for projects/apps may not be what we want to do. There are several cases of using route hijacking at the project level to alter the logic of core components.

arches/app/media/js/utils/create-vue-application.js
chrabyrd: ah my bad, didn't update after turning it into a named export

arches/app/utils/frontend_configuration_utils.py
chrabyrd: Mind fleshing out this issue a little more?

releases/8.0.0.md
chrabyrd: It's not required _yet_. Next major version we can remove support

arches/app/utils/frontend_configuration_utils.py
jacobtylerwalls: New projects get created with a settings.py template giving a string like "my_project" for APP_NAME. However, setup_db creates a system settings tile for the app name node with a value of "Arches" (#11702 suggests it should be "my_project"), and then `SystemSettings.update_from_db()` reads that tile, and merges it into the settings, with the result that settings.APP_NAME is now "Arches" even though it's the _project_ that's running. But this behavior does not happen when running the devserver, making this class of bugs easy to miss (#11700).

In sum, I'm suggesting that unless we fix those issues, we can't rely on APP_NAME to tell if the project being run is core arches versus a project. This also came up on the forum a couple days ago.

***
I guess if the only variation we're handling for here is just that the legacy core-arches folder structure has one more level of nesting, we can rely on that fact to derive whether the project being run is core or not like this:

```py
from pathlib import Path
running_core_arches_as_project = Path(settings.APP_ROOT).parent == Path(settings.ROOT_DIR)
```

Factor that out into a util, and we could fix #11701 which describes the same problem.

releases/8.0.0.md
jacobtylerwalls: I was thinking that since lingo breaks without updating to the new pattern, it's required if you inherit from base-root.htm, since the arches-urls include tag is gone now. No?

arches/app/utils/frontend_configuration_utils.py
jacobtylerwalls: I assigned #11701 to me and will stub out that util I had in mind

jacobtylerwalls: Interestingly, we were [already doing a version of this](https://github.com/archesproject/arches/blob/f2c2c14246b0d31b8ac2b67c231457d197614923/arches/settings_utils.py#L164) so I might wait on the util

chrabyrd:
> Actually it works without a breaking change if we do the import at the bottom of the file, but since this is 8.0 and some breaking changes are to be expected, I think it's fine to require this to be updated.

Opted to remove that instruction, as `jtw/fix-app-name-guard` supercedes

releases/8.0.0.md
chrabyrd: This appears to be local to your system, I was able to access it on this branch.

jacobtylerwalls: Thanks for checking 🙏

jacobtylerwalls (approved):

PR #11580, opened 2024-10-27
11579 ids filter

### LINKED ISSUE #11579: create a search filter for a list of resourceinstance ids
There are a few places where arches does search queries for specific resources, usually one query at a time using the search filter query paramter `id=`; obviously there would be performance improvements if we could batch those searches into a single search using `ids=`. The cleanest way to implement this would be to create a search filter to handle this parameter with the appropriate logic.

whatisgalen:
### Types of changes

### Issues Solved

Closes #11579 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Found by: @whatisgalen 
*   Tested by: @chiatt 
*   Designed by: @


arches/app/models/migrations/11579_ids_search_filter.py
```diff
@@ -0,0 +1,26 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("models", "10437_node_alias_not_null"),
```
chiatt: Looks like this dependency needs to be updated

whatisgalen: taking a look

arches/app/search/components/ids.py
```diff
@@ -0,0 +1,27 @@
...
+    "componentpath": "",
+    "componentname": "ids",
+    "config": {},
+}
+
+
+class IdsFilter(BaseSearchFilter):
+
+    def append_dsl(self, search_query_object, **kwargs):
+        ids = kwargs.get("querystring", None)
```
apeters, line 21: I think this will always come in as a string.  You might need to deserialize first then check 'isinstance'.

whatisgalen: just pushed a commit

arches/app/search/components/ids.py
```diff
@@ -0,0 +1,29 @@
+from arches.app.search.elasticsearch_dsl_builder import Bool, Ids
+from arches.app.search.components.base import BaseSearchFilter
+from arches.app.utils.betterJSONSerializer import JSONDeserializer
+
+details = {
+    "searchcomponentid": "f1856bfb-c3c4-4d67-8f23-0aa3eef3a160",
+    "name": "Ids Filter",
+    "icon": "",
+    "modulename": "ids.py",
+    "classname": "IdsFilter",
```
chiatt: Could you change the class name to 'ResourceIdsFilter'. Otherwise there isn't any place but the test that indicates the type of id.

chiatt: Or - add a doc string to the class to clarify

chiatt (approved):

Issue #11578, opened 2024-10-27
search filters should be request.method agnostic


whatisgalen:
currently some search-filters are hard-coded to look for the paramater/query-string (or other query args) on a GET request body. Instead, they should use the search request object created in the search-view which looks for either a GET or a POST request

bferguso:
I think this brings up a larger issue that I've been mulling over for a while, and will likely start to come up with the addition of the filter by feature functionality. I think we need to find a strategy to handle filters that exceed the maximum URL length. Filter requests are all currently submitted as GETs, which is convenient because the filter can easily be bookmarked / shared. However once the filter exceeds the maximum URL length (which varies with the browser) then the calls start to fail.

I've started an implementation that uses a POST if the URL length is exceeded, however it means that the filter can no longer be saved or shared. I'm wondering if we should have a discussion as a group to agree to an overall strategy to handle this.

whatisgalen:
@bferguso in the case of a search url being used in the arches /search UI, it would be pretty straightforward to dynamically set the request to POST regardless of whether it was a "saved-search" or an ad hoc one. For search requests sent from somewhere else on the backend I think it's probably more complicated but I'm guessing there's a path

bferguso:
@whatisgalen - Yes - it's pretty straight forward, however just changing the request type from GET to POST breaks the current functionality that you can just bookmark/share any arbitrary ad-hoc search as you lose the parameters when they're stored in the request body. You also lose the search context if you do a simple page refresh.

There's also some coordination necessary as the parameters are expected to be URL encoded in a number of places (general search, advanced search, etc).

whatisgalen:
@bferguso you could just modify this [code](https://github.com/archesproject/arches/blob/dev/7.6.x/arches/app/media/js/views/components/search/base-search-view.js#L50) in your own search-view to dynamically set POST vs GET based on json string char limit. This code gets executed for in-app searches as well as when saved-searches get loaded. I'm not seeing a blocker or a loss of functionality in saved-searches or parameters...maybe I'm missing something?

whatisgalen:
@bferguso sorry I think I understand now what you were alluding to -- requests that are too long for GET and thus must be handled as POST cannot store all their query parameters in the url, meaning you can't just copy and paste that url into a saved search. Yes -- you are totally right. 

I think a real solution to this problem might look like this for example:
[...]

Issue #10630, opened 2024-02-22
Large Search Filter exceeds maximum URL length


bferguso:
When filtering resources, if the search filter is complex the maximum URL length can be easily exceeded (by definition I think it's about 2000 characters, however milage varies by browser). In our case it was originally observed when we implemented a filter-by-feature function where any arbitrary geometry in an overlay (which can be complex) could be used as a map filter. The same error can be achieved by generating a complex filter using Advanced Search.

We have implemented a workaround for the interactive search map filter page which submits the filter in the body of a POST rather than a GET if it exceeds maximum URL length, and I'm currently working on resolving a similar issue when exporting the results by CSV.

The downside of the above POST strategy is that the search filter can no longer be saved as searches are currently just bookmarks.

It would be great for Arches to be able to both handle filters that exceed maximum URL length, as well as being able to save such filters for future use.

whatisgalen:
We ran into the same problem that became the genesis for #10529 . One thought I had was: do we really need to store every single search parameter in the URL? Is there an alternative where instead of passing them all into the URL, we hash them into a string value?

CWDamm-Kint:
We've also been having an issue with errors from long urls, especially with a very complex map-filter polygon. 

I can see there's quite a lot of discussion and activity around this, across here, https://github.com/archesproject/arches/issues/11592, https://github.com/archesproject/arches/issues/11578, and https://github.com/archesproject/arches/pull/11582

Just wondering what the latest state of play is? Appreciate there may also have been some discussions in stand ups. 

For what it's worth, I agree that it would be good to find a solution that moved content out of the url and into the body via a post request (especially map filter polygons), ideally while maintaining the ability to save and share those searches. https://github.com/archesproject/arches/pull/11582 looks like a step towards making this possible. 

We'd also be happy to discuss this further if people are interested? @whatisgalen @bferguso @SDScandrettKint @chiatt

jacobtylerwalls:
@robgaston also mentioned hashing an object (elastic dsl object, queryset, the args to create one) as a potential way to keep using GET

bferguso:
This has continued to be on my mind, but I haven't had the cycles to put into it yet... I'd like to be part of the discussion if it's sparked back up again.

CWDamm-Kint:
@robgaston @jacobtylerwalls - Could you elaborate a little on this would work please? I'm not sure I've understood.

robgaston:
[...]

PR #11582, opened 2024-10-27
11578 make search-filters request.method agnostic

whatisgalen:
### Types of changes

### Issues Solved

Partly addresses #11578

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


chiatt:
This change doesn't seem like it addresses a crashing bug or anything that wasn't already in previous versions of Arches. I could be wrong, but it seems like the target should be 8.0.

whatisgalen:
> This change doesn't seem like it addresses a crashing bug or anything that wasn't already in previous versions of Arches. I could be wrong, but it seems like the target should be 8.0.

One could argue that issues like #10630 are crashing bugs. Previously, to crash the vanilla map-filter on the frontend you would have had to draw an unrealistic number of vertices. However, given the new "feature by filter" functionality in 7.6.0, you can select an existing geometry that might have a very high vertex-count and this breaks the entire search request as it could exceed the number of characters permitted in the GET header. 

The vanilla map filter looks like this on the backend:
```.py
class MapFilter(BaseSearchFilter):
    def append_dsl(self, search_query_object, **kwargs):
        permitted_nodegroups = kwargs.get("permitted_nodegroups")
        include_provisional = kwargs.get("include_provisional")
        search_query = Bool()
        querystring_params = kwargs.get("querystring", "{}")
```
so it's actually **already** request-agnostic; a developer could make a search-view that sends a POST request instead of a GET request. The issue is that _other_ filters are hard-coded to expect a GET request, so if those other filters were used in combination with a map-filter POST request, only part of the search request would be fulfilled. Those other search filters are: `paging-filter`, `sort-results`, and `term-filter`. The standard search-view already looks for a POST request or a GET request so the scope is just limited to making sure that gets honored fully in all search filters.

whatisgalen:
Regarding my [comment](https://github.com/archesproject/arches/pull/10827#issuecomment-2450663793) on PR #10827:

> Per my [comment](https://github.com/archesproject/arches/pull/10827#issuecomment-2305886850), the featureid was removed from the request payload to the backend (map-filter) because if the user edited the drawn polygon on the map, (for example to cover a larger spatial query area) the feature identified by featureid would differ from the query geometry, and there wasn't a great way to confirm whether the geometry had actually changed (given potential precision differences). This would have yielded search results out of sync with the search query.

> However, because the only payload is the geojson feature itself, this means large coordinate sets can exceed the character limit on the GET request header (and url length in the browser). Unfortunately, those requests then fail, hence #11578 and the need for #11582

...There's now a solid category of search requests that will always fail whenever a high-vertex feature is used as a map filter. This means that the feature is inherently bugged until the search request can be sent as a POST request rather than GET.

arches/app/search/components/standard_search_view.py
```diff
@@ -132,15 +133,16 @@ def append_dsl(self, search_query_object, **kwargs):
         search_query_object["query"].include("map_popup")
         search_query_object["query"].include("provisional_resource")
         search_query_object["query"].include("permissions")
-        load_tiles = get_str_kwarg_as_bool("tiles", self.request.GET)
+        load_tiles = get_str_kwarg_as_bool("tiles", search_request_object)
```
apeters: for clarity, and to improve on something that should have been done in the first place, this should default to `False` and not to `search_request_object`

arches/app/search/components/standard_search_view.py
```diff
@@ -132,15 +133,16 @@ def append_dsl(self, search_query_object, **kwargs):
...
         if load_tiles:
             search_query_object["query"].include("tiles")
 
     def execute_query(self, search_query_object, response_object, **kwargs):
-        for_export = get_str_kwarg_as_bool("export", self.request.GET)
-        pages = self.request.GET.get("pages", None)
-        total = int(self.request.GET.get("total", "0"))
-        resourceinstanceid = self.request.GET.get("id", None)
+        search_request_object = kwargs.get("search_request_object", self.request.GET)
+        for_export = get_str_kwarg_as_bool("export", search_request_object)
```
apeters: same as above, should default to `False`

apeters (changes requested):
General comments about structure.

1.  I think we should push most of the logic to parse the request  into the `base_search_view.py` module and centralize it there.
2. Just like we put the the request object on `self`, we should put the `search_request` (currently called `search_request_object`) on `self` as well.
3. We should then not have to pass it into the methods on the `BaseSearchFilter`

Do that first and then I'll re-review.

arches/app/search/components/standard_search_view.py
whatisgalen: This actually already does default to `False`, see full method from `string_utils.py`:
```.py
def get_str_kwarg_as_bool(
    key, request_dict: Union[Dict[str, Any], QueryDict], default: bool = False
) -> bool:
    value = request_dict.get(key, str(default))
    if isinstance(value, bool):
        return value
    return str_to_bool(str(value))
```

arches/app/search/components/standard_search_view.py
whatisgalen: see comment above

whatisgalen:
> General comments about structure.
> 
> 1. I think we should push most of the logic to parse the request  into the `base_search_view.py` module and centralize it there.

✅ (moved it into `SearchFilterFactory` which leverages the `SearchView.create_query_dict()` method anyhow)

> 2. Just like we put the the request object on `self`, we should put the `search_request` (currently called `search_request_object`) on `self` as well.

✅ 

> 3. We should then not have to pass it into the methods on the `BaseSearchFilter`
> 

✅ 

> Do that first and then I'll re-review.

arches/app/media/js/views/components/search/base-search-view.js
```diff
@@ -50,8 +50,9 @@ define([
         doQuery: function() {
             const queryObj = JSON.parse(this.queryString());
             if (self.updateRequest) { self.updateRequest.abort(); }
+            const requestMethod = JSON.stringify(queryObj).length > 1800 ? "POST" : "GET";
```
apeters, line 53: where did this number come from?  2048 seems like the most Edge can handle but most other browsers much more.

arches/app/search/components/base_search_view.py
```diff
@@ -95,6 +102,7 @@ def __init__(self, request=None, user=None, componentname=None):
                 item.componentname, float("inf")
             ),
         )
+        self.search_request = self.create_query_dict(self.search_request)
```
apeters, line 105: I doubt this is neccessary.

whatisgalen:
making a shortlist of outstanding items on this PR to remind myself later and also lead in to further review:
- need to add csrf excemption decorator to `search_results()`
- need to reconfigure the `SearchResultsExporter` class' `__init__` method also be request-method-agnostic

arches/app/media/js/views/components/search/base-search-view.js
whatisgalen: This leaves a generous buffer in case the host is really long as can sometimes be the case with development hosts using Azure, for example: "https://longsubdomain.azure.cloudapp.westus3.developmentserverlongname." Is there a reason you think we should increase this limit + forego compatibility with Edge?

arches/app/search/components/base_search_view.py
whatisgalen: `create_query_dict` is a searchview class method, so if your search view has logic, for example, whitelisting/blacklisting certain query parameters, this method is where that's happening. I can see however that because this variable is called `search_request`, it should probably have some of the other properties of the `request` object like `user` etc.

whatisgalen: Also due to the inheritance logic of search-view classes, setting the correct search-view on a blank request here: 

```.py
def create_query_dict(self, query_dict):
        # check that all searchview required linkedSearchFilters are present
        query_dict[self.searchview_component.componentname] = True # set the database default search-view
```

... is what enables the python class of that search-view to be initialized. I know it looks circular but this came about as an intentional fix.

apeters:
I realize that we're going to need some tests for this.

arches/app/media/js/views/components/search/base-search-view.js
bferguso: Maybe another option is to build the full GET URL string and compare that to the 2048 limit? May need to URL Encode the query string to as that can add a large number of characters to the URL (eg: " " -> "%20", etc).

Something like this might work to get the full URL length?:
```
const urlLength = (window.location.origin+arches.urls.export_results+"?"+$.param(payload)).length
```

Also - there's at least one other place that we'll need to handle this on the front end. Maybe worth wrapping the logic in a separate util function as it feels like this may be implemented in a number of places?
https://github.com/archesproject/arches/blob/debacb941aef332762a606ac84e64523e50d95d5/arches/app/media/js/views/components/search/search-export.js#L77-L80

arches/app/search/components/base.py
```diff
@@ -83,10 +90,8 @@ def get_filter(self, componentname):
     def get_searchview_name(self):
         if not self.request:
             searchview_component_name = None
-        elif self.request.method == "POST":
-            searchview_component_name = self.request.POST.get("search-view", None)
         else:
-            searchview_component_name = self.request.GET.get("search-view", None)
+            searchview_component_name = self.search_request.get("search-view", None)
```
apeters, line 94: I think we only need the "else" portion of this statement.  self.request is already assumed to exist in the `__init__` method.  If it doesn't exist, then that method will fail.  Will self.request ever be None?

arches/app/search/search_export.py
```diff
@@ -46,16 +46,28 @@ class SearchResultsExporter(object):
...
+            search_request.POST = search_request.POST.copy()
+            search_request.POST["tiles"] = True
+            search_request.POST["export"] = True
+            self.report_link = search_request.POST.get("reportlink", False)
+            self.format = search_request.POST.get("format", "tilecsv")
+            self.export_system_values = get_str_kwarg_as_bool(
+                "exportsystemvalues", search_request.POST
+            )
+            self.compact = search_request.POST.get("compact", True)
+            self.precision = int(search_request.POST.get("precision", 5))
```
apeters, line 70: I think this if/else could be boiled down to a single chunk of code using something like this:
```
        the_request = getattr(search_request, method).copy()
        the_request["tiles"] = True
        the_request["export"] = True
        self.report_link = the_request.get("reportlink", False)
        self.format = the_request.get("format", "tilecsv")
        self.export_system_values = get_str_kwarg_as_bool(
            "exportsystemvalues", the_request
        )
        self.compact = the_request.get("compact", True)
        self.precision = int(the_request.get("precision", 5))
        setattr(search_request, method, the_request)
```

arches/app/views/search.py
```diff
@@ -349,6 +350,7 @@ def get_dsl_from_search_string(request):
     return JSONResponse(dsl)
 
 
+@csrf_exempt
```
apeters, line 353: what's the need for this?

arches/app/search/components/sort_results.py
```diff
@@ -16,7 +16,7 @@
 
 class SortResults(BaseSearchFilter):
     def append_dsl(self, search_query_object, **kwargs):
-        sort_param = self.request.GET.get(self.componentname, None)
+        sort_param = kwargs.get("querystring", None)
```
apeters, line 19: shouldn't this be 

`sort_param = self.search_request.get(self.componentname, None)`

arches/app/search/components/base.py
```diff
@@ -54,6 +57,10 @@ def __init__(self, request=None, user=None):
             for search_filter in SearchComponent.objects.all()
         }
         self.search_filters_instances = {}
+        request_object = (
+            self.request.GET if self.request.method == "GET" else self.request.POST
+        )
+        self.search_request = request_object.dict()
```
apeters, line 63: I think that we should rename this to "self.querystring" because essentially that's what it is.  This would reduce confusion with "self.request"

arches/app/views/search.py
whatisgalen: POST requests can fail for users who aren't authenticated.

chiatt:
Closing as stale. Also noting that now the right approach is probably to use the QUERY http method: https://tykok.medium.com/the-new-http-method-query-d1dc3efc35c4. That would allow us to use a single method rather than using GET in some cases and POST in others.

Issue #10079, opened 2023-09-27
Resource editor raises KeyError when switching languages


jacobtylerwalls:
1. Enable ~the language switcher~ an additional language
2. Save a new resource ~model~ instance
3. Change to that language

```py
Internal Server Error: /resource/44490b12-feb3-4b09-977c-8b15588baea4
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.10/dist-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.10/dist-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/django/utils/decorators.py", line 46, in _wrapper
    return bound_method(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/django/contrib/auth/decorators.py", line 23, in _wrapper_view
    return view_func(request, *args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/django/views/generic/base.py", line 143, in dispatch
    return handler(request, *args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/django/utils/decorators.py", line 46, in _wrapper
    return bound_method(*args, **kwargs)
  File "/web_root/arches/arches/app/utils/decorators.py", line 82, in wrapper
    return function(request, *args, **kwargs)
  File "/web_root/arches/arches/app/views/resource.py", line 277, in get
    serialized_graph = published_graph.serialized_graph
AttributeError: 'NoneType' object has no attribute 'serialized_graph'
```

Tested with afs
Version 7.5 dev

jacobtylerwalls:
Can still reproduce

chrabyrd:
Can you add the steps to reproduce?

jacobtylerwalls:
Looks like the OP was missing a step. (sorry!)

1. Publish a graph
2. Add a new language to settings.py that wasn't there before
3. Rebuild webpack and refresh
4. Create an instance of that graph
5. Switch to the language you added in step 2


Now it strikes me that the problem goes away if I run `manage.py graph publish --update` after adding the new language. If this is a requirement, we should document it, I figure?

I'll also update the traceback, which is a little different now.

[...]

PR #11601, opened 2024-11-01
Automate the changes to the github action template re #10079

jacobtylerwalls:
Follow-up to #10079.

The github actions only setup a test database, not a default database. The system check in #10079 runs against the default database. We need to make one conscious of the other, either introspect the database connection in the system check, which seems a little intense, or just simplify the github action to skip unnecessary checks, which is what #10079 did for core arches without adding any automation to `updateproject`.

This adds that automation to `updateproject` to reduce friction on upgrade.


chrabyrd (approved):
lgtm!

PR #11609, opened 2024-11-08
Hardens editable_future_graph against nodegroup changes #11570

### LINKED ISSUE #11570: Cannot republish graph after adding parent/child nodegroup configuration
dev/8.0.x
Graph Designer:
- Create graph
- Create one child node
- Save edits
- Create one grandchild node
- Save edits
- Create new publication and confirm dialog
- Grandchild node: toggle "Place node(s) in separate card" to ON
- Save edits
- Create new publication and confirm dialog
```py
[23/Oct/2024 09:45:04] "POST /graph_has_unpublished_changes/5defdbf2-914d-11ef-8101-1ea1ac521291 HTTP/1.1" 200 33
2024-10-23 09:45:12,810 arches.app.views.graph ERROR    NodeGroup matching query does not exist.
  File "/Users/jwalls/django/django/db/models/fields/related_descriptors.py", line 239, in __get__
    rel_obj = self.field.get_cached_value(instance)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jwalls/django/django/db/models/fields/mixins.py", line 37, in get_cached_value
    return instance._state.fields_cache[self.cache_name]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
KeyError: 'nodegroup'

During handling of the above exception, another exception occurred:

  File "/Users/jwalls/prj/arches/arches/app/views/graph.py", line 693, in post
    source_graph.update_from_editable_future_graph()
  File "/Users/jwalls/prj/arches/arches/app/models/graph.py", line 2582, in update_from_editable_future_graph
    _update_source_nodegroup_hierarchy(future_card.nodegroup)
                                       ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jwalls/django/django/db/models/fields/related_descriptors.py", line 257, in __get__
    rel_obj = self.get_object(instance)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jwalls/django/django/db/models/fields/related_descriptors.py", line 220, in get_object
    return qs.get(self.field.get_reverse_related_filter(instance))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jwalls/django/django/db/models/query.py", line 630, in get
arches.app.models.models.NodeGroup.DoesNotExist: NodeGroup matching query does not exist.
Internal Server Error: /graph/5defdbf2-914d-11ef-8101-1ea1ac521291/publish
[23/Oct/2024 09:45:12] "POST /graph/5defdbf2-914d-11ef-8101-1ea1ac521291/publish HTTP/1.1" 500 145
```

chrabyrd:
### Types of changes

### Description of Change

This handles `source_graph` nodegroup creation for when a node is set to `is_collector` after already being created. It includes general performance/QoL improvements, and has spawned a followup ticket: https://github.com/archesproject/arches/issues/11610

### Issues Solved

Closes #11570 

### Checklist


chrabyrd:
re-running all tests to ensure flakiness no longer occurs. 1 successful run isn't enough 😂

johnatawnclementawn (changes requested):
Dog-fooding this for converting string nodes into non-localized string nodes and found that if your graph has a slug, you get an error like this:

```
File "/Users/johnathanclementi/Documents/Arches/Projects/rdm/arches/arches/app/models/graph.py", line 2577, in update_from_editable_future_graph
    return self.restore_state_from_serialized_graph(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        serialized_editable_future_graph
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/johnathanclementi/Documents/Arches/Projects/rdm/arches/arches/app/models/graph.py", line 2689, in restore_state_from_serialized_graph
    updated_graph.save()
    ~~~~~~~~~~~~~~~~~~^^
  File "/Users/johnathanclementi/Documents/Arches/Projects/rdm/arches/arches/app/models/graph.py", line 599, in save
    self.validate()
    ~~~~~~~~~~~~~^^
  File "/Users/johnathanclementi/Documents/Arches/Projects/rdm/arches/arches/app/models/graph.py", line 2328, in validate
    raise GraphValidationError(
    ...<4 lines>...
    )
arches.app.models.graph.GraphValidationError: "Another resource model already uses the slug 'string_example_model'"
```

chrabyrd:
> Dog-fooding this for converting string nodes into non-localized string nodes and found that if your graph has a slug, you get an error like this:
> 
> ```
> File "/Users/johnathanclementi/Documents/Arches/Projects/rdm/arches/arches/app/models/graph.py", line 2577, in update_from_editable_future_graph
>     return self.restore_state_from_serialized_graph(
>            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
>         serialized_editable_future_graph
>         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     )
>     ^
>   File "/Users/johnathanclementi/Documents/Arches/Projects/rdm/arches/arches/app/models/graph.py", line 2689, in restore_state_from_serialized_graph
>     updated_graph.save()
>     ~~~~~~~~~~~~~~~~~~^^
>   File "/Users/johnathanclementi/Documents/Arches/Projects/rdm/arches/arches/app/models/graph.py", line 599, in save
>     self.validate()
>     ~~~~~~~~~~~~~^^
>   File "/Users/johnathanclementi/Documents/Arches/Projects/rdm/arches/arches/app/models/graph.py", line 2328, in validate
>     raise GraphValidationError(
>     ...<4 lines>...
>     )
> arches.app.models.graph.GraphValidationError: "Another resource model already uses the slug 'string_example_model'"
> ```


we should chat, I cannot reproduce this error in test or UX. I added a test anyway 👍

arches/app/media/js/viewmodels/graph-settings.js
```diff
@@ -15,9 +15,12 @@ define([
 
         self.resource_data = ko.observableArray([]);
         self.relatable_resources = ko.computed(function() {
-            return _.each(self.resource_data(), function(resource) {
```
jacobtylerwalls, line 18: Is https://github.com/archesproject/arches/pull/9689 something we can close now?

arches/app/models/migrations/11570_harden_editble_future_graphs.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 5.1.2 on 2024-11-13 18:17
```
jacobtylerwalls: filename: edtible -> editable

arches/app/models/models.py
```diff
@@ -855,37 +855,63 @@ def is_collector(self):
...
+        old_ids = set()
+        for res in self.get_relatable_resources():
+            if res.source_identifier_id is not None:
+                old_ids.add(res.source_identifier_id)
+            if res.nodeid is not None:
+                old_ids.add(res.nodeid)
+
+        self_ids = set(
+            id for id in (self.source_identifier_id, self.nodeid) if id is not None
+        )
```
jacobtylerwalls, line 799:
```suggestion
        self_ids = {self.source_identifier_id, self.nodeid} - {None}
```

arches/app/models/models.py
```diff
@@ -855,37 +855,63 @@ def is_collector(self):
...
+        if ids_to_create:
+            new_constraints = [
+                Resource2ResourceConstraint(
+                    resourceclassfrom_id=self.source_identifier_id or self.nodeid,
+                    resourceclassto_id=id_to_create,
                 )
-                new_r2r.save()
+                for id_to_create in ids_to_create
+            ]
+            Resource2ResourceConstraint.objects.bulk_create(new_constraints)
```
jacobtylerwalls, line 824: nice!

arches/app/models/graph.py
```diff
@@ -1706,9 +1706,15 @@ def update_permissions(self, serialized_graph):
                     "user_permissions"
                 ].values():
                     for serialized_user_permission in serialized_user_permission_list:
+                        serialized_user_permission["content_object"] = (
+                            models.NodeGroup.objects.get(
+                                pk=serialized_user_permission["object_pk"]
+                            )
+                        )
```
jacobtylerwalls: WDYT about prefetching these to avoid one-by-one queries?

arches/app/models/graph.py
```diff
@@ -1726,9 +1732,15 @@ def update_permissions(self, serialized_graph):
                     "group_permissions"
                 ].values():
                     for serialized_group_permission in serialized_group_permission_list:
+                        serialized_group_permission["content_object"] = (
+                            models.NodeGroup.objects.get(
```
jacobtylerwalls: ditto

arches/app/models/graph.py
```diff
@@ -2412,406 +2424,159 @@ def create_editable_future_graph(self):
...
-                        pk=nodegroup.parentnodegroup_id
-                    )
-
-                    if nodegroup_parent_node.source_identifier_id:
-                        source_nodegroup.parentnodegroup_id = (
-                            nodegroup_parent_node.source_identifier_id
-                        )
+        serialized_source_graph = JSONDeserializer().deserialize(
+            JSONSerializer().serialize(self)
+        )
```
jacobtylerwalls, line 2494: What does the round trip do? I would expect a model instance from a deserialize() method, but I get a dict.

arches/app/models/graph.py
```diff
@@ -2412,406 +2424,159 @@ def create_editable_future_graph(self):
...
+                serialized_nodegroup["nodegroupid"] = updated_nodegroup_id
 
-                    if future_edge.rangenode.source_identifier_id:
-                        future_edge.rangenode_id = (
-                            future_edge.rangenode.source_identifier_id
-                        )
+            updated_parent_nodegroup_id = node_id_to_node_source_identifier_id.get(
+                serialized_nodegroup["parentnodegroup_id"]
+            )
+            if updated_nodegroup_id:
```
jacobtylerwalls, line 2572: Is cardinality handled elsewhere?

arches/app/models/graph.py
```diff
@@ -2828,99 +2593,103 @@ def restore_state_from_serialized_graph(self, serialized_graph):
...
-                card_x_node_x_widget.pk
-                for card_x_node_x_widget in self.widgets.values()
-            ]
-        ).delete()
+        with transaction.atomic():
+            editable_future_graph = Graph.objects.filter(
+                source_identifier_id=self.pk
+            ).first()
+            if editable_future_graph:
+                editable_future_graph.delete()
```
jacobtylerwalls, line 2666: If you want to one-shot this, since we're not using the model instance.

Given that this is the Graph proxy, it probably does shave off all the `__init__()` queries.
```suggestion
            Graph.objects.filter(source_identifier_id=self.pk).delete()
```

tests/models/graph_tests.py
```diff
@@ -1481,3 +1271,1001 @@ def test_add_resource_instance_lifecycle(self):
...
+        edges_dict = {
+            "description": None,
+            "domainnode_id": "20000000-0000-0000-0000-100000000001",
+            "edgeid": "22200000-0000-0000-0000-000000000001",
+            "graph_id": cls.NODE_NODETYPE_GRAPHID,
+            "name": None,
+            "ontologyproperty": "http://www.cidoc-crm.org/cidoc-crm/P2_has_type",
+            "rangenode_id": "20000000-0000-0000-0000-100000000002",
+        }
+        models.Edge.objects.create(**edges_dict).save()
```
jacobtylerwalls, line 1405: I noticed you're calling save() after the create() call saves the object. Is that because there's custom save() logic on some of these models? Or are you guarding against the possibility that some of the models will have custom save() logic in the future?

tests/models/graph_tests.py
```diff
@@ -1481,3 +1271,1001 @@ def test_add_resource_instance_lifecycle(self):
...
+                serialized_updated_source_nodes,
+                serialized_editable_future_nodes,
+                [
+                    "graph_id",
+                    "nodeid",
+                    "nodegroup_id",
+                    "source_identifier_id",
+                ],
+            )
+        )
```
jacobtylerwalls: I'd suggest a tiny refactor to continue using assertEqual but just on the result of sorting/filtering. The reason being that `False is not True` is less useful than the output of assertEqual.

tests/models/graph_tests.py
```diff
@@ -1481,3 +1271,1001 @@ def test_add_resource_instance_lifecycle(self):
...
+        serialized_editable_future_edges = {
+            (
+                edge["source_identifier_id"]
+                if edge.get("source_identifier_id") not in [None, "None"]
+                else edge["edgeid"]
+            ): edge
+            for edge in serialized_editable_future_graph["edges"]
+        }
+
+        self.assertTrue(
```
jacobtylerwalls: Another nice to have would be putting these assertions under `with self.subTest(model=model_name)`, but I leave it to you if that would make the end result too indented/abstract. Just nicer to have all of the failures reported at once instead of find-one-fix-one-find-another-fix-another.

tests/models/graph_tests.py
```diff
@@ -1481,3 +1271,1001 @@ def test_add_resource_instance_lifecycle(self):
...
+                    "source_identifier",
+                    "source_identifier_id",
+                    "publication_id",
+                ],
+            )
+        )
+
+    def test_update_empty_graph_from_editable_future_graph(self):
+        source_graph = Graph.new(name="TEST RESOURCE", is_resource=True, author="TEST")
+        source_graph.save()
```
jacobtylerwalls: As a newer developer to this part of the code, tests like these help me understand the intended developer interface. Am I supposed to save() after calling new() because new() lacks some functionality? Would be nice to clarify with a comment here if so, or else open a follow-up ticket if the implementations need some attention.

tests/models/graph_tests.py
```diff
@@ -1481,3 +1271,1001 @@ def test_add_resource_instance_lifecycle(self):
...
+        node_count_after = models.Node.objects.count()
+        edge_count_after = models.Edge.objects.count()
+        card_count_after = models.CardModel.objects.count()
+        card_x_node_x_widget_count_after = models.CardXNodeXWidget.objects.count()
+
+        self.assertEqual(nodegroup_count_before, nodegroup_count_after)
+        self.assertEqual(node_count_before, node_count_after)
+        self.assertEqual(edge_count_before, edge_count_after)
+        self.assertEqual(card_count_before, card_count_after)
+        self.assertEqual(
```
jacobtylerwalls, line 2120: Nice. If we find corner cases we missed, it should be easy enough to slot them in here 👍

tests/models/graph_tests.py
```diff
@@ -1481,3 +1271,1001 @@ def test_add_resource_instance_lifecycle(self):
...
+        )
+
+    def test_update_from_editable_future_graph_does_not_affect_resources(self):
+        source_graph = Graph.new(name="TEST RESOURCE", is_resource=True, author="TEST")
+        source_graph.save()
+        editable_future_graph = source_graph.create_editable_future_graph()
+
+        nodegroup = models.NodeGroup.objects.create()
+        string_node = models.Node.objects.create(
+            graph=source_graph,
```
jacobtylerwalls, line 2131: Let's ensure there's a collector/root node:
```suggestion
            pk=nodegroup.pk,
            graph=source_graph,
```

tests/models/graph_tests.py
```diff
@@ -1481,3 +1271,1001 @@ def test_add_resource_instance_lifecycle(self):
...
+        updated_nodegroup_id = node.pk
+
+        models.NodeGroup.objects.create(
+            **{
+                "cardinality": "n",
+                "legacygroupid": "",
+                "nodegroupid": str(updated_nodegroup_id),
+                "parentnodegroup_id": None,
+            }
+        ).save()
```
jacobtylerwalls, line 2219: At first glance, it looks like this nodegroup is created with an arbitrary pk. If so, wouldn't that break the invariance it's supposed to have with the grouping node (same PK)? Or is that handled elsewhere in this test case?

arches/app/models/graph.py
```diff
@@ -2828,99 +2593,103 @@ def restore_state_from_serialized_graph(self, serialized_graph):
...
 
-        for serialized_nodegroup in serialized_graph["nodegroups"]:
-            for key, value in serialized_nodegroup.items():
-                try:
-                    serialized_nodegroup[key] = uuid.UUID(value)
-                except:
-                    pass
+            # ensures any resources that were related to the source graph are not deleted
+            self.pk = uuid.uuid4()
+            self.delete()
```
jacobtylerwalls: The chance of legit data loss from a UUID collision is infinitesimal, but I wonder if we should decompose `Graph.delete()` into `deleteRelatedObjects()` so we can call it without calling super().delete() and fiddling with the pk. By not hacking the pk, it will be a little clearer and hardened against other ORM internals (e.g., do we need to also fiddle with `._state.adding` and set it False, now that it's a nonexistent object?  Would rather not massage that...)

tests/models/graph_tests.py
```diff
@@ -1481,3 +1271,1001 @@ def test_add_resource_instance_lifecycle(self):
...
+
+        # the source_graph nodegroup and the editable_future_graph nodegroup have been deleted
+        self.assertEqual(nodegroup_count_before, models.NodeGroup.objects.count())
+
+    def test_can_update_graph_slug(self):
+        source_graph = Graph.new(name="TEST RESOURCE", is_resource=True, author="TEST")
+        source_graph.save()
+        editable_future_graph = source_graph.create_editable_future_graph()
+
+        editable_future_graph.slug = "test-resource"
```
johnatawnclementawn, line 2264: @chrabyrd Try setting the slug on the source graph to emulate a graph that already has a slug. I think this would be testing adding a slug to a graph that doesn't have a slug. This should cause the test fail. 

Sorry if this wasn't clear from my initial comment - the issue isn't so much that you can't _add_ a new slug, but graphs that already have a slug can't be updated.

jacobtylerwalls (changes requested):
Love the QoL improvements, and thanks for digging into this many-tentacled issue! A few questions for you.

tests/models/graph_tests.py
```diff
@@ -1481,3 +1271,1001 @@ def test_add_resource_instance_lifecycle(self):
...
+        editable_future_graph.save()
```
johnatawnclementawn: I was thinking about the scenarios in which I had this fail, and realized I was missing one critical detail - the source_graph should have a slug, but this also needs to be in conjunction with changes to the graph applied to the editable_future_graph. Here I've copied the node change from `test_update_node` to illustrate.

```suggestion
        source_graph = Graph.new(name="TEST RESOURCE", is_resource=True, author="TEST")
        source_graph.slug = "test-resource"
        source_graph.save()
        editable_future_graph = source_graph.create_editable_future_graph()

        editable_future_graph.append_branch(
            "http://www.cidoc-crm.org/cidoc-crm/P1_is_identified_by",
            graphid=self.NODE_NODETYPE_GRAPHID,
        )
        editable_future_graph.save()
```

If you run `python manage.py test tests.models.graph_tests --settings="tests.test_settings" -k test_can_update_graph_slug` now you should see what I see.

tests/models/graph_tests.py
chrabyrd: resolved! Looks like it was missing a str/uuid conversion in the graph validate method.

tests/models/graph_tests.py
chrabyrd: resolved! Looks like it was missing a str/uuid conversion in the graph validate method.

tests/models/graph_tests.py
chrabyrd: it shouldn't? `pk` == `nodegroupid` for the nodegroups table. Unless I'm misunderstanding what you're saying?

arches/app/models/graph.py
chrabyrd: but we also want to delete the graph. I'm unsure decomposing this will do anything for

```
graph = models.ForeignKey(GraphModel, db_column="graphid", on_delete=models.CASCADE)
```

on ResourceInstance.

arches/app/models/graph.py
chrabyrd: cardinality is de-facto handled because we're iterating `serialized_editable_future_graph["nodegroups"]`

arches/app/models/graph.py
chrabyrd: ya the round trip produces a dict of the serialized graph. At least a while ago ( and maybe still now ) for some models what it produces is different from calling the `serialize()` method on the model.

arches/app/models/models.py
chrabyrd: eh, 6-in-one. Do you feel strongly about this?

tests/models/graph_tests.py
jacobtylerwalls: No, my mistake, I glazed right over that.

arches/app/models/models.py
jacobtylerwalls: Nah.

arches/app/models/graph.py
jacobtylerwalls: Well from a code read, we're not deleting the graph, and I just checked the queries to be sure. We set the PK to a random value, so that super().delete() will shoot a bunch of blanks, firing off deletes on a bunch of tables that will never match on anything. Since we're not deleting the graph, I was thinking it would be clearer to decompose that out and avoid shooting the blanks:

```py
BEFORE:  a6cd3760-199b-4071-bd15-307fb04e53d2
AFTER:  de19d77d-4b59-41ae-bdee-aa7669cda961
```
Notice the deletes are for the random uuid (also why I was mentioning the very very rare risk of collision):
```
-> super(Graph, self).delete()
(Pdb) print(connection.queries)
[{'sql': 'SELECT "cards"."cardid" FROM "cards" WHERE "cards"."graphid" IN (\'de19d77d-4b59-41ae-bdee-aa7669cda961\'::uuid)', 'time': '0.003'}, {'sql': 'SELECT "edges"."edgeid" FROM "edges" WHERE "edges"."graphid" IN (\'de19d77d-4b59-41ae-bdee-aa7669cda961\'::uuid)', 'time': '0.002'}, {'sql': 'SELECT "graphs"."graphid" FROM "graphs" WHERE "graphs"."source_identifier" IN (\'de19d77d-4b59-41ae-bdee-aa7669cda961\'::uuid)', 'time': '0.001'}, {'sql': 'SELECT "graphs_x_published_graphs"."publicationid" FROM "graphs_x_published_graphs" WHERE "graphs_x_published_graphs"."graphid" IN (\'de19d77d-4b59-41ae-bdee-aa7669cda961\'::uuid)', 'time': '0.002'}, {'sql': 'SELECT "nodes"."nodeid" FROM "nodes" WHERE "nodes"."graphid" IN (\'de19d77d-4b59-41ae-bdee-aa7669cda961\'::uuid)', 'time': '0.001'}, {'sql': 'SELECT "resource_instances"."resourceinstanceid" FROM "resource_instances" WHERE "resource_instances"."graphid" IN (\'de19d77d-4b59-41ae-bdee-aa7669cda961\'::uuid)', 'time': '0.001'}, {'sql': 'DELETE FROM "functions_x_graphs" WHERE "functions_x_graphs"."graphid" IN (\'de19d77d-4b59-41ae-bdee-aa7669cda961\'::uuid)', 'time': '0.001'}, {'sql': 'DELETE FROM "resource_x_resource" WHERE ("resource_x_resource"."resourceinstancefrom_graphid" IN (\'de19d77d-4b59-41ae-bdee-aa7669cda961\'::uuid) OR "resource_x_resource"."resourceinstanceto_graphid" IN (\'de19d77d-4b59-41ae-bdee-aa7669cda961\'::uuid))', 'time': '0.001'}, {'sql': 'DELETE FROM "graphs_x_mapping_file" WHERE "graphs_x_mapping_file"."graphid" IN (\'de19d77d-4b59-41ae-bdee-aa7669cda961\'::uuid)', 'time': '0.001'}, {'sql': 'DELETE FROM "graphs" WHERE "graphs"."graphid" IN (\'de19d77d-4b59-41ae-bdee-aa7669cda961\'::uuid)', 'time': '0.001'}]
```

chrabyrd: I guess I'm talking about:

```
{'sql': 'DELETE FROM "graphs" WHERE "graphs"."graphid" IN (\'de19d77d-4b59-41ae-bdee-aa7669cda961\'::uuid)', 'time': '0.001'}
```

so that the current graph object gets deleted. I know the `Graph.delete` method manually handles cards, nodes, widgets, nodegroups, and edges in a way that wouldn't be affected by changing the graphid. `graph_x_published_graph` is interesting, I haven't checked that those are also deleted, but I will. Same with checking for strays with `functions_x_graphs` 

I'm not against the decomposition here, but at some point the graph itself needs to be deleted, so not I'm entirely sure that the decomposition will handle what we need it to.

And you're right about firing blanks, that's intentional for `resource_x_resource` and `resource_instances`.

jacobtylerwalls: But `de19d77d-4b59-41ae-bdee-aa7669cda961` isn't the current object, it's the random one, right?

BEFORE is self.pk, then
AFTER is after setting self.pk = uuid.uuid4()
```
BEFORE:  a6cd3760-199b-4071-bd15-307fb04e53d2
AFTER:  de19d77d-4b59-41ae-bdee-aa7669cda961
```

chrabyrd: ah. yeah I get it 😞 . Def something to handle for

tests/models/graph_tests.py
chrabyrd: ah -- this was just boilerplate copied from the `setUpTestData` on `GraphTests`. I'd like to keep them in sync if possible but I'm open to changing both as a followup

arches/app/models/graph.py
chrabyrd: ah -- this is this way because this calls the graph proxy model's `delete` method, while `Graph.objects.filter(source_identifier_id=self.pk).delete()` would skip it.

arches/app/models/graph.py
chrabyrd: Mind if this is a followup? Feels out of scope and I want to get this PR in

arches/app/models/migrations/11570_harden_editable_future_graphs.py
```diff
@@ -0,0 +1,38 @@
...
+        ),
+        migrations.AlterField(
+            model_name="resource2resourceconstraint",
+            name="resourceclassto",
+            field=models.ForeignKey(
+                blank=True,
+                db_column="resourceclassto",
+                null=True,
+                on_delete=django.db.models.deletion.CASCADE,
+                related_name="resxres_contstraint_classes_to",
```
chiatt, line 34: It looks like there's an extra 't' in constraint

chiatt (changes requested):

arches/app/models/graph.py
```diff
@@ -1689,51 +1689,106 @@ def update_permissions(self, serialized_graph):
...
+                    ].values():
+                        group_permissions.extend(group_permission_list)
+
+                    group_permission_nodegroups = models.NodeGroup.objects.filter(
+                        pk__in={
+                            group_permission["object_pk"]
+                            for group_permission in group_permissions
+                        }
+                    )
+                    group_permission_nodgroup_id_to_nodegroup = {
```
chiatt: nodegroup is missing an 'e' here: `group_permission_nodgroup_id_to_nodegroup`

arches/app/models/migrations/11570_harden_editable_future_graphs.py
chrabyrd: wow, looks like this goes back to migration 0001, and is a current issue in `models.py`. Mind if this is a followup?

chiatt (changes requested):

arches/app/models/migrations/11570_harden_editable_future_graphs.py
chiatt: No problem - a followup sounds like the way to go.

tests/models/graph_tests.py
jacobtylerwalls: sounds good

arches/app/models/graph.py
jacobtylerwalls: Alright, got it. It would be nice if these were vanilla model methods, but that's not a matter for today.

arches/app/models/graph.py
jacobtylerwalls: All in all I just want to grok what's going on -- happy to punt if I can get to a place of understanding first.

My understaning is:
- we set a random uuid to avoid deleting the graph when calling graph.delete()
- but we want to call the proxy model delete() method to delete the related instances that are still held in memory in the custom `self.nodes` and similar attributes, which fire off all kinds of cascade deletes on spatial views and stuff like that.

But the code comment says we're trying _not_ to delete related items. Hence my confusion

 Firing off a wasteful delete query with a random uuid for the graph itself is no big deal, just another performance penalty of having this functionality on the proxy models. Mostly I'm just unsure what this is here for.

It also leaves the model instance in an unstable state, since it now has a bogus pk. If you save() later, you'll create a duplicate row. I guess I am a little concerned about that.

This all goes away if we just delete whatever we want to delete like
```py
models.NodeGroup.objects.filter(node__graph=self).delete()
```
and with that the tests pass, but my guess is you are doing something more generic/robust, I'm just not understanding what it is yet. Sorry if I'm just not getting it.

jacobtylerwalls: (As part of moving the test from #11841 over to this branch on my local to make see if it passes, I did stumble over the bogus pk thing -- my code that did `Graph.objects.get(pk=source_graph.pk)` to try to refresh from the db failed because source_graph.pk was bogus after calling this method).

arches/app/media/js/viewmodels/graph-settings.js
```diff
@@ -15,9 +15,12 @@ define([
...
-                resource.isRelatable = ko.observable(resource.is_relatable);
-            }).filter(resource => !resource.graph.source_identifier_id);
+            return _.each(
```
chiatt, line 18: Certainly out of scope for this PR, but it would be nice to remove relatable_resources and ResourceXResourceConstraints as part of this: https://github.com/archesproject/arches/issues/11844

arches/app/models/graph.py
chrabyrd: no yeah, that comment is bad. we _want_ to delete associated entities. I've since abstracted out that bit from the Graph delete method and everything seems happy. Thanks for the persistence here ❤️

chrabyrd: ah yeah now I remember. We want to delete the associated entities, but keep the associated `ResourceInstances`, which get deleted if the graph gets deleted. Hence the uuid hack. IMO abstracting the thing was the way to go 👍

chiatt (approved):

tests/models/graph_tests.py
```diff
@@ -974,7 +974,7 @@ def test_derive_card_values(self):
             "http://www.cidoc-crm.org/cidoc-crm/P1_is_identified_by",
             graphid=graph.graphid,
         )
-        resource_graph.save()
+        resource_graph.save()  # `new()` doesn't persist to the database
```
jacobtylerwalls: Not a blocker, but to me it looks like new() _does_ persist? Is it that save() does extra magic for related objects?

arches/app/models/graph.py
```diff
@@ -724,6 +724,13 @@ def save(self, validate=True, nodeid=None):
         return self
 
     def delete(self):
+        self.delete_associated_entities()
+        super(Graph, self).delete()
```
jacobtylerwalls, line 728: While here, would you mind returning this value instead of swallowing it? #preexistingissue

arches/app/models/graph.py
```diff
@@ -724,6 +724,13 @@ def save(self, validate=True, nodeid=None):
...
+
+    def delete_associated_entities(self):
```
jacobtylerwalls, line 732: thanks 🥇

jacobtylerwalls (approved):
Thanks, this is great.

johnatawnclementawn (approved):

tests/models/graph_tests.py
chrabyrd: lol not sure what I was thinking there. Ya will change 👍

chrabyrd: turns out those were superfluous. Deleted!

PR #11614, opened 2024-11-11
Add `NodeGroup.grouping_node` #11613

### LINKED ISSUE #11613: Add `grouping_node` column to NodeGroup
> It just occurred to me could add a column `is_root_node` to the `Node` model ... so that we don't have to run these searches.

_Originally posted by @jacobtylerwalls in https://github.com/archesproject/arches/pull/11596#discussion_r1825085436_
***
Noticed while dev'ing #11595 that it would be more convenient to have a database column for the node at the root of a node's nodegroup--and have it be a self-referencing FK that returns `Node` instances--to avoid having to run a search that gathers siblings nodes and compares them to the nodegroup_id in Python.

jacobtylerwalls:
### Types of changes

### Description of Change
**Before**
From a node, to find the grouping node of that node's nodegroup, you would need to prefetch the sibling nodes and compare the results to the id of the nodegroup, or else just issue another Node.objects.get() to individually fetch the wanted node.

**After**
~There is now a self-referencing foreign key to the other node that serves as the nodegroup root.~
Per discussion, there is now a OneToOneField on the NodeGroup model pointing to the root node for the nodegroup.

Example query:
```py
In [22]: from django.db import connection, reset_queries

In [23]: reset_queries()

In [24]: ngs = NodeGroup.objects.filter(grouping_node__isnull=False).select_related("grouping_node").distinct()

In [25]: for ng in ngs:
    ...:     print(ng.grouping_node.alias)
    ...: 
app_names
basic_search_settings
sparql_endpoint_providers
system_settings
map_settings
project_extent
map_zoom
saved_searches
mapbox
temporal_search
analytics
search_results_grid

In [26]: len(connection.queries)
Out[26]: 1
```

### Issues Solved
Closes #11613

### Checklist

#### Ticket Background


arches/app/models/migrations/11613_node_nodegroup_root.py
```diff
@@ -0,0 +1,35 @@
...
+            field=models.ForeignKey(
+                blank=True,
+                null=True,
+                on_delete=django.db.models.deletion.CASCADE,
+                related_name="sibling_nodes",
+                related_query_name="sibling_node",
+                to="models.node",
+            ),
+        ),
+        migrations.RunPython(set_nodegroup_root, migrations.RunPython.noop),
```
jacobtylerwalls: I probably need to do a slightly fancier update/rewind on the serialized graphs here.

jacobtylerwalls: This occurred to me when I saw a failure in the UI (that may have just been my db in a bad state), but grepping for where we do `Node(**serialized_node)` showed that this should be pretty-okay as is, but I did add a little safety in [0304a86](https://github.com/archesproject/arches/pull/11614/commits/0304a860373763e1282850df1297056bf443c65d) if not.

chiatt: After reversing migrations, I found the `nodegroup_root_id` property on the serialized graph caused this error: `TypeError: Node() got unexpected keyword arguments: 'nodegroup_root_id'` here:

https://github.com/archesproject/arches/blob/0304a860373763e1282850df1297056bf443c65d/arches/app/models/graph.py#L224-L225

arches/app/models/models.py
```diff
@@ -907,15 +915,21 @@ def __init__(self, *args, **kwargs):
     def clean(self):
         if not self.alias:
             Graph.objects.get(pk=self.graph_id).create_node_alias(self)
+        self.grouping_node_id = self.nodegroup_id
```
jacobtylerwalls: Based on our conversation yesterday I decided to run this unconditionally instead of when empty. That would make the changes in a9b2f32162e7c7b15206b9728658f95a392a0b88 not strictly necessary, but I left them in. (They'll probably change anyway after #11609.)

arches/app/models/graph.py
```diff
@@ -251,6 +251,11 @@ def check_default_configs(default_configs, configs):
 
                 self.populate_null_nodegroups()
 
+    @classmethod
+    def check(cls, **kwargs):
+        """Bypass GraphModel.check()."""
```
chrabyrd: Is this bypassing the `GraphModel.check()` method? I see it's being `super`-ed below

chrabyrd (commented):
Overall lgtm 🚀 ! One question in the body of code. 

On a more meta note, if it's possible to add the `Node` relation to `NodeGroup` instead of this change and still power what you're trying to do I think there'd be an additional win gained there. There have been quite a few times I've needed to get the node from the nodegroup and end up running to the db again 💀 

If that can't be accommodated, how do you feel about changing `grouping_node` to `nodegroup_node`? idk `grouping_node` isn't really referenced elsewhere in code and IMO it's more descriptive to say what the thing actually is.

jacobtylerwalls:
Thanks for taking a look. I'm going to pump the brakes for a couple days since I just got a few 👍 's on [Slack](https://archesproject.slack.com/archives/C18GGP4RX/p1731605950940079) for the idea of just factoring out a subquery into a queryset method. If you have to remember `.select_related('grouping_node')` then `.with_grouping_node()` calling out to a subquery isn't much worse, as long as it performs roughly the same?

***
> If that can't be accommodated, how do you feel about changing grouping_node to nodegroup_node? idk grouping_node isn't really referenced elsewhere in code and IMO it's more descriptive to say what the thing actually is.

I first had `nodegroup_root` and called this branch `node_for_nodegroup`, so I was thinking along the same lines, but @chiatt encouraged me to consider grouping node. @chiatt, any thoughts?

chrabyrd:
> Thanks for taking a look. I'm going to pump the brakes for a couple days since I just got a few 👍 's on [Slack](https://archesproject.slack.com/archives/C18GGP4RX/p1731605950940079) for the idea of just factoring out a subquery into a queryset method. If you have to remember `.select_related('grouping_node')` then `.with_grouping_node()` calling out to a subquery isn't much worse, as long as it performs roughly the same?
> 
> > If that can't be accommodated, how do you feel about changing grouping_node to nodegroup_node? idk grouping_node isn't really referenced elsewhere in code and IMO it's more descriptive to say what the thing actually is.
> 
> I first had `nodegroup_root` and called this branch `node_for_nodegroup`, so I was thinking along the same lines, but @chiatt encouraged me to consider grouping node. @chiatt, any thoughts?

Cool cool! One thing re slack convo: `with_grouping_node_alias()` -> `with_grouping_node()` ? What if I wanted other info from the nodegroup node besides alias? And if you're going that route would it be a big lift to add `with_node()` to the nodegroup? For the same win mentioned earlier 😄 

And ++ for having the same thoughts about naming. Hopefully that makes it in 🤞

jacobtylerwalls:
Yeah I think if we want other stuff from the node then it should go back to an FK, but I can definitely explore putting the column on the nodegroup. I'll give it some thought.

chiatt:
> Thanks for taking a look. I'm going to pump the brakes for a couple days since I just got a few 👍 's on [Slack](https://archesproject.slack.com/archives/C18GGP4RX/p1731605950940079) for the idea of just factoring out a subquery into a queryset method. If you have to remember `.select_related('grouping_node')` then `.with_grouping_node()` calling out to a subquery isn't much worse, as long as it performs roughly the same?
> 
> > If that can't be accommodated, how do you feel about changing grouping_node to nodegroup_node? idk grouping_node isn't really referenced elsewhere in code and IMO it's more descriptive to say what the thing actually is.
> 
> I first had `nodegroup_root` and called this branch `node_for_nodegroup`, so I was thinking along the same lines, but @chiatt encouraged me to consider grouping node. @chiatt, any thoughts?

I don't have a strong opinion on this. My preference is `grouping_node` because it describes the node's role in the group. It's also a term already used in the label based graph tests. `nodegroup_node` would be fine, but it seems a bit awkward to me. I realize that's pretty subjective, so maybe others would like to weigh in (@apeters, @robgaston?)

jacobtylerwalls:
Also, `root` could be confused with the root node of the graph or the grouping node of the root nodegroup (in graph.py, this is defined as "(the first nodegroup that doesn't have a parentnodegroup)"

arches/app/models/graph.py
jacobtylerwalls: Good Q. It is, because I passed `GraphModel` as the first arg to super instead of self. This calls GraphModel's super, not self's.

jacobtylerwalls:
> On a more meta note, if it's possible to add the Node relation to NodeGroup instead of this change and still power what you're trying to do I think there'd be an additional win gained there. There have been quite a few times I've needed to get the node from the nodegroup and end up running to the db again 💀

I gave it some thought, and although it makes intuitive sense to put it on NodeGroup, I think there are some considerations going the other way:
- the relation would have to be nullable given the order we create nodegroups before nodes, plus the fact that nodegroups without nodes are allowed at a db level (although we do have a `manage.py validate` command to delete them). I'd rather it be not nullable.
- with the information in a different table, you can no longer have a database check constraint to enforce that the FK has a certain kind of relationship (e.g. you could enforce that the grouping_node pk == the pk of the nodegroup, but I don't think you could enforce that every node has a grouping node, e.g. that every not-root-node has a nodegroup with a grouping node set).
- In a simpler application, one could use django signals to keep tables in sync, but you know how much code doesn't go through django these days.

> get the node from the nodegroup and end up running to the db again 💀

This is actually my main motivation, and now there's a way to do that when fetching the data. On the other proposal (column on NodeGroup) you'd have to select_related() the grouping node when fetching data, so it's not like the developer gets a permission slip to not think about this ;)

```py
# Make sure to get grouping nodes before you need them
In [49]: with_grouping_nodes = NodeGroup.objects.filter(node__gt=0).prefetch_related('node_set__grouping_node').distinct()

In [50]: for ng in with_grouping_nodes:
    ...:     print(ng.node_set.all()[0].grouping_node.alias)
```

apeters:
> I don't have a strong opinion on this. My preference is `grouping_node` because it describes the node's role in the group. It's also a term already used in the label based graph tests. `nodegroup_node` would be fine, but it seems a bit awkward to me. I realize that's pretty subjective, so maybe others would like to weigh in (@apeters, @robgaston?)

I agree with @chiatt.  I think `grouping_node` describes its function nicely.

apeters:
I would also say that from a "what makes the most sense" perspective (and not to throw a monkey wrench 🐵  🔧  into this discussion) that having this on the "Nodegroup" table makes the most sense.  So, in that case you'd end with something like this `Node.nodegroup.rootnode` to get to the "grouping_node" (but in this case, I think "rootnode" makes most sense when it's hanging off of the Nodegroup table).

Now pragmatically speaking this may not net us less db calls if that's what we're trying to achieve with this, but it does feel very natural.

jacobtylerwalls:
Yeah, it makes good sense to move it. I can handle the missing safety with a Django system check.

jacobtylerwalls:
@chiatt ready for another look :-)

arches/app/models/models.py
```diff
@@ -768,6 +787,27 @@ class Meta:
...
 
+    @classmethod
+    def check(cls, **kwargs):
+        errors = super().check(**kwargs)
+        if kwargs["databases"]:
+            errors.extend(cls._check_root_node())
+        return errors
+
+    @classmethod
+    def _check_root_node(cls):
```
chiatt: This check ends up being run prior to migrations which throws an error because rootnode is not yet available.

```
django.db.utils.ProgrammingError: column node_groups.rootnodeid does not exist
LINE 1: ..."cardinality", "node_groups"."parentnodegroupid", "node_group"
```

arches/app/models/migrations/11613_nodegroup_root_node.py
```diff
@@ -0,0 +1,69 @@
...
+        published_graphs = PublishedGraph.objects.all()
+        for published_graph in published_graphs:
+            for node_dict in published_graph.serialized_graph["nodegroups"]:
+                node_dict.pop("root_node_id", None)
+        PublishedGraph.objects.bulk_update(published_graphs, ["serialized_graph"])
+
+    operations = [
+        migrations.AddField(
+            model_name="nodegroup",
+            name="root_node",
```
chrabyrd: I'd argue just `node` here. `root_node` is less clear and can have different connotations.

arches/app/models/migrations/11613_nodegroup_root_node_constraints.py
```diff
@@ -0,0 +1,33 @@
...
+    ]
+
+    operations = [
+        migrations.AddConstraint(
+            model_name="node",
+            constraint=models.CheckConstraint(
+                condition=models.Q(
+                    ("istopnode", True), ("nodegroup__isnull", False), _connector="OR"
+                ),
+                name="has_nodegroup_or_istopnode",
```
chrabyrd: I though nodes _always_ had a nodegroup? And the difference was that if a node is a top node its nodegroup shares an ID with it.

arches/app/models/migrations/7787_relational_data_model.py
```diff
@@ -9,7 +9,7 @@ class Migration(migrations.Migration):
 
     operations = [
         migrations.RunSQL(
-            """
+            r"""
```
chrabyrd, line 12: ? . If this isn't a mistake I assume it's not kosher to change a migration from a previous release?

arches/app/models/models.py
jacobtylerwalls: Yeah, I don't like telling people to `--skip-checks` with every migrate call, so I'm going to back out that release note advice and look into skipping this on the migrate command.

chrabyrd (changes requested):
I'm OOO today but saw these interesting updates and wanted to stick my nose in 👃

arches/app/models/migrations/11613_nodegroup_root_node_constraints.py
jacobtylerwalls: Top nodes don't, for better or worse!

arches/app/models/migrations/7787_relational_data_model.py
jacobtylerwalls: This resolves the SyntaxWarning on python 3.12+. Eventually it will get elevated to an error, so we have to do something eventually. [more than anyone wants to know](https://adamj.eu/tech/2022/11/04/why-does-python-deprecationwarning-invalid-escape-sequence/)

arches/app/models/migrations/11613_nodegroup_root_node.py
jacobtylerwalls: Happy to hold this until everyone's happy with the name, but `node` is the query name for a related item from `node_set`, so I don't think we can overload that. I think we need to land on something.

arches/app/models/migrations/11613_nodegroup_root_node_constraints.py
chrabyrd: I will double-check tomorrow morning, but I'm relatively sure they do? This was an issue with promoting/demoting nodes into their own nodegroups in the editable future graphs that eventually pushed me to do it all in JSON. Please let me double-check this before merging!

jacobtylerwalls: There's a lil [diagram](https://arches.readthedocs.io/en/stable/developing/reference/data-model/#resource-model-overview) showing a free floating top node ☁️

jacobtylerwalls: but @chiatt and I looked in the db, rather than just trusting the diagram ;)

chrabyrd: oh fun! I just checked on my local and see that it makes an extra nodegroup. We should touch base early tomorrow to verify each other's steps and see where the disconnect is 🔌

arches/app/models/migrations/7787_relational_data_model.py
chrabyrd: oh huh, interesting!

arches/app/models/migrations/11613_nodegroup_root_node_constraints.py
jacobtylerwalls: > Please let me double-check this before merging!

of course, no rush by any means! enjoy your time off!

arches/app/models/models.py
jacobtylerwalls: I'll just remove this and add it back if I get a positive response here before Django 5.2:
🤞 https://code.djangoproject.com/ticket/35920

arches/app/models/migrations/11613_nodegroup_root_node_constraints.py
chrabyrd: whoops yeah, just verified and you're right 👍

arches/app/models/models.py
```diff
@@ -726,6 +739,13 @@ def __init__(self, *args, **kwargs):
     class Meta:
         managed = True
         db_table = "node_groups"
+        constraints = [
+            models.CheckConstraint(
+                condition=Q(grouping_node=models.F("pk"))
+                | Q(grouping_node__isnull=True),
```
chrabyrd, line 745: sorry to potentially rehash but it's been a bit since I've looked at this. What's the condition for a NodeGroup to exist without a Node?

arches/app/models/models.py
```diff
@@ -874,15 +894,18 @@ def __init__(self, *args, **kwargs):
     def clean(self):
         if not self.alias:
             Graph.objects.get(pk=self.graph_id).create_node_alias(self)
+        if self.pk == self.source_identifier_id:
```
chrabyrd, line 897: This feels like part of a larger conversation. I would think that we should _never_ get to this point, and that behavior is entirely encapsulated in the `update_from_editable_future_graph` logic. 

And if for some reason it shouldn't be, then this should be added to all entities that have a `source_identifier`

BUT

This may be entirely moot with the editable_future_graph refactor that's in review, so perhaps the conversation should move there 🤔

arches/app/models/graph.py
```diff
@@ -367,6 +367,7 @@ def add_node(self, node, nodegroups=None):
                 node.nodegroup = self.get_or_create_nodegroup(
                     nodegroupid=node.nodegroup_id
                 )
+                node.nodegroup.grouping_node_id = node.nodegroup_id
```
chrabyrd: should this instead be part of `get_or_create_nodegroup`?

chrabyrd (commented):
looking good, just a few questions 👍

arches/app/models/models.py
jacobtylerwalls: Purely for convenience when creating them -- we have to create one (NodeGroup) and then the other (Node), unless we want to require that everybody create them in a single transaction and then run this as a deferred constraint, but that's a breaking change I didn't want to impose.

There's a validation check in manage.py validate that checks for nodeless nodegroups. Apparently we're still creating some on setup_db:
```
Arches integrity report
Prepared by Arches 8.0.0a0 on Thu Jan  2 15:41:29 2025

        Error   Rows    Fixable?        Description

PASS    1005    0       N/A             Nodes with ontologies found in graphs without ontologies

FAIL    1012    1       Yes             Node Groups without contained nodes

PASS    1013    0       N/A             Node Groups without a grouping node

PASS    1014    0       N/A             Publication missing for language
```

arches/app/models/models.py
jacobtylerwalls: I'm happy to sound this out, but I think it's out of scope, see next comment...

arches/app/models/models.py
```diff
@@ -874,15 +894,18 @@ def __init__(self, *args, **kwargs):
...
+            self.source_identifier_id = None
 
     def save(self, **kwargs):
         if not self.alias:
-            self.clean()
             add_to_update_fields(kwargs, "alias")
             add_to_update_fields(kwargs, "hascustomalias")
         if self.pk == self.source_identifier_id:
-            self.source_identifier_id = None
```
jacobtylerwalls, line 884: ...(continuing from last comment)

This is where the logic moved from. It was already in save(), and we already had a mixture of calling-clean-within-save and doing-more-cleaning-inside-save after that. If we want to move to model methods only, I'm all for it, and then we can stop overriding save()!

Anyway -- I can set this back if you don't like it, it's not strictly necessary, but while I was in the area I thought a little cleaner separation between clean and save would be good. Thoughts?

arches/app/models/graph.py
jacobtylerwalls: I like it 🧠

arches/app/models/models.py
chrabyrd: yeah no it's definitely a larger conversation but for sure out of scope for this PR 👍 . I won't block on it 😄

chiatt (commented):
I keep getting the following error when attempting to open an existing model in the graph designer or when trying to save a tile (on save the error is logged, but it doesn't prevent the save):
```
"arches/app/models/graph.py", line 224, in __init__
    nodes = [models.Node(**node_slug) for node_slug in node_slugs]
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "arches/app/models/models.py", line 890, in __init__
    super(Node, self).__init__(*args, **kwargs)
TypeError: Node() got unexpected keyword arguments: 'grouping_node_id'
```
And this error when trying to publish an existing model:
```
g_node_id'
  File "/arches/app/views/graph.py", line 688, in post
    source_graph.update_from_editable_future_graph()
  File "/arches/app/models/graph.py", line 2650, in update_from_editable_future_graph
    source_node.nodegroup.grouping_node_id = source_node.nodegroup_id
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'grouping_node_id'
```
I double checked my migrations, and they seem to have run successfully.

jacobtylerwalls:
For the first error, that looks like a graph publication dating from an earlier version of this PR when `grouping_node_id` was on `Node`. Unfortunately I think you'll need to edit it by hand. (Perhaps you could adjust `def remove_grouping_node` in the migration to also act on Nodes, and then run the migration backwards and forwards.)

For the second error, nice catch -- I'm going to remove the code that apparently I [thought](https://github.com/archesproject/arches/pull/11614#discussion_r1842359258) I didn't need two months ago 😄 . Graph publication works for me now.

arches/app/models/migrations/11613_nodegroup_grouping_node.py
```diff
@@ -0,0 +1,69 @@
...
+        nodegroups_with_nodes = NodeGroup.objects.filter(node__gt=0).distinct()
+        for nodegroup in nodegroups_with_nodes:
+            nodegroup.grouping_node_id = nodegroup.pk
+        NodeGroup.objects.bulk_update(nodegroups_with_nodes, ["grouping_node_id"])
+
+        PublishedGraph = apps.get_model("models", "PublishedGraph")
+        published_graphs = PublishedGraph.objects.all()
+        for published_graph in published_graphs:
+            for node_dict in published_graph.serialized_graph["nodes"]:
+                node_dict["grouping_node_id"] = node_dict["nodegroup_id"]
```
chiatt: I think the error that I am seeing is caused here. The "grouping_node_id" is getting assigned back to the node. Shouldn't it be assigned to the nodegroup?

jacobtylerwalls: Thanks, should have retested this. Fixed.

chiatt: Yep, that did the trick!

One other thing. I noticed that if I create a new graph and then create a new node, my new node does not have the groupingnodeid assigned. That may be fine because it's a nodegroup of 1, so there really isn't a group to be concerned about. However, if that first node is semantic, and I add another node, then the first semantic node still doesn't have a groupingnodeid. This seems like it could be an issue because it really does have a child node in its group.

image

jacobtylerwalls: Ouch, taking a look.

jacobtylerwalls: [383bfb2](https://github.com/archesproject/arches/pull/11614/commits/383bfb24f2c1c05850c32b8080c76eb0459c2d15)

chiatt (approved):
Looks good to me!

jacobtylerwalls:
Thanks for the reviews!

arches/app/models/migrations/7787_relational_data_model.py
jacobtylerwalls: Hello! So this change wasn't totally safe without auditing the backslashes in this huge function. Some postgres versions are apparently more forgiving, but when running Arches today on PG 17, I got an error on this migration. I'll open an issue & PR to resolve the SyntaxWarning more surgically.

PR #11626, opened 2024-11-15
[WIP] 10502 search layer

### LINKED ISSUE #10502: Discussion: Unpaginated Search Results and Supporting a Dynamic Search Results Layer on Map Filter component
The default behavior in the Arches SearchView is to paginate the search request to the appropriate django view, ``search.py``. While this UX is natural to a paginated UI component like the Search Results component, **it presents a fundamental constraint to an inherently unpaginated UI component like the Map Filter**. The available UX in the Map Filter component enables a user to _either_ see every single resource instance of a given resource model via the resource layer, a page-limit’s worth of search results from their query, or a combination of both.

Not having an unpaginated Map UX is a huge missed opportunity for Arches given that the map filter highlights the first-class status of mappable data, and given the robust power of the combinability of the other search filter components. Why hasn’t it already been done? Well there are technical challenges, of course.

We implemented the desired functionality in HistoricPlacesLA and there were a few core changes to make it run within a timely manner:

**Enable Backend Caching of Search results**
For the largest queries by result volume, many queries were repeated by a lot of the users. For “Anonymous” users in particular, given that they all had identical data permissions, it made sense to cache one set for all of them. The cache key would have to capture the whole query and also include some snapshot of the database in case search results had been mutated. We used the count from the EditLog model. Determining when these caches timeout (outside of the snapshot being embedded in the cache key) is a matter of preference and should be its own setting at the project level. Without embedding backend caching, the demand on Elasticsearch could increase significantly given the desired load (see next point).

**Enable Unpaginated Search results**
To paint all the features on the map, clustered or in whatever layer you style, a search query would need to request all search results instead of a single page’s worth. This means the default query in SearchView should be able to interrogate the instance settings for whether to paginate every search query. This means the search view should be able to handle the case of no paging-filter present on the query object.

**Enable Frontend & Backend Caching of Search Result Geometries**
Because the FeatureCollections of points and polygons tend to eat up compute during JSON conversion, there’s a lot of utility in caching a “master” lookup of geometry to lighten the payload of any subsequent query’s search results. In HistoricPlacesLA, a compressed geometry lookup for 50,000 mixed point/polygons  can be cached to the browser at about 3mb. There is a hard limit of 5mb since browsers don’t support larger browser caches and a real tradeoff to consider regarding whether it’s better to frontload first to save time later vs waiting a few seconds longer for those larger geom-heavy search results. This should be its own optional flow to be configured by the implementer, but it is a useful pattern that merits being available rather than reinvented.

**Create a Map Layer to cluster Search results**
The behavior of a dynamic search map differs from both the search-result layer and the resource-layer stylings. A functional map layer for unpaginated search results should by default have clustering behavior similar to the resource-layer. There isn’t a compelling reason *not* to enable UI-supported styling for this overlay map-layer, but in the near term it could be styled directly in the json similar to any other overlay.

**Example Implementation: HPLA**
HistoricPlacesLA has over 50,000 resources with geometries. For HPLA, our overall strategy was to cache both: A) a “master” set of search result geometries on the backend and B) literally every search result set. When a user lands on the home page (index.html), the page would request the cached geometry lookup from the backend via search_results and then compress and cache this set in the client’s browser localStorage cache. Typical wait time for this to complete was 3-8 seconds. Half of the runtime goes to compression of the geom lookup.

When the user proceeds to Search, their search query returns an unpaginated result set sans any point or polygons (best case from the backend search result cache). On the frontend, each search result gets compared against the frontend geom cache and if matched, gets “hydrated” with a cached geom. The ligher payload (even when no cached results exist) makes a noticeable difference in load time.

The end product is a dynamic, unpaginated map query that does what you expect it to. While the default functionality in SearchView runs in less than 1 second, for any project that has even half of the geoms HPLA has, they payoff of a dynamic map will likely merit the extra 2-3 seconds spent waiting.

I think there’s a strong case to be made that this functionality deserves to be _“opt-in”_ in core Arches.

whatisgalen:
### Types of changes

### Description of Change

dependencies: #11629 #11628

### Issues Solved

Closes #10502 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @ 

### Further comments

- caching
- ask the server to do computation, not end user machine
-

### DIFF
--- a/arches/app/media/js/viewmodels/map.js
+++ b/arches/app/media/js/viewmodels/map.js
@@ -317,6 +317,10 @@ define([
                 map.fitBounds(ko.unwrap(params.bounds), boundingOptions);
             }
 
+            // If searchQueryId is already available, add the search layer
+            if (self.searchQueryId()) {
+                self.addSearchLayer(self.searchQueryId());
+            }
         });
 
         this.bounds = ko.observable(ko.unwrap(params.bounds) || arches.hexBinBounds);
--- a/arches/app/media/js/viewmodels/map.js
+++ b/arches/app/media/js/viewmodels/map.js
@@ -12,6 +12,261 @@ define([
     const viewModel = function(params) {
         var self = this;
 
+        const searchLayerIds = [
+            'searchtiles-unclustered-polygon-fill',
+            'searchtiles-unclustered-point',
+            'searchtiles-clusters',
+            'searchtiles-clusters-halo',
+            'searchtiles-cluster-count',
+            'searchtiles-unclustered-polypoint'
+        ];
+        const searchLayerDefinitions = [
+            {
+              "id": "searchtiles-unclustered-polygon-fill",
+              "type": "fill",
+              "paint": {
+                "fill-color": "#fa6003",
+                "fill-opacity": 0.3,
+                "fill-outline-color": "#fa6003"
+              },
+              "filter": [
+                "==",
+                "$type",
+                "Polygon"
+              ],
+              "source": "search-layer-source",
+              "source-layer": "search_layer",
+              "minzoom": 10,
+              "tolerance": 0.75
+            },
+            {
+              "id": "searchtiles-unclustered-point",
+              "type": "circle",
+              "paint": {
+                "circle-color": "#fa6003",
+                "circle-radius": 6,
+                "circle-opacity": 1
+              },
+              "filter": [
+                "!",
+                [
+                  "has",
+                  "point_count"
+                ]
+              ],
+              "source": "search-layer-source",
+              "source-layer": "search_layer"
+            },
+            {
+              "id": "searchtiles-clusters",
+              "type": "circle",
+              "paint": {
+                "circle-color": "#fa6003",
+                "circle-radius": [
+                  "step",
+                  [
+                    "get",
+                    "point_count"
+                  ],
+                  10,
+                  100,
+                  20,
+                  750,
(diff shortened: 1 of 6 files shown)


robgaston:
Some off-the-cuff thoughts on a quick read of this code:
- the basic pattern here holds water *I think* but will likely need work in several areas
- am I reading properly that `GeoTileGridAgg` is being used to achieve geospatial clusters?  This is problematic as an approach to clustering for a number of reasons:
    - it doesn't seek to group nearby items when they are nearly coincident, but rather simply bins geometries into predefined cells (for example, two very spatially close records might not be clustered together if they are separated by a cell boundary).
    - Additionally, clusters should truly be generated across tile boundaries for much the same reasons (tiles are arbitrary pre-defined cells which may separate very spatially close geometries), which a tile based aggregation will never account for.
    - Grid cell centroids are almost always not the best proximate location for a clustered set of geometries
 - I would argue that the service class, route, etc. should not bear the name "search_layer" as layers are a presentation construct and these are defining data services.  The names here should be more descriptive of the API/response.
 - In my opinion there should not be a pre-defined layer in the map JS viewmodel and instead it should just set up the data source which can then be used in overlays freely.  This allows search layers to be defined, customized, and controlled by the end user as we do all other overlays.

chiatt:
Closing as stale

PR #11627, opened 2024-11-15
adds streaming test API

chrabyrd:
NOT FOR MERGING!

This is a proof of concept. Go here: http://localhost:8000/test_streaming

to see that wsgi supports streaming data ( at least locally ). Hopefully @aarongundel can test this on a production server to close the loop 🤞


aarongundel:
All seems good @chrabyrd - works just fine on our current prod setup.

PR #11638, opened 2024-11-19
Add tile sort order to bulk data manager

### LINKED ISSUE #11408: Feature: Capture tile sort order during bulk import modules
For cardinality-n cards, tiles can be sorted at the card level for a given resource. This is important for things like resource descriptors which derive the resource display name from the tile with the lowest sort order if using a cardinality-n card. 

Currently, the Bulk Data Manager import modules do not have a way of explicitly controlling the tile sort order for tiles being imported. The current behavior is to initialize tiles with a sort order of 0, and leave any sorting to the front end (if any). 

The primary goal should be to enable the ability to control sort order while creating new tiles through the bulk data importers.

chiatt:
### Types of changes

### Description of Change
Add support for tile sortorder in bulk data manager

### Issues Solved

Closes #11408

### Checklist


johnatawnclementawn (approved):
LGTM 👍🏼  - added a small change to the tile excel tests just to check that sortorder in the import exhibit is being respected on import.

PR #11342, opened 2024-08-15
adds more descriptive error if no published graph #10079

chrabyrd:
### Types of changes

-   [x] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Issues Solved

Closes #10079 

### Checklist

-   I targeted one of these branches:
    - [x] dev/7.6.x (under development): features, bugfixes not covered below
    - [ ] dev/7.5.x (main support): regressions, crashing bugs, security issues, major bugs in new features
    - [ ] dev/6.2.x (extended support): major security issues, data loss issues
-   [ ] I added a changelog in arches/releases
-   [ ] I submitted a PR to arches-docs (if appropriate)
-   [ ] Unit tests pass locally with my changes
-   [ ] I added tests that prove my fix is effective or that my feature works
-   [ ] My test fails on the target branch

#### Accessibility Checklist

[Developer Guide](https://arches.readthedocs.io/en/stable/developing/advanced/accessibility/)

| Topic            | Changed | Retested |
| ---------------- | ------- | -------- |
| Color contrast   |         |          |
| Form fields      |         |          |
| Headings         |         |          |
| Links            |         |          |
| Keyboard         |         |          |
| Responsive Design|         |          |
| HTML validation  |         |          |
| Screen reader    |         |          |

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

PR #11646, opened 2024-11-21
Rewrite model check as validation check re #10079

jacobtylerwalls:
### Types of changes

### Description of Change
When trying my hand at writing a django system check in #10079, I missed a pretty major [caveat](https://docs.djangoproject.com/en/5.1/topics/checks/#writing-your-own-checks), which is that if you use the database, you need to do something like:

https://github.com/archesproject/arches/blob/0f23a6cbc78a093e5a08cf2ab2abf2c542915df4/arches/app/models/models.py#L619-L620

... to avoid db hits during commands that don't generally expect a database to be configured. (My earlier workaround was was to run subsets of checks in different places, but that was kind of a code smell. This PR removes some of those workarounds.)

But to start observing `kwargs["databases"]` means that these checks only run on:
```
- manage.py migrate
- manage.py check --database default
```

So, given how infrequently this would be run, better to just implement this in our validation framework and call it out in the release notes?

#### Testing instructions
- Create and publish a graph
- Enable an additional system language, without republishing any graphs
- manage.py validate
- expect errros
- manage.py validate --fix
- expect errrors fixed
- ensure branches do not trigger the check
- ensure languages for which a publication exists but is not currently in settings.LANGUAGES doesn't alter the comparison and cause the check to unexpectedly pass

### DIFF
--- a/arches/app/models/models.py
+++ b/arches/app/models/models.py
@@ -613,39 +613,6 @@ class GraphModel(models.Model):
 
         super(GraphModel, self).save(*args, **kwargs)
 
-    @classmethod
-    def check(cls, **kwargs):
-        errors = super().check(**kwargs)
-        errors.extend(cls._check_publication_in_every_language())
-        return errors
-
-    @classmethod
-    def _check_publication_in_every_language(cls):
-        errors = []
-        system_languages = {lang[0] for lang in settings.LANGUAGES}
-
-        for graph in (
-            cls.objects.filter(publication__isnull=False)
-            .select_related("publication")
-            .prefetch_related("publication__publishedgraph_set")
-        ):
-            languages_with_a_publication = {
-                published_graph.language_id
-                for published_graph in graph.publication.publishedgraph_set.all()
-            }
-            missing_languages = system_languages - languages_with_a_publication
-            if missing_languages:
-                errors.append(
-                    checks.Error(
-                        "This graph is not published in all enabled languages.",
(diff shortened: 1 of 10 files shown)


arches/management/commands/validate.py
```diff
@@ -116,6 +120,36 @@ def handle(self, *args, **options):
             ),
             fix_action=DELETE_QUERYSET,
         )
+        self.check_integrity(
+            check=IntegrityCheck.PUBLICATION_MISSING_FOR_LANGUAGE,  # 1014
+            queryset=(
+                models.GraphModel.objects.filter(
+                    isresource=True,
```
jacobtylerwalls, line 127: The prior version of this was also checking branches when it shouldn't.

arches/app/models/models.py
```diff
@@ -613,39 +613,6 @@ def save(self, *args, **kwargs):
 
         super(GraphModel, self).save(*args, **kwargs)
 
-    @classmethod
```
jacobtylerwalls, line 616: Pretty sure this deletion is the main reason the coverage check is failing.

jacobtylerwalls:
> So, given how infrequently this would be run, better to just implement this in our validation framework and call it out in the release notes?

Some more info about the original motivation here: https://github.com/archesproject/arches/pull/11342#issuecomment-2491470794

aarongundel (approved):
LGTM

PR #11648, opened 2024-11-22
Fix duplicate results in `whatisthis` command

jacobtylerwalls:
### Types of changes

### Description of Change
Proxy models were included in the count of models in the `whatisthis` command, causing 2x the number of hits.


whatisgalen (approved):
tested, looks good!

PR #11653, opened 2024-11-25
Mark passwords as sensitive variables #11652

### LINKED ISSUE #11652: Filter out passwords from error reports in production
We can start masking passwords in error reports in production by using Django's `@sensitive_variables()` [decorator](https://docs.djangoproject.com/en/5.1/howto/error-reporting/#django.views.decorators.debug.sensitive_variables) in our auth views.

jacobtylerwalls:
### Types of changes

### Description of Change
Limit occurrences of passwords in error reports (if configured) in production.

### Issues Solved

Closes #11652

### DIFF
--- a/arches/app/views/auth.py
+++ b/arches/app/views/auth.py
@@ -19,7 +19,6 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 import base64
 import io
 
-from django.http import response
 from arches.app.utils.external_oauth_backend import ExternalOauthAuthenticationBackend
 import qrcode
 import pyotp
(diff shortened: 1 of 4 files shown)


jacobtylerwalls:
You can test with the standard django error view but, since DEBUG=True doesn't bother masking, the least realistic but fatest way to configure this to test would be hack this line in your django install to return True:

https://github.com/django/django/blob/857b1048d53ebf5fc5581c110e85c212b81ca83a/django/views/debug.py#L182

Then you should be able to throw errors inside any of these views and see the masking inside the django error view (but again, more realistically, we'd be talking about logs or emails to admins in prod environments, not the django error view):


Screenshot 2024-11-25 at 2 11 17 PM

chiatt (changes requested):
This is really great. Could you make the variables sensitive on the post method in the SignupView in `views/auth.py` as well?

arches/app/views/auth.py
```diff
@@ -223,6 +222,10 @@ def get(self, request):
             },
         )
 
+    @method_decorator(
+        sensitive_variables(),
+        sensitive_post_parameters(),
+    )
```
chiatt: This only seems to work when the methods are within an iterator (as I see you've done elsewhere)

chiatt (changes requested):

chiatt (approved):

Issue #11660, opened 2024-11-27
Unexpected SystemSettings interaction causing failure to run via wsgi.py


jacobtylerwalls:
@ekansa reported on the [forum](https://community.archesproject.org/t/arches-via-docker-installation-issues/2603/9?u=jacobtylerwalls) a failure to run an arches project via `wsgi.py`, i.e. not run with `runserver`. The TL;DR is that we have the default value of `APP_NAME` being stored in a system settings tile and overriding the project settings file, only under wsgi.py.

Traced it to this unexpectedly evaluating True:
https://github.com/archesproject/arches/blob/f3390eff2ff5fe8eca899d2b3f199868fcea1575/arches/urls.py#L789

... because wsgi.py loads system settings from the database:
https://github.com/archesproject/arches/blob/f3390eff2ff5fe8eca899d2b3f199868fcea1575/arches/wsgi.py#L39

... and we save the value "Arches" for `APP_NAME` in the database during `setup_db`, which is going to be incorrect for all cases (only applies to running "naked" core arches like a project):
https://github.com/archesproject/arches/blob/f3390eff2ff5fe8eca899d2b3f199868fcea1575/arches/db/system_settings/Arches_System_Settings.json#L153-L156
https://github.com/archesproject/arches/blob/f3390eff2ff5fe8eca899d2b3f199868fcea1575/arches/management/commands/setup_db.py#L234-L236

***
@chrabyrd and I didn't notice this when developing f569e097f02c1973a9b0c216bf9902a6f19d6e8c and related features because we were running the devserver, but arguably this is an orthogonal bug.

We have other cases where we need the value of APP_NAME to be correct:

https://github.com/archesproject/arches/blob/f3390eff2ff5fe8eca899d2b3f199868fcea1575/arches/settings_utils.py#L161-L162

***
I reproduced on 7.6.3 with Eric's instructions:

- clone opencontext/arches-via-docker
- checkout the 'local' branch
- cp edit_dot_env .env
- make DEBUG=False (in .env) and also change the SHOW_LANGUAGE_SWITCH to be True
- commented out `DEPLOY_HOST` in settings_local since there wasn't a fallback
- prepend `npm i typescript@5.6.3` to the build development command to work around https://github.com/vuejs/language-tools/issues/5018. (This will hopefully disappear in a few days.)
- visit the server at 0.0.0.0:8004

#### Next steps

[...]

PR #11661, opened 2024-11-28
Show the application name in the splash screen #6531

### LINKED ISSUE #6531: Application Name Setting does not apply to splash page
When I set the application name in settings, the browser tab updates with the name on pages except for the splash page.  Would be good it also updated the splash page given that is the intended landing page.

MikeGriniezakis:
### Types of changes

### Description of Change

The Application Name setting was not taken into account when rendering the splash screen.
This PR changes the default "Arches" text to use the Application Name setting.

### Issues Solved

Closes #6531 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/views/main.py
+++ b/arches/app/views/main.py
@@ -34,6 +34,7 @@ def index(request):
             "main_script": "index",
             "active_page": "Home",
             "app_title": settings.APP_TITLE,
+            "app_name": settings.APP_NAME,
             "copyright_text": settings.COPYRIGHT_TEXT,
             "copyright_year": settings.COPYRIGHT_YEAR,
             "app_version": settings.APP_VERSION,
--- a/arches/app/views/main.py
+++ b/arches/app/views/main.py
@@ -16,14 +16,17 @@ You should have received a copy of the GNU Affero General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>.
 """
 
-import re
-import urllib.request, urllib.error, urllib.parse
+import urllib.error
+import urllib.parse
+import urllib.request
 from urllib.parse import urlparse
+
+from django.conf import settings as project_settings
+from django.http import HttpResponse, HttpResponseNotFound
+from django.shortcuts import render
+
 from arches import __version__
 from arches.app.models.system_settings import settings
-from django.shortcuts import render, redirect
-from django.http import HttpResponseNotFound, HttpResponse, HttpResponseRedirect
-from django.utils import translation
 
 
 def index(request):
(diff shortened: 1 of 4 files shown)


arches/app/views/main.py
```diff
@@ -34,6 +34,7 @@ def index(request):
             "main_script": "index",
             "active_page": "Home",
             "app_title": settings.APP_TITLE,
+            "app_name": settings.APP_NAME,
```
chiatt: This doesn't seem to use the project settings, but rather only the Arches settings, so this value remains 'Arches'.

chiatt (changes requested):
Thanks for submitting this @MikeGriniezakis. It seems that the project settings aren't being used in the index view in main.py, but rather the Arches settings, so that will need to be fixed to get this working.

The template changes would also need to be made to the project template so that they are reflected in new projects. You can find that here: arches/arches/install/arches-templates/project_name/templates/index.htm.

arches/app/views/main.py
MikeGriniezakis: I updated the code to use the project settings when getting the APP_NAME in the commit d57bf025b5cd939458cc3a486e502695c3458de9

arches/app/views/main.py
```diff
@@ -16,14 +16,17 @@
...
 """
 
-import re
-import urllib.request, urllib.error, urllib.parse
+import urllib.error
+import urllib.parse
+import urllib.request
 from urllib.parse import urlparse
+
+from django.conf import settings as project_settings
```
jacobtylerwalls, line 24: Actually, I don't think we should be importing settings via the standard Django import until we address the behavior difference between dev and prod fleshed out in #11660.

arches/app/views/main.py
jacobtylerwalls: Isn't "Arches" the expected value if there has not been a system settings tile saved for this setting?

MikeGriniezakis: Upon running the setup_db, the value "Arches" is inserted as the default value into the Application Name setting. With my previous commit, the splash page only changes when the Application Name is changed through the settings.

chiatt:
Hi @MikeGriniezakis - I really appreciate your work on this. Your effort has been really valuable in exposing the importance of addressing #11660 as @jacobtylerwalls has pointed out. In light of this fact, and the fact that the index page is something that generally gets overwritten during project development anyhow, I don't think we should invest more time in this, so I'm going to close this PR and issue. Again, thanks very much for your help.

PR #11667, opened 2024-12-03
Add system dark mode detection for Vue apps #11624

### LINKED ISSUE #11624: Query browser setting for dark mode and provide it with base-root.htm
> I believe it ended up being that some of this logic gets moved to core. Essentially the logic to derive the user scheme preference and set it on the `<html>` element. I believe part of that work was to set a small entry in localStorage so the userPreference/toggleState is persisted through refreshes.

The rest -- supporting multiple scheme preferences and adding a column to the `UserProfile` table -- were punted

_Originally posted by @chrabyrd in https://github.com/archesproject/arches-lingo/issues/91#issuecomment-2477168769_

jacobtylerwalls:
### Types of changes

### Description of Change
A vue app no longer has to query for the system dark mode itself, and can depend on a user's persisted dark mode preference being hydrated (although a vue app's toggle will need to persist the choice, see example in archesproject/arches-lingo#91).

### Issues Solved

Closes #11624

### Checklist

#### Ticket Background

### DIFF
--- a/arches/app/templates/base-root.htm
+++ b/arches/app/templates/base-root.htm
@@ -35,6 +35,16 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
         gtag('config', '{{app_settings.GOOGLE_ANALYTICS_TRACKING_ID}}');
     </script>
     {% endif %}
+    <script>
+        const darkModeToggleState = localStorage.getItem("arches-dark");
+        if (
+            darkModeToggleState === "true" ||
+            (darkModeToggleState === null &&
+                window.matchMedia("(prefers-color-scheme: dark)").matches)
+        ) {
+            document.documentElement.classList.add("arches-dark");
+        }
+    </script>
     <title>
         {% block title %}{% endblock title %}
     </title>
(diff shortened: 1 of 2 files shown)


arches/app/templates/base-root.htm
```diff
@@ -35,6 +35,16 @@
         gtag('config', '{{app_settings.GOOGLE_ANALYTICS_TRACKING_ID}}');
     </script>
     {% endif %}
+    <script>
+        const darkModeToggleState = localStorage.getItem("arches-dark");
```
jacobtylerwalls, line 39: This class name is hard coded, because I don't see a way to read it dynamically from the DEFAULT_THEME js file. If project developers have a different use case, as a last resort, this template could be overridden.

chrabyrd (approved):
looks good! Once the toggle lives outside of lingo we'll probably need to revisit this. I can imagine keying value in localStorage on the username as well -- unless we go with adding a row to the userprofile table -- so that there isn't odd behavior on shared computers.

jacobtylerwalls:
Thanks for the review!

PR #11671, opened 2024-12-03
Handle custom selectors in dark mode query re #11624

jacobtylerwalls:
Follow-up to #11667, see [suggestion](https://github.com/archesproject/arches-lingo/pull/91#pullrequestreview-2476580442) to liberalize this. Agree we should avoid hard-coding the selector.

### DIFF
--- a/arches/app/media/js/utils/create-vue-application.js
+++ b/arches/app/media/js/utils/create-vue-application.js
@@ -13,7 +13,7 @@ import { createGettext } from "vue3-gettext";
 import arches from 'arches';
 import { DEFAULT_THEME } from "@/arches/themes/default.ts";
 
-export default async function createVueApplication(vueComponent, themeConfiguration) {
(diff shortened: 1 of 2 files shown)


arches/app/media/js/utils/create-vue-application.js
```diff
@@ -40,8 +40,19 @@ export default async function createVueApplication(vueComponent, themeConfigurat
         });
 
         const app = createApp(vueComponent);
+        const darkModeClass = themeConfiguration.theme.options.darkModeSelector.substring(1);
+        const darkModeStorageKey = `arches.${darkModeClass}`;
```
jacobtylerwalls, line 44: This will create a local storage key `arches.arches-dark`, which is good in case we end up with `arches.my-not-namespaced-dark-selector`?

chrabyrd (approved):
let's go 🪨 🇮🇹 !

PR #11673, opened 2024-12-04
Fix overflowing advanced styling panel in Map Layer Manager

### LINKED ISSUE #11649: Advanced styling panel overflows page for some resource models
Tested on Arches for HERs, Arches 7.5

On some models, the advanced styling code fails to properly format, resulting in a single long line of code. This causes the advanced styling panel to expand far beyond the page's normal boundaries. 

https://github.com/user-attachments/assets/cf02f7bf-3744-48c7-8881-b40497220033

The following error, sometimes (but not always) appears when toggling advanced search, but this does not seem to be determining  whether the json formats correctly or not. 

Advanced styling error

An incomplete solution is to add `lineWrapping: true` to line 177 of geojson-feature-collection.htm. This prevents the overflow, but it would better to fix whatever is sometimes stopping the json from formatting properly to begin with, as the result still isn't very readable:

Wrapped, unformatted json

CWDamm-Kint:
### Types of changes

### Description of Change
When the advanced styling config is not properly formatted, the advanced styling panel overflows the screen. There is an option for codemirror to wrap text, which stops this from happening. 

### Issues Solved
Closes #11649

### Checklist

    - [X] dev/7.6.x (main support): regressions, crashing bugs, security issues, major bugs in new features

#### Accessibility Checklist

#### Ticket Background
*   Found by: @CWDamm-Kint 
*   Tested by: @CWDamm-Kint
*   Designed by: @CWDamm-Kint

### Further comments

NA

### DIFF
--- a/arches/app/templates/views/components/datatypes/geojson-feature-collection.htm
+++ b/arches/app/templates/views/components/datatypes/geojson-feature-collection.htm
@@ -195,7 +195,7 @@
                                 <div id="style-panel" class="tab-pane fade active in">
                                     <div class="panel-body" style="padding: 0px;">
                                         <div class="row">
-                                            <div data-bind="codemirror: { value: config.advancedStyle, mode: { name: 'javascript', json: true }, theme: 'monokai' }"></div>
+                                            <div data-bind="codemirror: { value: config.advancedStyle, mode: { name: 'javascript', json: true }, theme: 'monokai', lineWrapping: true }"></div>
                                         </div>
                                     </div>
                                 </div>


chiatt:
@CWDamm-Kint, I've been unable to reproduce this bug in 7.6. Could you narrow down the conditions that cause it?

CWDamm-Kint:
@chiatt, I've updated the issue with a clip from 7.6 while pasting in improperly formatted json. Arguably, this is somewhat unlikely as Arches produces properly formatted json for new models by default, and you would likely then just edit what's there. But perhaps worth having the fix as a back up if something similar did recur?

chiatt:
Thanks @CWDamm-Kint - That's helpful. I'm wondering if this fix, although it certainly makes the data more readable, isn't masking the actual problem. My guess is that if you make a change to the style, like a change of a fill color, that change won't be reflected in the map. I suspect that the data is getting saved as a json-like string rather than json. If that's the case, then the fix that's really needed is ensuring that the data gets converted to true json on save.

samuelscandrett:
> Thanks @CWDamm-Kint - That's helpful. I'm wondering if this fix, although it certainly makes the data more readable, isn't masking the actual problem. My guess is that if you make a change to the style, like a change of a fill color, that change won't be reflected in the map. I suspect that the data is getting saved as a json-like string rather than json. If that's the case, then the fix that's really needed is ensuring that the data gets converted to true json on save.

It sounds like there are two issues at play here: the styling causing an overflow, and the incorrect AfHER data; but the overflow issue is caused by the json-like string in the AfHER data.

I can trace the json string down to the AfHER models, here: https://github.com/archesproject/arches-her/blob/a8e9144fc622921e02e3e0fb28a48bc711b58851/arches_her/pkg/graphs/resource_models/Application%20Area.json#L8184.  I can create this as an issue in AfHER.

The line wrapping seems like a small enough addition to fix the visual side of this issue, but the data is truly the culprit.

aj-he:
> there are two issues at play here: the styling causing an overflow, and the incorrect AfHER data; but the overflow issue is caused by the json-like string in the AfHER data.

@SDScandrettKint - This string will have been generated by the system on exporting the graph. The content of the string passes validation as JSON if you remove the escape characters. This seems to imply that there could be an underlying issue with how this json-like string is loaded from the model file? Or, how the string is loaded into codemirror?

samuelscandrett:
The AfHER json string must be failing to JSON stringify somewhere then?  

The AfHER `advancedStyle` JSON is functional.  Changing some of the fill colours is reflected on the map.

I've also noticed that any string can be added to the advanced styling box, it is not strict on JSON.  At first, this was tested by replacing the AfHER model `advancedStyle` line with regular text, but it looks as if this can also be done through the UI.  
As an example for a long string, I pasted one of the paragraphs from archesproject.org into the advanced styling and it produced the same issue (see below).

https://github.com/user-attachments/assets/dae8a6c0-d518-4d65-a837-7374a6fec1d1


EDIT: console logging `config.advancedStyle` returns a string always, regardless of if it is JSON formatted in the advanced style box.  Should we be converting the value to JSON somewhere?

chiatt (approved):
If a user modifies the Advanced Styling in the UI, the style is saved as a formatted string: 

image

It's not very pretty, but codemirror can load this and it render it as formatted 'json'. The issue for HER is that the tab and newline characters have all been removed, so codemirror is rendering the data exactly as expected - as a single line. 

I tested Advanced Styling with properly formatted json by exporting a graph, cleaning up the data and reloading it. The data loaded as valid json in the database, but codemirror threw an error when I tried to load it in Advanced Styling. I suspect the issue isn't with codemirror, but with how it's being used. Unfortunately a fix for this (so that only valid json is managed) might mean cleaning up data that is not formatted as valid json and most likely means a migration to do it. For the time being @CWDamm-Kint's solution addresses the problem of being unable able to edit Advanced styling in cases such as Arches for HERs.

samuelscandrett:
@chiatt I have raised a new ticket with this issue that you encountered - I'll try reproduce.

I think I encountered another issue from looking into this bug - changing advanced styling config doesn't reflect the changes in the UI user-input boxes e.g. changing polygon paint colour.  Now this could be due to me changing the wrong value, but it might need to be looked into.

chiatt:
> I think I encountered another issue from looking into this bug - changing advanced styling config doesn't reflect the changes in the UI user-input boxes e.g. changing polygon paint colour. Now this could be due to me changing the wrong value, but it might need to be looked into.

The two different types of styling are managed separately and there wasn't any intent to keep them in sync, so it's actually working as designed. I believe it's possible for users to make changes to the advanced styling that simply are not going to apply to the user-input boxes, so trying to keep them in sync could be very problematic.

samuelscandrett:
> > I think I encountered another issue from looking into this bug - changing advanced styling config doesn't reflect the changes in the UI user-input boxes e.g. changing polygon paint colour. Now this could be due to me changing the wrong value, but it might need to be looked into.
> 
> The two different types of styling are managed separately and there wasn't any intent to keep them in sync, so it's actually working as designed. I believe it's possible for users to make changes to the advanced styling that simply are not going to apply to the user-input boxes, so trying to keep them in sync could be very problematic.

Good to know, thanks for the info.

PR #11675, opened 2024-12-04
Remove memory leaks from GetFrontendI18NData

jacobtylerwalls:
### Types of changes

### Description of Change
Avoid leaving file handles open (increases memory usage).

Noticed while profiling the report view for a client that is prioritizing performance.

We may want to consider caching this view (by requested language) in the future, since this data is very unlikely to change often in a mature deployment, and file I/O can really add drag.


apeters (approved):
nice catch!

apeters:
I'm not going to worry about coverage decreasing.  If you round to 2 decimal places it stays the same!  😉

PR #11687, opened 2024-12-13
Fix enabling multiple languages if the `APP_NAME` in the db was "Arches" #11660

### LINKED ISSUE #11660: Unexpected SystemSettings interaction causing failure to run via wsgi.py

jacobtylerwalls:
### Types of changes

### Description of Change
setup_db creates a tile for the system setting `APP_NAME` with a value of "Arches". If the application was running under wsgi.py, then it's likely that the arches system settings object fetched tiles, thus merging the APP_NAME setting tile into the django settings, causing the check against the django settings in urls.py to unexpectedly evaluate True, breaking the ability of a project to enable multiple languages. This wasn't caught in development because the dev workflow doesn't run wsgi.py and thus update_from_db() never ran.

#### Two fixes:
- avoid checking `APP_NAME` in core urls.py. Instead use the same check that's working well in projects to determine "is this project the project being served"
- ~merge system settings into the django settings in more cases, not just under wsgi.py but also during development. This was discussed at a recent standup. Fix the docstring that implies that this is not wanted.~ EDIT: will defer this fix, caused issues with `setup_db`.

#### Testing instructions
- To reproduce the original issue, checkout the first commit (or undo its reversion, since it's no longer part of this PR), and you should be able to reproduce #11660 by adjusting your project level settings to enable the language switcher.
- Then, checkout the second commit, the issue should be fixed.

### Issues Solved

Closes #11660

### Checklist

#### Ticket Background
*   Found by: @ekansa

### DIFF
--- a/arches/app/models/system_settings.py
+++ b/arches/app/models/system_settings.py
@@ -22,9 +22,6 @@ from django.core.exceptions import ImproperlyConfigured
 from arches.app.models import models
 
 
-tiles_loaded = False
-
-
 class SystemSettings(LazySettings):
     """
     This class can be used just like you would use settings.py
@@ -36,12 +33,15 @@ class SystemSettings(LazySettings):
         settings.SEARCH_ITEMS_PER_PAGE
 
         # will list all settings
-        print(settings)
+        print settings
 
     """
 
     SYSTEM_SETTINGS_RESOURCE_MODEL_ID = "ff623370-fa12-11e6-b98b-6c4008b05c4c"
     RESOURCE_INSTANCE_ID = "a106c400-260c-11e7-a604-14109fd34195"
+
+    def __init__(self, *args, **kwargs):
+        super(SystemSettings, self).__init__(*args, **kwargs)
 
     def __str__(self):
         ret = []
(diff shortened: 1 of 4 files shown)


aarongundel:
@jacobtylerwalls the tests seem pretty angry here.  I'm guessing related?

jacobtylerwalls:
I'll back-burner that fix for bringing the dev experience into alignment with prod. was Worth a shot but too hot for a patch.

chiatt (approved):

PR #11715, opened 2025-01-06
Convert ANALYSIS_COORDINATE_SYSTEM_SRID to int in map_filter

### LINKED ISSUE #11714: Error in search map filter if ANALYSIS_COORDINATE_SYSTEM_SRID is string rather than int
I discovered (by accident) that if the settings.py ANALYSIS_COORDINATE_SYSTEM_SRID value is set to a string rather than an integer, the map filter can't be used due to the following error:

image

samuelscandrett:
### Types of changes

### Description of Change

Simple fix to convert ANALYSIS_COORDINATE_SYSTEM_SRID to int.

Not sure if this fix is sufficient, or whether I should go with the approach of below, just in case the settings value for ANALYSIS_COORDINATE_SYSTEM_SRID is incorrect and can't be converted to an int (as this fix could cause another problem).
```
    try:
        analysis_coord_system_srid = int(settings.ANALYSIS_COORDINATE_SYSTEM_SRID)
    except Exception:
        analysis_coord_system_srid = 3857 # revert to default value
```

### Issues Solved

Closes #11714 

### Checklist

#### Accessibility Checklist

#### Ticket Background

### DIFF
--- a/arches/app/search/components/map_filter.py
+++ b/arches/app/search/components/map_filter.py
@@ -62,6 +62,13 @@ def _buffer(geojson, width=0, unit="ft"):
     except Exception:
         width = 0
 
+    try:
+        int(settings.ANALYSIS_COORDINATE_SYSTEM_SRID)
+    except Exception:
+        raise ValueError(
+            "Settings ANALYSIS_COORDINATE_SYSTEM_SRID value is not a number."
+        )
+
     if width > 0:
         if unit == "ft":
             width = width / 3.28084
--- a/arches/app/search/components/map_filter.py
+++ b/arches/app/search/components/map_filter.py
@@ -73,7 +73,7 @@ def _buffer(geojson, width=0, unit="ft"):
                 4326)""",
                 (
                     geom.hex.decode("utf-8"),
-                    settings.ANALYSIS_COORDINATE_SYSTEM_SRID,
+                    int(settings.ANALYSIS_COORDINATE_SYSTEM_SRID),
                     width,
                 ),
             )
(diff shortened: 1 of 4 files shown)


chiatt (changes requested):
Could you include an entry in the changelog?

Also, regarding your comment about using 3857 if the SRID isn't an int - if it can't be converted to an integer, then I think we want it to break rather than do the calculation with a different SRID. Using 3857 instead of the SRID the user thinks they are using might cause unexpected behavior.

samuelscandrett:
@chiatt Changelog entry added in [94f34cc](https://github.com/archesproject/arches/pull/11715/commits/94f34cc549b737d744174e1d41a177b5e13522ce).

I agree with the reasons why the search should break rather than use the default SRID value.

Also in the commit, I added a raise error, as currently the error produced can come across as unhelpful:
image

The new error prevents the search and informs the user that their value isn't a number because cannot be converted to int:
image

This can be tested by setting ANALYSIS_COORDINATE_SYSTEM_SRID to a value like "test".

Does this seem like a helpful improvement?

samuelscandrett:
I see coverage has decreased because of this - should I write a test for it?

chiatt:
> I see coverage has decreased because of this - should I write a test for it?

That would be great. It should be an easy one since @whatisgalen has already blazed that trail. You could just copy his `test_spatial_search_by_point_buffered` test, add the following descriptor, and confirm that you get an error.

```@override_settings(ANALYSIS_COORDINATE_SYSTEM_SRID ="Not an SRID")```

samuelscandrett:
@chiatt Test added [fec5be1](https://github.com/archesproject/arches/pull/11715/commits/fec5be12b058b856aec8bb7d55a67252ecd5d725) :) 

and conflicts resolved ([7772749](https://github.com/archesproject/arches/pull/11715/commits/7772749678d73256e93b665f9223b564f6a2eaba))

chiatt:
Sorry @SDScandrettKint - Could you move your changelog entry to the 7.6.7 release notes?

chiatt (approved):

PR #11717, opened 2025-01-07
11716 - Fix usage of resolve(reverse()) to check if FORCE_SCRIPT_NAME…

### LINKED ISSUE #11716: resolve(reverse()) doesn't work when using WSGI / FORCE_SCRIPT_NAME
For WSGI applications hosted behind Apache in a sub-path,`resolve(reverse(<view name>))` doesn't work. This is currently used in `app/search/search_export.py` (where the issue was found) and in `app/etl_modules/save.py`.

The `reverse` function prepends the URL with the value of settings.FORCE_SCRIPT_NAME, and `resolve()` can't find it. 

For our implementation we have a reverse proxy directing traffic from `https://hostname/bc-fossil-management` to the Arches WSGI application. We have the FORCE_SCRIPT_NAME set to `/bc-fossil-management/`. All requests through the reverse proxy function as expected, but when running an export via Celery the search URL can't be found. The cause for this is that `reverse("search_results")` returns `/bc-fossil-management/search/results` but the path expected by the URL resolver is `/search/results`.

Fixed in our instance by change the following:
```
        func, args, kwargs = resolve(reverse("search_results"))
```
to
```
        results_path = reverse("search_results")
        if results_path.startswith(get_script_prefix()):
           results_path = results_path.replace(get_script_prefix(),"/")
        func, args, kwargs = resolve(results_path)
```

Looks like this was introduced in this commit: https://github.com/archesproject/arches/commit/a4d11bde0d77c6a99e84c28621286ecd61b99adc

The above seems to work however there may be a cleaner solution.

bferguso:
### Types of changes

### Description of Change
The `resolve(reverse(<view name))` pattern doesn't work if the FORCE_SCRIPT_NAME is setting is used when hosting behind a reverse proxy in a sub-path. This checks to see if the `FORCE_SCRIPT_NAME` is used to and removes the prefix to correct the issue.

### Issues Solved

Closes #11716

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: Self Funded
*   Found by: @bferguso
*   Tested by: @bferguso
*   Designed by: @bferguso

### DIFF
--- a/arches/app/etl_modules/save.py
+++ b/arches/app/etl_modules/save.py
@@ -7,7 +7,7 @@ from django.core.validators import URLValidator
 from django.db import connection
 from django.http import HttpRequest
 from django.utils.translation import gettext as _
-from django.urls import reverse, resolve
+from django.urls import reverse, resolve, get_script_prefix
 from arches.app.models.system_settings import settings
 from arches.app.utils.index_database import index_resources_by_transaction
 import logging
@@ -187,7 +187,10 @@ def get_resourceids_from_search_url(search_url, user=None):
     params = parse_qs(urlsplit(search_url).query)
     for k, v in params.items():
         request.GET.__setitem__(k, v[0])
-    func, args, kwargs = resolve(reverse("search_results"))
+    search_results_path = reverse("search_results")
+    if search_results_path.startswith(get_script_prefix()):
+        search_results_path = search_results_path.replace(get_script_prefix(), "/")
+    func, args, kwargs = resolve(search_results_path)
     kwargs["request"] = request
     response = func(*args, **kwargs)
     results = json.loads(response.content)["results"]["hits"]["hits"]
--- a/arches/app/search/search_export.py
+++ b/arches/app/search/search_export.py
@@ -27,9 +27,8 @@ from django.contrib.gis.geos import GeometryCollection, GEOSGeometry
 from django.core.files import File
 from django.core.files.storage import default_storage
 from django.utils.translation import gettext as _
-from django.urls import reverse, resolve
+from django.urls import reverse, resolve, get_script_prefix
 from arches.app.models import models
-from arches.app.models.system_settings import settings
 from arches.app.datatypes.datatypes import DataTypeFactory
 from arches.app.utils.flatten_dict import flatten_dict
 from arches.app.utils.betterJSONSerializer import JSONDeserializer
@@ -166,7 +165,10 @@ class SearchResultsExporter(object):
 
     def export(self, format, report_link):
         ret = []
-        func, args, kwargs = resolve(reverse("search_results"))
+        search_results_path = reverse("search_results")
+        if search_results_path.startswith(get_script_prefix()):
+            search_results_path = search_results_path.replace(get_script_prefix(), "/")
+        func, args, kwargs = resolve(search_results_path)
         kwargs["request"] = self.search_request
         search_res_json = func(*args, **kwargs)
         if search_res_json.status_code == 500:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
     "arcgis2geojson==2.0.0",
     "celery==5.3.6",
     "defusedxml==0.7.1",
-    "Django>=4.2.17,<5.0.0",
+    "Django>=4.2.18,<5.0.0",
     "django-celery-results==2.5.1",
     "django-cors-headers==4.2.0",
     "django-guardian==2.4.0",
--- a/releases/7.6.5.md
+++ b/releases/7.6.5.md
@@ -11,7 +11,7 @@
 ```
 Python:
     Upgraded:
-        Django == 4.2.17 (or <5.0.0)
+        Django == 4.2.18 (or <5.0.0)
 JavaScript:
     Upgraded:
         none
--- a/releases/7.6.5.md
+++ b/releases/7.6.5.md
@@ -4,7 +4,7 @@
 
 - Fix Activity Stream id value [#11727](https://github.com/archesproject/arches/issues/11727)
 - Prevent Arches templates from overriding App templates [#11731]https://github.com/archesproject/arches/pull/11731
-
+- Fix search export when `FORCE_SCRIPT_NAME` is used #[11716](https://github.com/archesproject/arches/issues/11716)
 
 ### Dependency changes:
 
--- a/tests/search/search_export_tests.py
+++ b/tests/search/search_export_tests.py
@@ -14,7 +14,7 @@ from arches.app.utils.skos import SKOSReader
 
 from django.contrib.auth.models import User
 from django.test.client import RequestFactory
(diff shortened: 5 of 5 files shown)


arches/app/etl_modules/save.py
```diff
@@ -187,7 +187,10 @@ def get_resourceids_from_search_url(search_url, user=None):
     params = parse_qs(urlsplit(search_url).query)
     for k, v in params.items():
         request.GET.__setitem__(k, v[0])
-    func, args, kwargs = resolve(reverse("search_results"))
+    search_results_path = reverse("search_results")
+    if search_results_path.startswith(get_script_prefix()):
```
jacobtylerwalls, line 191: Apparently this function is [not supported](https://docs.djangoproject.com/en/4.2/ref/urlresolvers/#get-script-prefix) outside the request-response cycle, e.g. like we're doing here for celery tasks.

Rather than test with celery, I quickly verified by adjusting my `manage.py shell` command to act like [Django's wsgi.py](https://github.com/django/django/blob/470e5545e56e8510f9b9d39a96d7094fedd9c45a/django/core/wsgi.py#L12) where `set_prefix=False`, and tested this patch against dev/8.0.x where we're removed the `django.setup()` call at the top of index_database.py.

Finally, set a breakpoint here and import the function. Voila, `get_script_prefix()` returns the wrong thing:

```py
In [1]: from arches.app.etl_modules.save import get_resourceids_from_search_url
INFO:rdflib:RDFLib Version: 4.2.2

In [2]: get_resourceids_from_search_url("https://127.0.0.1:8000/search_results")
> /Users/jwalls/prj/arches/arches/app/etl_modules/save.py(191)get_resourceids_from_search_url()
-> breakpoint()
(Pdb) n
> /Users/jwalls/prj/arches/arches/app/etl_modules/save.py(192)get_resourceids_from_search_url()
-> if search_results_path.startswith(get_script_prefix()):
(Pdb) get_script_prefix()
'/'
(Pdb) from django.conf import settings; settings.FORCE_SCRIPT_NAME
'/bc/'
```

What do you think about explicitly checking `settings.FORCE_SCRIPT_NAME` also?

Depending on complexity, a util for this might be nice.

jacobtylerwalls (changes requested):

arches/app/etl_modules/save.py
jacobtylerwalls: Then a unit test would be nice, e.g with:
```py
    @override_settings(FORCE_SCRIPT_NAME='/bc/')
    def ...
        with sync_overridden_test_settings_to_arches():
            do_thing()
```

bferguso: @jacobtylerwalls - thanks for flagging that. I should have looked into that function a bit closer - didn't realize it didn't work outside of the request-response cycle. I'll update the logic, and maybe pull it out into a separate util function, which will make it easier to unit test anyway.

bferguso: @jacobtylerwalls  - Unless I'm missing something (entirely possible) I think the use of `get_script_prefix()` may be valid since that's what's used in the `django.urls.base.reverse()` [here](https://github.com/django/django/blob/470e5545e56e8510f9b9d39a96d7094fedd9c45a/django/urls/base.py#L44) to generate the prefix in the first place.

If we're not in a request-response cycle that function always returns "/" which should be OK as the reverse() won't add a prefix, and if we are in a request-response cycle it should work as expected based on the value of the settings.FORCE_SCRIPT_NAME. For the Celery issue I was encountering in our app, I did a test with our application and it appears to work as expected, because it's creating an HttpRequest object to submit the export the results.

Maybe worth removing the change in the etl_modules since it's not an issue there - I wasn't sure where that was called from - thought that was initiated from the UI but I'm not familiar with that part of the app.

jacobtylerwalls: Hey Brett, thanks for the additional details. I still think there's value in being more careful here, even if mostly just to set a precedent when this gets copy/pasted forward.

> If we're not in a request-response cycle that function always returns "/" which should be OK

I'm not sure that's the case: apparently recent versions of Django [started caching](https://code.djangoproject.com/timeline?from=2022-09-24T12%3A34%3A12-05%3A00&precision=second) the first value of `SCRIPT_NAME`, which could be:
- the `{FORCE_}?SCRIPT_NAME` on the first HTTP request to your server
- _or_ `"/"` if `get_script_prefix()` was called first during middleware initialization (or anywhere outside request/response, e.g. celery)
- _or_ `FORCE_SCRIPT_NAME` if the devserver was booted (maybe?)

Given that we've run into issues before that went undetected because we were using the devserver (recent `APP_NAME` troubles), and given that Django has discouraged this use -- I'm still wary about doing it here. Does that make sense?

> For the Celery issue I was encountering in our app, I did a test with our application and it appears to work as expected

Did you test against 8.0? I mentioned the behavior is different there because we removed a call to `django.setup()` in index_database.py.

> because it's creating an HttpRequest object to submit the export the results.

Also not sure about this one, because if it's a dummy request that we're using just to avoid issues with the factoring of the underlying implementation, then we won't get or set prefixes, which is done by the request handlers.

bferguso: Hi Jacob -

> I'm not sure that's the case: apparently recent versions of Django [started caching](https://code.djangoproject.com/timeline?from=2022-09-24T12%3A34%3A12-05%3A00&precision=second) the first value of SCRIPT_NAME, which could be:

I guess my thinking here is that if we use the same call to `get_script_prefix()` that is used in the `reverse()` function, and it's all being executed in the same call stack, that would be the the most consistent call to use when handling the return value of `reverse()`. Are you concerned about an implementation change in the Django `reverse()` function going forward?

For the creation of the HttpRequest object in the Celery task - that was where we observed the problem, (in version 7.6.3), and based on earlier comments that the `get_script_prefix()` only adds the value of `{FORCE_}?SCRIPT_NAME` within an HttpRequest context, I think that it follows that it must add that value. Otherwise I don't think we would have observed that issue there.

I haven't been able to test against 8.0 as I've been struggling to get a functioning Docker image of that going, and haven't had time to track down the issue yet.

Having said all that, I'm happy to use the `FORCE_SCRIPT_NAME` value directly if you still think that's the best solution. In what you read - should we be checking both the `FORCE_SCRIPT_NAME` and `SCRIPT_NAME` when looking at the prefix using the resolve(reverse()) pattern?

jacobtylerwalls (approved):
Ah ah I get it now -- you're saying all we need to do is undo the call to get_script_prefix() inside `reverse()`, and we're in the same thread. Sorry for being slow on the uptake.

PR looks good!

jacobtylerwalls:
Added coverage, test fails on main. The missing coverage is a code path mostly (only?) called from async paths, where we have basically zero test coverage. Would be a worthy challenge in some cases, but looking closer, this is code I think we want to improve on anyway (we have a module level function outside the request/response cycle creating a dummy request). Better to remove this code than test it.

PR #11724, opened 2025-01-10
11670 Make webpack add filehashes to all (read: entrypoint) js files in build

### LINKED ISSUE #11670: some webpack-built files are named with their file hash while others aren't
A common use case is that a developer has updated their project and pushes these updates to a production server (run webpack build, collectstatic, etc.) but end-users' browsers are still caching the old files. There's a cache-bursting strategy built into the logic of how webpack names these files:
```.js
resolve({
            entry: {
                ...archesCoreEntryPointConfiguration,
                ...archesApplicationsEntrypointConfiguration,
                ...projectEntryPointConfiguration,
                ...CSSFilepathLookup,
            },
            devServer: {
                port: WEBPACK_DEVELOPMENT_SERVER_PORT,
            },
            output: {
                path: Path.resolve(__dirname, APP_ROOT, 'media', 'build'),
                publicPath: STATIC_URL,
                filename: '[name].[contenthash].js', // Adds a hash to all entry points
                chunkFilename: '[id].[contenthash].js', // Adds a hash to all dynamically imported chunks
                clean: true,
                assetModuleFilename: 'img/[hash][ext][query]',
            },
            ...
```

The `contenthash` is used in creating the name of each file, capturing its updated state which then forces browsers to get the new files from the server. However when `npm run build_production` gets run, only so-called `chunkFiles` (`.chunk.js`) get their content hash mixed into their file names. Files like `search.js` or `graph.js` do not contain any hash values in the filenames, despite the explicit logic above in the webpack config.

whatisgalen:
### Types of changes

### Description of Change

This PR creates a webpack_manifest (essentially just a lookup for files to their actual webpack asset filename) so that all `.js` files can be saved using a filehash in the filename, thereby forcing a cache burst on end-user browser clients whenever novel changes get pushed to a deployment. This is effective on both dev and prod builds.

A context processor makes this manifest available to all templates. The `base-root.htm` template now looks up the asset corresponding to the `main_script` value.

See [webpack-manifest-plugin](https://www.npmjs.com/package/webpack-manifest-plugin) for more docs

_Note that there is an assumption in where the manifest should be looked for:_
```.py
...
manifest_path = os.path.join(settings.APP_ROOT, "media/build", "manifest.json")
```
...and it's possible this should search a list of locations instead.

### Issues Solved

Closes #11670 

### Testing this PR
update your package.json to include:
```.js
"webpack-manifest-plugin": "5.0.0"
```
...then delete `package-lock.json' and run `npm install` to install webpack_manifest. You'll also want to run both a production and development build of webpack. All "main_views" should be tested, e.g.
- rdm
- views/graph
- views/graph-designer
- views/graph/model-history
- views/graph/function-manager
- views/map-layer-manager
- views/plugin
- views/resource
- views/resource-editor
- views/edit-history
- views/edit-log
- views/report
- views/search
- views/user-profile

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Found by: @whatisgalen 
*   Tested by: @chrabyrd 
*   Designed by: @whatisgalen


whatisgalen:
~~making a note to self and others that when ultimately approved, the changes to the frontend webpack utils files (like build-filepath and webpack.common) need to be duplicated in the arches-templates directory's versions of those files.~~
done ✅

webpack/webpack.config.dev.js
chrabyrd, line 1: If this is being added to both `webpack.config.dev.js` and `webpack.config.prod.js`, it instead belongs in `webpack.common.js`

package.json
```diff
@@ -88,7 +88,8 @@
         "uuidjs": "^3.3.0",
         "vue": "^3.5.8",
         "vue3-gettext": "^3.0.0-beta.6",
-        "webpack-bundle-tracker": "^3.1.0"
+        "webpack-bundle-tracker": "^3.1.0",
+        "webpack-manifest-plugin": "5.0.0"
```
chrabyrd, line 92: this should instead be a dev dependency

webpack/webpack.common.js
```diff
@@ -41,6 +42,18 @@ module.exports = () => {
             PROJECT_RELATIVE_NODE_MODULES_PATH = Path.resolve(APP_ROOT, '..', 'node_modules');
         }
 
+        const manifestPlugin = new WebpackManifestPlugin({
```
chrabyrd: nit: inline in `plugins`

chrabyrd: There appears to be a large overlap between what this generates and what's already created in `webpack/webpack-stats.json`. Is there a chance to combine them?

arches/app/utils/context_processors.py
```diff
@@ -109,3 +113,25 @@ def app_settings(request=None):
             "DEBUG": settings.DEBUG,
         }
     }
+
+
+def webpack_manifest(request):
```
chrabyrd: why make this a context_processor?

arches/app/views/base.py
```diff
@@ -93,6 +94,12 @@ def get_context_data(self, **kwargs):
             )
             > 0
         )
+        context["webpack_manifest"] = webpack_manifest(self.request)
+        main_script = kwargs.get("main_script", None)
+        if main_script:
+            context["main_script_webpack_asset"] = context["webpack_manifest"].get(
+                f"{main_script}.js", ""
+            )
```
chrabyrd: doesn't this need error handling as well?

arches/app/views/base.py
chrabyrd, line 1: The hard part here is that there's no guarantee and endpoint will use `BaseManager` anymore.

chrabyrd (changes requested):
Overall it looks pretty good. I like the concept and I get why this needs to be a things. There are a few pattern issues to address, but I'm glad you've got this ball rolling! 🍙

webpack/webpack.common.js
whatisgalen: I might need to connect with you about that...been trying to do so but I'm struggling with getting a clean path key for each asset under webpack-stats.json "assets"

chrabyrd:
@whatisgalen it looks like the build is failing? Keeping 👀 on this && happy to get a quick review in once it builds 👍

whatisgalen:
bump

chrabyrd (changes requested):
all in all this looks sane, but there's still no guarantee a route will inherit baseManager. This will break in that case. Why not include your new context processor in `build_templates_config` like the others?

arches/templatetags/template_tags.py
```diff
@@ -9,6 +9,14 @@
 register = template.Library()
 
 
+@register.filter
+def get(dictionary, key):
```
whatisgalen: @jacobtylerwalls I checked the django [builtin template tags](https://docs.djangoproject.com/en/5.1/ref/templates/builtins/) and I feel like this ought to be in there. Do you think I should post in the django forum?

package.json
whatisgalen: it's being used in both prod and dev builds though

chrabyrd: just following the doc guidance:

https://www.npmjs.com/package/webpack-manifest-plugin

> `npm install webpack-nano webpack-manifest-plugin --save-dev`

whatisgalen: Sorry I did not realize that had any significance with regard to dependencies vs devDependencies in Arches. With that said, do I need to make a PR into the `arches-dev-dependencies` repo?

arches/templatetags/template_tags.py
jacobtylerwalls: I did a little search and couldn't find a discussion about it, other than this [ticket](https://code.djangoproject.com/ticket/577) from 19 years ago that looks like it got lost in triage.

I know there are some that feel very strongly about keeping programming concepts out of templates, so to anticipate that feedback be sure to explain why you can't just calculate the manifest path in your view/context processor, which is more like what you had before your last push.

I see that we're loosening the dependency on BaseManager in favor of a dependency on a template rendered with a `main_script` variable, so I have a feeling someone will ask if you considered making that explicit, which is also more like your previous version.

All in all, I say go for it, I'll definitely learn something if you post.

whatisgalen: @jacobtylerwalls [curious to see what happens](https://forum.djangoproject.com/t/argument-against-pythonic-dict-get-template-tag/39341)

chrabyrd:
👋 @whatisgalen !

Just wanted to make you aware of https://github.com/archesproject/arches/pull/11881

It's a webpack overhaul that folds in cache poisoning... Just wanted to make you aware before you pushed more on this

chrabyrd:
superceded by #11881

PR #11732, opened 2025-01-17
Fix  any value and no value complements in advanced search

### LINKED ISSUE #11424: Search not correctly returning complements when using has any value and has no value
The search functionality of Arches has a function:

Has Any Value
Has No Value

In principle if the card that this search function points at is single value, then asking for 'has any value' and 'has no value' should return results which = the total value. Because the sum of everything of one thing that has a property and everything of one thing that doesn't have a property = all the things. I explain this again below. Anyhow the search doesn't return this. So is this a bug or is there a weird logic of what has any value and no value means? It leads to a lack of trust in the results from the search because it doesn't add up.

Here is a real world example:

In this example the card is not multi cardinality, it is single cardinality and uses a concept list. So as far as I can process that means that that card has either a value in it or it doesn't. So I think it should create complementarity in search ie the sum of 'show me all with something' and 'show me all with nothing' should equal 'the sum of all those things'. If there are 1000 records and 600 have a value in that card then show me 'has any value' should produce 600 and logically (given that it is not a multicardinality card) the inverse show me 'has no value' should equal 400 and then the total of a and b should be 1000, ie the number of records in the system.

I created a spreadsheet with the results from a real system showing that the numbers don't work out that way. We end up with a difference of 73 fewer records than the total number of records for group when we sum the 'has any value', 'has no value' searches. Since it is single cardinality, classic logic says it either does or doesn't have a value. One inference one could make here is that the 73 records difference means there are 73 records that do and do not have a value. This might make sense in the multicardinality card setup situation but I don't see how it does in a non multi.

https://docs.google.com/spreadsheets/d/1ufSi6SBa_OkJhnIrTc4XICkyo7Fc50MrMCJPu8Y8C2Y/edit?usp=sharing

MikeGriniezakis:
### Types of changes

### Description of Change

In some cases when indexing resources, a resource might not have tiles. This results in ES indexing an empty list for the tiles of that resource. The nested query for the tiles ensures that at least one indexed tile must be present. This was the issue where the "Has any value" and "Has no value" did not complement each other. By removing the at least one indexed tile requirement, the two options complement each other and sum up to the correct count.

### Issues Solved

Closes #11424 

### Checklist

### DIFF
--- a/arches/app/search/components/advanced_search.py
+++ b/arches/app/search/components/advanced_search.py
@@ -1,15 +1,15 @@
+from arches.app.datatypes.datatypes import DataTypeFactory
 from arches.app.models.models import (
-    Node,
-    DDataType,
-    GraphModel,
     CardModel,
     CardXNodeXWidget,
+    DDataType,
+    GraphModel,
+    Node,
 )
 from arches.app.models.system_settings import settings
-from arches.app.datatypes.datatypes import DataTypeFactory
-from arches.app.utils.betterJSONSerializer import JSONDeserializer
-from arches.app.search.elasticsearch_dsl_builder import Bool, Nested
 from arches.app.search.components.base import BaseSearchFilter
+from arches.app.search.elasticsearch_dsl_builder import Bool
+from arches.app.utils.betterJSONSerializer import JSONDeserializer
 
 details = {
     "searchcomponentid": "",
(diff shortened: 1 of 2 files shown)


chiatt:
I'm closing this because I'm not able to reproduce this bug in the latest version of dev/7.6.x. If You're still seeing it @MikeGriniezakis, please let me know and we'll reopen this.

PR #11737, opened 2025-01-21
Fix node sortorder in reorder nodes API response #11651

### LINKED ISSUE #11651: Graph Nodes inexplicably re-ordered on refresh
Some new ordering is put in place when you refresh a graph in the graph designer (which is pretty annoying after you've ordered them alphabetically), see screen recording:

https://github.com/user-attachments/assets/e2c54858-4d09-4d63-8741-243d62b09ade

seen on dev/7.6.x

jacobtylerwalls:
### Types of changes

### Description of Change
Before, changes to node order not reflected in the API response for a reorder nodes call.

### Issues Solved
Closes #11651

### Checklist

#### Ticket Background
*   Sponsored by: Getty Digital in part, Farallon in part
*   Found by: @whatisgalen 

### DIFF
--- a/arches/app/views/graph.py
+++ b/arches/app/views/graph.py
@@ -22,6 +22,9 @@ from slugify import slugify
 import uuid
 import logging
 from dateutil import tz
+from http import HTTPStatus
+from io import BytesIO
+
 from django.db import transaction
 from django.shortcuts import redirect, render
 from django.db.models import F, Func, Q
@@ -56,8 +59,6 @@ from arches.app.utils.permission_backend import (
     get_group_perms,
     get_user_perms,
 )
-from io import BytesIO
-
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


tests/views/graph_manager_tests.py
```diff
@@ -189,8 +189,6 @@ def setUpTestData(cls):
         cls.GRAPH_ID = str(graph.pk)
         cls.NODE_COUNT = 5
 
-        cls.graph = graph
```
jacobtylerwalls: Django cleans up the db for us, but not the model instances -- we had some tests publishing and unpublishing the same graph instance, which could introduce some test isolation issues, so I just removed this reference to make it harder to do. I think I was the one that added this (whoops)

jacobtylerwalls: Lies. TIL django [cleans up](https://docs.djangoproject.com/en/5.1/topics/testing/tools/#django.test.TestCase.setUpTestData) the model instances for you. i'll revert these unnecessary changes.

arches/app/views/graph.py
```diff
@@ -584,16 +585,34 @@ def post(self, request, graphid=None):
...
-                                    no.sortorder = sortorder
-                                    no.save()
-                                    sortorder = sortorder + 1
-                            ret = data
+                        incoming_node_data = data.get("nodes", [])
+                        ordered_node_ids = []
+                        graph_ids = []
+                        for json_node in incoming_node_data:
+                            ordered_node_ids.append(json_node["nodeid"])
+                            graph_ids.append(json_node["graph_id"])
```
samuelscandrett: Do we want to make `graph_ids` a set from the start (or do a check if the id is already present in the list)?  If we have a huge graph this list could end up being in the hundreds of just the same uuid.  I suppose it doesn't matter massively, as it's unlikely to cause performance issues, but if we could avoid it should we?

tests/views/graph_manager_tests.py
```diff
@@ -292,6 +292,21 @@ def test_node_update(self):
         self.assertEqual(node_.name, "new node name")
         self.assertTrue(node_.is_collector)
 
+    def test_node_reorder(self):
+        self.client.login(username="admin", password="admin")
+        url = reverse("reorder_nodes")
+        reversed_nodes = list(reversed(self.graph.nodes.values()))
```
samuelscandrett: I can't seem to get the test to run, it hits here and fails on self.graph with the error 
```
ERROR: test_node_reorder (tests.views.graph_manager_tests.GraphManagerViewTests.test_node_reorder)
  File "/home/samuel/arches_dev/v8_development/arches/tests/views/graph_manager_tests.py", line 283, in test_node_reorder
    print(self.graph)
          ^^^^^^^^^^
AttributeError: 'GraphManagerViewTests' object has no attribute 'graph'
```

Do you see the same?

samuelscandrett (commented):
The PR successfully reflects the sortorder of reordered nodes, whereas before the JSON response would still show the old order.

Just a couple minor questions / issues.

tests/views/graph_manager_tests.py
jacobtylerwalls: Thanks, I hadn't upgraded this branch in a while, should be `self.test_graph`

jacobtylerwalls: (once rebased, which I'll do now, you'll need to hard reset this branch if you pull)

arches/app/views/graph.py
jacobtylerwalls: nice idea!

tests/views/graph_manager_tests.py
samuelscandrett: Thanks!  Test now passes.

samuelscandrett (approved):
Looks good!

PR #11775, opened 2025-02-07
ES Snapshotting Utility

aarongundel:
### Types of changes

### Description of Change

Add elasticsearch snapshotting commands.  REQUIRES EXISTING SNAPSHOT REPOSITORY!

Also includes "true-up" - that is, guarantees consistency of an ES snapshot that is older than the current database - reindexes items from edit log that have changed since the snapshot was taken.

### DIFF
--- a/arches/app/utils/index_database.py
+++ b/arches/app/utils/index_database.py
@@ -20,6 +20,8 @@ from arches.app.datatypes.datatypes import DataTypeFactory
 from arches.app.utils import import_class_from_string
 from typing import Iterable
 
+from arches.app.models.models import EditLog
+
 
 logger = logging.getLogger(__name__)
 serialized_graphs = {}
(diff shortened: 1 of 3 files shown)


arches/app/utils/index_database.py
```diff
@@ -605,3 +607,47 @@ def index_resources_by_transaction(
...
+    Keyword Arguments:
+    quiet -- Silences the status bar output during certain operations, use in celery operations for example
+    recalculate_descriptors - forces the primary descriptors to be recalculated before (re)indexing
+
+    """
+    start = datetime.now()
+
+    with connection.cursor() as cursor:
+        cursor.execute(
+            """SELECT DISTINCT resourceinstanceid FROM edit_log WHERE timestamp >= %s;""",
```
whatisgalen, line 632: I may be misreading the logic here, but won't this attempt to re/index resources that have been deleted?

aarongundel: That's a great point; I think it will reindex those resources.  So this needs to be more sensitive to the edittype of delete.

chiatt: Given the isolated nature of this feature, I think it makes sense just to get this merged in before the feature freeze and fix the issue that @whatisgalen has identified in a separate PR.

chiatt (approved):

PR #11778, opened 2025-02-10
Make non-existent resource report raise 404

### LINKED ISSUE #11772: Return a 404 when /report/ or /resource/ doesn't lead to an existing resource
In debug mode, the error you see when you try to go to a report for a non-existent resource is:

`DoesNotExist at /report/edb83d95-543e-4d01-98b9-abdbd3212763`

...which shows up as a 500 in production because the `DoesNotExist` exception is unhandled by the Report view. It should actually be a 404 since the resource sought was not found.

whatisgalen:
…re #11772

### Types of changes

### Issues Solved

Closes #

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/tests/views/resource_tests.py
+++ b/tests/views/resource_tests.py
@@ -363,12 +363,14 @@ class CommandLineTests(ArchesTestCase):
         response = self.client.get(url)
         self.assertEqual(response.status_code, 200)
 
-    def test_resource_report_bad(self):
+    def test_resource_report_for_missing_resource(self):
         self.client.login(username="admin", password="admin")
         resource = Resource.objects.get(resourceinstanceid=self.resource_instance_id)
         resource.delete()
-        url = reverse(
-            "resource_report", kwargs={"resourceid": self.resource_instance_id}
-        )
-        response = self.client.get(url)
+        with self.assertLogs("django.request", level="WARNING"):
+            response = self.client.get(
+                reverse(
+                    "resource_report", kwargs={"resourceid": self.resource_instance_id}
+                )
+            )
         self.assertEqual(response.status_code, 404)
--- a/tests/views/resource_tests.py
+++ b/tests/views/resource_tests.py
@@ -365,7 +365,8 @@ class CommandLineTests(ArchesTestCase):
 
     def test_resource_report_bad(self):
         self.client.login(username="admin", password="admin")
-        ResourceInstance.objects.delete(resourceinstanceid=self.resource_instance_id)
+        resource = Resource.objects.get(resourceinstanceid=self.resource_instance_id)
+        resource.delete()
         url = reverse(
             "resource_report", kwargs={"resourceid": self.resource_instance_id}
         )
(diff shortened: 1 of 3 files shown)


tests/views/resource_tests.py
```diff
@@ -354,3 +354,21 @@ def test_get_related_resource(self):
...
+
+    def test_resource_report_good(self):
+        self.client.login(username="admin", password="admin")
+        url = reverse(
+            "resource_report", kwargs={"resourceid": self.resource_instance_id}
+        )
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+
+    def test_resource_report_bad(self):
```
chiatt: Because `test_resource_report_good` tests if the status is 200, that name makes sense. In this case though, I think what's being tested isn't whether the response is bad (which I would take to mean a 500) but rather whether it's missing. So I think more intuitive name would be `test_resource_report_for_missing_resource` or something like that.

chiatt (changes requested):
Could you add a changelog entry?

chiatt (commented):
There are other places where this should probably be applied as well such as the resource editor, but that could be done is a different PR.

jacobtylerwalls:
This issue is already fixed in 8.0. Does it qualify for a 7.6 fix?

chiatt:
> This issue is already fixed in 8.0. Does it qualify for a 7.6 fix?

@jacobtylerwalls Are you referring to #11784? This seems a bit different, but you may be thinking of a different PR. I was a little on the fence about whether this belongs in 7.6, but it made some sense that a 500 error is arguably a 'crash'. I could be convinced otherwise though.

jacobtylerwalls:
Ah no, it was in #11637, but I didn't bother to mention it in the changelog.

> it made some sense that a 500 error is arguably a 'crash'

Of course, that's a good point.

Maybe @whatisgalen you'll want to take the lead on solving the merge conflict when pulling 7.6 -> 8.0 after this fix?

tests/views/resource_tests.py
```diff
@@ -354,3 +354,21 @@ def test_get_related_resource(self):
...
+        self.client.login(username="admin", password="admin")
+        resource = Resource.objects.get(resourceinstanceid=self.resource_instance_id)
+        resource.delete()
+        url = reverse(
+            "resource_report", kwargs={"resourceid": self.resource_instance_id}
+        )
+        response = self.client.get(url)
```
jacobtylerwalls: This clutters the test output, unfortunately:
```py
Not Found: /report/f562c2fa-48d3-4798-a723-10209806c068
.
Ran 1 test in 7.857s
```

The test in #11637 shows a way to silence this with assertLogs. I [asked](https://forum.djangoproject.com/t/best-patterns-for-handling-logging-of-4xx-and-5xx-status-codes-in-test-suite/35521) on the Django forum if there's a better way to do this with out removing all logging from tests, but I didn't get any bites.

jacobtylerwalls (changes requested):

tests/views/resource_tests.py
whatisgalen: change in latest

tests/views/resource_tests.py
whatisgalen: change in latest

jacobtylerwalls (approved):

PR #11784, opened 2025-02-11
Restore custom error views (but return JSON where desired) #11722

### LINKED ISSUE #11722: Add error handlers for api routes
***Problem***
The routes in core arches serve a mixture of HTML and JSON. We've generally been using the `api/` prefix to clarify which ones should return JSON. Without further configuration, a view or decorator raising an error will return Django's default HTML error responses, which can confuse a client expecting to always parse JSON, as shown below.

Recently written API views have been trying to catch and transform errors to JSON, but this is both verbose and not bullet-proof, e.g. see `@group_required` decorator below.

Suggest to add [custom](https://docs.djangoproject.com/en/5.1/topics/http/views/#customizing-error-views) 400, 403, 404 (maybe) and 500 handlers that sniff for whether it's an api route (prefixed with `api/` or `api.`, I guess?) and return JSON instead.

> Log in to lingo as tester3/Test12345!
See:
html-no-parse

> This is because the `@group_required` decorator on `ConceptTreeView` (and the others) returns HTML, and we crash on calling .parse()

_Originally posted by @jacobtylerwalls in https://github.com/archesproject/arches-lingo/pull/86#discussion_r1907720275_

***
An alternative to inspecting the route prefix would be to segregate the routes into two python modules and switch on that. We could look into how backward compatible this would be.

jacobtylerwalls:
### Types of changes

### Description of Change
This PR addresses two issues:
- Arches used to have generic error views, but they were removed from the core URL config in [f8a5045](https://github.com/archesproject/arches/commit/f8a50459340cc9ff98b12b4b792685e070a30a56). This PR restores them to the config and models in the release notes how to make sure your project uses them.
- Not every view that is supposed to return JSON returns JSON in every single error circumstance. One scenario with a view decorator is described in #11722, but there is no bulletproof way to catch every error without leveraging Django's built-in error views. If we do this, then front-end parsers that expect to call `.json()` on a response body won't throw when they get an HTML response from the generic (or customized) error view.

### Issues Solved
Closes #11722

### Checklist

#### Ticket Background

### DIFF
--- a/arches/app/utils/request.py
+++ b/arches/app/utils/request.py
@@ -7,7 +7,7 @@ def looks_like_api_call(request):
         return True
     if not request.resolver_match:
         return False
-    if "api." in request.resolver_match.func.__module__:
+    if ".api." in request.resolver_match.func.__module__:
         return True
     if (
         request.resolver_match.func.__module__
--- a/arches/app/utils/request.py
+++ b/arches/app/utils/request.py
@@ -1,0 +1,17 @@
+def looks_like_api_call(request):
+    known_api_paths_without_api_signposting = [
+        # Until these are renamed for consistency, just list them.
+        "arches.app.views.workflow_history",
+    ]
+    if "/api/" in request.path:
+        return True
+    if not request.resolver_match:
+        return False
+    if "api." in request.resolver_match.func.__module__:
+        return True
+    if (
+        request.resolver_match.func.__module__
+        in known_api_paths_without_api_signposting
+    ):
+        return True
+    return False
--- a/arches/urls.py
+++ b/arches/urls.py
@@ -781,8 +781,12 @@ urlpatterns = [
         api.SpatialView.as_view(),
         name="spatialview_api",
     ),
-    re_path("^api", api.API404.as_view(), name="api_404"),
 ]
+
+handler400 = "arches.app.views.main.custom_400"
+handler403 = "arches.app.views.main.custom_403"
+handler404 = "arches.app.views.main.custom_404"
+handler500 = "arches.app.views.main.custom_500"
 
 # This must be included in core to keep webpack happy, but cannot be appended when running a project.
 # See https://github.com/archesproject/arches/pull/10754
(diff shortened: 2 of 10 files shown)


arches/app/utils/request.py
```diff
@@ -0,0 +1,17 @@
+def looks_like_api_call(request):
+    known_api_paths_without_api_signposting = [
+        # Until these are renamed for consistency, just list them.
+        "arches.app.views.workflow_history",
```
jacobtylerwalls, line 4: We can add more as we find them, but I'm considering it out of scope to audit everything right now: that's a lot of views to analyze. See next comment for why this one is here.

tests/views/workflow_tests.py
```diff
@@ -83,8 +84,9 @@ def test_get_workflow_history(self):
                 )
             )
 
-        self.assertEqual(response.status_code, 403)
-        self.assertIn(b"Forbidden", response.content)
+        self.assertContains(
+            response, "Permission Denied", status_code=HTTPStatus.FORBIDDEN
+        )
```
jacobtylerwalls, line 89: Because this error condition was already tested, we caught an example of HTML where we should have sent JSON.

Before:

```html
\n<!doctype html>\n<html lang="en">\n<head>\n  <title>403 Forbidden</title>\n</head>\n<body>\n  <h1>403 Forbidden</h1><p></p>\n</body>\n</html>\n
```
After:
```json
{"message": "Permission Denied", "status": "false", "success": false, "title": "Request Failed"}
```

arches/urls.py
```diff
@@ -781,9 +781,13 @@
         api.SpatialView.as_view(),
         name="spatialview_api",
     ),
-    re_path("^api", api.API404.as_view(), name="api_404"),
```
jacobtylerwalls, line 784: Now we don't need this anymore to handle errors, which is good, because it was too fragile in practice (sneaky failures depending on the order in which you included your other URLs)

njkim (approved):
lgtm

Issue #11595, opened 2024-10-31
Interface for querying & updating resource models by node & nodegroup aliases


jacobtylerwalls:
### User story
As an arches application developer, I would like to be able to query and update data for a resource model by a semantic label like a node alias (guaranteed unique) or widget label rather than constructing tiles directly and having to know their UUIDs.

### Precedents
- [relational views](https://arches.readthedocs.io/en/stable/developing/reference/import-export/#sql-import)
- [Arches ORM](https://github.com/flaxandteal/arches-orm) (experimental)

### Proposal
Another approach would be to provide, in core arches, a solution that:
- annotates resource instance tile data onto the `ResourceInstance` model instances via Django querysets for Django-native querying, updating, related-object selecting, prefetching, paginating, signalling, etc.
- dogfoods the existing python application methods (thereby reducing drift) for:
  - tile validation
  - edit log, indexing side effects
  - function triggers
 - generates (customizable) django rest framework serializers from your resource models (similar to the way django rest framework can [generate serializers](https://www.django-rest-framework.org/api-guide/serializers/#modelserializer) from your django models in general)
 - reads from dynamic graph data, avoiding the need to take snapshots of "well-known" models or write migrations or deploy database views

### Demo
Example queries might look like this:
https://github.com/archesproject/arches/blob/d05d0c15cf97c2787554fa6ef3858b4e94636830/arches/app/models/querysets.py#L100-L132

jacobtylerwalls:
Development now taking place in archesproject/arches-querysets

PR #8143, opened 2022-02-05
Upgrade django to v3.2, re #8140

chiatt:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
Upgrades django to v3.2, re #8140

### Issues Solved

#8140

PR #11792, opened 2025-02-18
Restore primary key field defaults and support for `create()` #10958

### LINKED ISSUE #10958: Use db_default for auto-UUID fields in Django 5
When Django 5 is the minimum required version, we can just use [db_default](https://docs.djangoproject.com/en/5.0/ref/models/fields/#db-default) for database-generated UUIDs and drop our [custom implementation](https://github.com/archesproject/arches/blob/760d4826201e5ad6d87f6fd3350a86d10a318578/arches/db/migration_operations/extras.py#L109-L126), which currently throws errors if you try to `manage.py sqlmigrate`:

```py
$ manage.py sqlmigrate models 0001
  File "/Users/jwalls/prj/arches/manage.py", line 27, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/jwalls/django/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/Users/jwalls/django/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/jwalls/django/django/core/management/base.py", line 413, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/jwalls/django/django/core/management/commands/sqlmigrate.py", line 38, in execute
    return super().execute(*args, **options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jwalls/django/django/core/management/base.py", line 459, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jwalls/django/django/core/management/commands/sqlmigrate.py", line 80, in handle
    sql_statements = loader.collect_sql(plan)
                     ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jwalls/django/django/db/migrations/loader.py", line 381, in collect_sql
    state = migration.apply(state, schema_editor, collect_sql=True)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jwalls/django/django/db/migrations/migration.py", line 108, in apply
    schema_editor.collected_sql.append("-- %s" % operation.describe())
                                                 ^^^^^^^^^^^^^^^^^^^^
  File "/Users/jwalls/prj/arches/arches/db/migration_operations/extras.py", line 126, in describe
    return "Sets default value for uuid column(s) %s in %s" % self.columns, self.table
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
TypeError: not enough arguments for format string
```

jacobtylerwalls:
### Types of changes

### Description of Change
Restore model field-level defaults for primary keys.
Closes #10958

**Before**
Django 3.0 introduced an [optimization](https://docs.djangoproject.com/en/5.2/releases/3.0/#model-save-when-providing-a-default-for-the-primary-key) for models with defaults defined on their primary key fields in a way that prevents "blind overwrites". By that, I mean creating an instance without fetching it from the database, assigning a primary key attribute manually, and saving it back, intending to potentially overwrite data. Arches does blind overwrites (e.g. in management commands with an `--overwrite` flag).

The stopgap as part of Arches' upgrade to Django 3.2 was to [remove](https://github.com/archesproject/arches/pull/8143/commits/9c2cf65201ec4c820a6c314099c0bcbd89310216) field-level defaults altogether and add code to `__init__()` to set a value, but the problems with that are:
- you can no longer use a model manager `create()` method without specifying the primary key, which is redundant and surprising. Support for `create()` is blocking #11595.
- any transformations after the instance are created are not captured, e.g. if a model instance is later edited to have `.pk = None`

**After**
Field-level defaults are back, plus a solution to support blind overwrites (possible now on Django 5.1+, which [supports](https://code.djangoproject.com/ticket/35425) `force_update=True` for a blind overwrite). Left a comment discouraging copying this pattern forward to future models.

**While here**
- Prefer `db_default` (Django 5+) over our custom migration operations that were auto-creating UUID v1's.
- This migration operation was causing `sqlmigrate` to crash, see #10958
- It also needed to be updated to use uuidv4, so we can just drop our custom operation.
- Change other UUID fields to use uuid v4.

### Issues Solved
Closes #10958

### Checklist
-   [n/a] My test fails on the target branch

#### Ticket Background

### Further comments
The postgres rule checking for bidirectional concept relation uniqueness was preventing using db_default, giving this:
```py
ERROR: setUpClass (tests.exporter.datatype_to_rdf_tests.RDFExportUnitTests)
  File "/Users/jwalls/django/django/db/models/query.py", line 934, in get_or_create
    return self.get(**kwargs), False
           ~~~~~~~~^^^^^^^^^^
  File "/Users/jwalls/django/django/db/models/query.py", line 632, in get
        "%s matching query does not exist." % self.model._meta.object_name
    )
arches.app.models.models.Relation.DoesNotExist: Relation matching query does not exist.

During handling of the above exception, another exception occurred:

  File "/Users/jwalls/django/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
psycopg2.errors.FeatureNotSupported: cannot perform INSERT RETURNING on relation "relations"
HINT:  You need an unconditional ON INSERT DO INSTEAD rule with a RETURNING clause.
```

I worked around that by rewriting the rule as a functional `UniqueConstraint`, replacing the unique constraint we already had. Replacing 1 constraint and 1 rule with 1 constraint is presumably more efficient, but I haven't benched it. I added a test.


jacobtylerwalls:
Looks like the constraint I added causes instability in the migrations autodetector. This is fixed in 5.2 beta which comes out tomorrow, so I'll just bump that separately tomorrow and then rebase 👍

chiatt:
> Looks like the constraint I added causes instability in the migrations autodetector. This is fixed in 5.2 beta which comes out tomorrow, so I'll just bump that separately tomorrow and then rebase 👍

I noticed that the mixin isn't working with Django 5.1, but it worked fine with 5.2. Should be good to rebase now that #11797 is ready.

jacobtylerwalls:
> I noticed that the mixin isn't working with Django 5.1, but it worked fine with 5.2.

Thanks, that's probably because I'm using the `.pk_fields` property that was added in 5.2 to be resilient against composite primary keys.

Rebased, ready for another look!

chiatt (approved):
LGTM!

PR #8472, opened 2022-05-26
Fixes request object not being passed to tile save from nodevalue api causing unexpected function behaviour

aj-he:
### Types of changes

-   [x] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

Passes request object from nodevalue api call to the tile save so that functions expecting a request object, that is usually passed when using the web UI, doesn't error.

### Issues Solved

fixes #8471 

### Checklist

-   [ ] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

PR #11807, opened 2025-02-21
Let TypeError propagate from tile functions #11805

### LINKED ISSUE #11805: Tile functions catch TypeError too widely
Any TypeError emitted during a tile function is caught by the caller and logged in such a way that it breaks the loop and causes other functions to not run.

https://github.com/archesproject/arches/blob/6d065afae35ab38d6c569a49d2e3dd06e1606ab8/arches/app/models/tile.py#L798-L808

jacobtylerwalls:
Catching this could cause other functions to not run by breaking out of the loop.
Closes #11805

:wave: @aj-he you might be interested as I see in #8472 you elevated the info log to a warning. A hard fail would be nice here, no?

### DIFF
--- a/arches/app/models/tile.py
+++ b/arches/app/models/tile.py
@@ -795,30 +795,18 @@ class Tile(models.TileModel):
         context -- string e.g. "copy" indicating conditions under which a resource is saved and how functions should behave.
         """
 
-        try:
-            for function in self._getFunctionClassInstances():
-                try:
-                    function.save(self, request, context=context)
-                except NotImplementedError:
-                    pass
-        except TypeError as e:
-            logger.warning(
-                _("No associated functions or other TypeError raised by a function")
-            )
-            logger.warning(e)
+        for function in self._getFunctionClassInstances():
+            try:
+                function.save(self, request, context=context)
+            except NotImplementedError:
+                pass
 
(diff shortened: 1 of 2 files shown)


aj-he:
@jacobtylerwalls - A key thing for us it to be able to trace function failure through logging as we have quite a few functions that interface other internal services. I suspect that this was changed to ensure we could see this clearer in our monitoring when the ArcGIS add-in failed to write through the nodevalue api.

I'm happy for the typeerror to be explicitly removed, but could there be a level of logging on unhandled errors to allow the issue to be traced?

jacobtylerwalls:
Great idea. We could just `logger.error(exception)` inside the custom 500 view added in #11784.

jacobtylerwalls:
I'll test it, I have a feeling it might be duplicative with the existing logging to arches.log configured in LOGGING setting.

jacobtylerwalls:
I'll bounce this idea to a new issue, I'm surprised we didn't have it configured by default.

jacobtylerwalls:
See #11813

aj-he (approved):
@jacobtylerwalls, given the associated change to catch the errors I'm happy to approve. Once we upgrade to 8 we'll see if there are any issues that reappear and address appropriately.

PR #11809, opened 2025-02-21
Restore ability to join TileModel to NodeGroup via ORM #11808

### LINKED ISSUE #11808: Avoid repetitive nodegroup queries by using ORM to join Tile to Nodegroup
**Background**
The foreign key from TileModel -> NodeGroup was removed in v8 (under development) to support graph versioning, specifically so that graphs could migrate away to new versions while leaving tile data behind.

A property was added to add compatibility with `.nodegroup` calls, but unlike a Django model field which caches when you access it, or lets you `select_related()` to avoid the query in the first place, in Lingo we're seeing avoidable queries on every access.

See [discussion](https://github.com/archesproject/arches/pull/9613#issuecomment-2027457468).

**Proposal**
One of the goodies in Django 5.2 is that an API for expressing joins without foreign keys was made more public. See the section on [ForeignObject](https://docs.djangoproject.com/en/dev/topics/composite-primary-key/#composite-primary-keys-and-relations). It's newly documented but has been stable for many versions.

I have a little proof of concept that lets us restore the prior functionality for things like:

```py
.nodegroup = 
.filter(nodegroup=
.select_related("nodegroup")
```
etc. I can PR it so we can take a decision about it. The performance is one aspect and then there's avoiding a breaking change if we can.

This will also help with writing arches applications compatible with both 7.6 and 8.0.

cc/ @chrabyrd @chiatt

jacobtylerwalls:
### Types of changes
-   [x] _un-Breaking change_

### Description of Change
Restore the ability to traverse from `TileModel` <--> `NodeGroup` efficiently via the Django ORM:
- improve performance (keep prior caching behavior of .nodegroup access, ability to `select_related("nodegroup")`)
- avoid having to write code in different ways on 7.6 & 8.0 (other than allowing for nodegroup to be null, potentially)

### Issues Solved
Closes #11808, see more background on the proposal there.

### Checklist

#### Ticket Background
*   Sponsored by: Getty Digital


jacobtylerwalls:
I pushed a commit to keep the prior behavior of `.nodegroup` returning `None` instead of raising when there is no referenced nodegroup row, mostly just to demonstrate that it's possible with this approach.

I think we need to decide whether we want `.nodegroup` to return None or raise in that case -- and then fix the call sites, since we [still have](https://github.com/archesproject/arches/pull/9613#issuecomment-2024001859) code doing things like `.nodegroup.cardinality == "n"`, which will crash either way.

In the last discussion there was a [note](https://github.com/archesproject/arches/pull/9613#issuecomment-2043702540) about tile serialization, so I added an assertion for that. With this approach it passes either way (with or without the None fallback).

EDIT: this proof of concept was reverted after discussion, see below

chiatt:
It makes sense to me that if related nodegroup isn't available an error should be raised. That's how I would generally expect Django to behave if an object doesn't exist.

chrabyrd (changes requested):
ooo fancy. Yeah it looks good IMO 👍 

~~looks like the builds are failing though. Once those are healthy this is g2g~~

whoops looks like a stale page. Approving!

chrabyrd (approved):

jacobtylerwalls:
Thanks for the reviews!

PR #9602, opened 2023-06-06
Reordering of overlays & resource layers via front end

SDScandrettKint:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
This change introduces the functionality of moving overlays and resource layers by dragging them around in the map overlay tab.

In order to have both overlays and resource layers sorting within the same array, the "sortorder" field that was added to the  MapLayers model in #9291 was renamed to "overlaysortorder" and the same field was added to the Node model. This was done because "sortorder" already exists in the Node model and I didn't want to create confusion or have two differently named variables referring to the same thing. All other changes referencing map "sortorder" have been replaced with "overlaysortorder"

### Issues Solved

#9309

### Checklist

-   [ ] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: Knowledge Integration
*   Found by: @ 
*   Tested by: @SDScandrettKint  +  @KacperSzyf
*   Designed by: @SDScandrettKint  +  @KacperSzyf

### Further comments

This is currently a draft PR as it has not been thoroughly tested and there is an issue where the map_layer_manager overlay sort order input can be the same as another overlay, or the same as the overlaysortorder of a resource layer. 
I plan on adding a input for resource layer sort order, the same as is in the overlays tab, and some sorting (probably in map_layer_manager.js) to ensure no layers have the same integer. Further, it's currently possible to add an overlay sort order that is greater than the number of existing overlays.

Issue #9309, opened 2023-03-07
Add ability to order resource overlays


bferguso:
As part of issue #9291 it was noted that Resource Overlays cannot currently be sorted either within themselves or within the entire set of overlays, including those provided by tileserver services.

User Story
Would like to add the ability to sort all overlays, including Resource Overlays in any order. Currently Resource Overlays are at the bottom of the list and presented in an arbitrary order.

Acceptance Criteria
- [ ] Map overlay order can be configured in the map layer manager
- [ ] Resource overlays can be interleaved with Tileserver overlays
- [ ] Ordering can be achieved by drag and drop. This may require the "Resource Layers" and "Overlays" tabs in the Layer Manger to be merged
- [ ] Overlays are presented in the configured order in both search and resource report pages
- [ ] What should the "natural" order of overlays be? Currently it is Tileserver overlays by name, and then Resource overlays.
- [ ] Unit Tests passed
- [ ] Documentation updated

Definition of Done
Layer sort order is configurable by the user and defaults to sort by ascending layer name if order is not defined. Tileserver overlays are presented before Resource overlays (to be confirmed?).

Ticket Background
Sponsored by: 
Suggested by:
Designed by:

PR #11810, opened 2025-02-24
Rearrange map overlays in the search front-end

samuelscandrett:
### Types of changes

### Description of Change

A remake of https://github.com/archesproject/arches/pull/9602 for V8 using the map_layer SortOrder field.  Also adds the ability to reorder map overlays using keyboard shortcuts (holding Ctrl + up/down arrow key).

### Issues Solved

The bullet point "Ordering can be achieved by drag and drop." from #9309

### Checklist

#### Accessibility Checklist

| Keyboard         |     x    |    x      |

#### Ticket Background

### Further comments

Demonstration of moving overlays using keyboard control:

https://github.com/user-attachments/assets/c7c2b575-3382-4dd0-8a8b-6b7225a846b1

Of course this feature uses the SortOrder field from MapLayers, meaning that moving the resource overlays does nothing because they come from the Node, which doesn't have a sortorder for the layer.  The original PR https://github.com/archesproject/arches/pull/9602 added a layersortorder to both map_layers and the resource layers (from node), which meant that both could be sorted in the same.  Are there plans to move the resource layers into MapLayers?  

Final question, at the moment this only affects overlays, should we extend the sorting to basemaps too?

### DIFF
--- a/arches/app/views/map.py
+++ b/arches/app/views/map.py
@@ -163,3 +163,21 @@ class TileserverProxyView(ProxyView):
         if settings.TILESERVER_URL is None:
             raise Http404(_("Tileserver proxy not configured"))
         return headers
+
+
+@method_decorator(group_required("Resource Editor"), name="dispatch")
+class OverlayOrderView(BaseManagerView):
+    def put(self, request):
+        json = request.body
+        data = JSONDeserializer().deserialize(json)
+        map_layers = models.MapLayer.objects.all()
+
+        for layer in data["map_order"]:
+            try:
+                db_layer = map_layers.get(maplayerid=layer["maplayerid"])
+                db_layer.sortorder = layer["sortorder"]
+                db_layer.save()
+            except models.MapLayer.DoesNotExist:
+                pass  # will be resource overlay
+
+        return JSONResponse(data)
--- a/arches/app/media/js/viewmodels/map.js
+++ b/arches/app/media/js/viewmodels/map.js
@@ -531,7 +531,9 @@ define([
                     } else if (direction == "down") {
                         newIndex++
                     }
-                    self.overlays.splice(newIndex, 0, self.overlays.splice(index, 1)[0]);
+                    if (newIndex != -1 && newIndex != self.overlays().length) {
+                        self.overlays.splice(newIndex, 0, self.overlays.splice(index, 1)[0]);
+                    }
                 }
             }
 
--- a/arches/app/media/js/viewmodels/map.js
+++ b/arches/app/media/js/viewmodels/map.js
@@ -7,7 +7,8 @@ define([
     'utils/map-popup-provider',
     'utils/map-configurator',
     'utils/aria',
-    'templates/views/components/map-popup.htm'
+    'templates/views/components/map-popup.htm',
+    'bindings/sortable'
 ], function($, _, arches, ko, koMapping, mapPopupProvider, mapConfigurator, ariaUtils) {
     const viewModel = function(params) {
         var self = this;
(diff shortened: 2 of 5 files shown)


arches/app/views/map.py
```diff
@@ -163,3 +163,21 @@ def get_request_headers(self):
         if settings.TILESERVER_URL is None:
             raise Http404(_("Tileserver proxy not configured"))
         return headers
+
+
+@method_decorator(group_required("Resource Editor"), name="dispatch")
+class OverlayOrderView(BaseManagerView):
```
chiatt, line 169: Doesn't look like you actually need the BaseManagerView

arches/app/views/map.py
```diff
@@ -163,3 +163,21 @@ def get_request_headers(self):
...
+    def put(self, request):
+        json = request.body
+        data = JSONDeserializer().deserialize(json)
+        map_layers = models.MapLayer.objects.all()
+
+        for layer in data["map_order"]:
+            try:
+                db_layer = map_layers.get(maplayerid=layer["maplayerid"])
+                db_layer.sortorder = layer["sortorder"]
+                db_layer.save()
```
chiatt, line 179: It doesn't seem like we would want everyone in the resource editor group to be able to persist layer order changes to the database. This could be frustrating for some users if the default ordering is changing in unexpected ways. Rather than committing changes to the db, a front-end approach that saves the order to local storage in the browser might allow for layer sorting by user while leaving the default order defined by the admin. I could also see storing user preferences like this in the database - along with things like saved searches, but that would be a much heavier lift and require more planning.

chiatt (changes requested):

arches/app/views/map.py
samuelscandrett: That's a good point.  Perhaps we set it to the Application Administrator group, as this is the only group with direct access to the Map Layer Manager?  Users with this group have the ability to change the order in the MLM in current versions, so it wouldn't be any different to how it is now.

I agree it'd be a cool feature if users could save their own preference of orders, as I'm sure not everyone would agree on an order, additionally certain users (e.g. on different departments/teams) may require certain overlays at the top of the list that others wouldn't need.  Perhaps using the browser storage would be the way to do it for the time being.

I also agree that saved searches per user would be another good feature, for similar reasons, plus it would allow for 'internal' saved searches.

chiatt: I think it would be better if it was limited to Admins, my concern would be though that they wouldn't know that they were changing it for everyone, whereas in the MLM, that's pretty clear.

chiatt:
Regarding whether there are plans to move resource layers into MapLayers - this has been discussed and it's certainly something we'd like to see happen. @bferguso points to it a bit here: https://github.com/archesproject/arches/issues/9309
He doesn't explicitly mention moving resources into the maplayers table, but I recall discussing it with him and @robgaston, and I think we agreed that that would be the way forward.

arches/app/views/map.py
samuelscandrett: This does raise a very good point.  One idea would be to make moving the overlays raise a "dirty" state, and introduce a save button with text that states they're changing the global order.  Having a save button in the search page does feel wrong though...

Not to undo the work in this PR, but it does seem that defining order would be best in the MLM.

chiatt:
@SDScandrettKint and I agreed after discussion with the dev team that it makes sense to create a new user preferences feature in Arches v8 that will persist user specific configs such as overlay sort order and perhaps saved searches. However, the accessibility improvement in this PR that allows users to change overlay ordering using the keyboard will be submitted in a new PR targeting a 7.6 patch.

PR #11817, opened 2025-02-25
Fix missing case in check_cache_backend_for_production()

jacobtylerwalls:
### Types of changes

### Description of Change
Before, this check for inappropriate cache backends in production was only checking DummyCache, not LocMemCache.

### Further comments
This change makes `check_cache_backend_for_production` identical to `check_cache_backend` except for the warning vs. error severity (and the error id). This means they can be silenced separately with `SILENCED_SYSTEM_CHECKS`.

Discussed in standup 2025-02-25, for someone to see this, they'd have to be either:
- running `manage.py check` with their production settings as mentioned in the release notes
- running `runserver` in production, which is against best practice

### DIFF
--- a/arches/apps.py
+++ b/arches/apps.py
@@ -32,17 +32,22 @@ if settings.FILE_TYPE_CHECKING not in (None, "lenient", "strict"):
 
 
 ### SYSTEM CHECKS ###
+supported_by_django_ratelimit = (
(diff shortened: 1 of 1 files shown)


aarongundel (approved):
Very helpful!  Thanks for this!

PR #11822, opened 2025-02-26
Assorted query reductions - Feb 2025

jacobtylerwalls:
### Types of changes

### Description of Change
Assorted fixes to reduce Django ORM queries, see commit messages for notes if any.

### DIFF
--- a/arches/app/utils/betterJSONSerializer.py
+++ b/arches/app/utils/betterJSONSerializer.py
@@ -114,7 +114,7 @@ class JSONSerializer(object):
             or isinstance(object, tuple)
             or isinstance(object, set)
         ):
-            return self.handle_list(object)
+            return self.handle_list(object, **kwargs)
         elif isinstance(object, Model):
             if hasattr(object, "serialize"):
                 serialize_function = getattr(object, "serialize")
@@ -187,15 +187,15 @@ class JSONSerializer(object):
 
         return obj
 
-    def handle_list(self, l):
+    def handle_list(self, l, **kwargs):
         """Called to handle a list"""
         arr = []
         for item in l:
-            arr.append(self.handle_object(item))
+            arr.append(self.handle_object(item, **kwargs))
 
         return arr
 
-    # a slighty modified version of django.forms.models.model_to_dict
+    # a slightly modified version of django.forms.models.model_to_dict
     def handle_model(self, instance, **kwargs):
         """
         Returns a dict containing the data in ``instance``.
--- a/arches/app/utils/data_management/resources/formats/format.py
+++ b/arches/app/utils/data_management/resources/formats/format.py
@@ -320,13 +320,17 @@ class Writer(object):
             filters = {"resourceinstance__graph_id": graph_id}
             if user:
                 filters["nodegroup_id__in"] = permitted_nodegroups
-            self.tiles = models.TileModel.objects.filter(**filters)
+            self.tiles = models.TileModel.objects.filter(**filters).select_related(
+                "parenttile", "nodegroup"
+            )
             self.graph_id = graph_id
         else:
             filters = {"resourceinstance_id__in": resourceinstanceids}
             if user:
                 filters["nodegroup_id__in"] = permitted_nodegroups
-            self.tiles = models.TileModel.objects.filter(**filters)
+            self.tiles = models.TileModel.objects.filter(**filters).select_related(
+                "parenttile", "nodegroup"
+            )
             try:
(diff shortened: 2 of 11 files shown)


arches/app/utils/data_management/resources/formats/format.py
```diff
@@ -320,13 +321,17 @@ def get_tiles(self, graph_id=None, resourceinstanceids=None, **kwargs):
             filters = {"resourceinstance__graph_id": graph_id}
             if user:
                 filters["nodegroup_id__in"] = permitted_nodegroups
-            self.tiles = models.TileModel.objects.filter(**filters)
+            self.tiles = models.TileModel.objects.filter(**filters).select_related(
+                "parenttile"
```
jacobtylerwalls: If #11809 lands, we can add "nodegroup", too.

arches/app/utils/betterJSONSerializer.py
```diff
@@ -114,7 +114,7 @@ def handle_object(self, object, **kwargs):
             or isinstance(object, tuple)
             or isinstance(object, set)
         ):
-            return self.handle_list(object)
+            return self.handle_list(object, **kwargs)
```
jacobtylerwalls, line 117: Spotted this this morning, every template render is unnecessarily fetching graph functions, since `value_from_object` fetches objects that we're throwing away if they're included in "exclude", which we're doing here:

https://github.com/archesproject/arches/blob/d27ba10b53f8ca05d9c51ed77c9aab1516661b34/arches/app/views/base.py#L50-L54

```
/Users/jwalls/prj/arches/arches/app/views/base.py in get_context_data(50)
  context["createable_resources"] = JSONSerializer().serialize(

/Users/jwalls/prj/arches/arches/app/utils/betterJSONSerializer.py in serialize(77)
  obj = self.serializeToPython(obj, **options)

/Users/jwalls/prj/arches/arches/app/utils/betterJSONSerializer.py in serializeToPython(74)
  return self.handle_object(obj, **self.options)

/Users/jwalls/prj/arches/arches/app/utils/betterJSONSerializer.py in handle_object(117)
  return self.handle_list(object)

/Users/jwalls/prj/arches/arches/app/utils/betterJSONSerializer.py in handle_list(194)
  arr.append(self.handle_object(item))

/Users/jwalls/prj/arches/arches/app/utils/betterJSONSerializer.py in handle_object(134)
  return self.handle_model(object, **kwargs)

/Users/jwalls/prj/arches/arches/app/utils/betterJSONSerializer.py in handle_model(249)
  qs = f.value_from_object(instance)
```

jacobtylerwalls: It's not clear to me what's supposed to happen if the model being serialized and its related model also being serialized share a field name and that field name is in `exclude`.

chiatt: It looks like the intent was to skip `value_from_object` if an object is in the exclude list: https://github.com/archesproject/arches/blob/d27ba10b53f8ca05d9c51ed77c9aab1516661b34/arches/app/utils/betterJSONSerializer.py#L233-L234, but maybe that's not working as intended.

jacobtylerwalls: Yep that's the crux of the fix I made -- exclude is None because we swallowed **kwargs during the chain, see the stacktrace I pasted, you'll see where kwargs get swallowed.

I think this fix is tenable, I'm just noting it might lead to fields on several related models all being excluded if they share a field name. Seems okay for now.

arches/app/views/notifications.py
```diff
@@ -12,7 +11,7 @@ class NotificationView(View):
     def get(self, request):
         if request.user.is_authenticated:
             if self.action == "get_types":
-                default_types = list(models.NotificationType.objects.all())
+                default_types = models.NotificationType.objects.all()
```
chiatt: This change causes an error because .index, which is a list method and not a queryset method, is used to find the matching notification_types between the user and default querysets: ```i = default_types.index(user_type.notiftype)```. It seems like the logic in this if block could be refactored a bit. I'm not sure if it's worth it though. There are never very many notification types and I doubt this `get` is called very frequently, so any improvement in performance would probably be pretty small.

chiatt (changes requested):

arches/app/views/notifications.py
jacobtylerwalls: Oh good catch, I definitely got carried away. I was seeing that this was called on every page load in every base template, but I think you're right it's not worth it to understand and refactor this for one really tiny query.

chiatt (approved):

PR #11824, opened 2025-02-28
11823 - Allow packages to be loaded as part of an initial migration

### LINKED ISSUE #11823: Packages cannot be loaded as part of an initial migration
There are 2 issues that cause the failure of a package to be loaded as part of a setup_db call:

1. The `load_package` option of the `packages` management command makes a call to the `refresh_geojson_geometries()` postgres function, which fails if it is within a transaction. This causes the migration to fail from a migration as it is in a transaction.
2. the `management.call_command("createcachetable")` call happens after the `management.call_command("migrate")` call in the `setup_db` management command. This causes errors when loading the package graphs because the process tries to delete rows from the `user_permission_cache` table which is not present until after the migration completes.

Fixing these two issues will allow a package to be loaded as part of an initial migration.

bferguso:
### Types of changes

### Description of Change

Resolves 2 issues that cause package load to fail when part of an initial migration.

### Issues Solved

Closes #11823

### Checklist

#### Accessibility Checklist

n/a

#### Ticket Background
*   Sponsored by: Self Funded
*   Found by: @bferguso
*   Tested by: @ 
*   Designed by: @bferguso @jacobtylerwalls

### DIFF
--- a/arches/management/commands/packages.py
+++ b/arches/management/commands/packages.py
@@ -669,7 +669,7 @@ class Command(BaseCommand):
             with connection.cursor() as cursor:
                 cursor.execute("SELECT * FROM refresh_geojson_geometries();")
         else:
-            print("WARNING: Not refreshing geometries - transaction active")
+            self.stdout.write("WARNING: Not refreshing geometries - transaction active")
 
     def load_package(
         self,
@@ -862,7 +862,7 @@ class Command(BaseCommand):
             collection_data = []
             for file_type in file_types:
                 collection_data.extend(
-                    glob.glob(
+                glob.glob(
                         os.path.join(
                             package_dir, "reference_data", "collections", file_type
                         )
--- a/arches/management/commands/packages.py
+++ b/arches/management/commands/packages.py
@@ -665,8 +665,11 @@ class Command(BaseCommand):
 
     @staticmethod
     def update_resource_geojson_geometries():
-        with connection.cursor() as cursor:
-            cursor.execute("SELECT * FROM refresh_geojson_geometries();")
+        if not connection.in_atomic_block:
+            with connection.cursor() as cursor:
+                cursor.execute("SELECT * FROM refresh_geojson_geometries();")
+        else:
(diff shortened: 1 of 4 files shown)


arches/management/commands/packages.py
```diff
@@ -665,8 +665,11 @@ def create_package(self, dest_dir):
 
     @staticmethod
     def update_resource_geojson_geometries():
-        with connection.cursor() as cursor:
-            cursor.execute("SELECT * FROM refresh_geojson_geometries();")
+        if not connection.in_atomic_block:
+            with connection.cursor() as cursor:
+                cursor.execute("SELECT * FROM refresh_geojson_geometries();")
+        else:
+            print("WARNING: Not refreshing geometries - transaction active")
```
jacobtylerwalls: `self.stdout.write()`/`self.stderr.write()` is preferred to `print` in commands for ease of redirecting, esp. during tests

arches/management/commands/packages.py
```diff
@@ -665,8 +665,11 @@ def create_package(self, dest_dir):
 
     @staticmethod
     def update_resource_geojson_geometries():
-        with connection.cursor() as cursor:
-            cursor.execute("SELECT * FROM refresh_geojson_geometries();")
+        if not connection.in_atomic_block:
+            with connection.cursor() as cursor:
+                cursor.execute("SELECT * FROM refresh_geojson_geometries();")
```
jacobtylerwalls, line 669: Can you try removing this workaround in the test framework now?
https://github.com/archesproject/arches/blob/d27ba10b53f8ca05d9c51ed77c9aab1516661b34/tests/base_test.py#L149-L150

That will add some implicit test coverage.

arches/management/commands/packages.py
bferguso: I can change all the `print()` calls to `self.stdout.write()` as part of this PR if that makes sense.

jacobtylerwalls (approved):

chiatt (approved):

PR #11831, opened 2025-03-04
Allow arches apps to depend on either captcha version

jacobtylerwalls:
Follow-up to 4ae44e0.
This is blocking arches application development that intends to support both 7.6 & 8


jacobtylerwalls:
@chiatt -- An alpha of v8 with this in it would be helpful, to allow arches-controlled-lists to loosen its pin on arches.

chiatt (approved):
Looks good!

chiatt:
Disregarding test coverage check because changes don't effect any methods

PR #11832, opened 2025-03-05
Enforce nodegroup permissions against `search_term` results #11826

### LINKED ISSUE #11826: enforce permissions against search_terms
In 7.6, an anonymous user on an instance using default deny can type a phrase in term search and get search terms hints for nodegroups/models they don't have read access on. The `search_terms()` method should inspect nodegroup permissions.

whatisgalen:
### Types of changes

### Description of Change

This PR enforces a filter on the main query in search_terms so that only values under permitted_nodegroups can be searched. ~As a minor improvement, it also updates the logic so that if two different nodegroups both matched the same term, each will get their own listing in the results (previously they were conflated into one of the nodegroups)~ (removed in d0bfbac462190c6999441e3a93a8285b729c97ae).

### Issues Solved

Closes #11826 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/views/search.py
```diff
@@ -180,6 +181,11 @@ def search_terms(request):
 
         if user_is_reviewer is False and index == "terms":
             boolquery.filter(Terms(field="provisional", terms=["false"]))
+            boolquery.filter(
+                Terms(
+                    field="nodegroup_id", terms=[str(ng) for ng in permitted_nodegroups]
+                )
+            )
```
chiatt: This only gets applied if a user isn't a reviewer, but it should probably be applied either way. Otherwise users that are reviewers will get to see terms from restricted tiles.

chiatt (dismissed):

arches/app/views/search.py
whatisgalen: fixed in latest

jacobtylerwalls:
Taking a look

arches/app/views/search.py
```diff
@@ -178,8 +179,15 @@ def search_terms(request):
             Match(field="displayname.value", query=searchString, fuzziness=2, boost=2)
         )
 
-        if user_is_reviewer is False and index == "terms":
-            boolquery.filter(Terms(field="provisional", terms=["false"]))
+        if index == "terms":
+            boolquery.filter(
+                Terms(
+                    field="nodegroupid", terms=[str(ng) for ng in permitted_nodegroups]
```
jacobtylerwalls, line 187: Even 1000 nodegroups shouldn't be a burden on the limit, looks like the term limit is 65536 by default 👍

arches/app/views/search.py
```diff
@@ -205,6 +213,8 @@ def search_terms(request):
         query.add_aggregation(base_agg)
 
         ret[index] = []
+        if len(permitted_nodegroups) == 0:
+            continue
```
jacobtylerwalls: Won't this short-circuit a concepts search when it shouldn't? If it's intentional, let's move it outside the loop to clarify.

releases/7.6.10.md
```diff
@@ -2,6 +2,8 @@
 
 ### Bug Fixes and Enhancements
 
+-   Search Term queries now enforce nodegroup permissions [#11832](https://github.com/archesproject/arches/pull/11832)
+-
```
jacobtylerwalls:
```suggestion
```

tests/search/search_tests.py
```diff
@@ -169,6 +176,52 @@ def test_search_terms(self):
         request.user = self.tester
         response = search_terms(request)
         result = {}
+        try:
+            result = json.loads(response.content)
+        except json.decoder.JSONDecodeError:
+            print("Failed to parse search result")
```
jacobtylerwalls, line 182: What failure condition is being tested here? Maybe remove the except and let it fail so we know it's flaky? EDIT: oh, it's blending it with surrounding tests, I leave it to you then 👍

tests/search/search_tests.py
```diff
@@ -169,6 +176,52 @@ def test_search_terms(self):
...
+        self.assertTrue("terms" in result and len(result["terms"]) == 0)
+        remove_perm("no_access_to_nodegroup", self.tester, nodegroup)
+        new_tile.delete()
+        resource.delete()
+        self.sync_es(se)
```
jacobtylerwalls: We sync_es in the setUp() before each test, can we remove all these calls, or is there a nuance (then please add a comment)?
```suggestion
```

tests/search/search_tests.py
```diff
@@ -169,6 +176,52 @@ def test_search_terms(self):
...
+        resource.graph_id = "c9b37a14-17b3-11eb-a708-acde48001122"
+
+        resource.save(user=admin_user, transaction_id=uuid.uuid4())
+        tile_data = {}
+        tile_data[nodeid] = {
+            "en": {"value": "Alpha Gamma Delta Eta Epsilon", "direction": "ltr"}
+        }
+        new_tile = Tile(
+            tileid=uuid.UUID(tileid),
+            resourceinstance_id=resourceinstanceid,
```
jacobtylerwalls, line 205: Thanks for using a UUID for tile -- let's do that for resource and nodegroup also, so we don't add tech debt on top of #10877 (I'm hoping we'll get a CI baseline for that and "plug the leak before bailing out the leak")

jacobtylerwalls (commented):
Tests are super helpful, thanks. One substantive question about concept queries, then some very minor feedback.

arches/app/views/search.py
chiatt: I agree - if a user doesn't have access to any nodegroups, then it seems like this should be moved out of the loop.

apeters: I think we should check this first thing and if it's empty immediately return an empty array to the front end

arches/app/views/search.py
whatisgalen: Perhaps this should interrogate which permission backend is in use to pick an allow vs a deny list?

arches/app/views/search.py
whatisgalen: change in latest

jacobtylerwalls (approved):

arches/app/views/search.py
jacobtylerwalls: Could always look into that if it becomes an issue.

PR #11834, opened 2025-03-05
updates project github postgis image to match Arches core #11833

### LINKED ISSUE #11833: Update `.github/workflows/main.yml` to use latest postgis image
Otherwise the build fails. the latest postgis image is: `postgis:15-3.3`

chrabyrd:
### Types of changes

### Issues Solved

Closes #11833 

### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
 
     services:
       postgres:
-        image: postgis/postgis:14-3.4
+        image: postgis/postgis:15-3.3
         env:
           POSTGRES_PASSWORD: postgis
           POSTGRES_DB: arches
@@ -52,7 +52,7 @@ jobs:
 
     services:
       postgres:
-        image: postgis/postgis:14-3.4
+        image: postgis/postgis:15-3.3
         env:
           POSTGRES_PASSWORD: postgis
           POSTGRES_DB: arches
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
 
     services:
       postgres:
-        image: postgis/postgis:15-3.3
+        image: postgis/postgis:14-3.4
         env:
           POSTGRES_PASSWORD: postgis
           POSTGRES_DB: arches
@@ -52,7 +52,7 @@ jobs:
 
     services:
       postgres:
-        image: postgis/postgis:15-3.3
+        image: postgis/postgis:14-3.4
         env:
           POSTGRES_PASSWORD: postgis
           POSTGRES_DB: arches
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -52,7 +52,7 @@ jobs:
 
     services:
       postgres:
-        image: postgis/postgis:14-3.4
+        image: postgis/postgis:15-3.3
```
aarongundel: What is the latest version of postgres we (arches) support?  IIRC it's 16 for 8 - I know that versions I've deployed in AWS use 16 without issue.  Might be worth upgrading to there.  17 is available too.  https://hub.docker.com/r/postgis/postgis/

releases/8.0.0.md
```diff
@@ -42,6 +42,7 @@ Arches 8.0.0 Release Notes
 - Fix issue that produced an error when cycling between the dropdown options in Advanced search [#11442](https://github.com/archesproject/arches/issues/11442)
 - Add message when copying geojson url to clipboard [#11076](https://github.com/archesproject/arches/issues/11076)
 - Hide "view report" button in Arches system settings [#11659](https://github.com/archesproject/arches/issues/11659)
+- Updates github workflows postgis image to latest [#11833](https://github.com/archesproject/arches/issues/11833)
```
aarongundel: if not upgrading to latest, we should probably change latest to 15.

aarongundel (changes requested):
nits re: latest

.github/workflows/main.yml
jacobtylerwalls: If we have to make a choice between testing the oldest supported and the newest supported, I would suggest testing older, just to catch pull requests accidentally making use of 16+ functionality, which has happened before.

aarongundel: @jacobtylerwalls good point - do we have an official oldest supported version?

aarongundel: postgres roadmap suggests nothing earlier than 13 (as do the available dockerhub images)
https://www.postgresql.org/support/versioning/

jacobtylerwalls: [I think it's 14](https://github.com/archesproject/arches/blob/4f69be0d1805840d432a61600c44880155df02a6/releases/7.6.0.md#upgrading-an-arches-project), but if we can't get that on the CI runners, then 15 is fine.

jacobtylerwalls (approved):

.github/workflows/main.yml
chrabyrd: actually that's fair -- happy to update this PR to not update the Arches postgis image. When it got updated previously the `main.yml` file in `install/arches-templates` didn't , and v13 was causing build failures for projects

aarongundel (approved):
LGTM

PR #11840, opened 2025-03-10
Support rule perms for guest user

chiatt:
### Types of changes

### Description of Change
Removes calling `append` on resource_ids because it errors if resource_ids is a queryset.


aarongundel (approved):
LGTM

PR #11850, opened 2025-03-13
Fix spatial view slug validator message #11849

### LINKED ISSUE #11849: Spatial view slug validator doesn't match help text
The spatial view slug field validator help text says hyphens are accepted, but they aren't. It should probably say underscore.

jacobtylerwalls:
### Types of changes

### Issues Solved

Closes #11849

### Checklist


samuelscandrett (approved):
Looks good to me.  
Are we happy to exclude hyphens from being used, with the reasoning being that if a space-like character is required then it should be an underscore?

jacobtylerwalls:
Thanks for having a look. Yes, I assume that was the intent. But interestingly, our other slug fields use Django's [validate_slug](https://docs.djangoproject.com/en/5.1/ref/validators/#validate-slug) and therefore accept hyphens too. I just wouldn't want to make a change here without seeing if it breaks the underlying integration with postgres.

I just tested node aliases, and hyphens are (at the postgres level?) silently converted to underscores.

PR #11813, opened 2025-02-24
Model how to log failed requests #11812

jacobtylerwalls:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
Show how to configure LOGGING to log failed web requests, e.g. to see the full python traceback. Include a note for upgrading, and update the project template.

### Issues Solved

Closes #11812

#### Ticket Background
*   Sponsored by: Farallon
*   Found by: @jacobtylerwalls

Issue #11739, opened 2025-01-22
Add middleware to patch user.has_perm


jacobtylerwalls:
As of 7.6, the permissions framework is accessed via a "backend", not via the django-native `has_perm()` method on the `User`.

If existing code or new code calls that `has_perm()` method on the User, though, the permissions framework will not be consulted.

We could add a middleware class that intercepts `request.user` and wraps `has_perm()` so that it either:
- raises
or
- does the right thing, i.e. calls permission backend's `has_perm()` method

We could also add a system check that checks that this middleware is configured. (If not configured, the severity could be error or perhaps only error if PERMISSION_FRAMEWORK is anything other than the default, assuming default behaves the same way as the django-native has_perm().)

jacobtylerwalls:
Would need to do this in a way affecting arches permissions only, not other 3rd party apps.

PR #11853, opened 2025-03-13
User preferences

### LINKED ISSUE #11842: User Preferences feature
Create a User Preferences model for storing custom user config.  

Originally considered as a part of https://github.com/archesproject/arches/pull/11810, but it also has applications in user specific saved searches, and for integration with Arches applications.

The user preferences table needs the user, potentially a key with the name of the application (e.g. arches), a key (such as 'saved_search' or 'layer_order'), and a value or config column (JSON) that stores the preference.

samuelscandrett:
### Types of changes

### Description of Change

A model for storing preferences for core (or application) functionality as JSON config for users.  

### Issues Solved

Closes #11842

### Checklist

#### Accessibility Checklist

#### Ticket Background


arches/app/models/models.py
```diff
@@ -2265,6 +2265,17 @@ def to_json(self):
         }
 
 
+class UserPreferences(models.Model):
+    userpreferenceid = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
```
chiatt: App developers will likely want to allow users to define preferences for their app. If there is only one preference record per user, then all apps would have to edit the same JSON column. This doesn't seem terribly risky, but conflicts could arise. I think it would be safer to make this a foreign key rather than a OneToOne. Then each app could have its own preferences record per user.

samuelscandrett: Good spot, that's something I overlooked.  I have now changed this to foreign key.

arches/urls.py
```diff
@@ -781,6 +781,11 @@
         api.SpatialView.as_view(),
         name="spatialview_api",
     ),
+    re_path(
```
jacobtylerwalls, line 777: Don't feel the need to change this, but I'm hoping we'll stop using `re_path` in favor of a simpler `path` for new URLs, and avoid having this optional trailing slash. This is probably a larger discussion, so just raising awareness 🥁

tests/models/user_preference_model_tests.py
```diff
@@ -0,0 +1,61 @@
...
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+"""
+
+import random
```
jacobtylerwalls: Several unused imports here you might remove.

tests/models/user_preference_model_tests.py
```diff
@@ -0,0 +1,61 @@
...
+from arches.app.models.models import UserPreference
+from arches.app.utils.data_management.resources.importer import BusinessDataImporter
+from tests import test_settings
+
+# these tests can be run from the command line via
+# python manage.py test tests.models.user_preference_model_tests --settings="tests.test_settings"
+
+
+class UserPreferenceTests(ArchesTestCase):
+    def generate_user_preference(self):
```
jacobtylerwalls: Test data reused in multiple tests is usually created in `setUpTestData()` so that it runs once per class.
```suggestion
    @classmethod
    def setUpTestData(cls):
        cls.user_preference = UserPreference.objects.create(
            ...
            
```

tests/models/user_preference_model_tests.py
```diff
@@ -0,0 +1,61 @@
...
+        user_preference.userpreferenceid = uuid.uuid4()
+        user_preference.user = models.User.objects.get(username="admin")
+        user_preference.preferencename = "test preference"
+        user_preference.config = [
+            {"overlayid": "7d0dffba-5bcf-4694-a5d7-3425ad97fa2b", "sortorder": 1},
+            {"overlayid": "85c0cdd4-95d0-4350-bd9d-729f326fe1d5", "sortorder": 2},
+        ]
+        return user_preference
+
+    def test_user_preference_passes(self):
```
jacobtylerwalls: Then you can remove this

tests/models/user_preference_model_tests.py
```diff
@@ -0,0 +1,61 @@
...
+        self.assertTrue(
+            UserPreference.objects.filter(
+                userpreferenceid=user_pref.userpreferenceid
+            ).exists()
+        )
+
+    def test_user_preference_invalid_user(self):
+        user_pref = self.generate_user_preference()
+
+        with self.assertRaises(Exception):
```
jacobtylerwalls: More specific exception types make these assertions more readable, but looking closer, this just seems to test generic Django model functionality, so I'd probably remove it?

tests/models/user_preference_model_tests.py
jacobtylerwalls: But see below, we may not need this module unless it's a placeholder for future work.

arches/app/models/models.py
```diff
@@ -2271,6 +2271,22 @@ def to_json(self):
         }
 
 
+class UserPreference(models.Model):
+    userpreferenceid = models.UUIDField(primary_key=True, default=uuid.uuid4)
```
jacobtylerwalls: You might add db_default for consistency with other models.

arches/app/models/models.py
```diff
@@ -2271,6 +2271,22 @@ def to_json(self):
...
+    user = models.ForeignKey(
+        User,
+        to_field="username",
+        on_delete=models.PROTECT,
```
jacobtylerwalls: @chiatt should we block User deletes like this?

arches/app/models/models.py
```diff
@@ -2271,6 +2271,22 @@ def to_json(self):
...
+        null=False,
+    )
+    preferencename = models.TextField(blank=True, null=True)
+    config = JSONField(blank=False, null=False)
```
jacobtylerwalls, line 2441: Thanks for making this not nullable.

arches/app/models/models.py
```diff
@@ -2271,6 +2271,22 @@ def to_json(self):
...
 
+class UserPreference(models.Model):
+    userpreferenceid = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    user = models.ForeignKey(
+        User,
+        to_field="username",
+        on_delete=models.PROTECT,
+        null=False,
+    )
+    preferencename = models.TextField(blank=True, null=True)
```
jacobtylerwalls: Django recommends making CharFields not nullable to avoid two representations, empty and null--unless you need null to support multiple nulls when all other values otherwise satisfy a unique constraint. What do you think, is there a use case for this being null and blank=True?

jacobtylerwalls: I'd use CharField with a max_length, TextField signifies larger blobs

jacobtylerwalls: Speaking of which, should we have a models.UniqueConstraint on the combination of preferencename and user, or should we expect callers to sort them and know which one to prefer?

jacobtylerwalls: I'm wondering if we should also have a column for project name / arches application to prevent namespace collisions? And make that one not nullable?

arches/app/models/models.py
```diff
@@ -2271,6 +2271,22 @@ def to_json(self):
         }
 
 
+class UserPreference(models.Model):
+    userpreferenceid = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    user = models.ForeignKey(
+        User,
+        to_field="username",
+        on_delete=models.PROTECT,
+        null=False,
```
jacobtylerwalls, line 2435: You might consider setting `related_name` and `related_query_name` from the get-go to be able to replace:

```py
User.objects.exclude(userpreference__config={})
```
with
```py
User.objects.exclude(preference__config={})
```
And
```py
admin.userpreference_set.all()
```
with
```py
admin.preferences.all()
```

Here it seems nice because "user" is repetitive, but that's just one opinion, I leave it to you :-)

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
                 status=400,
             )
         return JSONResponse(status=204)
+
+
+class UserPreference(APIBase):
```
jacobtylerwalls: I would get the word view in this name somehow. See next comment, maybe this becomes UserPreferenceListView.

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+    def get(self, request, identifier=None):
```
jacobtylerwalls: This is handling both list and detail views. I would normally see that decomposed into two separate views. Then you have less if/else switching inside the view.

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+    @method_decorator(group_required("Application Administrator", raise_exception=True))
+    def post(self, request, identifier=None):
+        if identifier:
+            return JSONErrorResponse(
+                _("User Preference creation failed"),
+                _(
+                    "POST request should not have a userpreferenceid provided in the URL"
+                ),
+                status=400,
+            )
```
jacobtylerwalls: And if the detail view is the only view implementing a post, then you don't have to check for this error condition.

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+class UserPreference(APIBase):
+    def get(self, request, identifier=None):
+        """
+        Returns specific user preference when given uuid
+        If no id specified, and depending on user permissions, returns a list of all user preferences
+        """
+
+        # check user permission
+        overall_access = False
+        if request.user.groups.filter(name="Application Administrator").exists():
```
jacobtylerwalls: I think we have to start using the permissions framework for this, unless/until archesproject/arches#11739

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+            try:
+                returned_user_preference = models.UserPreference.objects.get(
+                    pk=identifier
+                )
+            except:
+                return JSONErrorResponse(
+                    _("User Preference GET request failed"),
+                    _("No User Preference with this id"),
+                    status=404,
+                )
```
jacobtylerwalls: Interested if @chiatt has a view on whether adding these strings and localizing them is necessary versus just only providing the error code? I'm concerned we're bloating the number of strings to translate.

If we want a message, we could prefer implementing a generic one, possibly parameterized to take a model name, inside our JSONErrorResponse class? (This is kind of the burden of implementing our own REST framework.)

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+                response_data = returned_user_preference
+            else:
+                return JSONErrorResponse(
+                    _("User Preference GET request failed"),
+                    _("You do not have access to view this userpreferenceid"),
+                    status=403,
+                )
+        else:
+            if overall_access:  # give view to everything
+                response_data = [x for x in models.UserPreference.objects.all()]
```
jacobtylerwalls: You can just pass a queryset to JSONResponse.

arches/app/views/api/__init__.py
jacobtylerwalls: I'd suggest moving this out of `__init__.py`, which is getting too large.

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+        try:
+            user_pref_json = JSONDeserializer().deserialize(request.body)
+        except ValueError:
+            return JSONErrorResponse(
+                _("Invalid JSON data"),
+                _("The User Preference API was sent invalid JSON"),
+                status=400,
+            )
+
+        if user_pref_json:
```
jacobtylerwalls: I would invert this if to reduce indentation below.

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+            if user_pref_json["userpreferenceid"]:
```
jacobtylerwalls: Handle KeyError?

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+                    ),
+                    status=400,
+                )
+            else:
+                try:
+                    models.User.objects.get(username=user_pref_json["user"])
+                except:
+                    return JSONErrorResponse(
+                        _("Invalid user"),
+                        _("The User Preference API includes an invalid user."),
```
jacobtylerwalls: Handle KeyError separately so the message is accurate?

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+                        status=400,
+                    )
+                new_user_preference = models.UserPreference()
+                new_user_preference.userpreferenceid = uuid.uuid4()
```
jacobtylerwalls:
```suggestion
```

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+                new_user_preference.user = models.User.objects.get(
+                    username=user_pref_json["user"]
+                )
```
jacobtylerwalls: Either just set `.user_id` or save the object you retrieved above to avoid re-querying.

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+                new_user_preference.preferencename = user_pref_json["preferencename"]
+                new_user_preference.config = user_pref_json["config"]
```
jacobtylerwalls: Might be nice to centralize the handling of KeyError up front somewhere.

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+                return JSONErrorResponse(
+                    _("User Preference delete failed"),
+                    _("User Preference does not exist"),
+                    status=404,
+                )
+
+            try:
+                user_preference.delete()
+            except Exception as e:
+                logger.error(e)
```
jacobtylerwalls: I don't think we need to catch errors here, I can't think of a realistic error condition.

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+                return JSONErrorResponse(
+                    _("User Preference delete failed"),
+                    _("An error occurred when trying to delete the spatialview"),
```
jacobtylerwalls: heads up in case you choose to preserve the try/catch: says spatialview

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,128 @@ def delete(self, request, identifier=None):
...
+                    status=500,
+                )
+
+        else:
```
jacobtylerwalls: Same as above, no need to handle for this if this is just a method on UserPreferenceDetailView with a mandatory identifier component

tests/views/user_preference_api_tests.py
```diff
@@ -0,0 +1,181 @@
...
+from arches.app.models import models
+from tests import test_settings
+from django.test.client import Client
+
+
+# these tests can be run from the command line via
+# python manage.py test tests.views.user_preference_api_tests --settings="tests.test_settings"
+
+
+class UserPrefApiTest(TransactionTestCase):
```
jacobtylerwalls:
```suggestion
class UserPrefApiTest(ArchesTestCase):
```
I don't think we need TransactionTestCase. That's a last resort if the functionality under test uses transactions / triggers / anything that would disrupt the django test runner from using transactions to restore state between tests. Spatial views ran triggers.

tests/views/user_preference_api_tests.py
```diff
@@ -0,0 +1,181 @@
...
+
+    serialized_rollback = True
+
+    def generate_user_preference(self):
```
jacobtylerwalls: See above: re: setUpTestData

tests/views/user_preference_api_tests.py
```diff
@@ -0,0 +1,181 @@
...
+            "user": username,
+            "preferencename": "test preference",
+            "config": [
+                {"overlayid": "7d0dffba-5bcf-4694-a5d7-3425ad97fa2b", "sortorder": 1},
+                {"overlayid": "85c0cdd4-95d0-4350-bd9d-729f326fe1d5", "sortorder": 2},
+            ],
+        }
+        return user_pref
+
+    def test_user_preference_api(self):
```
jacobtylerwalls: Thanks for providing such thorough tests. Kind of a shame we don't have a framework to delegate more of this to. I can take a closer look on the next iteration once some of these error conditions go away, but you might consider breaking some of these out into discrete tests.

jacobtylerwalls (commented):
Thanks @SDScandrettKint, this looks great! Appreciate the really close attention to detail.

I can tell this was closely modeled on the spatial views api, but I think we can keep iterating to improve the patterns and set a great bar for the next one, so I left some comments, let me know what you think :-)

arches/app/models/models.py
```diff
@@ -2271,6 +2271,22 @@ def to_json(self):
         }
 
 
+class UserPreference(models.Model):
+    userpreferenceid = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    user = models.ForeignKey(
+        User,
+        to_field="username",
```
jacobtylerwalls, line 2433: On first pass I didn't notice that this had a to_field. Can you share why? If we keep it, I'd suggest adjusting the name of this field.

arches/urls.py
samuelscandrett: That is a good point.  Perhaps an issue can be raised to discuss this?

arches/app/models/models.py
samuelscandrett: Good point on the nullable TextField, there is no reason why it should be null.  Empty names would be best off with an empty string.

Having a unique constraint between the preference name and user is probably wise.  It depends how we plan to use this model in core, e.g. if we had a built-in preference set to the anonymous user, or dynamic preferences to active users (overlay orders, search result defaults etc), we wouldn't want the possibility of another preference with the same name messing with things.

tests/models/user_preference_model_tests.py
samuelscandrett: Good spot - this was from using an existing test as a template.  Now removed

arches/app/models/models.py
samuelscandrett: Added in [6bf2332](https://github.com/archesproject/arches/pull/11853/commits/6bf2332e60c293768314187a62fae817fad35227)

arches/app/models/models.py
samuelscandrett: Changes added in [6bf2332](https://github.com/archesproject/arches/pull/11853/commits/6bf2332e60c293768314187a62fae817fad35227).

I have noticed that the constraint is case sensitive, so a preferences with names of 'Test' and 'test' are both allowed.  Not sure if we should be strict on this? @jacobtylerwalls

samuelscandrett: Have changed to CharField and set a max length of 255 (same as the workflow model).

samuelscandrett: I agree.  It was something I planned on adding from the start, but I was interested in the thoughts from others first.  I'd suggest it is called `appname`, not nullable and maybe also not blank?  The default appname could be 'arches'.

@chiatt what are your thoughts?

arches/app/views/api/__init__.py
samuelscandrett: I agree `__init.py__` is getting very long.  Might be wise to restructure `__init__.py` in a separate PR?  I'm not sure what the best structure would be?  Perhaps directories for models, views, search etc.

jacobtylerwalls: That's totally fair. I opened #11894 for it. 👍

arches/app/views/api/__init__.py
jacobtylerwalls: (never mind about #11739, that wouldn't help because we're not using has_perm() here, but my understanding is we need to find the right util from arches.app.utils.permission_backend to call here)

arches/app/models/models.py
jacobtylerwalls: Smart idea. There's an [example](https://docs.djangoproject.com/en/5.1/ref/models/constraints/#django.db.models.UniqueConstraint.expressions) in the docs for adding a case insensitive constraint via `Lower()`

arches/app/views/api/__init__.py
jacobtylerwalls: Checked with @aarongundel and I was mistaken -- it's fine to check Django groups here. The permissions framework doesn't replace / stand in for django user groups.

arches/app/models/models.py
samuelscandrett: Added in [77a369e](https://github.com/archesproject/arches/pull/11853/commits/77a369ea9fa604b14fa898e21b4550ec81de4941), also added a test here [06ce771](https://github.com/archesproject/arches/pull/11853/commits/06ce77165f9b9cf7a5169fe959273d5b678704d0).  Interestingly I had to change the testcase to `TransactionTestCase` in order for the db UniqueConstraint trigger to not break and allow the third, case insensitive test to work.

arches/app/views/api/__init__.py
samuelscandrett: Separated into two separate views in [570db54](https://github.com/archesproject/arches/pull/11853/commits/570db542dd7810ce68e90e9f160f2b81852215f4)

tests/models/user_preference_model_tests.py
samuelscandrett: I have added a setUpTestData() function in the API tests, as it makes sense to have some test cases in the db for testing GET and DELETE.  I couldn't fit this function in the model tests, as they are testing saving new UserPreferences with potential issues.

arches/app/views/api/__init__.py
samuelscandrett: I can see the need for deleting with an invalid identifier, thus an exception on `.get(pk=identifier)`.  The latter try `.delete()` might be useful to catch any internal errors that may occur if delete fails (for whatever reason).  I can't see this one being used often, but probably good to have?

arches/urls.py
samuelscandrett: Have raised here https://github.com/archesproject/arches/issues/11919

arches/app/views/api/__init__.py
```diff
@@ -2408,73 +2489,21 @@ def get(self, request, identifier=None):
...
-                )
-                new_user_preference.preferencename = user_pref_json["preferencename"]
-                new_user_preference.config = user_pref_json["config"]
-                new_user_preference.save()
-                return JSONResponse(new_user_preference, status=201)
-
-        return JSONErrorResponse(_("No json request payload"), status=400)
-
     @method_decorator(group_required("Application Administrator", raise_exception=True))
     def delete(self, request, identifier=None):
```
jacobtylerwalls: I would make identifier a required param and reduce indentation below. (forgive me if you were already in the process of making this change!)

arches/app/views/api/__init__.py
```diff
@@ -2385,17 +2385,98 @@ def delete(self, request, identifier=None):
...
+                status=400,
+            )
+
+
+class UserPreferenceDetailView(APIBase):
+    """
+    Takes an identifier.  Use for GET or DELETE requests to a specific userpreferenceid.
+    """
+
     def get(self, request, identifier=None):
```
jacobtylerwalls: Same here, remove the `=None` and deindent below (and remove the unreachable error condition)

arches/app/views/api/__init__.py
```diff
@@ -2385,17 +2385,98 @@ def delete(self, request, identifier=None):
...
+                _("The User Preference API includes an invalid user."),
+                status=400,
+            )
+
+        try:
+            new_user_preference = models.UserPreference()
+            new_user_preference.user = preference_user
+            new_user_preference.preferencename = user_pref_json["preferencename"]
+            new_user_preference.config = user_pref_json["config"]
+            new_user_preference.save()
```
jacobtylerwalls: Because we're not using a form or a serializer (we're kind of doing a serializer on-the-fly here by manually drilling into all these key-value pairs), we need to call .`full_clean()` once we're done parsing and before `save()` to run the Django model validation like checking max_length on preferencename. I wouldn't expect `save()` to raise a ValidationError.

arches/app/views/api/__init__.py
```diff
@@ -2385,17 +2385,98 @@ def delete(self, request, identifier=None):
...
-class UserPreference(APIBase):
+class UserPreferenceListCreateView(APIBase):
+    """
+    Does not take an identifier, and represents response as a list.  Use for GET and POST requests.
+    """
+
+    def get(self, request):
+        """
+        Returns a list of all current User's preferences.
+        Depending on permissions, returns a of all user preferences.
```
jacobtylerwalls: a -> a list

arches/app/models/models.py
jacobtylerwalls: Discussed on Slack: makes it easier for the application admin to retrieve a list of all preferences and have the results serialize with the username instead of id.

That works for me. If keeping it, then I suggest changing the name of this field to `username` and adjusting the keys in the post payload to `"username"`.

arches/app/views/api/__init__.py
jacobtylerwalls: Django has a global error handler. We made it behave a little better with our arches patterns in #11784, and we modeled how to configure logging naturally in #11813. I think we should delegate to that.

If we eagerly catch and log here, we will duplicate log entries, we won't get the Django debug view during development, and we will create additional strings for translators.

tests/models/user_preference_model_tests.py
```diff
@@ -0,0 +1,66 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+"""
+
+import uuid
+from django.test import TransactionTestCase
```
jacobtylerwalls: unused

tests/views/user_preference_api_tests.py
```diff
@@ -0,0 +1,193 @@
...
+
+        user_pref = self.user_preference_json_data("admin")
+        response = self.client.post(
+            reverse("api_user_preference_list_view"),
+            data=user_pref,
+            content_type="application/json",
+        )
+        self.assertEqual(response.status_code, 201)
+        response_json = json.loads(response.content)
+        self.assertTrue("userpreferenceid" in response_json.keys())
```
jacobtylerwalls: assertIn provides nicer output if this fails, rather than `False is not true`

tests/views/user_preference_api_tests.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def test_list_get_administrator_response(self):
+        # Tests admin users can see all user preferences, not just their users
+        self.client.login(username="admin", password="admin")
+        response = self.client.get(
+            reverse("api_user_preference_list_view"),
+        )
+        self.assertEqual(response.status_code, 200)
+        response_json = json.loads(response.content)
+        self.assertTrue(isinstance(response_json, list))
```
jacobtylerwalls: same, I almost always avoid assertTrue for that reason

tests/models/user_preference_model_tests.py
```diff
@@ -0,0 +1,66 @@
...
+            {"overlayid": "85c0cdd4-95d0-4350-bd9d-729f326fe1d5", "sortorder": 2},
+        ]
+        return user_preference
+
+    def test_user_preference_invalid_user(self):
+        user_pref = self.generate_user_preference()
+
+        with self.assertRaises(ValueError):
+            user_pref.user = "admin"
+            user_pref.save()
```
jacobtylerwalls: I think I'd still remove tests like this -- we don't need to test the Django framework.

tests/models/user_preference_model_tests.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def test_unique_constraint(self):
+        # User preference number 1 (valid and no errors)
```
jacobtylerwalls: I think I'd also say that here, but I could could be convinced to keep it if you think it serves a documentation purpose (but I think the model serves that documentation purpose).

jacobtylerwalls (commented):
Looking really great. Just piping up with responses to your latest questions.

arches/app/views/api/__init__.py
samuelscandrett: Good spot, was clearly typing away without thinking!

arches/app/views/api/__init__.py
samuelscandrett: Would we be better off doing the creation in one go, i.e. using `UserPreference.objects.create()` rather than adding each of the components on a new line, or would that not make any difference?

jacobtylerwalls: You could achieve the same effect by instantiating the model instance with keyword args:
```py
UserPreference(
    preference_name=user_pref_json(["preferencename"]),
    ...
```

But my point was less about how we're doing the one-by-one drilling and more about the _fact_ that we're rolling our own serializer instead of depending on a framework that will have already run the equivalent of all the django model validations on the incoming data. And how that fact obligates us to call `full_clean()` ourselves.

jacobtylerwalls: (The spatial views api calls full_clean() for this same reason. I can get you some links to read more if you like.)

arches/app/views/api/__init__.py
samuelscandrett: Identifier made a requirement.  I noticed that the API can be called without the identifier, but it logs the error of `No User Preference with this id` which isn't exactly accurate.  I've kept `if not identifier` with a log message of `No userpreferenceid specified`.  Does this seem sensible?

[dd13781](https://github.com/archesproject/arches/pull/11853/commits/dd13781ea7ff60dcb0e6ddfa9f8980d81b9ddd54)

tests/models/user_preference_model_tests.py
samuelscandrett: Yeah that is a fair point.  It makes me wonder what aspects of the model should we be testing?  There are only two at the moment (for invalid user and db constraints) and you are correct that they're covered by the model.  Does this feature need any model tests, or does it only need API tests?

jacobtylerwalls: I think only api tests. Add tests that exercise our custom code. Django model functionality is tested by them. :-)

arches/app/views/api/__init__.py
jacobtylerwalls: My thought was that the API shouldn't be called without the identifier. I was assuming you would remove the `?` from the route that resolves to the detail view.

samuelscandrett: Yes, you're correct.  Previously, the url was `re_path(r"^api/user_preference$",` which meant going to `api/user_preference/` would hit the detailView and the respective exception because an identifier was not provided.  Changing the url to `re_path(r"^api/user_preference/$",` now means that going to `api/user_preference/` hits the listView which is the expected route.  
This also means I can delete the `if not identifier` clauses.

arches/app/models/models.py
chiatt: I think we would actually want deletion of users to cascade to user preferences rather than block user deletes.

chiatt (changes requested):

tests/views/user_preference_api_tests.py
samuelscandrett: Replaced with `self.assertIsInstance(response_json, list)`

samuelscandrett: Also replaced the one below with `self.assertEqual(len(response_json), 2)`

tests/models/user_preference_model_tests.py
samuelscandrett: File now deleted.

tests/models/user_preference_model_tests.py
samuelscandrett: File deleted.

arches/app/models/models.py
jacobtylerwalls: Fine to skip this for now

arches/app/views/api/__init__.py
```diff
@@ -2383,3 +2383,155 @@ def delete(self, request, identifier=None):
...
+                        "POST REST request should not have a userpreferenceid provided in the JSON data."
+                    ),
+                    status=400,
+                )
+        except KeyError:
+            pass
+
+        try:
+            user_pref_json.get("user")
+        except KeyError:
```
jacobtylerwalls: get() won't raise KeyError

arches/app/models/models.py
samuelscandrett: Changes from user to username in [d76dcfc](https://github.com/archesproject/arches/pull/11853/commits/d76dcfcf196ff69bcb039586a4853199387ad5c5)

arches/app/models/models.py
samuelscandrett: Changed to cascade [d76dcfc](https://github.com/archesproject/arches/pull/11853/commits/d76dcfcf196ff69bcb039586a4853199387ad5c5)

arches/app/views/api/__init__.py
samuelscandrett: Ah I see.  So if we just include `user_preference.delete()` without the try block, and an error occurs, the global Django handler will raise the error?

arches/app/views/api/__init__.py
samuelscandrett: I also spotted this when renaming all occurences of `user` with `username`.  I've added an inner try catch block for this and a test to cover it [497f75a](https://github.com/archesproject/arches/pull/11853/commits/497f75a484fc13d6d63f1b16585d991f7e0aa452).

arches/app/views/api/__init__.py
jacobtylerwalls: Yes, it will log the error depending on your LOGGING setting, and it will transform the python error to an HTML error response depending on your project's default error views -- we added instructions to 8.0 to set a default error view that's smart enough to return JSON for api views, and a sane LOGGING setting.

(#11784 was the fix for the errors you see with garbage like "Doctype is not valid ..." where a client is expecting to parse JSON but Django returned HTML instead.)

arches/app/models/models.py
chiatt: That sounds good to me

arches/app/views/api/__init__.py
jacobtylerwalls: (Happy to punt this to a new issue.)

arches/app/models/models.py
chiatt: Does seem fine to skip it, but if it's not much trouble, I support the change.

arches/app/views/api/__init__.py
chiatt: Yeah, the 404 status seems sufficient

arches/app/models/models.py
samuelscandrett: Added in [59c2d22](https://github.com/archesproject/arches/pull/11853/commits/59c2d22f089e19640a899c39b97e84d316279033).
```
        related_name="preference",
        related_query_name="preferences",
```

arches/app/models/models.py
samuelscandrett: Added the appname field in [e81dd24](https://github.com/archesproject/arches/pull/11853/commits/e81dd2468f8580b1e8ae5c4ee3c03e4f99a7e696).  
This got me wondering whether the constraints between user and preferencename should be application specific?  Could there be scenarios where an `arches` preference and an app preference could have the same name and be correct?  For example, a core `arches` preference and a `lingo` preference with the name `default_search_results`, both concerning different functionality but requiring the same name.

Also, should `preferencename` allow blank, or should a name be required?

arches/app/views/api/__init__.py
samuelscandrett: Messages are often nice to see, so I agree with the idea of implementing a generic, parameter-based response.  Putting this into a new issue seems like a good idea.

arches/app/models/models.py
jacobtylerwalls: Good point, I would expect the constraint to be loosened to accept the triple of user, preference name, app name.

I'd expect preferencename to be blank only if we were inserting a default value ourselves on save. Since we're not, let's make it blank=False (or just remove that kwarg entirely) so as to require it when interacting with the Django admin or when calling `full_clean()` in the API.

arches/app/views/api/__init__.py
jacobtylerwalls: Just went with something flat for now in #11903.

chiatt (changes requested):
Now that @jacobtylerwalls has moved the API views out of __init__.py and into their own files, I think it makes sense to change this PR to follow that pattern.

arches/app/views/api/__init__.py
```diff
@@ -1,2498 +1,24 @@
...
+from .ontology import *
+from .plugin import *
+from .resource import *
+from .search import *
+from .spatialview import *
+from .tile import *
+from .user import *
+from .validator import *
+from .workflow import *
+from .userpreference import *
```
jacobtylerwalls: (Do you mind alphabetizing this addition?)

arches/app/models/models.py
samuelscandrett: I've removed `blank=False` from both `appname` and `preferencename`.  I've also adjusted the constraint to include all three of `appname`, `preferencename` and `username`.

arches/app/views/api/__init__.py
samuelscandrett: Oops yes I see that now!  Now userpreference is now below the .user import

arches/app/views/api/__init__.py
samuelscandrett: Added here: [599b603](https://github.com/archesproject/arches/pull/11853/commits/599b6034560694edf1eda568fc52926be8b5047a)

arches/app/models/models.py
```diff
@@ -2424,6 +2424,35 @@ def to_json(self):
...
+    userpreferenceid = models.UUIDField(
+        primary_key=True, default=uuid.uuid4, db_default=UUID4()
+    )
+    username = models.ForeignKey(
+        User,
+        to_field="username",
+        on_delete=models.CASCADE,
+        null=False,
+        related_name="preference",
+        related_query_name="preferences",
```
jacobtylerwalls: These are backwards. related_name is the set, and related_query_name is for filtering on an item of the set.

jacobtylerwalls (changes requested):
Awesome work @SDScandrettKint, one teeny fix and this will be good to go.

arches/app/models/models.py
samuelscandrett: Ah my bad.  Corrected [here](https://github.com/archesproject/arches/pull/11853/commits/7622ec297710ff3db9d82c618f86dec28b6bdf62)

jacobtylerwalls (approved):

chiatt (approved):

jacobtylerwalls:
Thanks for the updates, @SDScandrettKint!

PR #11855, opened 2025-03-13
Rename functions.py -> query_expressions.py

jacobtylerwalls:
On second thought I don't think `functions.py` is the best name for a module to hold ORM expressions. Something like "expressions" or "query_expressions" seems better to avoid confusion with arches functions or function triggers.

For example arches-lingo currently holds ORM wrappers for database functions in [query_expressions.py](https://github.com/archesproject/arches-lingo/blob/main/arches_lingo/query_expressions.py), to avoid having both `/functions` (arches functions) and `functions.py` at root.

Open to any input!


chiatt (approved):
Nice change. I agree - query_expressions is a more suitable name.

PR #11858, opened 2025-03-14
Spatial view geometrynode validation fixes and v8 filter additions

### LINKED ISSUE #11856: No geometrynode in spatial view returns django error rather than validation message
Creating a spatial view with no geometry returns the following (if in debug=True mode, otherwise it'd be a 500).

Image

### LINKED ISSUE #11857: Spatial view includes double geometrynodes (V8)
In V8, the spatial view geometrynode dropdown shows extra nodes that aren't geojson-feature-collection datatypes.  My guess would be that this issue stems from changes to graph versioning/nodes.

e.g.

Image

Image

The bottom two are from the system settings resource model - is this required in the the spatial views geometrynode dropdown?

samuelscandrett:
### Types of changes

### Description of Change

Added a try catch for geometry node - this introduces the validation error if no geometrynode is selected.
Added an extra option to `limit_choices_to` for nodes that have `source_identifier` as Null - this removes the additional/duplicate nodes.  @jacobtylerwalls  is this a good solution to the issue?    

### Issues Solved

Closes #11856 & #11857

### Checklist

#### Accessibility Checklist

#### Ticket Background


arches/app/models/migrations/11857_spatial_view_source_identifier_filter.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 5.2b1 on 2025-03-14 08:38
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("models", "11570_harden_editable_future_graphs"),
```
jacobtylerwalls: I just merged #11850, so it's a good time to update this.

samuelscandrett:
Might be good to add tests here too

arches/app/models/models.py
```diff
@@ -2218,7 +2221,11 @@ def clean(self):
         """
         Validate the spatial view before saving it to the database as the database triggers have proved hard to test.
         """
-        graph = self.geometrynode.graph
+        try:
+            graph = self.geometrynode.graph
+        except ObjectDoesNotExist:
+            raise ValidationError("A geometry node must be selected")
```
jacobtylerwalls: I nudged Andy toward using Django-native `clean()` for validation in large part because we were supporting a Django-admin based workflow. Given that there are no other callers of clean() besides the admin--in other words, that we know that all "validation traffic" is going through the Django models--what do you think about just letting the model field handle this by adding `blank=False`?

jacobtylerwalls: Then no code to add here and no test coverage to add :-)

jacobtylerwalls (approved):
Thanks for jumping on this!

arches/app/models/models.py
jacobtylerwalls: Oh wait, blank=False is the default. Let me figure out a better suggestion...

jacobtylerwalls: Alright, I refreshed myself on what's up here. In a django admin/ModelForm pattern, `full_clean()` runs four layers of validation and aggregates the errors together. The two that are relevant here are `clean_fields()`, which runs first and will raise ValidationError for blank=False fields, then `clean()`, where model-instance-as-a-whole validation is implemented.

So if you just did:
```py
        if not self.geometrynode_id:
            return
        graph = self.geometrynode.graph
```
That would be enough to prevent `clean()` from failing and would also avoid having to create a duplicative error message (which I notice wasn't localized, and which was appearing at the top of the form instead of inline with the dropdown).

jacobtylerwalls: Short-circuiting like that does mean the user would get feedback in two steps instead of one if `clean()` reveals other problems, but that was already an issue if `clean()` found multiple problems itself. (Out of scope here, but if this were a concern, this could be refactored to raise a compound ValidationError with all the issues.)

jacobtylerwalls: > Given that there are no other callers of clean() besides the admin

Should have clarified: the api follows the same pattern, calling full_clean() at the input boundary, which is what it's designed for:

https://github.com/archesproject/arches/blob/1a2c96f1160fa4dabce4763d4029b4a1c005e1e3/arches/app/views/api/__init__.py#L2296-L2297

arches/app/models/migrations/11857_spatial_view_source_identifier_filter.py
samuelscandrett: Resolved in [1d40cca](https://github.com/archesproject/arches/pull/11858/commits/1d40cca2c518c542136e2cf0e6bd8e2a531fb5f5)

arches/app/models/models.py
samuelscandrett: > up here. In a django admin/ModelForm pattern, `full_clean()` runs four layers of validation and aggregates the errors together

Unless I'm looking at the wrong thing, all of the Validation errors raised show at the top, rather than inline of where the error is occuring, e.g.
image

Good point that they're not localised though.

The following snippet works, and rather than a top level validation message we just get directed to the geometrynode dropdown with the notice "This field is required."
```
        if not self.geometrynode_id:
            return
        graph = self.geometrynode.graph
```
Happy to replace the existing try, except block with it.  Should a test be added too?

jacobtylerwalls: > Unless I'm looking at the wrong thing, all of the Validation errors raised show at the top, rather than inline of where the error is occuring, e.g.

When errors are not associated with a specific field, they just float to the top like you see there. That's what's happening in `clean()`, because we're checking the model as a whole and not instantiating ValidationError with a key/val pair naming the field.

This is out of scope, but you could imagine rewriting the field specific logic in clean() into either field specific validators on the model fields or by instantiating ValidationError with a key/val pair.

> Happy to replace the existing try, except block with it. Should a test be added too?

Yeah, that's what I would do. A test is always appreciated 👍 . We have some cases for validation in SpatialViewTests already.

samuelscandrett: @jacobtylerwalls I've replaced the try/except block and added a test.

tests/models/spatialview_model_tests.py
```diff
@@ -283,6 +283,16 @@ def test_spatial_view_isactive_set_to_false_removes_views(self):
...
+    def test_spatial_view_without_geometrynode(self):
+        spatialview = self.generate_valid_spatiatview()
+        spatialview.geometrynode = None
+        with self.assertRaises(Exception):
+            spatialview.full_clean()
+            spatialview.save()
+
+        with self.assertRaises(SpatialView.DoesNotExist):
+            fetched_spatialview = SpatialView.objects.get(pk=spatialview.spatialviewid)
+
```
jacobtylerwalls: I would just chop the save() and the last assertion and let the test fail naturally if the validation doesn't fire. We don't need to test the test helper itself.
```suggestion
```

jacobtylerwalls (approved):
Looks great, feel free to merge 👍

Issue #11851, opened 2025-03-13
Remove `Graph.new()`


chrabyrd:
Graph.new() is a non-standard method that diverges from typical Django conventions and is used in 1 place in Arches-core, outside of testing. Refactoring to standard Django practices will improve code maintainability and readability.

jacobtylerwalls:
Discussed in standup, makes sense, but be aware there are arches apps using this in tests that will need updating, too.

Also forget what I said in standup about `is_resource` doing the wrong thing, I glanced again and can't see why I thought that before 🤷‍♂

jacobtylerwalls:
I guess an alternative apporoach would be an improved method that creates the initial graph & editable-future-graph pair, including any necessary root nodes and nodegroups. I can see _that_ being super useful to avoid test failures like in #11852

chrabyrd:
I hear you there, but IMO it should remain explicit so that the Graph proxy model falls a little more in-line with traditional Django models. Or maybe a hook somewhere to better support other methods of graph proxy model generation.

I hate to see the codebase polluted with `Graph.objects.create()`,  `Graph.objects.create().save()`, `Graph(**).save()`, `Graph.new()`, etc etc. I'd like to pare that down as much as possible

Ideally the stuff in `save()` would also get moved to something else to better support all the ways to create a new graph proxy model, but that's a different kettle of fish entirely

jacobtylerwalls:
Yeah some rails would be nice.

The way I think about it is that:
- `create()`/`bulk_create()` etc. -- native queryset methods, shouldn't be overridden. cumbersome to use in tests given how many related instances need to be bootstrapped
- `create().save()` -- should never happen
- `save()` -- shouldn't only be overridden to schedule generic side effects like indexing (you're right that's a different kettle of fish). use this if you want signals and other side effects.
- `new()`/`new_with_editable_future_graph()` -- this is where we should be doing business logic to put some rails around  application or test code, ideally on the vanilla model but understandably might have to be on proxy until #11367 

Open to any/all feedback, just wanted to see if we're on the same rough page!

chrabyrd:
resolved in #11864

PR #11864, opened 2025-03-19
Hardens graph versioning #11851

chrabyrd:
### Types of changes

### Issues Solved

Closes #

### Checklist

### Further comments

This is a major hardening of the graph versioning feature.


arches/app/models/managers/graph.py
```diff
@@ -0,0 +1,60 @@
+import uuid
+from django.utils.translation import gettext as _
+from django.db import models
+
+
+class GraphManager(models.Manager):
+    def create(self, *args, **kwargs):
+        raise NotImplementedError(
+            _(
```
jacobtylerwalls: I'd be okay not localizing this. This is a low level error, and I wouldn't expect an API to transform it to a response.

arches/app/models/managers/graph.py
```diff
@@ -0,0 +1,60 @@
+import uuid
+from django.utils.translation import gettext as _
+from django.db import models
+
+
+class GraphManager(models.Manager):
+    def create(self, *args, **kwargs):
+        raise NotImplementedError(
```
jacobtylerwalls: This is smart, I like it.

I think I would consider subclassing QuerySet potentially in addition to Manager. Here's why.

1 -- `create()` is a method on QuerySet. What does this mean in practice? It means your guard here only throws on a call to `create()`, not other calls like `get_or_create()` or `update_or_create()`, because _those_ methods delegate to a base QuerySet and call `create()` on that, evading your guard. So we'll get more safety without any more code.

2 -- it gives us a more flexible base to compose out other business logic. Putting logic on Manager traps you to only calling it from the beginning of a query. e.g. you can't do `Graph.objects.filter().my_method()` if it's on Manager, but you can if it's on QuerySet. There are use cases where you want to enforce that your special query is called first, and then you need it to be on the manager (EDIT: and I found one, and mentioned it below), but I think that should be intentional & opt-in, not "let me just add to what's already here".

Point 1 is stronger I think. create_graph can probably stay on manager either way, it's the create() that works better on queryset.


***
If you agree, then this just becomes in the model
```py
         objects = GraphQuerySet.as_manager()
```

but see later comment about prefetching for another suggestion to have both a custom manager (entry point only) and a custom queryset (accessible anywhere)

I don't really care whether the file is called managers.py or querysets.py.

arches/app/models/managers/graph.py
```diff
@@ -0,0 +1,60 @@
...
+            )
+        )
+
+    def create_graph(self, author="", name="", is_resource=False):
+        from arches.app.models import models as arches_models
+
+        """
+        Create a new Graph and related objects, encapsulating all creation side effects.
+        """
+        new_id = uuid.uuid1()
```
jacobtylerwalls: We've updated all these models to uuid4 already, so I think we should use this opportunity to fix things like this here

arches/app/models/managers/graph.py
```diff
@@ -0,0 +1,60 @@
...
+            subtitle="",
+            author=author,
+            description="",
+            version="",
+            isresource=is_resource,
+            iconclass="",
+            ontology=None,
+            slug=None,
+        )
+        graph_model.save()  # to access side-effects declared in save method
```
jacobtylerwalls: Perfect, this will help everyone who touches this next 👍

arches/app/models/managers/graph.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # root node
+        arches_models.Node.objects.create(
+            pk=new_id,
+            name=name,
+            description="",
+            istopnode=True,
+            ontologyclass=None,
+            datatype="semantic",
+            nodegroup=nodegroup,
```
jacobtylerwalls: I thought top nodes didn't have nodegroups.

arches/app/models/managers/graph.py
```diff
@@ -0,0 +1,60 @@
...
+            graph=graph_model,
+        )
+
+        graph = self.get(pk=graph_model.graphid)
+
+        graph.publish()
+        graph.create_draft_graph()
+
+        # ensures entity returned matches database entity
+        return self.get(pk=graph_model.graphid)
```
jacobtylerwalls: You've made me love the Graph proxy model now!

arches/app/models/managers/graph.py
```diff
@@ -0,0 +1,60 @@
...
+            istopnode=True,
+            ontologyclass=None,
+            datatype="semantic",
+            nodegroup=nodegroup,
+            graph=graph_model,
+        )
+
+        graph = self.get(pk=graph_model.graphid)
+
+        graph.publish()
```
jacobtylerwalls: Should we expose a user param and pass it through here? In GraphDataView we have the user, we stringify it to get the author, but if we had the user and could pass it through...

arches/app/models/graph.py
```diff
@@ -2759,7 +2630,10 @@ def publish(self, user=None, notes=None):
         Adds a corresponding entry to the GraphXPublishedGraph table,
         and creates a PublishedGraph entry for every active language
         """
-        self.publication = None
+        if self.source_identifier_id:
+            raise RuntimeError("Publishing an draft_graph is prohibited.")
```
jacobtylerwalls:
```suggestion
            raise RuntimeError("Publishing a draft_graph is prohibited.")
```

arches/app/models/models.py
```diff
@@ -523,20 +523,149 @@ def is_editable(self):
         else:
             return True
 
-    def get_published_graph(self, language=None, raise_if_missing=False):
+    def should_use_published_graph(self):
+        return bool(
+            self.publication
+            and not self.source_identifier
```
jacobtylerwalls: To avoid eagerly fetching:
```suggestion
            self.publication_id
            and not self.source_identifier_id
```

arches/app/models/models.py
```diff
@@ -523,20 +523,149 @@ def is_editable(self):
...
+            and not self.has_unpublished_changes
+        )
+
+    def get_published_graph(self, language=None):
```
jacobtylerwalls, line 533: Let's doc the removed param as a minor breaking change.

arches/app/models/models.py
```diff
@@ -523,20 +523,149 @@ def is_editable(self):
...
             language = translation.get_language()
 
-        try:
-            graph = PublishedGraph.objects.get(
-                publication=self.publication, language=language
+        published_graph = PublishedGraph.objects.filter(
+            publication=self.publication, language=language
+        ).first()
+
+        return published_graph
```
jacobtylerwalls: This eagerly does a query. I try to avoid starting from another manager. If you start with `self`, you make it possible for the caller to prefetch related objects and have that used here.

```suggestion
        if not self.publication_id:
            return None
        
        return self.publication.publishedgraph_set.filter(language=language).first()
```

Now, my example about prefetching doesn't totally work because of the filter() and the first(). You can return from a loop to handle that. Trade fetching a couple extra rows for the ability to avoid later fetches on subsequent calls (and we are calling this over and over -- so it might be worth being a little picky.)

```suggestion
        if not self.publication_id:
            return None
        
        for published_graph in self.publication.publishedgraph_set.all():
            if published_graph.language_id == language:
                return published_graph
        return None
```

This won't just help picky callers who are in love with `prefetch_related()` -- looks like there's motion in Django 6 to get "prefetching by default" as a settings bit 🥳

arches/app/models/models.py
```diff
@@ -523,20 +523,149 @@ def is_editable(self):
...
+
+            return [
+                models.CardXNodeXWidget(**cards_x_nodes_x_widgets_dict)
+                for cards_x_nodes_x_widgets_dict in cards_x_nodes_x_widgets_slugs
+            ]
+        else:
+            return [
+                card_x_node_x_widget
+                for card in self.cardmodel_set.all()
+                for card_x_node_x_widget in card.cardxnodexwidget_set.all()
```
jacobtylerwalls, line 670: This is great -- everything is accessed on self, so any prefetches can be reused. But now there's a nested loop here, so we should check to see where the caller is and have that do a prefetch. I walked backward, and ended up at `GraphModel.__init__()`. So what we want is a way to force anybody fetching a Graph to do the prefetches that we'll need by the time we get here.

That sounds like a use case for GraphManager, because we want to enforce that for all traffic coming here.

You could express that as:

```py
class GraphQuerySet(models.QuerySet):
    def prefetch_related_entities(self):
        return self.prefetch_related("cardmodel_set", "cardmodel_set__cardxnodexwidget_set", ...)

class GraphManager(models.Manager):
    def get_queryset(self):
         return super().get_queryset().prefetch_related_entities()

class GraphModel(...
    objects = GraphManager.from_queryset(GraphQuerySet)()
```

And then anybody that didn't want that (lol) could still cancel it with `Graph.objects.prefetch_related(None)`.

But we'd be doing the right thing basically all the time.

This is what I was alluding to the other day in our huddle when I mentioned to @chiatt that we can still get fast lookups without serialized graphs by being a little more heavy-handed about forcing traffic to go through these prefetches.

arches/app/models/managers/graph.py
```diff
@@ -0,0 +1,60 @@
...
+    def create_graph(self, author="", name="", is_resource=False):
+        from arches.app.models import models as arches_models
+
+        """
+        Create a new Graph and related objects, encapsulating all creation side effects.
+        """
+        new_id = uuid.uuid1()
+        nodegroup = None
+
+        graph_model = arches_models.GraphModel(
```
jacobtylerwalls: I think self.model would be more resilient against subclassing. This comes up in projects from time to time. And it's easy to subclass since it's already a proxy model.

arches/app/models/models.py
jacobtylerwalls: This should dramatically reduce the queries to do a `Graph.objects.all()`. Happy to eyeball performance once you have a chance to weigh in.

arches/app/models/tile.py
```diff
@@ -110,14 +110,14 @@ def __init__(self, *args, **kwargs):
         self.serialized_graph = None
         self.load_serialized_graph()
 
-    def load_serialized_graph(self, raise_if_missing=False):
+    def load_serialized_graph(self):
```
jacobtylerwalls, line 113: Let's doc this.

arches/app/signals.py
```diff
@@ -284,3 +284,49 @@ def ensure_single_default_searchview(sender, instance, **kwargs):
             raise ValidationError(
                 "Only one search logic component can be default at a time."
             )
+
+
+@receiver(post_save, sender=models.Card)
+@receiver(post_delete, sender=models.Card)
+@receiver(post_save, sender=models.CardModel)
+@receiver(post_delete, sender=models.CardModel)
```
jacobtylerwalls, line 292: Ouch, nice catch that we have to double-up with the proxy models here. Looks like there's a draft PR against django to fix this.

arches/app/signals.py
```diff
@@ -284,3 +284,49 @@ def ensure_single_default_searchview(sender, instance, **kwargs):
...
+    )
+
+
+@receiver(post_save, sender=models.NodeGroup)
+@receiver(post_delete, sender=models.NodeGroup)
+def set_related_graph_has_unpublished_changes_to_true(sender, instance, **kwargs):
+    # NodeGroups have no direct relation to the GraphModel objects,
+    # so this signal can fail to find the node when deleting a Graphs
+    try:
+        node = models.Node.objects.filter(pk=instance.pk).first()
```
jacobtylerwalls: In case we already have it:
```suggestion
        node = instance.grouping_node
```

arches/app/signals.py
```diff
@@ -284,3 +284,49 @@ def ensure_single_default_searchview(sender, instance, **kwargs):
...
+        pass
+
+
+@receiver(post_save, sender=models.CardXNodeXWidget)
+@receiver(post_delete, sender=models.CardXNodeXWidget)
+def set_related_graph_has_unpublished_changes_to_true(sender, instance, **kwargs):
+    # CardXNodeXWidgets have no direct relation to the GraphModel objects,
+    # so this signal can fail to find the node when deleting a Graphs
+    try:
+        node = models.Node.objects.filter(pk=instance.node_id).first()
```
jacobtylerwalls:
```suggestion
        node = instance.node
```

tests/models/graph_tests.py
```diff
@@ -2221,118 +2456,122 @@ def test_placing_node_in_separate_card_does_not_pollute_database(self):
...
         serialized_updated_source_graph = JSONDeserializer().deserialize(
             JSONSerializer().serialize(updated_source_graph)
         )
 
         self.assertEqual(serialized_updated_source_graph["name"], "TEST RESOURCE TWO")
 
-        self._compare_serialized_updated_source_graph_and_serialized_editable_future_graph(
-            serialized_updated_source_graph, serialized_editable_future_graph
+        self._compare_serialized_updated_source_graph_and_serialized_draft_graph(
+            serialized_updated_source_graph, serialized_draft_graph
```
jacobtylerwalls, line 2608: Thanks for doing this thankless work

tests/models/managers/graph_manager_tests.py
```diff
@@ -0,0 +1,110 @@
...
+
+class GraphManagerTests(ArchesTestCase):
+    def test_create_raises_not_implemented_for_create(self):
+        """
+        Calling create() should raise NotImplementedError.
+        """
+        with self.assertRaises(NotImplementedError):
+            Graph.objects.create(author="test", name="test")
+
+    @patch("uuid.uuid1")
```
jacobtylerwalls: Just leaving myself a reminder to sync up on uuid1 vs uuid4

tests/signals_tests.py
```diff
@@ -0,0 +1,414 @@
...
+            "description": "Represents a single node in a graph",
+            "graphid": cls.SINGLE_NODE_GRAPHID,
+            "iconclass": "fa fa-circle",
+            "isresource": False,
+            "name": "Node",
+            "ontology_id": "e6e8db47-2ccf-11e6-927e-b8f6b115d7dd",
+            "subtitle": "Represents a single node in a graph.",
+            "version": "v1",
+        }
+        graph_model = models.GraphModel.objects.create(**graph_data)
```
jacobtylerwalls, line 33: Nice to know this escape hatch is still here.

jacobtylerwalls (approved):
This looks great @chrabyrd. I left some minor feedback just to make sure we stay focused on performance. But this is a big win for developers IMO.

arches/app/models/managers/graph.py
chrabyrd: aren't we using uuid1 because it's inherently ordered?

jacobtylerwalls: Haven't heard of that. The Django model field already has uuid4 as both python and postgres level defaults.

chrabyrd: oh huh. yeah I guess I misread https://docs.python.org/3/library/uuid.html#uuid.uuid1

happy to make the change 👍

arches/app/models/managers/graph.py
chrabyrd: ya, good suggestion here.

arches/app/models/managers/graph.py
chrabyrd: agreed. This is just copypasta from the `Graph.new()` method so as not to unintentionally break anything.

arches/app/models/models.py
chrabyrd: this feels like it could get involved, booting to a followup: https://github.com/archesproject/arches/issues/11910

arches/app/models/managers/graph.py
chrabyrd: this causes some tests to fail unexpectedly

jacobtylerwalls: Oh I get it -- yeah, we have to leave it like this because you're calling into the vanilla model. And anyone subclassing the proxy model is necessarily downstream of that. Whoops.

PR #11865, opened 2025-03-20
Requiring graph slug, #11799

### LINKED ISSUE #11799: Make graph slug required
As an arches app developer, I would like to be able to write code against models using their slug rather than their primary key, and without having to write fallback code if the slug is missing.

Suggest to make slug not-nullable in v8 and include a data migration to populate this on upgrade.

njkim:
### Types of changes

### Description of Change

Make the graph slug required so that graph can be requested by the slug, #11799 

### Issues Solved

Closes #11799 

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/models/utils.py
```diff
@@ -16,3 +16,20 @@ def add_to_update_fields(kwargs, field_name):
...
+
+    Arguments:
+    name -- the name to check and modfiy to make unique in the list of "names_to_check"
+    names_to_check -- a list of names that "name" should be unique among
+    """
+
+    i = 1
+    temp_node_name = name
+    while temp_node_name in names_to_check:
+        temp_node_name = "{0}{1}{2}".format(name, suffix_delimiter, i)
```
chiatt: Could you change this variable to `temp_name` given that it no longer applies to only nodes.

chiatt (dismissed):

arches/app/models/graph.py
```diff
@@ -2270,6 +2254,10 @@ def validate_fieldname(fieldname, fieldnames):
                         ).format(slug=self.slug),
                         1007,
                     )
+        else:
+            raise GraphValidationError(
+                _("You must supply a slug for your graph."), 1005
```
jacobtylerwalls: Let's use a fresh number instead of one we already use for something else (1005 is `IntegrityCheck.NODE_HAS_ONTOLOGY_GRAPH_DOES_NOT`)

It would be nice to start putting all of these in one enum, we could continue using `IntegrityCheck` for that, which is a little more general than just graphs.

arches/app/models/querysets/graph.py
```diff
@@ -9,7 +12,20 @@ def create(self, *args, **kwargs):
...
-    def create_graph(self, name="", user=None, is_resource=False):
+    def get_slug(self, name, is_resource):
+        if name:
+            slug = slugify(name, separator="_")
+        else:
+            if is_resource:
+                slug = "new_resource_model"
+            else:
+                slug = "new_branch"
+        existing_slugs = GraphModel.objects.all().values_list("slug", flat=True)
```
jacobtylerwalls: self is already a queryset, which may have already been filtered. not likely that it's been filtered here, but still good to do IMO in case people copy this pattern for other methods.
```suggestion
        existing_slugs = self.values_list("slug", flat=True)
```

arches/app/models/querysets/graph.py
```diff
@@ -18,6 +34,9 @@ def create_graph(self, name="", user=None, is_resource=False):
         new_id = uuid.uuid4()
         nodegroup = None
 
+        if not slug:
+            slug = self.get_slug(name, is_resource)
```
jacobtylerwalls: I love how @chrabyrd's changes in #11864 make this so natural now!

arches/app/models/utils.py
```diff
@@ -16,3 +16,20 @@ def add_to_update_fields(kwargs, field_name):
...
 def field_names(instance_or_class):
     return {f.name for f in instance_or_class._meta.fields}
+
+
+def make_name_unique(name, names_to_check, suffix_delimiter="_"):
+    """
+    Makes a name unique among a list of names
+
+    Arguments:
+    name -- the name to check and modfiy to make unique in the list of "names_to_check"
```
jacobtylerwalls: modfiy -> modify

arches/app/views/graph.py
```diff
@@ -562,7 +566,9 @@ def post(self, request, graphid=None):
 
                     clone_data = graph.copy()
                     ret = clone_data["copy"]
-                    ret.slug = None
+                    ret.slug = self.get_slug(
+                        str(clone_data["copy"].name), clone_data["copy"].isresource
+                    )
```
jacobtylerwalls: Should we copy the name or the slug?

arches/app/views/graph.py
```diff
@@ -546,7 +548,9 @@ def post(self, request, graphid=None):
 
                 elif self.action == "export_branch":
                     clone_data = graph.copy(root=data)
-                    clone_data["copy"].slug = None
+                    clone_data["copy"].slug = self.get_slug(
+                        str(clone_data["copy"].name), False
+                    )
```
jacobtylerwalls: Also here.

arches/app/views/graph.py
```diff
@@ -660,6 +666,19 @@ def validate_images_only_config(old_node_data, new_node_data):
                     return False
         return True
 
+    def get_slug(self, name, is_resource):
```
jacobtylerwalls: We should be able to remove this and rely on the method you added to GraphQuerySet

arches/app/views/graph.py
```diff
@@ -562,7 +566,9 @@ def post(self, request, graphid=None):
 
                     clone_data = graph.copy()
                     ret = clone_data["copy"]
-                    ret.slug = None
+                    ret.slug = self.get_slug(
```
jacobtylerwalls: This should do it, plus a tweak to supply the correct name for "exported_branch" etc
```suggestion
                    ret.slug = Graph.objects.get_slug(
```

arches/app/models/querysets/graph.py
```diff
@@ -9,7 +12,20 @@ def create(self, *args, **kwargs):
             "Use create_graph() to create new Graph instances with proper business logic."
         )
 
-    def create_graph(self, name="", user=None, is_resource=False):
+    def get_slug(self, name, is_resource):
```
jacobtylerwalls: maybe generate_slug() ?

tests/fixtures/jsonld_base/models/test_1_basic_object_cardinality_n.json
```diff
@@ -195,7 +195,7 @@
                 "ontologyclass": "http://www.cidoc-crm.org/cidoc-crm/E1_CRM_Entity",
                 "sortorder": 0
             },
-            "slug": "taco",
+            "slug": "test_1_basic_object_cardinality_n",
```
jacobtylerwalls, line 198: 🌮

arches/app/models/models.py
```diff
@@ -481,7 +481,7 @@ class GraphModel(SaveSupportsBlindOverwriteMixin, models.Model):
         on_delete=models.SET_DEFAULT,
     )
     config = JSONField(db_column="config", default=dict)
-    slug = models.TextField(validators=[validate_slug], null=True)
+    slug = models.TextField(validators=[validate_slug])
```
jacobtylerwalls, line 484: This validation is totally ignored because we're not in the habit of calling `full_clean()` anywhere when writing our custom views. I can create a very perky slug ending in `!` ;)

I'm happy to bounce this to a new issue, given the feature freeze is imminent and it's a pre-existing bug.

releases/8.0.0.md
```diff
@@ -50,6 +50,7 @@ Arches 8.0.0 Release Notes
 - Allow datatypes to transform their default values for use on the front-end[#11904](https://github.com/archesproject/arches/issues/11904) & [#11944](https://github.com/archesproject/arches/issues/11944)
 - Add validation error for geometry node used in spatial view attributes [#11888](https://github.com/archesproject/arches/issues/11888)
 - Add source mapping to webpack development configuration [#11906](https://github.com/archesproject/arches/issues/11906)
+- Update the graph model to make the slug required [#11799](https://github.com/archesproject/arches/issues/11799)
```
jacobtylerwalls, line 61: Let's also mention this in the breaking changes (we have a section "The following fields are no longer nullable ..."). And I think even calling out graph imports will no longer work if they're missing slugs would be good.

Unless we want to auto-generate slugs when importing? /cc @chiatt

arches/app/models/querysets/graph.py
```diff
@@ -9,7 +12,20 @@ def create(self, *args, **kwargs):
...
+            if is_resource:
+                slug = "new_resource_model"
+            else:
+                slug = "new_branch"
+        existing_slugs = GraphModel.objects.all().values_list("slug", flat=True)
+        slug = make_name_unique(slug, existing_slugs, "_")
+
+        return slug
+
+    def create_graph(self, name="", slug=None, user=None, is_resource=False):
```
jacobtylerwalls: Since we have the opportunity before v8 goes out, let's force everything except name to be keyword-only. If I looked at a call like `create_graph("foo", None, None, True)`, I would have no idea what was going on.
```suggestion
    def create_graph(self, name="", *, slug=None, user=None, is_resource=False):
```

arches/app/models/migrations/11799_require_graph_slug.py
```diff
@@ -0,0 +1,40 @@
...
+
+    dependencies = [
+        ("models", "11882_principal_user_read_only"),
+    ]
+
+    def add_missing_graph_slug(apps, schema_editor):
+        Graph = apps.get_model("models", "GraphModel")
+        graphs_missing_slug = Graph.objects.filter(slug__isnull=True)
+        for graph in graphs_missing_slug:
+            graph.slug = slugify(str(graph.name), separator="_")
```
jacobtylerwalls: Names aren't necessarily unique, so I can envision this failing. Maybe we want to use `add_slug()` here? But then we would need to set `use_in_migrations = True` on the GraphManager. (And that means not using the shortcut `objects = GraphQuerySet.as_manager()` anymore, following the second example [here](https://docs.djangoproject.com/en/5.1/topics/migrations/#model-managers).) Maybe you'd prefer a simpler solution just inline here.

arches/app/models/migrations/11799_require_graph_slug.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("models", "11882_principal_user_read_only"),
+    ]
+
+    def add_missing_graph_slug(apps, schema_editor):
+        Graph = apps.get_model("models", "GraphModel")
```
jacobtylerwalls: I'd use either Graph or GraphModel but not mix the two (this is importing GraphModel but calling it Graph) -- I guess it should be Graph on both sides to get Graph.save() validation checks.

jacobtylerwalls (commented):
Thanks, this is a huge help.

releases/8.0.0.md
chiatt: I don't think v7 graphs can be imported into a v8 instance anyway. That's almost always the case in a feature or major release. Generally the expectation is that after graphs are migrated during the project upgrade steps, they need to be exported out to the package as part of upgrading the package. We should remind people that they must export their resource models and reference data in order to ensure that their packages are v8 compatible, but that could be in a separate PR.

arches/app/models/migrations/11799_require_graph_slug.py
jacobtylerwalls: Just tried to sketch this out and didn't get it working, so this is probably way overkill. Let's do a simpler solution of just avoiding using any existing slugs.

jacobtylerwalls (approved):
Looks good!

PR #11872, opened 2025-03-24
11871 adding sort-by feature for search results on the main search page

### LINKED ISSUE #11871: [Potential Feature] sort-by drop-down for search results on the main search page
As the first stage of implementing custom search functionality (#11780) we are aiming to add a ‘sort-by’ dropdown to the main search page, alongside the existing ‘search-order’ options. 

Search results in Arches are currently returned sorted by resource name. The user can specify whether these should be in ascending or descending order. In addition to 'Resource Name', we intend to add 'Date created', and ‘Date last updated’. As part of this work, these variables will be added to the ES index for resources.

CWDamm-Kint:
### Types of changes

### Description of Change
Date created and date edited have been added to the index. The front-end now has a 'sort-by' dropdown, as well as an button to switch between sort-order ascending and descending. These control two query parameters in the URL, which are then passed as an array to the backend 'sort' search-component. These are then used to modify the DSL passed to Elasticsearch. 

### Issues Solved

Closes #11871 

### Checklist
    - [X] dev/8.0.x (under development): features, bugfixes not covered below

#### Accessibility Checklist

#### Ticket Background
Sponsored by: Knowledge Integration
Found by: @CWDamm-Kint @SDScandrettKint
Tested by: @CWDamm-Kint @SDScandrettKint
Designed by: @CWDamm-Kint @SDScandrettKint


aj-he:
@SDScandrettKint @CWDamm-Kint - one of our earlier Arches instances had it's Editlog truncated after the data migration (I can't remember the reason why!) so can you make sure you handle where a resource may not have any editlog entries?

CWDamm-Kint:
For info, this is the current state of play for the feature's front-end

https://github.com/user-attachments/assets/bd4a3a55-8872-4f6a-9bd2-f1d4804ab1b4

samuelscandrett (commented):
Currently the button has a noticeable border and background when clicking quickly or clicking and holding:

https://github.com/user-attachments/assets/e8cf46cf-d5e9-435a-8180-a2e2511cd195

There is also a bit of a delay/lag between selecting an option that can be sorted and going back to unsorted:


https://github.com/user-attachments/assets/5626a5ed-5ddf-4b7d-b9b6-4c6714b100eb

samuelscandrett:
> @SDScandrettKint @CWDamm-Kint - one of our earlier Arches instances had it's Editlog truncated after the data migration (I can't remember the reason why!) so can you make sure you handle where a resource may not have any editlog entries?

@aj-he this is an example of the sort for date created in action, where two resources have had their edit_log data removed.  The resources with no edit log history are sat at the bottom in an unordered manner, despite whether the sort is set to ascending or descending. 

https://github.com/user-attachments/assets/9e158391-a879-4df4-8062-35c3235b8e98

I suppose this makes sense, as if the resources with no editlog information were to go at the top in a descending sort, it'd essentially be "unsorted" until resources with dates appeared.  
Is this the sort of thing you were expecting to see?

aj-he:
> > @SDScandrettKint @CWDamm-Kint - one of our earlier Arches instances had it's Editlog truncated after the data migration (I can't remember the reason why!) so can you make sure you handle where a resource may not have any editlog entries?
> 
> @aj-he this is an example of the sort for date created in action, where two resources have had their edit_log data removed. The resources with no edit log history are sat at the bottom in an unordered manner, despite whether the sort is set to ascending or descending.
> 
>  sort-current-date-order.mp4 
> I suppose this makes sense, as if the resources with no editlog information were to go at the top in a descending sort, it'd essentially be "unsorted" until resources with dates appeared. Is this the sort of thing you were expecting to see?

@SDScandrettKint - I think that is reasonable behaviour given a lack of edit log data!

arches/app/models/resource.py
```diff
@@ -500,6 +500,26 @@ def get_documents_to_index(
...
+            document["date_created"] = None
+        try:
+            document["date_edited"] = (
+                models.EditLog.objects.filter(
+                    resourceinstanceid=str(self.resourceinstanceid)
+                )
+                .latest("timestamp")
+                .timestamp
+            )
+        except:
```
chiatt: Could you add the error types that you are expecting to except?

arches/app/search/mappings.py
```diff
@@ -281,6 +281,8 @@ def prepare_search_index(create=False):
                         "provisional": {"type": "boolean"},
                     },
                 },
+                "date_created": {"type": "keyword"},
+                "date_edited": {"type": "keyword"},
```
whatisgalen: would it be possible to rename `"date_edited"` to `"date_last_edited"` for clarity?

samuelscandrett: I agree that is clearer.  Added in [0ba2602](https://github.com/archesproject/arches/pull/11872/commits/0ba2602114e314381eee0796c0b626f0e2a10cce).

tests/search/search_sort_tests.py
```diff
@@ -0,0 +1,221 @@
...
+            "a94c81c5-2047-4b53-8341-495f12bfad95",  # RC (1st updated)
+            "8ec120f4-61cd-4fa1-8f9a-3d12cff3176f",  # RB (2nd updated)
+            "09ba8b82-b6db-4ece-9a46-98f6d381a7b2",  # RA (3rd updated)
+        ]
+
+        self.assertListEqual(
+            hit_ids,
+            ids_ordered_by_date_last_edited,
+            "The results are not ordered by date edited ascending",
+        )
```
samuelscandrett: @CWDamm-Kint Is this test needed?  If the sort order is not present In the /search URL, it will just be added back when loaded.
If so, at the moment it fails so it should probably be changed to `assertNotEqual`?

CWDamm-Kint: @SDScandrettKint I think I changed how the sort-order is handled in the frontend since writing this test, so now it's put there in the url by default when you select anything other than unsorted. I agree that does make this test feel a bit of an unusual edge case now, as you'd have to manually remove it to get at request like this in the backend. 

Though I guess it still clarifies / tests what behaviour to expect if sort-order is manually removed from the url (e.g. it defaults to asc). Seems to be passing for me - is it failing for you?

samuelscandrett: I see, this test is checking that without a specified sort order in the request, the default order of ascending is enforced.  I'm happy for the test to remain, I think all that needs changing is the name of the test to the agreed name of `test_default_ascending_sort_order`.

Not sure why my tests were failing previously, but they seem to be fine now... will keep an eye on it but can assume it having a moment.

arches/app/models/resource.py
```diff
@@ -500,6 +500,26 @@ def get_documents_to_index(
...
+        try:
+            document["date_created"] = (
+                models.EditLog.objects.filter(
+                    resourceinstanceid=str(self.resourceinstanceid)
+                )
+                .earliest("timestamp")
+                .timestamp
+            )
+        except:
+            document["date_created"] = None
```
chiatt: The date created is a property on the resource instance itself, so there shouldn't be a need to go to the edit log for this. 

```suggestion
        document["date_created"] = self.createdtime
```

arches/app/models/resource.py
```diff
@@ -500,6 +500,26 @@ def get_documents_to_index(
...
+        try:
+            document["date_last_edited"] = (
+                models.EditLog.objects.filter(
+                    resourceinstanceid=str(self.resourceinstanceid)
+                )
+                .latest("timestamp")
+                .timestamp
+            )
+        except:
+            document["date_last_edited"] = None
```
chiatt: I'm not sure how others feel, but I would not object to the 'lasteditedtime' also being added to the resource instance model. Otherwise this query is going to slow down indexing and is vulnerable to admins deleting their edit history. I think updating the last edited timestamp on the resource could easily be added to the `resource.save_edit` method.

chiatt (changes requested):

tests/search/search_sort_tests.py
```diff
@@ -0,0 +1,221 @@
...
+            "resource_graphs",
+            "Search Sort Test Model.json",
+        )
+
+        management.call_command(
+            "packages",
+            operation="import_graphs",
+            source=search_sort_test_model_path,
+            verbosity=0,
+        )
```
jacobtylerwalls: Instead I think you can just add this at the top of your class: `graph_fixtures = ["Search Sort Test Model"]`
```suggestion
```

arches/app/models/resource.py
```diff
@@ -500,6 +500,26 @@ def get_documents_to_index(
         document["displayname"] = []
         document["displaydescription"] = []
         document["map_popup"] = []
+        try:
+            document["date_created"] = (
+                models.EditLog.objects.filter(
+                    resourceinstanceid=str(self.resourceinstanceid)
```
jacobtylerwalls:
```suggestion
                    resourceinstanceid=self.resourceinstanceid
```
Stringifying None probably would crash when attempting to cast this back to UUID.

Also, although strings are acceptable in a call like this, I'm concerned that it makes it "look safe" to use strings for UUID fields with the ORM, when in general we shouldn't, see #10877.

jacobtylerwalls: We should filter out rows where the timestamp is null.

jacobtylerwalls (commented):
:wave: left some minor feedback, might be moot points, however, if we take Cyrus's suggestion to add a new model field!

chiatt:
> 👋 left some minor feedback, might be moot points, however, if we take Cyrus's suggestion to add a new model field!

We have agreement from the team to add the new field, but I think we may be getting a bit close to feature freeze to add that into this PR. I've created a separate [issue](https://github.com/archesproject/arches/pull/11872) for it. So, it makes sense to me to follow @jacobtylerwalls's suggestions here, and then once the new field is added (whether that's in 8.0 or 8.1) we can use that.

bferguso:
> @SDScandrettKint @CWDamm-Kint - one of our earlier Arches instances had it's Editlog truncated after the data migration (I can't remember the reason why!) so can you make sure you handle where a resource may not have any editlog entries?

@aj-he - Just curious if you remember what type of a migration caused the truncation. This would be of concern to our business areas as they need a full audit log to support their decision making as part of their legislated mandate.

arches/app/models/resource.py
samuelscandrett: Refined the exception to ObjectDoesNotExist.  Since date created can use `self.createdtime`, I don't think there's any need to include a try catch for that one.

arches/app/models/resource.py
samuelscandrett: Added `timestamp__isnull=False` to the filter in [06eb00d](https://github.com/archesproject/arches/pull/11872/commits/06eb00d28b8bc6f4bd791d731f8918af925cfa56) so we're only fetching resourceinstance matches + valid timestamps

aj-he:
> > @SDScandrettKint @CWDamm-Kint - one of our earlier Arches instances had it's Editlog truncated after the data migration (I can't remember the reason why!) so can you make sure you handle where a resource may not have any editlog entries?
> 
> @aj-he - Just curious if you remember what type of a migration caused the truncation. This would be of concern to our business areas as they need a full audit log to support their decision making as part of their legislated mandate.

@bferguso - I believe this was done directly to the database in our early days on Arches to minimise the size of the database. It was not the result of a migration.

At the time it was treated like a log, and we did not need to persist ~1million rows stating a record was created, when we know the data was loaded into the new system on a particular date.  The idea was that it would begin recording edits from the loading date.

With hindsight, a bit of foresight would have been useful!

jacobtylerwalls (approved):

tests/search/search_sort_tests.py
CWDamm-Kint: Thanks Jacob.

chiatt (approved):

arches/app/search/components/standard_search_view.py
```diff
@@ -209,6 +209,13 @@ def handle_search_results_query(
         sorted_query_obj = search_filter_factory.create_search_query_dict(
             list(self.request.GET.items()) + list(self.request.POST.items())
         )
+
```
whatisgalen, line 212: It seems like folks who use a different search view would miss out on this functionality -- I understand the simplicity of not having `sort-by` and `sort-order` as json, however is there a way in which this logic could live within the search component `sort_results.py` file instead? That would make it more portable

PR #11875, opened 2025-03-24
Add default ordering on additional models #11873

### LINKED ISSUE #11873: Add default ordering to Django models with a sortorder column
Using `first()` executes a query regardless if the instances have been prefetched unless the related descriptor is ordered, either via a custom `Prefetch()`, which is verbose, or via a default ordering on the related model. See Django [forum](https://forum.djangoproject.com/t/first-method-does-not-use-prefetch-related/39452/2) for more details. Suggesting we add default ordering for models with obvious candidates like a `sortorder` column.

Without either of the above, it then becomes difficult/error-prone to access anything you have prefetched, because you have to remember to avoid `first()` and instead handle the null case explicitly:

E.g.

Instead of
```py
cross_record = (
    node.cardxnodexwidget_set.all()[0]
    if node.cardxnodexwidget_set.all()
    else None
)
```

it could just be this:
```py
cross_record = node.cardxnodexwidget_set.first()
```

And then this becomes one fewer thing to check during code review.

jacobtylerwalls:
### Types of changes

### Description of Change
Before, it was cumbersome and arguably error-prone to have to manually order cards and ~tiles~ and other models with a sortorder field, especially when attempting to leverage `prefetch_related()`.

### Issues Solved
Closes #11873

### Checklist


releases/8.0.0.md
```diff
@@ -117,6 +118,16 @@ JavaScript:
     - From TileModel, traversing a child tile: `tilemodel` -> `child`
     - From TileModel, collecting child tiles: `tilemodel_set` -> `children`
 
+- The following models now have a default ordering by `sortorder`:
+    - CardModel (and Card)
+    - TileModel (and Tile)
+    - CardXNodeXWidget
+    - Node
+    - Plugin
+    - ETLModule
```
jacobtylerwalls, line 149: Interestingly, MapLayer already had this feature.

arches/app/views/api/__init__.py
```diff
@@ -233,7 +230,6 @@ def get(self, request):
             tiles = tiles.filter(resourceinstance_id__in=resourceid.split(","))
         if tileid is not None:
             tiles = tiles.filter(tileid=tileid)
-        tiles = tiles.order_by("sortorder")
```
jacobtylerwalls: We can leave these in if people prefer the explicitness. (Although there is something to be said for know-your-model 😉 )

arches/app/views/api/__init__.py
```diff
@@ -1069,16 +1062,9 @@ def get(self, request, resourceid):
...
         else:
-            cards = (
-                graph.cardmodel_set.order_by("sortorder")
-                .filter(nodegroup__in=permitted_nodegroups)
-                .prefetch_related(
-                    Prefetch(
-                        "cardxnodexwidget_set",
-                        queryset=models.CardXNodeXWidget.objects.order_by("sortorder"),
-                    )
-                )
```
jacobtylerwalls: This is how verbose it was to do this before.

arches/app/views/api/__init__.py
```diff
@@ -1662,10 +1644,7 @@ def get(self, request):
 
             cardwidgets = [
                 widget
-                for widgets in [
-                    card.cardxnodexwidget_set.order_by("sortorder").all()
-                    for card in graph_cards
-                ]
+                for widgets in [card.cardxnodexwidget_set.all() for card in graph_cards]
```
jacobtylerwalls: So removing this stuff is I think good because it reinforces that you don't have to do the custom `Prefetch` as highlighted in the previous comment.

arches/app/utils/data_management/resources/formats/csvfile.py
```diff
@@ -423,9 +423,8 @@ def write_resources(self, graph_id=None, resourceinstanceids=None, **kwargs):
         csvs_for_export = []
 
         tiles = self.group_tiles(
-            self.tiles.order_by(
-                "nodegroup_id"
-            ).values(),  # TODO: refactor to avoid going to the db again
+            # TODO: refactor to avoid going to the db again
+            self.tiles.order_by("nodegroup_id", "sortorder").values(),
```
jacobtylerwalls, line 427: Behavior change: shouldn't this sort by sort order? Added. Can pull this out into a separate PR if preferable.

chiatt: Yeah - I think this change makes sense. Let's leave it in.

chiatt (commented):
I think these changes are really nice, but is seems like they could result in a decrease in performance. 

This is the result of `models.TileModel.objects.all().explain()` on dev/8.0.x
```'Seq Scan on tiles  (cost=0.00..3777.56 rows=48856 width=553)'```

and this is the result on the `default-ordering` branch
```'Gather Merge  (cost=11031.46..15781.76 rows=40714 width=553)\n  Workers Planned: 2\n  ->  Sort  (cost=10031.44..10082.33 rows=20357 width=553)\n        Sort Key: sortorder\n        ->  Parallel Seq Scan on tiles  (cost=0.00..3492.57 rows=20357 width=553)'```

Maybe this isn't a concern. I guess it may depend on how often we query without any interest in having results sorted, but it doesn't seem uncommon.

As you mention, people could just call `order_by()` to avoid ordering, but it might be easy to forget that order_by needs to be called in order not to not sort.

**Comment by jacobtylerwalls** [comment]
> Maybe this isn't a concern. I guess it may depend on how often we query without any interest in having results sorted, but it doesn't seem uncommon.

Yeah, this is the question. I'd like to weigh the performance not only in terms of the queries we're doing today, but how many we'll be doing (or can avoid doing) when continue on the path we're taking to start using the ORM in more natural ways.

One of the problems with our ORM habits is that we almost always default to starting new queries from a new model manager, like for a resource, get `models.TileModel.objects.filter(resource...` instead of `self.tilemodel_set.all()` so we can reuse `self.tilemodel_set` over and over, especially if it's been prefetched.

If we're going to start using "self" more and reuse querysets like that, it becomes more difficult to predict precisely which use cases will need to grok order versus which don't. If you need ordering most of the time, it seems better to optimistically order, rather than cause most of your downstream callers to need to run to the database again.

From the query plan you showed, yes, ordering the entire tiles table would be expensive! But I think it would be very rare to query that many tiles, and that was the source of the severe performance problem we fixed in #11898.

I poked around and looked for instances where we don't care about tile order. Deletion is one. I can sprinkle in some `order_by()` to remove default ordering, but some of them would better be rewritten as `self.tilemodel_set` anyway, and I'd think our efforts would be better focused on that.

***

When developing arches-querysets, it's felt a little error-prone to remember to `order_by("sortorder")` every time I query for tiles and child tiles. I guess I'd selfishly prefer to do the right thing most of the time by default and defer the optimization burden to the caller who has a use case for an unordered query.

Thoughts?

**Comment by chiatt** [comment]
> From the query plan you showed, yes, ordering the entire tiles table would be expensive! But I think it would be very rare to query that many tiles, and that was the source of the severe performance problem we fixed in #11898.
> 
True, it probably would be pretty rare. 
> I poked around and looked for instances where we don't care about tile order. Deletion is one. I can sprinkle in some `order_by()` to remove default ordering, but some of them would better be rewritten as `self.tilemodel_set` anyway, and I'd think our efforts would be better focused on that.
> 
Agreed
> When developing arches-querysets, it's felt a little error-prone to remember to `order_by("sortorder")` every time I query for tiles and child tiles. I guess I'd selfishly prefer to do the right thing most of the time by default and defer the optimization burden to the caller who has a use case for an unordered query.
> 
> Thoughts?

Okay, It does seem like defaulting to sorted results makes sense and probably isn't going to have a significant performance cost as long as we use the orm wisely.

**Comment by apeters** [comment]
One thing that seems really confusing to me is that when you don't want to incur the overhead of default sorting you add the `order_by()` clause.  Not super intuitive.

**Comment by jacobtylerwalls** [comment]
Fair. All I would add is that it sort of(*) rhymes with the pattern for canceling an earlier `select/prefetch_related` (that might be out of your control, like here) by chaining `prefetch_related(None)`.

(*) not quite, because it's no-args instead of None

***
If the tilemodel change is controversial, we can take a baby step here and just do the other models.

Mostly I'm trying to make it easier/less verbose/less tricky than having to remember to do [this](https://github.com/archesproject/arches/pull/11875#discussion_r2010964023) to efficiently query widget configs.

**Comment by jacobtylerwalls** [comment]
I think this is too hasty to add to tiles (too presumptive to suggest not ordering by pk or parenttile is not also just as wise).

I'll revert the tiles changes and keep the changes for the models I was really interested in (making easier prefetches of CardXNodeXWidget)

File: arches/app/models/migrations/11873_model_ordering.py [hunk 6]
```diff
@@ -0,0 +1,37 @@
+# Generated by Django 5.2rc1 on 2025-03-24 16:26
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("models", "11842_userpreferences"),
```
chiatt:
```suggestion
        ("models", "11579_ids_search_filter"),
```

chiatt (approved):

PR #11879, opened 2025-03-25
Allow import of values formatted as date in Excel

### LINKED ISSUE #11878: Bulkloader fails to import values from Excel cells formatted as 'date'
When openpyxl, which is used to process Excel files in the Bulkloader, reads a value from a cell that is formatted as a date, it creates a datetime object rather than a string of the date. Currently, the `transform_value_for_tile` method on the date datatype does not expect a datetime object and will fail to import that data.

chiatt:
### Types of changes

### Description of Change
Fixes the date datatype's `transform_value_for_tile` method to import dates from excel files regardless whether they are formatted as strings or dates.

### Issues Solved

Closes #11878

### Checklist

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -717,6 +717,7 @@ class DateDataType(BaseDataType):
         try:
             value = value.astimezone()
         except:
+            # The .astimezone function throws an error on Windows for dates before 1970
             value = self.backup_astimezone(value)
         return value.isoformat(timespec="milliseconds")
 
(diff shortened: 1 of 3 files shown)


aarongundel:
Tested this out with procedure from Cyrus - 

- Create a simple resource with a date node and maybe a string node and then create a couple of instances.
- Export the resources from search in tile excel format.
- Delete the resources from Arches.
- Then change the date cell to be a date type in excel (or LibreOffice - I haven't tested Numbers, but it's probably okay too) and reimport the data with the bulk loader.

Created three resources matching description - they all reimported as expected.

aarongundel (approved):
LGTM

PR #11883, opened 2025-03-26
Make principaluser read-only #11882

### LINKED ISSUE #11882: Set editable=False on ResourceInstance.principaluser
ResourceInstance.principaluser should always be read-only when creating forms or JSON interfaces (serializers).

At the Django level you can set this with editable=False.

That would allow other Django tooling like Django REST Framework's `ModelSerializer` that generate serializers by inspecting your model definitions to recognize that the field should be read-only without further configuration.

jacobtylerwalls:
### Types of changes

### Description of Change
Tooling that generates forms or JSON interfaces (serializers) from Django model definitions can rely on the value of `editable=False` to derive that the field should be read-only. We should leverage this for the `principaluser` field to prevent overposting attacks.

### Issues Solved
Closes #11882

### Checklist

#### Ticket Background

### Further comments
To test, you could check out archesproject/arches-lingo main and verify that any "principaluser" key is ignored when posting back. Before, you could post back changes to this field value.

I'll open a follow-up in arches-querysets for 7.6 compat -- we'll need to redefine the field explicitly on ArchesResourceSerializer.


jacobtylerwalls:
/cc @chiatt -- discussed this with Aaron yesterday. This field option is really just a hint for other tooling, e.g. Django Admin or Django REST Framework, or any other custom tooling. If we end up writing our own form to allow superusers to edit this field, we can always ignore this "hint" and build a form or serializer ourselves.

Basically, this just adds some safety to avoid accidentally exposing this field when using Django tools.

aarongundel (approved):
LGTM

PR #11890, opened 2025-03-27
Remove references to related resource index re #8578

jacobtylerwalls:
Remove this code as promised.

The first commit included a little note about a proactive deletion on upgrade, but I realized any reindex since arches 6.2 would have killed this index anyway, so I removed it.

### DIFF
--- a/arches/app/search/mappings.py
+++ b/arches/app/search/mappings.py
@@ -26,7 +26,6 @@ from arches.app.search.es_mapping_modifier import EsMappingModifierFactory
 CONCEPTS_INDEX = "concepts"
 TERMS_INDEX = "terms"
 RESOURCES_INDEX = "resources"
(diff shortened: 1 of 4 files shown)


whatisgalen:
honestly I think about this all the time and an emerging use case is traversing the relations graph ("metagraph"?) created by related resources. For example if I want to know an Actor instance's related Event and what Place instances are related to those Event instances, I need to traverse Actor --> Event --> Place one edge at a time (but for N instances of each). If I could envision a related_resources index mapping on the Resources index, it would map relations within 1, 2, and mayyyybe 3 degrees of a given origin.

FYI @bferguso @apeters @chiatt

jacobtylerwalls:
100% agree (subway take™), but I wouldn't necessarily build such a lookup in elastic 😉

chiatt (approved):

PR #11899, opened 2025-03-31
Limit restricted nodegroups query to single graph

jacobtylerwalls:
Follow-up to 31dae7e.

To maintain parity with the prior query, this should filter out irrelevant nodegroups belonging to other graphs. Although I don't think it really would make a difference on the frontend, which consults this information in the context of a single graph anyway. Still, better for debugging to not introduce confusion.


aarongundel (approved):
LGTM

PR #11900, opened 2025-03-31
Remove cruft re: restricted nodegroups

jacobtylerwalls:
Follow-up to c83da46.

`restricted_nodegroups` is always an empty array in v8 after we obsoleted `OVERRIDE_RESOURCE_MODEL_LOCK`, so remove the code that was interrogating it (and remove the empty array altogether).


chrabyrd (approved):
lgtm!

PR #11905, opened 2025-04-01
Allow datatypes to transform default values for use in dropdowns

### LINKED ISSUE #11904: Allow datatypes to transform their default value
#### Background
The tile representation of the reference datatype was designed to be more flexible than the concept datatype by a uri, list item values (to provide human-readable display values) and the controlled list_id, all wrapped as an array with this form:
```
[
    {
      "uri": "https://vocab.getty.edu/aat/300456597",
      "labels": [
        {
          "id": "24559510-c6c6-4558-b411-7cd127642657",
          "value": "warrant assertion event",
          "language_id": "en",
          "list_item_id": "14751cb1-f7af-45d9-b32a-59d150bdac0f",
          "valuetype_id": "prefLabel"
        }
      ],
      "list_id": "6eaa2c6f-af83-464c-9200-051c4cfe7e8e"
    }
  ]
```

To be more friendly to the front-end, [Arches Controlled Lists #68](https://github.com/archesproject/arches-controlled-lists/pull/68) adds a method, [build_select_option](https://github.com/archesproject/arches-controlled-lists/pull/68/files#diff-5fb61b74a19567bd526df184a7dfc487d5ea4d03a4e2997f73d4f4543accc5d0R227), to transform the db representation of a list item into something that is more useful for stocking a dropdown picker on the front end, with this form:
```
[
    {
        "children": [],
        "display_value": "warrant assertion event",
        "list_item_id": "14751cb1-f7af-45d9-b32a-59d150bdac0f",
        "list_item_values": [
            {
                "id": "24559510-c6c6-4558-b411-7cd127642657",
                "language_id": "en",
                "list_item_id": "14751cb1-f7af-45d9-b32a-59d150bdac0f",
                "value": "warrant assertion event",
                "valuetype_id": "prefLabel"
            }
        ],
        "sortorder": 0,
        "uri": "https://vocab.getty.edu/aat/300456597"
    }
]
```
This is adequate (or at least not in question in this PR) for the transforming the options that will stock the dropdown, but what if you want to handle the default value for a reference datatype node and use a lazy loading dropdown (as we do with Arches Lingo).

#### Proposed Change
Arches datatype should provide a method to allow transformations from the db/tile representation to something useable on the front-end.

johnatawnclementawn:
### Types of changes

### Description of Change
Please see description in ticket #11904 for context and proposed solution.

### Issues Solved

Closes #11904

### Checklist

### DIFF
--- a/arches/app/datatypes/base.py
+++ b/arches/app/datatypes/base.py
@@ -533,8 +533,8 @@ class BaseDataType(object):
         """
         pass
 
-    def build_dropdown_option(self, default_tile_value):
+    def transform_value_from_tile(self, value, **kwargs):
         """
-        Transform db value to simpler representation for dropdown-style widgets
+        Transform db value to simpler representation for use on the front-end
         """
-        return default_tile_value
+        return value
--- a/arches/app/datatypes/base.py
+++ b/arches/app/datatypes/base.py
@@ -75,6 +75,12 @@ class BaseDataType(object):
         """
         return value
 
+    def transform_value_from_tile(self, value, **kwargs):
+        """
+        Transform db value to simpler representation for use on the front-end
+        """
+        return value
+
     def update(self, tile, data, nodeid, action):
         """
         Updates the tile.data value of a given datatype and returns the updated
@@ -532,9 +538,3 @@ class BaseDataType(object):
         a GraphValidationError
         """
         pass
-
-    def transform_value_from_tile(self, value, **kwargs):
-        """
-        Transform db value to simpler representation for use on the front-end
-        """
-        return value
--- a/arches/app/datatypes/base.py
+++ b/arches/app/datatypes/base.py
@@ -532,3 +532,9 @@ class BaseDataType(object):
         a GraphValidationError
         """
         pass
+
+    def build_dropdown_option(self, default_tile_value):
+        """
+        Transform db value to simpler representation for dropdown-style widgets
+        """
+        return default_tile_value
--- a/releases/8.0.0.md
+++ b/releases/8.0.0.md
@@ -46,6 +46,7 @@ Arches 8.0.0 Release Notes
 - Add message when copying geojson url to clipboard [#11076](https://github.com/archesproject/arches/issues/11076)
 - Hide "view report" button in Arches system settings [#11659](https://github.com/archesproject/arches/issues/11659)
 - Updates github workflows postgis image to version 14-3.4 to match Arches core [#11833](https://github.com/archesproject/arches/issues/11833)
+- Allow datatypes to transform their default values for use on the front-end[#11904](https://github.com/archesproject/arches/issues/11904)
 - Add validation error for geometry node used in spatial view attributes [#11888](https://github.com/archesproject/arches/issues/11888)
 - Add source mapping to webpack development configuration [#11906](https://github.com/archesproject/arches/issues/11906)
 


arches/app/datatypes/base.py
```diff
@@ -532,3 +532,9 @@ def validate_node(self, node):
         a GraphValidationError
         """
         pass
+
+    def build_dropdown_option(self, default_tile_value):
```
chiatt: It seems like this should have a more generic name in case something similar is needed for other datatypes. If there are possibly different kinds of transformations, then maybe this method needs an optional argument.

chrabyrd:
I think we should also wait to hear from @apeters , I remember he has strong opinions about the layer in which this logic lives.

arches/app/datatypes/base.py
```diff
@@ -532,3 +532,9 @@ def validate_node(self, node):
         a GraphValidationError
         """
         pass
+
+    def transform_value_from_tile(self, value, **kwargs):
+        """
+        Transform db value to simpler representation for use on the front-end
+        """
+        return value
```
apeters: can you put this method underneath the `transform_value_for_tile` method as they are essentially matched pairs.

arches/app/datatypes/base.py
apeters, line 1: As this method is not implemented anywhere in Arches currently is this really needed?  As you have shown, custom methods like this can be implemented in custom datatype w/o having to rely on them being on the Base class.

apeters (changes requested):

arches/app/datatypes/base.py
johnatawnclementawn: For more context, [here](https://github.com/archesproject/arches-component-lab/pull/28/files#diff-42171eb4a1a2bbbf9381bfd614c22e6a45b3ba29669c14408b647d39b539e0bdR49) is where this is being implemented in the endpoint that returns widget configuration in Arches Component Lab.

I suppose an alternative to needing to add this method on the Base class would be to key on node.datatype and just perform `transform_value_from_tile` if the datatype is reference. This would push us back to a more reference-datatype specific implementation as I initially stubbed out. @jacobtylerwalls @chrabyrd thoughts?

jacobtylerwalls: My thought was that arches-component-lab is basically core arches in all but name -- it shouldn't mix in knowledge of custom hooks in custom datatypes.

> just perform transform_value_from_tile if the datatype is reference.

Right, this is what I was encouraging Johnathan not to do. Alexei, what do you think? One (and only one) generic hook on the base datatype so that we don't mix in knowledge of project datatypes into arches-component-lab?

chrabyrd: > This would push us back to a more reference-datatype specific implementation 

That's what I'd like to avoid. Impromptu soapbox diatribe incoming:

Whatever pattern we land on here will be used else. It must be generic. It must be different for each datatype. It must take place on the backend. Let's look at the current patterns for a similar core component: the resource instance select.

There's 3 places where we're dealing with values/potential data transformation:

defaultValue
Data is fetched directly from the widget config. For example:

```
[
  {
    "resourceXresourceId": "8777c9c5-33cf-4e21-adae-88ca3af2d807"
  }
]
```
initialValue
This data comes from the layer above the widget (from the tile). For example:

```
[
  {
    "resourceId": "69af178a-6a04-4bb0-b501-00105923a3d4",
    "ontologyProperty": "",
    "resourceXresourceId": "86bf87ce-66a8-4867-a802-30ca15a2bfd9",
    "inverseOntologyProperty": ""
  },
  {
    "resourceId": "c5318a5a-6c4f-45d7-9763-fcc3904eab3c",
    "ontologyProperty": "",
    "resourceXresourceId": "8777c9c5-33cf-4e21-adae-88ca3af2d807",
    "inverseOntologyProperty": ""
  }
]
```
value/valueOptions
Data is fetched from custom endpoints (ES), which requires additional transformation. For example:

```
[
  {
    "_id": "c5318a5a-6c4f-45d7-9763-fcc3904eab3c",
    "_index": "arches_lingo_resources",
    "_score": null,
    "_source": {
      "displaydescription": "Undefined",
      "displayname": "BOB",
      "displayname_language": "en",
      "geometries": [],
      "graph_id": "9ffb6fcc-b4b4-11ea-84f7-3af9d3b32b71",
      "map_popup": [],
      "permissions": {
        "principal_user": [1],
        "users_with_no_access": [],
        "users_without_delete_perm": [],
        "users_without_edit_perm": [],
        "users_without_read_perm": []
      },
      "points": [],
      "provisional_resource": "false",
      "resourceinstanceid": "c5318a5a-6c4f-45d7-9763-fcc3904eab3c",
      "root_ontology_class": "http://www.cidoc-crm.org/cidoc-crm/E21_Person"
    },
    "can_edit": true,
    "can_read": true,
    "is_principal": true,
    "sort": ["BOB"]
  },
  {
    "_id": "69af178a-6a04-4bb0-b501-00105923a3d4",
    "_index": "arches_lingo_resources",
    "_score": null,
    "_source": {
      "displaydescription": "Undefined",
      "displayname": "MIKE",
      "displayname_language": "en",
      "geometries": [],
      "graph_id": "9ffb6fcc-b4b4-11ea-84f7-3af9d3b32b71",
      "map_popup": [],
      "permissions": {
        "principal_user": [1],
        "users_with_no_access": [],
        "users_without_delete_perm": [],
        "users_without_edit_perm": [],
        "users_without_read_perm": []
      },
      "points": [],
      "provisional_resource": "false",
      "resourceinstanceid": "69af178a-6a04-4bb0-b501-00105923a3d4",
      "root_ontology_class": "http://www.cidoc-crm.org/cidoc-crm/E21_Person"
    },
    "can_edit": true,
    "can_read": true,
    "is_principal": true,
    "sort": ["MIKE"]
  }
]
```

Slight tangent: currently the defaultValue and initialValue formats do not match. Ideally they would. That is, the same format would exist both in the tile and the card_x_node_x_widget field. eg:

```
[
  {
    "resourceId": "c5318a5a-6c4f-45d7-9763-fcc3904eab3c",
    "ontologyProperty": "",
    "resourceXresourceId": "8777c9c5-33cf-4e21-adae-88ca3af2d807",
    "inverseOntologyProperty": ""
  }
]
```

And while that works for the tile and config, we're not able to do anything with it on the frontend without another call to trade the UUID for a human-readable string. Hence core using elastic, which we're trying to move away from with the widget rewrites.

From the perspective of a frontend component, these 3 places should all have the same format. There's really two options to choose from, the minimalist approach:

```
[
  {
    "resourceXresourceId": "8777c9c5-33cf-4e21-adae-88ca3af2d807",
    "displayname": "MIKE",
  }
]
```

and the as-close-to-tiledata-as-possible approach:

```
[
  {
    "resourceId": "c5318a5a-6c4f-45d7-9763-fcc3904eab3c",
    "ontologyProperty": "",
    "resourceXresourceId": "8777c9c5-33cf-4e21-adae-88ca3af2d807",
    "inverseOntologyProperty": ""
    "displayname": "MIKE",
  }
]
```

And while I don't have a strong opinion about which one we choose, it serves to have a single source of truth to do this. And that source of truth must be on the backend to accommodate all scenarios. In a perfect world we'd already have a `display_value` table that could be leveraged in this. But as it stands I believe this _should_ live on the generic datatype.

chrabyrd: Alternatively, I suppose, this could be a separate util on the backend. But it'd be chonky and delineated by datatype -- essentially a glorified switch statement. Feels more cohesive to just put this on the generic datatype.

apeters (approved):
Ok, I can see the need for something like this on the base class that is future leaning.
I'll just say here for posterity that whatever datatype developers implement, that the output shouldn't reflect the UI requirements of any specific library (eg: PrimeVue).

chrabyrd:
@johnatawnclementawn I've spun this up for the naming change:

https://github.com/archesproject/arches/issues/11944

PR #11911, opened 2025-04-02
Bump Django to 5.2 final, celery to 5.5

jacobtylerwalls:


pyproject.toml
```diff
@@ -64,7 +64,6 @@ dependencies = [
     "requests[security]>=2.31.0",
     "semantic-version==2.10.0",
     "SPARQLWrapper==1.8.5",
-    "tzdata==2023.3",
```
jacobtylerwalls, line 67: Arches doesn't use `tzdata` itself, it's a dependency of celery via kombu, was getting this:

```
ERROR: Cannot install arches==8.0.0a2 and celery because these package versions have conflicting dependencies.

The conflict is caused by:
    arches 8.0.0a2 depends on tzdata==2023.3
    kombu 5.5.2 depends on tzdata==2025.2; python_version >= "3.9"
```

pyproject.toml
```diff
@@ -96,7 +95,6 @@ exclude = ["*.tests", "*.tests.*", "tests.*", "tests"]
 namespaces = false
 
 [tool.black]
-target-version = ['py310']
```
jacobtylerwalls, line 99: This is cruft from before we had python support made explicit via `requires-python`

chiatt (approved):

PR #11913, opened 2025-04-02
fixes caching issue causing frontend breakage

aarongundel:
Needed for fix in Lingo

### DIFF
--- a/arches/app/media/js/utils/resource.js
+++ b/arches/app/media/js/utils/resource.js
@@ -1,5 +1,4 @@
 import arches from 'arches';
-import resourceLookup from 'viewmodels/map'
 
 const resourceUtils = {
     /**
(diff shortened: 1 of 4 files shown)


arches/app/media/js/utils/resource.js
```diff
@@ -1,4 +1,5 @@
 import arches from 'arches';
+import resourceLookup from 'viewmodels/map'
```
chrabyrd: instead of import from viewmodels/map, the function should be refactored to take `resourceLookup` as an arg. `resource-instance-select.js` will then need to be updated to have it declared before the call

chrabyrd (changes requested):

chrabyrd (approved):
lol branch name && PR name are funky but not gonna block on that. LGTM

aarongundel:
This is certainly not what it started as!

PR #11931, opened 2025-04-04
Handle for numeric layer opacity #11930

### LINKED ISSUE #11930: Search map fails when layer opacity is numeric
According to the mapbox spec, opacity should be a numeric value: 
https://docs.mapbox.com/help/glossary/layout-paint-property/

For example: 
```
map.addLayer({
  id: 'trees-point',
  type: 'circle',
  source: 'trees',
  paint: {
    'circle-radius': 3,
    'circle-color': '#223b53',
    'circle-stroke-color': 'white',
    'circle-stroke-width': 1,
    'circle-opacity': 0.5
  }
});
```

However, in the map.js viewmodel, we are treating it like an object, assigning a value to a property called `base`.

https://github.com/archesproject/arches/blob/6b4ed2e6f849e12ad84adbc66e26725cc043a602/arches/app/media/js/viewmodels/map.js#L255

There are cases in which the mapbox style spec allows a property to be a primitive or an object, and this is likely the case here. The problem is that if the opacity is numeric and we try to assign a value to a `base` property, the page crashes complaining that a property can't be assigned to a number. Interestingly, an error is not thrown in v7.6. Even if the opacity is numeric, no error is thrown. This issue only seems to affect v8 which is much less forgiving.

chiatt:
### Types of changes

### Description of Change
Allows layers with numeric opacity values (as opposed to objects) to load in the search map without error.

### Issues Solved

Closes #11930

### Checklist

### DIFF
--- a/arches/app/media/js/viewmodels/map.js
+++ b/arches/app/media/js/viewmodels/map.js
@@ -251,7 +251,10 @@ const viewModel = function (params) {
             var startVal = layer.paint ? layer.paint[opacityType + '-opacity'] : null;
 
             if (startVal) {
-                if (parseFloat(startVal)) {
+                if (parseFloat(startVal) && parseFloat(layer.paint[opacityType + '-opacity'])) { // verify startVal and opacity can be numbers
+                    layer.paint[opacityType + '-opacity'] = startVal * opacityVal;
+                } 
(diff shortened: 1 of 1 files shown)


aarongundel (approved):
@chiatt demonstrated fix, LGTM

PR #11946, opened 2025-04-09
Rename transform value from tile

### LINKED ISSUE #11944: Change `transform_value_from_tile` to more descriptive name

johnatawnclementawn:
### Types of changes

### Description of Change

Rename base datatype method created in #11905

### Issues Solved

Closes #11944

### Checklist

### DIFF
--- a/arches/app/datatypes/base.py
+++ b/arches/app/datatypes/base.py
@@ -75,9 +75,11 @@ class BaseDataType(object):
         """
         return value
 
-    def transform_default_value(self, value, **kwargs):
+    def get_interchange_value(self, value, **kwargs):
         """
-        Transform db value to simpler representation for use on the front-end
+        Transforms values for use in api's that will provide data for programmatic use on the frontend.
+        The interchange value is a shape that is more useful on the frontend than the node value shape.
+        The return value can be more or less complex than the value passed in, depending on the datatype.
         """
         return value
 
--- a/arches/app/datatypes/base.py
+++ b/arches/app/datatypes/base.py
@@ -75,7 +75,7 @@ class BaseDataType(object):
         """
         return value
 
-    def transform_value_from_tile(self, value, **kwargs):
+    def transform_value_for_frontend(self, value, **kwargs):
         """
         Transform db value to simpler representation for use on the front-end
         """
--- a/arches/app/datatypes/base.py
+++ b/arches/app/datatypes/base.py
@@ -75,7 +75,7 @@ class BaseDataType(object):
         """
         return value
 
-    def transform_value_for_frontend(self, value, **kwargs):
+    def transform_default_value(self, value, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


arches/app/datatypes/base.py
```diff
@@ -75,7 +75,7 @@ def transform_value_for_tile(self, value, **kwargs):
         """
         return value
 
-    def transform_value_from_tile(self, value, **kwargs):
+    def transform_value_for_frontend(self, value, **kwargs):
```
chiatt: If the purpose is to transform the default value, why can't this method be `transform_default_value`? Although I think this is an improvement, as it is, its purpose is unclear and it could have other uses that aren't necessarily the frontend.

chiatt: This is also a bit too similar to `get_display_value` which is also the value for the frontend.

johnatawnclementawn: Thanks Cyrus, this is a great point and I like `transform_default_value` much better.

chrabyrd: The issue for me is that I can see this being used in scenarios outside of the transforming the default value. Essentially any time we need to massage the value for the frontend ( eg dropdown, data from the tile, etc etc )

johnatawnclementawn: I wrote up a response arguing for keeping `transform_default_value`, but ultimately talked myself out of arguing for it. Ultimately, this method should be available to transform tile-like values, whether they're default-values or tiles for use on the frontend. 

In standup on 2025/04/10, we landed on `get_interchange_value` as an appropriate method name. This indicates that the method should be used in api's that will provide data for programmatic use on the frontend (differentiated from `get_display_value` which is used to simplify values for display).

tests/utils/datatypes/datatype_tests.py
```diff
@@ -38,6 +38,30 @@
 # python manage.py test tests.utils.datatypes.datatype_tests --settings="tests.test_settings"
 
 
+class BaseDataTypeTests(ArchesTestCase):
+    def test_get_tile_data_only_none(self):
```
chrabyrd, line 42: Adding this test seems out of scope for this ticket?

chrabyrd (approved):
LGTM! one note but won't block on it

tests/utils/datatypes/datatype_tests.py
johnatawnclementawn: Yeah, sorry for the potentially confusing diff - I just moved that test up to the top because IMO it makes sense to have the tests for the Base class first, then datatype-specific tests after. Previously that test was under `BooleanDataTypeTests`.

johnatawnclementawn: https://github.com/archesproject/arches/pull/11946/commits/e149e70ed812e7439fe4afc6d616116abb27de56

Issue #11909, opened 2025-04-01
Add isort to pre-commit


jacobtylerwalls:
I used isort to make it easier to alphabetize imports when doing #11903. We could add that to the pre-commit hooks so that no one has to expend effort sorting imports themselves.

PR #11962, opened 2025-04-13
Speed up serialization by ~50%

jacobtylerwalls:
### Types of changes

### Description of Change
Speed up serialization by ~50% by using structural pattern matching (`match` statement) instead of if/elif/else and repeating isinstance()

### Benchmark
in arches/app/management
```py
import timeit

from django.core.management.base import BaseCommand

class Command(BaseCommand):
        t = timeit.Timer(
            stmt="serializer.handle_object(obj)",
            setup="from arches.app.utils.betterJSONSerializer import JSONSerializer; import uuid; obj = uuid.uuid4(); serializer = JSONSerializer()",
        )
        print("Best of 5:", min(t.repeat()))
```

**Before**
Best of 5: 1.4309943750267848
**After**
Best of 5: 0.8083759160363115

### DIFF
--- a/arches/app/utils/betterJSONSerializer.py
+++ b/arches/app/utils/betterJSONSerializer.py
@@ -1,19 +1,16 @@
 import datetime
 import decimal
-import types
-import json
 import inspect
+import json
 import uuid
 from io import StringIO
 from itertools import chain
-from django.db import models, DEFAULT_DB_ALIAS
+
 from django.db.models import Model
 from django.db.models.query import QuerySet
 from django.utils.encoding import smart_str
-from django.core.serializers.python import Serializer as PythonSerializer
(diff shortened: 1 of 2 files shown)


arches/app/utils/betterJSONSerializer.py
```diff
@@ -1,19 +1,16 @@
 import datetime
 import decimal
-import types
```
jacobtylerwalls, line 3: I could revert this to get the python test coverage to pass, but really we should just #11909

apeters (approved):
I didn't actually confirm a speed improvement, so I'll take your word, but I had wondered if doing all these "isinstance" checks were maybe a source of slowness.

It would be nice to see this in 7.6.x as well.

Issue #11702, opened 2024-12-19
Bootstrap new projects with more appropriate APP_NAME tiles


jacobtylerwalls:
A new project gets created with an `APP_NAME` in their project settings matching the project name used in the creation command. 👍 

On setup_db, however, we create system settings tiles that will override this back to "Arches", which isn't necessarily desired.

We could avoid creating this tile entirely, or create it with the value that matches settings.py, or any solution for removing drift.

The system settings tile is generated from this data:
https://github.com/archesproject/arches/blob/f3390eff2ff5fe8eca899d2b3f199868fcea1575/arches/db/system_settings/Arches_System_Settings.json#L152-L157

Issue #11700, opened 2024-12-19
System settings tiles not consulted when running the development server


jacobtylerwalls:
As discussed in #11660 and #11697, the [documented inheritance order](https://arches.readthedocs.io/en/stable/configuring/settings-beyond-the-ui/#settings-inheritance) for settings configurable through the UI is that they trump everything, including python settings modules.

However, this only works correctly when running a WSGI application, not in dev, since wsgi.py calls `update_from_db()`, and nothing in the dev server does.

A first [attempt](https://github.com/archesproject/arches/pull/11687/commits/fbc966a266d987af497639ccbd8f2d00dd9b9a05) at fixing this that didn't account for `setup_db`.

PR #11963, opened 2025-04-14
Remove project ready() methods #11701

### LINKED ISSUE #11701: Find a better guard for generate_frontend_configuration()
In both core and arches app templates, we have a guard around the call to `generate_frontend_configuration()` that tries to avoid running this multiple times for every arches application:

https://github.com/archesproject/arches/blob/02f2d8cf3c976c10f8f5793556dd528d4c8bb693/arches/install/arches-templates/project_name/apps.py-tpl#L11-L13

However, it checks `APP_NAME`, which isn't very reliable, since it is configurable through the UI, and comes with a default of "Arches" that might not even match the `self.name` in the AppConfig class.

In v8 we should find a better guard or remove the guard entirely.

jacobtylerwalls:
### Types of changes

### Description of Change
Checking APP_NAME was not very reliable, since it may be "Arches" on projects anyway (#11702), and is different when running in dev/prod (#11700).

Also, it's duplicative to do this multiple times, we can just do this once by delegating it to core.

We could target 7.6 instead if we remove the updateproject change.

### Issues Solved

Closes #11701

### Checklist

### DIFF
--- a/arches/management/commands/updateproject.py
+++ b/arches/management/commands/updateproject.py
@@ -17,6 +17,7 @@ class Command(BaseCommand):  # pragma: no cover
         answer = input(
             "This operation will upgrade your project to version 8.0\n"
             "This will replace the following files in your project:\n"
+            "  - <project>/apps.py\n"
             "  - .github/workflows/main.yml\n"
             "  - webpack/webpack-utils/build-filepath-lookup.js\n"
             "  - webpack/webpack.common.js\n"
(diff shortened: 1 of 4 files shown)


arches/management/commands/updateproject.py
```diff
@@ -31,6 +32,29 @@ def handle(self, *args, **options):
             self.stdout.write("Operation aborted.")
 
     def update_to_v8(self):
+        # Update apps.py to remove generate_frontend_configuration
```
jacobtylerwalls, line 35: Maybe this is a little presumptive, but it seems unlikely projects have _innovated_ all that much with their apps.py

arches/management/commands/updateproject.py
```diff
@@ -31,6 +32,29 @@ def handle(self, *args, **options):
...
+        to_replace_1 = """
+from django.conf import settings
+
+from arches.settings_utils import generate_frontend_configuration"""
```
chrabyrd, line 39: In the (albeit unlikely) event that this file has been updated with additional imports, or even that space between imports removed, this will fail. Should it just be bumped to 8.0.0.md instructions?

chrabyrd (commented):
👀 This was originally built this way because of where the files/folder are generated, eg the command being run in core arches generates the folder in core arches. Testing locally to ensure this is no longer the case.

arches/management/commands/updateproject.py
jacobtylerwalls: Right, I never know how to make this call. I guess a loud fail with instructions is another option?

chrabyrd:
tested locally and this works! Good change 👍 

I don't mind whichever way the updateproject bit gets updated

jacobtylerwalls:
Thanks for spot-checking me there!

arches/management/commands/updateproject.py
jacobtylerwalls: >  this will fail.

I guess this is already a loud fail. I can add a note to the 8.0.0.md to verify that `ready()` is no longer present. It's not too important either way -- it's just cruft.

releases/8.0.0.md
```diff
@@ -223,6 +224,8 @@ JavaScript:
     python manage.py updateproject
     ```
 
+    Verify in `<project_name>/apps.py` that you no longer have a ``ready()`` method.
```
chrabyrd: seems like this should have it's own number? also double-ticked?

chrabyrd (commented):
sorry for being nitty 😅

releases/8.0.0.md
jacobtylerwalls: for whatever reason the vs code preview treats single & double the same way, but I think you're right 👍 

Sounds good on a new numbered bullet.

chrabyrd (approved):
lgtm!

PR #11966, opened 2025-04-14
Add unique constraint on lifecycle XRef tables

jacobtylerwalls:
See [docs](https://docs.djangoproject.com/en/5.1/topics/db/models/#extra-fields-on-many-to-many-relationships):

> If you don’t want multiple associations between the same instances, add a [UniqueConstraint](https://docs.djangoproject.com/en/5.1/ref/models/constraints/#django.db.models.UniqueConstraint) including the from and to fields. Django’s automatically generated many-to-many tables include such a constraint.

Given that we chose not to use the automatic generated m2m tables here, it's on us to add the unique constraint.


chrabyrd:
But we _don't_ want a unique constraint I believe? A resource instance lifecycle state could potentially have many `to` or `from` options. Fleshed out the user could define logic forks for such things.

jacobtylerwalls:
For sure, but there would be no need to record the pair of a single to and a single from more than once. That doesn't capture any additional information.

Really just trying to add some extra safety against any weirdsies when using the Django admin or ETL that would create dupes and cause confusion if only some were deleted afterward.

chrabyrd (approved):
OH! oh yeah duh. Chalk that up to pre-coffee PR reviews 😅

PR #11967, opened 2025-04-14
Make graph slug unique per source identifier re #11565

### LINKED ISSUE #11565: Add UniqueConstraint for Graph.slug where Graph.source_identifier = None
There are four ways to get JSON-LD data into arches:

1. the POST api
2. the PUT api
3. the `load_jsonld` management command
4. the json-ld bulk import module, which is a wrapper around no. 3

All four allow you to pass a graph by slug. In dev/8.0.x we currently assume the user means the original graph. But if your tile data has already migrated away to some future graph version, that assumption would leave you working against an outdated version. 

Should we assume when these four places are given a graph slug, that the caller wants to work against the latest graph?

cc/ @robgaston

***
Noticed while dev'ing archesproject/arches-lingo#103, which might create a fifth place to look up graphs by slug.

jacobtylerwalls:
Restore unique constraint on graph slugs. This was unique=True in 7.6 and modified as part of graph versioning in v8, but we can still keep functional uniqueness across versions with a multi-field unique constraint.

Closes #11565


chrabyrd (approved):
lgtm!

PR #11976, opened 2025-04-15
Improve ResourceXResource field and related names #11869

### LINKED ISSUE #11869: rename ResourceXResource model fields
In the `ResourceXResource` [model](https://github.com/archesproject/arches/blob/dev/8.0.x/arches/app/models/models.py#L1093-L1118), the foreign key fields are misleadingly and inconsistently named:

```.py
resourceinstanceidfrom = models.ForeignKey(
        "ResourceInstance",
        db_column="resourceinstanceidfrom",
        blank=True,
        null=True,
        related_name="resxres_resource_instance_ids_from",
        on_delete=models.CASCADE,
        db_constraint=False,
    )
```

...this field references a resourceinstance, not its id, so it actually shouldn't be named `resourceinstanceidfrom` because then you end up with query filter expressions like this `ResourceXResource.objects.filter(resourceinstanceidfrom_id=...)` which unfortunately is an unending source of confusion. Indeed, the graphmodel foreign key fields have a similar issue:

```.py
resourceinstancefrom_graphid = models.ForeignKey(
        "GraphModel",
        db_column="resourceinstancefrom_graphid",
        blank=True,
        null=True,
        related_name="resxres_resource_instance_fom_graph_id",
        on_delete=models.CASCADE,
        db_constraint=False,
    )
```

We should remove the `id` portion of the model fieldnames and leave it as is for the field's `db_column` attribute. Same goes for the `tileid` --> `tile` and `nodeid` --> `node` fields.

jacobtylerwalls:
### Types of changes

### Description of Change
Before, these fields and related names were difficult to use (or had typos).

### Issues Solved

Closes #11869

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @whatisgalen
*   Tested by: @ 
*   Designed by: @


chiatt:
Could we get rid of the 'instance' from the fieldnames as well? e.g. `resourceinstancefrom` -> `resourcefrom`.

jacobtylerwalls:
Is `resource_from` ok?

whatisgalen:
hot take: can we do prefix instead: `from_resource` (which would then be referrable as (`from_resource_id`)

jacobtylerwalls:
I like it 🔥

jacobtylerwalls:
@whatisgalen is there any chance you could help me get this over the finish line from here?

whatisgalen:
@jacobtylerwalls do you know if was there a consensus on changing the db columns in postgres, as well?

whatisgalen:
@chiatt @apeters @robgaston @jacobtylerwalls  Can we get rid of these fields, too? 
```.py
class ResourceXResource(SaveSupportsBlindOverwriteMixin, models.Model):
    ...
    datestarted = models.DateField(blank=True, null=True)
    dateended = models.DateField(blank=True, null=True)
```
My understanding is that` resource_x_resources` either exist or they don't and the query logic isn't checking anywhere for this field: `dateended`. It seems like these fields' function overlaps with tiledata

apeters:
My 2 cents.  Can the db column names keep the "_id" suffix because in the db, that's all they are.

whatisgalen:
> My 2 cents. Can the db column names keep the "_id" suffix because in the db, that's all they are.

@apeters I believe this PR leaves the db column names unchanged. They're not entirely consistent as they are, eg: `"resourceinstanceidfrom"`, `"resourceinstancefrom_graphid"`

arches/app/models/models.py
```diff
@@ -1238,93 +1238,91 @@ class ResourceXResource(SaveSupportsBlindOverwriteMixin, models.Model):
...
+            data = self.tile.data[str(self.node_id)]
             if type(data) != list:
                 data = [data]
             for relatedresourceItem in data:
                 if relatedresourceItem:
                     if relatedresourceItem["resourceId"] != str(deletedResourceId):
                         newTileData.append(relatedresourceItem)
-            self.tileid.data[str(self.nodeid_id)] = newTileData
-            self.tileid.save()
+            self.tile.data[str(self.node_id)] = newTileData
```
whatisgalen, line 1315: FYI @jacobtylerwalls I changed this from `str(self.node)` --> `str(self.node_id)` as I don't believe calling `str` on the Node model instance would have yielded the `pk`/`nodeid`

jacobtylerwalls:
Leaving the db columns alone is nice for imposing fewer breaking changes on folks with sql scripts when really all we're trying to do is make the python orm nicer.

arches/app/models/models.py
jacobtylerwalls: Good catch!

arches/app/models/migrations/11869_rename_resxres_fields.py
```diff
@@ -0,0 +1,128 @@
+# Generated by Django 5.2 on 2025-04-15 09:08
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("models", "11842_userpreferences"),
```
chiatt:
```suggestion
        ("models", "11579_ids_search_filter"),
```

arches/app/models/migrations/11869_rename_resxres_fields.py
```diff
@@ -0,0 +1,128 @@
+# Generated by Django 5.2 on 2025-04-15 09:08
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("models", "11579_ids_search_filter"),
```
chiatt:
```suggestion
        ("models", "11873_model_ordering"),
```

chiatt (approved):
I think with the migration dependency update and the merge conflict resolved, this seems good to go.

jacobtylerwalls:
Thanks all!

arches/app/models/migrations/11869_rename_resxres_fields.py
```diff
@@ -0,0 +1,128 @@
...
+                to="models.tilemodel",
+            ),
+        ),
+        migrations.RemoveField(
+            model_name="resourcexresource",
+            name="datestarted",
+        ),
+        migrations.RemoveField(
+            model_name="resourcexresource",
+            name="dateended",
```
chiatt, line 126: I think removing these is good. They're really holdovers from the older generic resource relationship approach. People should use nodes to manage that data.

chiatt (approved):

PR #12018, opened 2025-05-03
Fix failure when loading duplicate relations from thesaurus

jacobtylerwalls:
Regression in 0834197.


robgaston (approved):
lgtm - thanks!

PR #12021, opened 2025-05-03
Remove further queries [May 2025]

### LINKED ISSUE #11910: `GraphQuerySet` should be updated to prefetch related entities.
This is a followup to https://github.com/archesproject/arches/pull/11864

> This is great -- everything is accessed on self, so any prefetches can be reused. But now there's a nested loop here, so we should check to see where the caller is and have that do a prefetch. I walked backward, and ended up at `GraphModel.__init__()`. So what we want is a way to force anybody fetching a Graph to do the prefetches that we'll need by the time we get here.
> 
> That sounds like a use case for GraphManager, because we want to enforce that for all traffic coming here.
> 
> You could express that as:
> 
> ```py
> class GraphQuerySet(models.QuerySet):
>     def prefetch_related_entities(self):
>         return self.prefetch_related("cardmodel_set", "cardmodel_set__cardxnodexwidget_set", ...)
> 
> class GraphManager(models.Manager):
>     def get_queryset(self):
>          return super().get_queryset().prefetch_related_entities()
> 
> class GraphModel(...
>     objects = GraphManager.from_queryset(GraphQuerySet)()
> ```
> 
> And then anybody that didn't want that (lol) could still cancel it with `Graph.objects.prefetch_related(None)`.
> 
> But we'd be doing the right thing basically all the time.
> 
> This is what I was alluding to the other day in our huddle when I mentioned to @chiatt that we can still get fast lookups without serialized graphs by being a little more heavy-handed about forcing traffic to go through these prefetches.

_Originally posted by @jacobtylerwalls in https://github.com/archesproject/arches/pull/11864#discussion_r2019230195_

jacobtylerwalls:
I was testing the version of the django-auto-prefetch that's got a pull request against django/django, which comes with a debug tool similar to charettes/django-seal. I used it to identify queries to remove.

A few reasons I stopped here:
- I'm in a JetBlue window seat ✈️ 
- if the django-auto-prefetch work lands in Django 6 we can just use that and remove a lot of these fiddly-prefetches.
- otherwise, it might be nice to leave this work for someone who wants to get up to speed with the Django ORM.
- some of these ORM calls might be better rewritten as starting from `self` rather than starting from a fresh model manager, to allow reusing a prefetched cache

We should probably look into a CI job that runs charettes/django-seal and compares it to some sort of baseline so that we can "plug the leak before bailing out the leak"

### DIFF
--- a/arches/app/datatypes/concept_types.py
+++ b/arches/app/datatypes/concept_types.py
@@ -81,7 +81,9 @@ class BaseConceptDataType(BaseDataType):
             return self.value_lookup[valueid]
         except:
             try:
-                self.value_lookup[valueid] = models.Value.objects.get(pk=valueid)
+                self.value_lookup[valueid] = models.Value.objects.select_related(
+                    "concept", "valuetype"
+                ).get(pk=valueid)
                 return self.value_lookup[valueid]
             except ObjectDoesNotExist:
                 return models.Value()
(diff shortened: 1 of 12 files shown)


chiatt (approved):
Looks good!

PR #12022, opened 2025-05-03
BDM bulk delete speed improvements

### LINKED ISSUE #12015: speed improvement to bulk deletion of resources
There are a few bottlenecks with bulk deletion of resources (whether from the graph designer or the bulk delete ETL module)
- EditLog objects are saved one at a time instead of in bulk; these could be bulk_inserted by batch
- `resource.save_descriptors()` gets called from every `tile.delete()` no matter what
- the ResourceProxy model deletes an instance's related resources which leads to an _ORM query against ResourceXResource for each individual resource_ to be deleted; these could also be batched akin to how resources get bulk indexed using a prefetch

whatisgalen:
### Types of changes

### Description of Change

Deleting data should be the fastest CRUD operation, so why does it take as long as (or longer than) Load inserts? The short answer is: 
- too much stored in-memory which can lead to costly swaps
- various unnecessary indexing operations
- a few too many ORM queries

A few highlights:
- in the `ResourceXResource.delete` method, deletes a tile that has no related resources in tile.data array
- creates a new kwarg for `tile.save` and `tile.delete`: `recalculate_descriptors=True` since most of the time it's not needed, especially when bulk indexing usually handles `recalculate_descriptors` in its own block of code
- `Resource` proxy model now has `self.torelations = []` and `self.fromrelations = []` properties for storing prefetched relations
- `Resource` proxy model `save` method now calls `self.save_descriptors` and passes `recalculate_descriptors=False` to local `tile.save`s
- `index_database.optimize_resource_iteration` now prefetches a resource's related resources
- `BulkDataDeletion.delete_resources` uses `optimize_resource_iteration` instead of doing its own thing
- `reverse_edit_log_entries` now uses `iterators`, bulk indexes tile deletion, and doesn't index until necessary
- failed `load.reverse` will reset the load status to the original status instead of being stuck in "unloading"

### Issues Solved

Addresses #12015

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/etl_modules/base_data_editor.py
+++ b/arches/app/etl_modules/base_data_editor.py
@@ -3,6 +3,7 @@ import json
 import logging
 from urllib.parse import urlsplit, parse_qs
 import uuid
+from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
 from django.core.validators import URLValidator
 from django.db import connection
@@ -10,7 +11,7 @@ from django.http import HttpRequest
 from django.utils.decorators import method_decorator
 from django.utils.translation import gettext as _
 from arches.app.datatypes.datatypes import DataTypeFactory
-from arches.app.models.models import GraphModel, Node, ETLModule, LoadStaging
+from arches.app.models.models import GraphModel, Node, ETLModule, LoadStaging, LoadEvent
 from arches.app.models.system_settings import settings
 from arches.app.search.elasticsearch_dsl_builder import (
     Bool,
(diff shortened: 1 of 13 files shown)


whatisgalen:
I didn't do a numerical comparison but reversing loads is significantly faster, somewhere between 10-50x faster

arches/app/etl_modules/base_data_editor.py
```diff
@@ -51,12 +52,17 @@ def __init__(self, request=None, loadid=None):
         self.node_lookup = {}
 
     def reverse_load(self, loadid):
+        user = (
+            User.objects.get(id=self.userid)
```
jacobtylerwalls, line 56: Seems like we should be storing the `request.user` instead of re-querying

jacobtylerwalls: But maybe this is not the time to do that if there are too many downstream impacts, it was just an impulse. WIll try to find time to do a real review :-)

PR #12023, opened 2025-05-03
Add multicolumn index on TileModel

jacobtylerwalls:
### Types of changes

### Description of Change
v8 changed `TileModel.nodegroup_id` from ForeignKey to UUIDField. This means the column is no longer indexed. This could be considered a regression. (EDIT: we changed it back to `ForeignKey(db_constraint=False, db_index=False)`, which amounts to the same thing just to make it easier to use with the ORM in #12033.)

We can fix that, but we can also use it as an opportunity to index the most common lookup, which is a lookup with _both_ resourceinstance and nodegroup. Ordering nodegroup _first_ ensures it can also be used for searches using that column alone.

Preliminary testing on a large subset from the provenance dataset showed this to be superior to indexing nodegroup alone.

arches-querysets in particular has a use case.


arches/app/models/models.py
```diff
@@ -1702,7 +1702,7 @@ class TileModel(SaveSupportsBlindOverwriteMixin, models.Model):  # Tile
         related_query_name="child",
     )
     data = JSONField(blank=True, default=dict, db_column="tiledata")
-    nodegroup_id = models.UUIDField(db_column="nodegroupid", null=True)
+    nodegroup_id = models.UUIDField(db_column="nodegroupid", null=True, db_index=False)
```
jacobtylerwalls: No change here (db_index is False by default--notice no migration was generated), but I left it in to reinforce the comment below.

jacobtylerwalls:
@apeters to add a little more detail, there are 2 questions here.

1 - Should we restore the index on `TileModel.nodegroup` we had in v7? I think this is an easy yes.
2 - While here, do we want to add a multicolumn index to further optimize for tile lookups by nodegroup + resource? I think this is still yes.

Postgres says multicolumn indexes [should be used sparingly](https://www.postgresql.org/docs/current/indexes-multicolumn.html), but it seems the reason they exist is when your leftmost column (nodegroup) is checked for equality (or containment), which is exactly what we have in our most common lookups (certainly with arches-querysets [here](https://github.com/archesproject/arches-querysets/blob/47e8dbb2fe1af0d16b6a66c1c0e1f65302a9d76e/arches_querysets/utils/models.py#L98-L101) and [here](https://github.com/archesproject/arches-querysets/blob/47e8dbb2fe1af0d16b6a66c1c0e1f65302a9d76e/arches_querysets/querysets.py#L101))

**Option 1** (parity with 7.6 -- single column index on nodegroup):
```py
In [5]: TileModel.objects.filter(resourceinstance=ri, nodegroup_id=CONCEPT_NAME_NOD
   ...: EGROUP).explain()
Out[5]: "Bitmap Heap Scan on tiles  (cost=4.33..24.02 rows=1 width=934)\n  Recheck Cond: (resourceinstanceid = '0005820e-8c89-35f0-9560-7663abdb392f'::uuid)\n  Filter: (nodegroupid = 'ab9fee9c-0eb6-11ef-93db-0a58a9feac02'::uuid)\n  ->  Bitmap Index Scan on tiles_resourceinstanceid_e41126e6  (cost=0.00..4.33 rows=5 width=0)\n        Index Cond: (resourceinstanceid = '0005820e-8c89-35f0-9560-7663abdb392f'::uuid)"
```

**Option 2** (multicolumn index):
```py
In [5]: TileModel.objects.filter(resourceinstance=ri, nodegroup_id=CONCEPT_NAME_NOD
   ...: EGROUP).explain()
Out[5]: "Index Scan using nodegroup_and_resource on tiles  (cost=0.41..8.43 rows=1 width=934)\n  Index Cond: ((nodegroupid = 'ab9fee9c-0eb6-11ef-93db-0a58a9feac02'::uuid) AND (resourceinstanceid = '0005820e-8c89-35f0-9560-7663abdb392f'::uuid))"
```

Notice the much lower query cost with option 2. I don't have the numbers for provenance, but it was much more dramatic than these numbers with lingo when I checked last time.

apeters:
@jacobtylerwalls This PR looks like it only adds the composite index, but the not the `nodegroup only` index.  Like you mentioned, I think we'd want both, and maybe the composite index can do that, but I don't know, do you?.

jacobtylerwalls:
> maybe the composite index can do that, but I don't know, do you?.

Yep that's it -- the multicol index is also used if you're querying the first column only (here, nodegroup). This avoids having to maintain 2 indexes.

Notice this query uses the "nodegroup_and_resource" index from this PR:
```py
In [6]: TileModel.objects.filter(nodegroup_id=uuid.uuid4()).explain()
Out[6]: "Index Scan using nodegroup_and_resource on tiles  (cost=0.41..12.35 rows=2 width=934)\n  Index Cond: (nodegroupid = 'a5b74c05-6f87-4cb3-b15c-f88570ef9151'::uuid)"
```

apeters (approved):
looks great

PR #12024, opened 2025-05-05
Remove user query deluge during indexing #11553

### LINKED ISSUE #11553: Permissions framework classes aren't stateful
The permissions framework classes inheriting from `ArchesPermissionBase` don't have any state, e.g. instance attributes. They get instantiated as singletons and attached as a module-level constant. They're really just a collection of pure functions, probably to rhyme with the pure function shortcuts from django-guardian.

What this means in practice is that there's no way to avoid repetitive queries like `User.objects.all()` inside `get_restricted_users()`, a pure function. @chiatt profiled indexing and found that almost all of the time was spent inside `get_restricted_users()`. Much of that in turn is due to [design flaws](https://github.com/django-guardian/django-guardian/blob/a61df706b060b1ca45addae9febd5274d482f0b7/guardian/shortcuts.py#L325) in django-guardian, but some of it is avoidable on our side. The time spent iterating the Users actually takes longer than iterating the ResourceInstances!

https://github.com/archesproject/arches/blob/a8e97834ff8e6e9658e8e82d1bc7c8871258cc3d/arches/app/permissions/arches_default_allow.py#L134

For 1 million resources, that means 1 million avoidable `User.objects.all()` queries if we could just provide the users somehow to `get_users_with_perms()`.

If we can't provide state at the instance attribute level because we want to continue treating the instances as singletons, maybe we can provide some sort of dictionary holding state and adjust the signatures of all the pure functions to take that dict as an optional keyword argument.

jacobtylerwalls:
### Types of changes

### Description of Change
Indexing tasks queried the User table once per resource. (For 1m resources, that was 999,999 wasteful queries.)

Further, we were also querying for groups individually instead of fetching a user's set of groups and testing that cache.

### Issues Solved
Closes #11553

### Checklist

### DIFF
--- a/arches/app/permissions/arches_permission_base.py
+++ b/arches/app/permissions/arches_permission_base.py
@@ -365,9 +365,9 @@ class ArchesPermissionBase(PermissionFramework, metaclass=ABCMeta):
                 )
                 if result is not None:
                     if result["permitted"] == "unknown":
-                        return user.groups.filter(
-                            name__in=settings.RESOURCE_EDITOR_GROUPS
-                        ).exists() or self.user_can_edit_model_nodegroups(
+                        return self.user_in_group_by_name(
+                            user, settings.RESOURCE_EDITOR_GROUPS
+                        ) or self.user_can_edit_model_nodegroups(
                             user, result["resource"]
                         )
                     else:
@@ -376,7 +376,7 @@ class ArchesPermissionBase(PermissionFramework, metaclass=ABCMeta):
                     return None
 
             return (
-                user.groups.filter(name__in=settings.RESOURCE_EDITOR_GROUPS).exists()
+                self.user_in_group_by_name(user, settings.RESOURCE_EDITOR_GROUPS)
                 or len(self.get_editable_resource_types(user)) > 0
             )
         return False
@@ -410,9 +410,9 @@ class ArchesPermissionBase(PermissionFramework, metaclass=ABCMeta):
                         )
                         if len(protected_tiles) > 0:
                             return False
-                        return user.groups.filter(
-                            name__in=settings.RESOURCE_EDITOR_GROUPS
-                        ).exists() or self.user_can_delete_model_nodegroups(
+                        return self.user_in_group_by_name(
+                            user, settings.RESOURCE_EDITOR_GROUPS
+                        ) or self.user_can_delete_model_nodegroups(
                             user, result["resource"]
                         )
                     else:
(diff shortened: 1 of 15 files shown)


arches/app/permissions/arches_permission_base.py
```diff
@@ -503,32 +503,35 @@ def user_can_read_concepts(self, user: User) -> bool:
...
 
-        return user.groups.filter(name="Resource Exporter").exists()
+        return self.user_in_group_by_name(user, ["Resource Exporter"])
 
     def user_in_group_by_name(self, user: User, names: Iterable[str]) -> bool:
-        return bool(user.groups.filter(name__in=names))
+        for group in user.groups.all():
+            if group.name in names:
+                return True
+        return False
```
jacobtylerwalls, line 534: This populates the cache (if not prefetched already) so that we avoid going to the db every time to test this.

arches/app/utils/index_database.py
```diff
@@ -247,6 +246,7 @@ def index_resources_using_singleprocessing(
         str(nodeid): datatype
         for nodeid, datatype in models.Node.objects.values_list("nodeid", "datatype")
     }
+    all_users = User.objects.prefetch_related("groups")
```
jacobtylerwalls, line 249: Doing this outside the loop saves millions(?) of queries.

tests/models/resource_test.py
```diff
@@ -486,6 +486,13 @@ def test_recalculate_descriptors_prefetch_related_objects(self):
                 ]
                 self.assertEqual(len(tile_selects), 1)
 
+                non_guardian_user_selects = [
+                    q
+                    for q in queries
+                    if q["sql"].endswith('FROM "auth_user"') and "guardian" not in q
+                ]
+                self.assertEqual(len(non_guardian_user_selects), 1)
```
jacobtylerwalls, line 494: This is 1 of 2 new assertions -- the rest is cleanup (more obvious if you review by commit)

tests/views/resource_tests.py
```diff
@@ -162,47 +156,45 @@ def test_user_cannot_edit_without_permission(self):
...
-            )
+        self.assertGreater(len(group_dict["system_permissions"]), 0)
 
-            self.assertGreater(len(group_dict["system_permissions"]), 0)
+        # Groups should not be individually queried. Instead,
+        # membership should be tested against all prefetched groups.
+        resource_editor_query = [
+            query for query in queries if "Resource Editor" in query["sql"]
+        ]
+        self.assertEqual(resource_editor_query, [])
```
jacobtylerwalls, line 199: New assertion 2 of 2

aarongundel (approved):
LGTM

PR #12029, opened 2025-05-06
Nudge authors toward hyphenated package names #12028

### LINKED ISSUE #12028: Nudge arches applications to use hyphens in their package name
Hyphens are the convention for python package names, not underscores, see [PEP 8](https://peps.python.org/pep-0008/#package-and-module-names):

> Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

Also arches' own metadata:

https://github.com/archesproject/arches/blob/4a961ec24413a79a6c35bafdc6c8f36142093aef/pyproject.toml#L34-L42

We can nudge authors toward this by fixing the project template.

jacobtylerwalls:
### Types of changes

### Issues Solved

Closes #12028

### DIFF
--- a/arches/install/arches-templates/pyproject.toml
+++ b/arches/install/arches-templates/pyproject.toml
@@ -3,7 +3,7 @@ requires = ["setuptools>=77", "setuptools-scm"]
 build-backend = "setuptools.build_meta"
 
 [project]
-name = "{{ project_name }}"
+name = "{{ project_name_kebab_case }}"
 readme = "README.md"
 authors = []
 license = "AGPL-3.0-or-later"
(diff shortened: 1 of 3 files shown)


johnatawnclementawn (approved):
Would it also make sense to nudge the ~~project~~ app directory name in the same direction? e.g. if I run `arches-admin startproject my_project`, it may be a little confusing if my new project looks like this:

- my_project
    - my_project
    - pyproject.toml (which defines name as `my-project`)

I would tend to expect this:
- my-project
    - my_project
    - pyproject.toml (which defines name as `my-project`)

johnatawnclementawn (approved):
LGTM 👍🏼

jacobtylerwalls:
@ekansa how about we add a little debug output to the terminal when a `_` -> `-` replacement takes place?

PR #12037, opened 2025-05-07
fix graph caching issue, re #12036

### LINKED ISSUE #12036: Graphs updated after server startup don't reflect those changed in the RelatedResourcesView class
Here's the offending code

```
@method_decorator(can_read_resource_instance, name="dispatch")
class RelatedResourcesView(BaseManagerView):
    action = None
    graphs = (
        models.GraphModel.objects.all()
        .exclude(pk=settings.SYSTEM_SETTINGS_RESOURCE_MODEL_ID)
        .exclude(isresource=False)
        .exclude(publication=None)
    )

    def paginate_related_resources(self, related_resources, page, request):
        total = related_resources["total"]["value"]

.....

```

The issue is if `publication` is None when the server starts up then it is never included in methods of the RelatedResourcesView even if it is updated afterwards via the UI.  Any references to `self.graphs` later in the code are affected by this.

apeters:
### Types of changes

### Description of Change

This fixes an issue where graphs are cached on server start up and updates to those graphs aren't reflected in running code which can cause an issue in the related resources view.

### Issues Solved

Closes #12036

### Checklist

### DIFF
--- a/arches/app/views/resource.py
+++ b/arches/app/views/resource.py
@@ -975,12 +975,6 @@ class ResourceReportView(MapBaseManagerView):
 @method_decorator(can_read_resource_instance, name="dispatch")
 class RelatedResourcesView(BaseManagerView):
     action = None
-    graphs = (
-        models.GraphModel.objects.all()
-        .exclude(pk=settings.SYSTEM_SETTINGS_RESOURCE_MODEL_ID)
-        .exclude(isresource=False)
-        .exclude(publication=None)
-    )
 
     def paginate_related_resources(self, related_resources, page, request):
         total = related_resources["total"]["value"]
@@ -1020,7 +1014,7 @@ class RelatedResourcesView(BaseManagerView):
 
         return ret
 
-    def get(self, request, resourceid=None, include_rr_count=True):
+    def get(self, request, resourceid=None, include_rr_count=True, graphs=None):
         ret = {}
 
         if self.action == "get_candidates":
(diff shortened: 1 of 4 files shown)


apeters:
To test the failure, do this:

1. Select any graph and unpublish it.
2. Restart the django server.
3. Go to the search page and select the "Related Resources" link from any search result.
4. Notice the error (this is actually not the issue, but a separate one) and that the Cytoscape graph doesn't render
5. Now, republish the graph you unpublished (do NOT restart the server)
6. Go back to the search page and select any other "Related Resources" link from the search results (make sure it's a different one from the one previously selected)
7. Notice the error still and the Cytoscape graph still doesn't render.

After applying the fix you should notice that the Cytoscape graph now renders in step 7.

apeters:
This is also an issue in v8, except that instead of "unpublishing" the graph you should "deactivate" it.   
This should be cherry picked to v8 as well

arches/app/views/resource.py
```diff
@@ -1051,6 +1045,13 @@ def get(self, request, resourceid=None, include_rr_count=True):
                 request.GET.get("paginate", "true")
             )  # default to true
             resource = Resource.objects.get(pk=resourceid)
+            if graphs is None:
+                graphs = (
+                    models.GraphModel.objects.all()
+                    .exclude(pk=settings.SYSTEM_SETTINGS_RESOURCE_MODEL_ID)
+                    .exclude(isresource=False)
+                    .exclude(publication=None)
+                )
```
jacobtylerwalls: I think we can just delete this, looks like `get_related_resources()` defaults to this same query if `graphs` is None.

apeters: good call

jacobtylerwalls (approved):

PR #12039, opened 2025-05-07
Use dependency-groups for dev dependencies #12038

### LINKED ISSUE #12038: Replace project.optional-dependencies with dependency-groups
Now that pip 25.1 [supports dependency-groups](https://ichard26.github.io/blog/2025/04/whats-new-in-pip-25.1/#key-features-), we can do the "right" thing (see below) and move our dev dependencies to dependency-groups.

This is:
- more correct (the dev dependencies are not an optional part of arches functionality, e.g. `arches[iiif]`)
- and avoids the clunky `'.[dev]'` syntax in favor of `--group dev`

jacobtylerwalls:
The `'.[dev]'` is clunky, with its apostrophes for shell-escaping, and wasn't really designed for packages that are not part of your application functionality (as opposed to IIIF which would be). **pip 25.1** has something better now, so use that.

Closes #12038

### DIFF
--- a/.github/actions/build-and-test-branch/action.yml
+++ b/.github/actions/build-and-test-branch/action.yml
@@ -27,7 +27,7 @@ runs:
     - name: Install Python packages
       run: |
         python -m pip install --upgrade pip
-        pip install '.[dev]'
+        pip install . --group dev
         echo Python packages installed
       shell: bash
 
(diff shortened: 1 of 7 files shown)


chrabyrd (approved):
LGTM!

PR #12043, opened 2025-05-08
Replace stray uses of uuidv1

jacobtylerwalls:
These seem to be older inconsistent uses -- all models are using uuidv4 as their defaults.


samuelscandrett (approved):
This all seems fine to me, and I've often wondered whether we need uuid1's in Arches.

I don't suppose we need to migrate any existing uuids, as they'll just exist fine (plus that could get messy); then any new nodes, widgets etc will be given a uuid4?

I've tested modifying graphs, creating new ones and the same for resources - all seems normal.

jacobtylerwalls:
Thanks for checking!

PR #12049, opened 2025-05-09
Fix --clear_index parameter unable to be set to False in command line

### LINKED ISSUE #12048: --clear_index parameter for the es management commands cannot not be set to false
There appear to be no way to stop indexes from being cleared when running any of the `es` indexing commands.

Using the `--clear_index` option does nothing when trying to set it to `False` as it is `True` by default, thus always clears the index.

See below, I have used `-c False` and this is converted to a string and when tested resolves as `True`! All documents for this model got deleted and then slowly repopulated.

Image

As a developer, and seeing that this option/flag was available suggests the `index_*` commands reindexes the documents in place and that you add the `--clear_index`/`-c` flag when you DO want to deleted all documents first.

The default behaviour seems to be for indexes to be cleared, but we need a way to clear this. We need this to work in 7.6 as we want to be able to run an indexing process in production with no loss of data for users as the process runs.

So propose to adjust for now the logic to allow `"False"` to be passed as a value and evaluted as `False`.

aj-he:
### Types of changes

### Description of Change

`--clear_index` parameter of `es` command has no way of being set to `False`, and therefore will always clear the resources out of the index before indexing the resources.

For production systems that may need to use these commands, it removes data from search and causes significant reduction of availability during the process.

Its current behaviour meant that by default it must continue to clear indexes if not explicitly set, but the parameter name must remain the same in case users were explicitly using it (despite it not working as described if using False). Therefore the simplest approach was to use the `type` action to evaluate a string value provided into a boolean using value.lower() == 'true'.

Ideally in the future this would be correctly implemented as an optional flag (store_true) and the default behavior would be non-destructive, but this would be a breaking change.

### Issues Solved

Closes #12048 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Found by: @aj-he 
*   Tested by: @aj-he
*   Designed by: @aj-he

### DIFF
--- a/arches/management/commands/es.py
+++ b/arches/management/commands/es.py
@@ -125,7 +125,7 @@ class Command(BaseCommand):
             action="store",
             dest="clear_index",
             default=True,
-            type=lambda x: str(x).lower() == 'true',
+            type=lambda x: str(x).lower() == "true",
             help="Set to True (default) to remove all the resources from the index before the reindexing operation. Set to False to overwrite in place the existing resources in the index.",
         )
 
--- a/arches/management/commands/es.py
+++ b/arches/management/commands/es.py
@@ -33,6 +33,7 @@ from arches.app.search.mappings import (
     delete_resource_relations_index,
 )
 import arches.app.utils.index_database as index_database_util
+from arches.app.utils.string_utils import str_to_bool
 
 logger = logging.getLogger(__name__)
 
@@ -125,8 +126,8 @@ class Command(BaseCommand):
             action="store",
             dest="clear_index",
             default=True,
-            type=lambda x: str(x).lower() == "true",
-            help="Set to True (default) to remove all the resources from the index before the reindexing operation. Set to False to overwrite in place the existing resources in the index.",
+            type=lambda x: str_to_bool(x),
+            help="Whether to remove all target resources from the index before reindexing. Accepts 'true'/'false', 'yes'/'no', 'y'/'n', 't'/'f', 'on'/'off', or '1'/'0' (case-insensitive). Default is True.",
         )
 
         parser.add_argument(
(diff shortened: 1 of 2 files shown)


arches/management/commands/es.py
```diff
@@ -125,7 +125,8 @@ def add_arguments(self, parser):
             action="store",
             dest="clear_index",
             default=True,
-            help="Set to True(default) to remove all the resources from the index before the reindexing operation",
+            type=lambda x: str(x).lower() == "true",
```
jacobtylerwalls: We have a helper `str_to_bool()` that's a little more permissive (accepts "t", "y", and "yes"), but this works too and I'm all for just changing this to be more sane in the next version.

jacobtylerwalls (approved):
Nice find, +1 tested

I did find that there was a way to achieve a false value before, with an explicit empty string (yuck) `--clear_index ''`, but I take your point we should design something better where the flag undoes the default behavior. Could you open an issue?

arches/management/commands/es.py
```diff
@@ -125,7 +126,8 @@ def add_arguments(self, parser):
             action="store",
             dest="clear_index",
             default=True,
-            help="Set to True(default) to remove all the resources from the index before the reindexing operation",
+            type=lambda x: str_to_bool(x),
```
jacobtylerwalls:
```suggestion
            type=str_to_bool,
```

aj-he: even better 😄

arches/management/commands/es.py
aj-he: done

jacobtylerwalls (approved):

arches/management/commands/es.py
aj-he: done

aj-he:
Future breaking change issue... #12052

PR #12054, opened 2025-05-10
12053 fix reorder widget bug

### LINKED ISSUE #11651: Graph Nodes inexplicably re-ordered on refresh

### LINKED ISSUE #12053: widgets in resource editor dont respect ordering within card defined in graph editor
Found this problem using the CASF application with core Arches v7.6.9

https://github.com/user-attachments/assets/d17e712e-8980-4b65-b086-30cbde970810

May be related to #11651

apeters:
### Types of changes

### Issues Solved

Closes #12053

### Checklist

### DIFF
--- a/arches/app/media/js/viewmodels/card-component.js
+++ b/arches/app/media/js/viewmodels/card-component.js
@@ -125,7 +125,29 @@ define([
         };
 
         this.beforeMove = function(e) {
-            e.cancelDrop = (e.sourceParent!==e.targetParent);
+            if(e.sourceParent === e.targetParent) {
+                const orignalWidgetOrder = [...self.card.widgets()];
+                e.item.card.save(function(response, status, card) {
+                    if(status === 'error') {
+                        params.pageVm.alert(
+                            new AlertViewModel(
+                                'ep-alert-red',
+                                response.responseJSON.title,
+                                response.responseJSON.message,
+                                null,
+                                function(){}
+                            )
+                        );
+                        // we can't use e.cancelDrop because of the async nature of the save
+                        // so we need to manually reset the order of the widgets
+                        // and set the selected widget to the original position
+                        // this is a bit of a hack, but it works
+                        self.card.widgets(orignalWidgetOrder);
+                    }
+                });
+            } else {
+                e.cancelDrop = true;
+            }
         };
 
         this.startDrag = function(e, ui) {
--- a/arches/app/media/js/viewmodels/card-component.js
+++ b/arches/app/media/js/viewmodels/card-component.js
@@ -129,7 +129,6 @@ define([
         };
 
         this.reorderWidgets = function(e) {
-            const orignalWidgetOrder = [...self.card.widgets()];
             e.item.card.save(function(response, status, card) {
                 if(status === 'error') {
                     params.pageVm.alert(
@@ -144,7 +143,11 @@ define([
                     // we can't use e.cancelDrop because of the async nature of the save
                     // so we need to manually reset the order of the widgets
                     // and set the selected widget to the original position
-                    self.card.widgets(orignalWidgetOrder);
+                    const undoSort = (array, sourceIndex, targetIndex) => {
+                        const [movedItem] = array.splice(targetIndex, 1);
+                        array.splice(sourceIndex, 0, movedItem);
+                    };
+                    undoSort(self.card.widgets, e.sourceIndex, e.targetIndex);
                 }
             });
         };
--- a/arches/app/media/js/viewmodels/card-component.js
+++ b/arches/app/media/js/viewmodels/card-component.js
@@ -125,29 +125,28 @@ define([
         };
 
         this.beforeMove = function(e) {
-            if(e.sourceParent === e.targetParent) {
-                const orignalWidgetOrder = [...self.card.widgets()];
-                e.item.card.save(function(response, status, card) {
-                    if(status === 'error') {
-                        params.pageVm.alert(
-                            new AlertViewModel(
-                                'ep-alert-red',
-                                response.responseJSON.title,
-                                response.responseJSON.message,
-                                null,
-                                function(){}
-                            )
-                        );
-                        // we can't use e.cancelDrop because of the async nature of the save
-                        // so we need to manually reset the order of the widgets
-                        // and set the selected widget to the original position
-                        // this is a bit of a hack, but it works
-                        self.card.widgets(orignalWidgetOrder);
-                    }
-                });
-            } else {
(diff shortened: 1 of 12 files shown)


arches/app/models/card.py
```diff
@@ -202,6 +204,14 @@ def save(self):
 
         """
         with transaction.atomic():
+            if self.graph.publication is not None:
```
jacobtylerwalls: To make this a little lazier (avoid fetching the publication)
```suggestion
            if self.graph.publication_id:
```

arches/app/views/graph.py
```diff
@@ -678,8 +678,11 @@ def post(self, request, cardid=None):
         if self.action == "update_card":
             if data:
                 card = Card(data)
-                card.save()
-                return JSONResponse(card)
+                try:
+                    card.save()
+                    return JSONResponse(card)
+                except Exception as e:
+                    return JSONErrorResponse(content=e.args[0], status=403)
```
jacobtylerwalls, line 699: I'd return 400 instead of a permssions-related error code

arches/app/models/card.py
```diff
@@ -202,6 +204,14 @@ def save(self):
...
+                ret = {
+                    "title": _("Unable to Save Card"),
+                    "message": _(
+                        "You cannot edit a card in a published graph.  Please unpublish the graph before editing."
+                    ),
+                }
+                raise Exception(ret)
```
jacobtylerwalls: Maybe something slightly more specific like `ValidationError`? That way you can be sure all other exceptions get transformed to 500 by the global error hander.

tests/views/graph_manager_tests.py
```diff
@@ -470,3 +471,18 @@ def test_graph_import(self):
         self.assertEqual(imported_json[0], [])
         self.assertEqual(imported_json[1]["graphs_saved"], 1)
         self.assertEqual(imported_json[1]["name"], "Cardinality Test Model")
+
+    def test_prevent_card_save_of_published_graph(self):
+        """
+        test to make sure that a card cannot be saved if the graph is published
```
jacobtylerwalls, line 477: I usually chop these if the test title is good enough (it's A+).

While here, it seems weird that the coverage check failed even though you added a test. I'll run the coverage check locally and see if I get the same result.

tests/views/graph_manager_tests.py
```diff
@@ -470,3 +471,18 @@ def test_graph_import(self):
...
+
+        """
+
+        self.client.login(username="admin", password="admin")
+        user_id = self.client.session["_auth_user_id"]
+        logged_in_user = get_user_model().objects.get(pk=user_id)
```
jacobtylerwalls: If we're logging in as admin anyway, seems less roundabout:
```suggestion
        logged_in_user = get_user_model().objects.get(username="admin")
```

jacobtylerwalls: Oh I see it's that way in surrounding tests. No big deal then

arches/app/media/js/viewmodels/card-component.js
```diff
@@ -125,7 +125,29 @@ define([
         };
 
         this.beforeMove = function(e) {
-            e.cancelDrop = (e.sourceParent!==e.targetParent);
+            if(e.sourceParent === e.targetParent) {
```
jacobtylerwalls: We have the same problem in the graph tab, so maybe we should keep the underlying issue open until we tackle that, or did you want to roll in a fix?

arches/app/media/js/viewmodels/card-component.js
```diff
@@ -125,7 +125,29 @@ define([
         };
 
         this.beforeMove = function(e) {
```
jacobtylerwalls: I don't get a banner / error when trying to reorder an entire card (makes sense, that's a separate endpoint)

arches/app/views/graph.py
apeters: I specifically chose 403 because it specifies that "The request failure is tied to application logic", whereas 400 "is typically due to malformed request syntax, invalid request message framing, or deceptive request routing".

tests/views/graph_manager_tests.py
apeters: Actually, since I'm not using the client to perform any kind of action, I'll use your suggestion.

arches/app/views/graph.py
jacobtylerwalls: Glad I asked, I didn't realize the spec had broad language about application logic. Everywhere else but the spec seems to suggest in practice 403 is only used for authorization issues, e.g [postman blog](https://blog.postman.com/what-are-http-status-codes/), [wikipedia](https://en.wikipedia.org/wiki/HTTP_403), [so](https://stackoverflow.com/questions/3290182/which-status-code-should-i-use-for-failed-validations-or-invalid-duplicates), but we can leave this as is.

(Some interesting discussion on that SO post about using 422 instead of 400 for validation, but that's too niche I think, we'd want to set up a global handler for 422's in that case if we started using it.)

tests/views/graph_manager_tests.py
jacobtylerwalls: Ah, it's because the view is not tested, and this is only testing the model. Since this test file is under `tests/views`, do you think it makes more sense to test the view? I find it handy to `self.assertContains, response, "error msg", status_code=403`.

arches/app/media/js/viewmodels/card-component.js
apeters: I'll look to see if it makes sense to do that here

apeters: Yeah, I took care of the other situations as well related to node sort and card sort.

jacobtylerwalls (commented):
I'm getting error responses from the backend consistently now, but the banner only appears sometimes:

https://github.com/user-attachments/assets/be3c1088-68e9-489c-8f4d-fc1d3afaaaa3

jacobtylerwalls (approved):
Vastly improved!

apeters:
Ok fixed that issue you found.

As a side note, when a user tries to move a node to a different parent, we don't even go to the server and just disallow it, but we don't pop up an alert either (this is how it currently works in the UI).

jacobtylerwalls (approved):
Works great!

PR #12059, opened 2025-05-13
Fix usage of startproject command without underscores re #12028

jacobtylerwalls:
### Types of changes

### Description of Change
Regression in #12029 for project names without underscores.

Issue: by creating the directory name eagerly instead of letting the Django command do it, it was failing the check in the Django command that checks for existing python modules on your path.

Solution: only create the directory if the underscore -> hyphen transform took place.

Future: Django 6 allows deleting all this code, see [djangoproject/django#18387](https://github.com/django/django/pull/18387)

#### Ticket Background
*   Sponsored by: 
*   Found by: @ekansa
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/install/arches_admin.py
+++ b/arches/install/arches_admin.py
@@ -124,6 +124,16 @@ class ArchesProjectCommand(TemplateCommand):
         options["project_name_title_case"] = project_name.title().replace("_", "")
         options["project_name_kebab_case"] = project_name.replace("_", "-")
 
+        if options["project_name_kebab_case"] != project_name:
+            self.stdout.write(
+                self.style.NOTICE(
+                    f"Renamed the directory from {project_name} to {options['project_name_kebab_case']}. "
+                    "If this is not desired, use the --directory option to create "
+                    "a directory with the name you want. Consider using a name "
+                    "distinct from your project name."
+                )
+            )
+
         super(ArchesProjectCommand, self).handle(
             "project", project_name, target, **options
         )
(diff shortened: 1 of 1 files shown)


arches/install/arches_admin.py
```diff
@@ -124,6 +124,16 @@ def handle(self, options):
         options["project_name_title_case"] = project_name.title().replace("_", "")
         options["project_name_kebab_case"] = project_name.replace("_", "-")
 
+        if options["project_name_kebab_case"] != project_name:
```
johnatawnclementawn:
```suggestion
        if options["project_name_kebab_case"] != project_name and not target:
```

johnatawnclementawn (changes requested):
I ran 
```mkdir blank80```
then
```python arches/arches/install/arches_admin.py startproject blank_80 --directory blank80```

and still saw the notice about renaming the directory. Suggested code should fix that

File: arches/install/arches_admin.py (hunk 1 above)
Comment (jacobtylerwalls) [reply arches/install/arches_admin.py]: Nice catch!

Comment (jacobtylerwalls) [reply arches/install/arches_admin.py]: Nice catch!

File: arches/install/arches_admin.py [hunk 2]
```diff
@@ -170,13 +180,19 @@ def handle(self, options):
 def command_startproject(args):
     options = vars(args)
     name = options["name"]
-    if not options["directory"]:
+    make_directory = False
+    if not options["directory"] and "_" in name:
+        # The user is supposed to create the --directory themselves. But we
+        # should do it for them if the command is the one that invented the
+        # --directory argument.
+        # TODO(v8.1): remove (Django 6 creates target dir automatically)
```
jacobtylerwalls:
```suggestion
        # RemovedInArches81Warning (Django 6 creates target dir automatically)
```

johnatawnclementawn (approved):
Nice! LGTM 👍🏼

PR #12070, opened 2025-05-14
Remove obsoleted tests re #11808

jacobtylerwalls:
Follow-up to e8996b3.

These tests aren't relevant now that we don't have to demonstrate that joining via the ORM still works -- we're back on the rails and no longer using a quasi-private Django api.


johnatawnclementawn (approved):
👍🏼

PR #12050, opened 2025-05-09
Fix importing relations command when foreign keys are used

MikeGriniezakis:
### Types of changes

-   [x] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

When importing relations using the `import_business_data_relations` package and the imported CSV 
has the column nodeid or tileid and valid UUIDs, an error would be thrown that the appropriate model must be used
instead of UUID. This PR casts the UUIDs to the appropriate model so the models can be created.

### Issues Solved

Closes #12045

### Checklist

-   I targeted one of these branches:
    - [ ] dev/8.1.x (under development): features, bugfixes not covered below
    - [ ] dev/8.0.x (main support): regressions, crashing bugs, security issues, major bugs in new features
    - [x] dev/7.6.x (extended support): major security issues, data loss issues
-   [x] I added a changelog in arches/releases
-   [ ] I submitted a PR to arches-docs (if appropriate)
-   [ ] Unit tests pass locally with my changes
-   [ ] I added tests that prove my fix is effective or that my feature works
-   [ ] My test fails on the target branch

#### Ticket Background
*   Sponsored by: Getty Conservation Institute 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

PR #12072, opened 2025-05-14
Fix importing relations when foreign keys are used

jacobtylerwalls:
Follow-up to 98c11c6.

We would have needed to look at this when merging 7.6 -> 8.0, but just getting out ahead of this to aid whoever does the merge, and since it was ID'd as a follow-up [here](https://github.com/archesproject/arches/pull/12050#discussion_r2081561548).

No changelog since this is fixed in the 7.6.10 changelog.


whatisgalen:
Looks good, how did this one sneak through?

whatisgalen (approved):

jacobtylerwalls:
Thanks for the review!

Issue #10645, opened 2024-03-04
Unpublished graph resources show in search 


SDScandrettKint:
Found in V7.

Resources that are under a graph that is unpublished will still remain in the search results.  Clicking on one of the resources results in endless loading.  The Arches server log will show the following error:
```
Internal Server Error: /api/resource_report/e7067530-1f70-4ee3-a53d-629b8f9e1ea6
Traceback (most recent call last):
  File "/mnt/c/arches_dev/v7_development/env3.11_v7dev/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/arches_dev/v7_development/env3.11_v7dev/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/arches_dev/v7_development/env3.11_v7dev/lib/python3.11/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/arches_dev/v7_development/arches/arches/app/views/api.py", line 108, in dispatch
    return super(APIBase, self).dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/arches_dev/v7_development/env3.11_v7dev/lib/python3.11/site-packages/django/views/generic/base.py", line 143, in dispatch
    return handler(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/arches_dev/v7_development/arches/arches/app/views/api.py", line 1111, in get
    "report_json": resource.to_json(compact=compact, version=version),
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/arches_dev/v7_development/arches/arches/app/models/resource.py", line 783, in to_json
    return LabelBasedGraphV2.from_resource(
[...]

PR #12095, opened 2025-05-20
Remove resources from unpublished graphs from search results

CWDamm-Kint:
### Types of changes

### Description of Change
Resources from unpublished graphs still show in search results. This PR filters them out within arches_permission_base.py, upstream of most uses within search components. 

### Issues Solved

Partially solves #10645, but a PR for v8 will also be needed. 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Found by: @CWDamm-Kint @SDScandrettKint 
*   Tested by: @CWDamm-Kint @SDScandrettKint 
*   Designed by: @CWDamm-Kint


whatisgalen:
@CWDamm-Kint you may want to check out the method `get_nodegroups_by_perm` which is implemented by the permission backends as that might be the best place to add the publication not null filter since it's upstream of all other `get_permitted_nodegroups` calls

CWDamm-Kint:
Thanks @whatisgalen. Seems like a good idea, unless you can think of any scenarios in search where you might want to use nodegroups that are permitted, but not published? (I can't think of any). I've pretty much finished it locally, but I'm about to go on leave until Tuesday, so I'll commit it when back.

arches/app/permissions/arches_permission_base.py
```diff
@@ -186,15 +186,39 @@ def get_nodegroups_by_perm(
...
+            if node.graph_id in published_permitted_graph_uids
+        ]
+
+        published_permitted_nodegroups = [
+            nodegroup_id
+            for nodegroup_id in permitted_nodegroup_uids
+            if nodegroup_id in published_permitted_node_uids
+        ]
+
+        return published_permitted_nodegroups
```
samuelscandrett: I'm not sure if I like the amount of chaining that's happening here.  I get that looking purely at db tables to get nodegroups from published graphs we have to jump from nodegroup > node > graph and then backwards, but this way we end up with a lot of lists that can get quite long (possible performance implications).

Could we do something like this instead which uses the nodes table to find nodegroups (as all nodegroups will exist within nodes too)?
```
        permitted_nodegroups = list(
            set(
                nodegroup.pk
                for nodegroup in get_nodegroups_by_perm_for_user_or_group(
                    user, perms, any_perm=any_perm
                )
            )
        )
        
        permitted_graphids = set()
        for node in Node.objects.filter(nodegroup__in=permitted_nodegroups):
            permitted_graphids.add(str(node.graph_id))

        published_permitted_graphids = set(
            str(graphid)
            for graphid in GraphModel.objects.filter(graphid__in=permitted_graphids)
            .exclude(publication=None)
            .values_list("graphid", flat=True)
        )

        published_permitted_nodegroups = set()
        for node in Node.objects.filter(nodeid__in=permitted_nodegroups, graph_id__in=published_permitted_graphids):
            published_permitted_nodegroups.add(str(node.nodeid))

        return list(published_permitted_nodegroups)
```

arches/app/permissions/arches_permission_base.py
```diff
@@ -37,7 +37,7 @@
 from arches.app.models.models import Node, NodeGroup, TileModel
 from django.db.models import Q
 from arches.app.models.system_settings import settings
-from arches.app.models.models import ResourceInstance, MapLayer
+from arches.app.models.models import ResourceInstance, MapLayer, GraphModel
```
samuelscandrett: Interesting that we have two lines for `from arches.app.models.models import...` (L37 and L40).
Can we merge them into one line?

samuelscandrett (commented):
I'm still cautious that removing unpublished nodegroups/graphs from `get_nodegroups_by_perm` could have downstream effects on some permissions, especially as `get_nodegroups_by_perm` is called by many other files/functions.

Unpublished resources can't be viewed/edited anyway, so perhaps this won't actually make a difference?  I didn't notice any issues when viewing the nodegroups in the graph either.

arches/app/permissions/arches_permission_base.py
jacobtylerwalls: I think you can push all of that into the db like (adjust to break out into statements as desired):
```py
return set(NodeGroup.objects.filter(
    pk__in=permitted_nodegroups,
    node__graph__publication__isnull=False,
).values_list("pk", flat=True))
```

CWDamm-Kint: Thanks, didn't know you could link them like that.

arches/app/permissions/arches_permission_base.py
```diff
@@ -186,15 +191,26 @@ def get_nodegroups_by_perm(
...
         """
-        return list(
+
+        permitted_nodegroups = get_nodegroups_by_perm_for_user_or_group(
+            user, perms, any_perm=any_perm
+        )
+
+        permitted_nodegroup_ids = list(
+            set(nodegroup.pk for nodegroup in permitted_nodegroups)
+        )
```
samuelscandrett: Do we need both lists, or can they be combined?

chiatt: I think permitted_nodegroups is actually a dictionary, so it's still necessary to convert that to a list of ids. I don't think there's any need to cast things to a set and back to a list though. Everything is already unique because the nodegroups are dictionary keys.

I think this would be more explicit: 
```
permitted_nodegroups = get_nodegroups_by_perm_for_user_or_group(
            user, perms, any_perm=any_perm
        ).keys()
        
permitted_nodegroup_ids = [nodegroup.pk for nodegroup in permitted_nodegroups]
```

arches/app/permissions/arches_permission_base.py
```diff
@@ -34,10 +34,15 @@
...
+from arches.app.models.models import (
+    Node,
+    NodeGroup,
+    TileModel,
+    ResourceInstance,
+    MapLayer,
+)
 from django.db.models import Q
 from arches.app.models.system_settings import settings
-from arches.app.models.models import ResourceInstance, MapLayer
```
chiatt: I understand the temptation to fix this, but I think it's best to avoid making formatting changes unrelated to the PRs target issue. It makes reviewing code a bit easier and helps avoid unnecessary merge conflicts.

arches/app/search/components/resource_type_filter.py
```diff
@@ -19,8 +19,10 @@
 
 def get_permitted_graphids(permitted_nodegroups):
     permitted_graphids = set()
+
     for node in Node.objects.filter(nodegroup__in=permitted_nodegroups):
         permitted_graphids.add(str(node.graph_id))
+
```
chiatt: Probably best to avoid format fixes unrelated to the PR's issue.

arches/app/permissions/arches_permission_base.py
jacobtylerwalls: Yeah I'm expecting us to just batch fix these in #11909

samuelscandrett: That's fair, it was just something I noticed when @CWDamm-Kint extended the second import from models.models.

apeters:
@CWDamm-Kint 
I do fear that there may be unintended side effects with this.   Specifically, the name of the method get_nodegroups_by_perm only implies the permissions that have been applied to node groups.  Whether a graph is published or not, in that sense, doesn’t have anything to do with permissions.

If the goal of this is to filter from search results instances from unpublished graphs, I would probably just add that additional logic to the search to filter out those instances (which is what you did at first).   Or alternatively, and in an effort to maintain backward compatibility, add a flag to get_nodegroups_by_perm called include_unpublished_graphs=True (or something like it, you get the idea). (edited)

tests/search/search_tests.py
```diff
@@ -488,3 +490,31 @@ def test_adv_search_on_null_file_node(self):
...
         results = JSONDeserializer().deserialize(results.content)["results"]["hits"]
         self.assertEqual(1, len(results["hits"]))
+
+    def test_no_resources_from_unpublished_graphs(self):
+
+        test_graph = Graph.objects.get(name="Resource Test Model")
+
+        Resource.objects.create(
+            graph=test_graph,
+            name="Resource from unpublished graph",
```
whatisgalen, line 500:
```suggestion
            name={"en": "Resource from unpublished graph"},
```

whatisgalen: fyi the name field on a Resource is json

jacobtylerwalls: The I18n field is "smart" enough to wrap this as JSON, keying off the current language during the trip to the database. I think since that's supposed to be a usability feature, keeping the tests modeling ordinary usage (as it is here) might be good, but I don't have a strong opinion on what's clearer.

CWDamm-Kint:
> @CWDamm-Kint I do fear that there may be unintended side effects with this. Specifically, the name of the method get_nodegroups_by_perm only implies the permissions that have been applied to node groups. Whether a graph is published or not, in that sense, doesn’t have anything to do with permissions.
> 
> If the goal of this is to filter from search results instances from unpublished graphs, I would probably just add that additional logic to the search to filter out those instances (which is what you did at first). Or alternatively, and in an effort to maintain backward compatibility, add a flag to get_nodegroups_by_perm called include_unpublished_graphs=True (or something like it, you get the idea). (edited)

Thanks @apeters. On reflection, I think I'd probably agree with moving it back to where it was, for the reasons discussed. Apologies for the flip-flopping. Any other views on this @whatisgalen, @SDScandrettKint , @chiatt, @jacobtylerwalls? I'd bring it to stand up, but unfortunately I can't make any of the ones coming up.

arches/app/search/components/resource_type_filter.py
```diff
@@ -18,12 +18,19 @@
...
+        permitted_graphids.add(str(node.graph_id))
+
+    permitted_graphids = set(
+        str(graphid)
+        for graphid in GraphModel.objects.filter(graphid__in=permitted_graphids)
+        .exclude(publication=None)
+        .values_list("graphid", flat=True)
+    )
+
+    return permitted_graphids
```
chiatt: I think it would be clearer to use different variable names rather than to reassign a different value to the same variable name. Something like: 
```suggestion
    permitted_and_published_graphids = set(
        str(graphid)
        for graphid in GraphModel.objects.filter(graphid__in=permitted_graphids)
        .exclude(publication=None)
        .values_list("graphid", flat=True)
    )

    return permitted_and_published_graphids
```

chiatt (changes requested):

chiatt (approved):
Looks good to me. I think this is a nice solution that addresses @apeters [comment](https://github.com/archesproject/arches/pull/12095#issuecomment-2932978760)

PR #12097, opened 2025-05-20
fixes #11848 - deleted spatial views

### LINKED ISSUE #11848: Graph publication workflow in v8 deletes spatial views
The graph publication workflow in v8 implements graph versioning by allowing _draft_ (aka "editable future") graphs. The publishing and reverting workflows then involve deleting the draft graph and associated entities (nodes/nodegroups/cards/widgets etc) and recreating the target version graph from the serialized representation in the database. Spatial view model instances cascade delete if their geometry nodes are deleted, with the result that spatial views get wiped on any graph publish/revert.

Perhaps we should:
- Change SpatialView.geometrynode to on_delete=models.PROTECT
- store a reference in memory to the target geometry node before doing the delete
- re-establish the link after the publishing operation is done

#### Reproduction
dev/8.0.x
1. Create a model with a string node and a geojson node. Save edits.
2. Publish the graph.
3. Click show ids and keep this tab open to be able to copy easily.
4. In a new tab: Django admin > Spatial views > Add
5. In the form:
- add a slug
- pick geojson node from geometry node dropdown
- pick a language
- add this json array to attributenodes: `[{"nodeid": <string node pk>; "description": "string"}]`
- save
6. In graph editor tab: add description "added spatial view"
7. Save edits
8. Create new publication
9. Refresh admin

Result: no spatial views exist

/cc @chrabyrd

aarongundel:
### Types of changes

### Description of Change

To make this work, 
* I changed the geometrynode in spatialviews to PROTECT and allowed it to be nullable.
* I save the geometry nodes at the last possible second and set geometrynode to null.
* After publishing is complete, I restore the geometrynode to the view.  To do this, I had to tack on some state to the graph class.  I don't love doing this - happy to hear alternatives.  One option was to move this whole procedure outside of update_from_draft_graph and publish, but the downside is that whenever you call update_from_draft_graph followed by publish you'd need to wrap it in this spatial view geometrynode swapper.  That seemed less good than where I landed here.  Maybe I'm missing an obvious option.  
* I had to modify the spatial view creation trigger.  In the event of a null geometry node, it will will skip attempting to recreate the view and bail.  As part of this (and as discussed in stand up) I added django pgtrigger.  This should make future changes easier to deal with.  It also suggests that models.py is going to be onerously long and we should consider breaking it up (similar to how we're starting to break up datatypes.py)

### Issues Solved

Closes #11848 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


aarongundel:
Test failures appear to be unrelated to this PR.

jacobtylerwalls:
> It also suggests that models.py is going to be onerously long and we should consider breaking it up (similar to how we're starting to break up datatypes.py)

100% agree, we did the same in v8 with api.py

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "django-guardian==2.4.0",
     "django-hosts==6.0.0",
     "django-oauth-toolkit==1.7.0",
+    "django-pgtrigger==4.15.2",
```
jacobtylerwalls, line 39: We usually call out these changes in the v8 release notes.

arches/settings.py
```diff
@@ -375,6 +375,7 @@
     "corsheaders",
     "oauth2_provider",
     "django_celery_results",
+    "pgtrigger",
```
jacobtylerwalls, line 378: - needs adding to template project settings file
- needs adding to upgrade instructions

arches/app/models/models.py
```diff
@@ -2397,6 +2527,11 @@ def clean(self):
         except (KeyError, TypeError):
             raise ValidationError("attributenodes must be a list of node objects")
 
+        if not self.geometrynode:
+            raise ValidationError(
+                "Geometry node must be set to a valid geojson-feature-collection node"
+            )
```
jacobtylerwalls, line 2533: Seems to duplicate the check on L2521, we could probably move this there.

arches/app/models/models.py
```diff
@@ -2383,6 +2385,134 @@ def __str__(self):
...
+                when=pgtrigger.After,
+                operation=pgtrigger.Update | pgtrigger.Delete | pgtrigger.Insert,
+                declare=[
+                    ("sv_perform", "text"),
+                    ("valid_geom_nodeid", "boolean"),
+                    ("has_att_nodes", "integer"),
+                    ("valid_att_nodeids", "boolean"),
+                    ("valid_language_count", "integer"),
+                ],
+                func=pgtrigger.Func(
```
jacobtylerwalls, line 2400: In arches-for-science, we worked around this by putting the SQL string in a [separate file](https://github.com/archesproject/arches-for-science/blob/4bc234956b3e1b3c3d4b6767202f198421a58203/arches_for_science/trigger_functions.py#L4): I'd suggest trying that here. Having the trigger wrapper here still provides some "discoverability" IMO without having to have the monster string right here.

arches/app/models/migrations/11801_alter_spatialview_geometrynode.py
```diff
@@ -0,0 +1,174 @@
+# Generated by Django 5.2.1 on 2025-05-19 16:11
+
+import django.db.models.deletion
+from django.db import migrations, models
+import pgtrigger
+
+
+drop_legacy_trigger = "drop trigger if exists __arches_trg_update_spatial_views on spatial_views; drop function if exists __arches_trg_fnc_update_spatial_views;"
```
jacobtylerwalls, line 8: Alexei has expressed before preferring to see these indented under the Migration class so that the dependencies are always near the top of the file.

arches/app/models/migrations/11801_alter_spatialview_geometrynode.py
```diff
@@ -0,0 +1,174 @@
+# Generated by Django 5.2.1 on 2025-05-19 16:11
```
jacobtylerwalls, line 1: You might rename the file to use the issue number, if that's still something the dev team cares about? (I do find the zigzag numbering confusing myself...)

arches/app/models/graph.py
```diff
@@ -2523,6 +2524,15 @@ def restore_state_from_serialized_graph(self, serialized_graph):
             if draft_graph:
                 draft_graph.delete()
 
+            views = models.SpatialView.objects.filter(geometrynode__graph=self.pk)
```
jacobtylerwalls, line 2527: In general, I'm pushing for us to keep things lazy by using `self` to do queries, for the reason that the nodes might already be fetched, eg.

```py
for node in self.node_set.all():  # could have been populated already -- caller in charge
    for spatial_view in node.spatialview_set.all():
```

Double loop not so great, but that feels like a sign this could be implemented in `Node.xxx()` instead of `Graph.save()`. What do you think?

Then the geometry node is in charge of shuffling the spatial view around, or returning it to the caller. `restore_state_from_serialized_graph()` could call that `Node` method and know what to do without storing state on `Graph`.

arches/app/models/graph.py
```diff
@@ -50,6 +50,7 @@ class Graph(models.GraphModel):
     """
 
     objects = GraphQuerySet.as_manager()
+    saved_spatial_view_geometry_nodes = None
```
jacobtylerwalls, line 53: I don't think this is going to work as a place to stash state, as a typical deployment uses multiple workers.

arches/app/models/graph.py
```diff
@@ -2523,6 +2524,15 @@ def restore_state_from_serialized_graph(self, serialized_graph):
             if draft_graph:
                 draft_graph.delete()
 
+            views = models.SpatialView.objects.filter(geometrynode__graph=self.pk)
+            orphaned_spatial_view_geometry_nodes = {}
+            for view in views:
```
jacobtylerwalls, line 2529: I think it's worth adding a test, maybe in the spatialview_model_tests.py, eg.:

```py

    def test_restore_state_from_serialized_graph(self):
        spatialview = self.generate_valid_spatialview()
        spatialview.full_clean()
        spatialview.save()

        from arches.app.models.graph import Graph  # move this to top
        graph = Graph(pk=spatialview.geometrynode.graph.pk)  # alas...
        graph.publish()
        published_graph = graph.get_published_graph()
        graph.restore_state_from_serialized_graph(published_graph.serialized_graph)
        # assert the spatialview still exists
```

jacobtylerwalls (changes requested):
Thanks! My main concern is just finding a solution without state on the Graph class

arches/app/models/models.py
aarongundel: much nicer, thanks for the suggestion.

chrabyrd:
Per a chat with @aarongundel , we're instead going to add spatial views to the serialized graph && rehydrate like all other assets. That should be enough to not need to munge the graph update method or change db-level protections on the SpatialView model

aarongundel:
closed in favor of https://github.com/archesproject/arches/pull/12102

Issue #11965, opened 2025-04-14
Graph slug model field validation is ignored


jacobtylerwalls:
> This validation is totally ignored because we're not in the habit of calling `full_clean()` anywhere when writing our custom views. I can create a very perky slug ending in `!` ;)

I'm happy to bounce this to a new issue, given the feature freeze is imminent and it's a pre-existing bug.

_Originally posted by @jacobtylerwalls in https://github.com/archesproject/arches/pull/11865#discussion_r2040746550_

Issue #11696, opened 2024-12-17
Proposal: Change migration strategy for database object SQL


bferguso:
### Background
With the current patterm for deploying database objects  (table, view, function, etc.), where the SQL is directly embedded in the migration `.py` files, often with several objects embedded in a single file, it can be quite difficult to trace the changes made to those objects over time.

It would be nice to externalize SQL for those database objects (not 1-off SQL statements, but the SQL that represents database objects) so that it can:
1. be read clearly in its own file, not embedded as a Python string
2. can be diff'd to previous versions of the same object and
3. creates an easily visible history of the database object as it changes through time

### Proposed Solution
1. Create a sub-directory of the migrations directory, say `migrations/sql`
2. Any SQL that is used to create a database object not be directly included in the migration file, but as a separate file in the `sql/` directory, one object per file
3. SQL filename could be something ala Liquibase or Flyway pattern , something liike `vYYYY.MM.DD.SEQ__object_name.sql`, eg: `v2024.12.17.001__my_function_call.sql`
4. For other projects, we've broken down the SQL by year so the list of files doesn't grow to be too long over time (eg `sql/2024`, `sql/2025`..)
5. Write a Python migration utility to allow those files (1 or more) to be easily digested by the migrations. First cut of ours is something like this:
```
def format_files_into_sql(files: [str], sql_dir: str):
    sql_string = ""
    for filename in files:
        file_path = os.path.join(sql_dir, filename)
        with open(file_path) as file:
            sql_string = sql_string + "\n" + file.read()
    return sql_string
```
6. When referencing those file(s) from the migration it would look something like this:
```
    sql_dir = os.path.join(os.path.dirname(__file__), "sql")
    files = [
        "2024-07-25_fossil_sample.fossil_name_vw.sql",
        "2024-07-25_fossil_sample.fossil_name_mv.sql",
...
        "2024-08-15_databc_grants.sql"
    ]
    sql_statement = format_files_into_sql(files, sql_dir)
    
    operations = [
        migrations.RunSQL(create_materialized_views, migrations.RunSQL.noop),
    ]
```

A PR for this could target v8.

bferguso:
fyi @chiatt, @chrabyrd - Here's a starting ticket from our conversation today.

jacobtylerwalls:
I really like this @bferguso.

[...]

PR #12102, opened 2025-05-21
fix for #11848 modifying spatial views

### LINKED ISSUE #11848: Graph publication workflow in v8 deletes spatial views

aarongundel:
### Types of changes

### Description of Change

* Serializes the spatial views and rehydrates them with the rest of the graph.  
* note the substantial change to the trigger here was making sure it doesn't compare to published graphs - now it compares to languages, which seems more correct and negates the problems associated with rehydrating the views after publish.

### Issues Solved

Closes #11848 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


aarongundel:
This feels much better.

arches/install/arches-templates/project_name/settings.py-tpl
```diff
@@ -144,6 +144,7 @@ INSTALLED_APPS = (
     "corsheaders",
     "oauth2_provider",
     "django_celery_results",
+    "pgtrigger"
```
jacobtylerwalls: Tests are failing b/c we need this in core arches settings.py also

jacobtylerwalls: Also needs trailing comma

arches/app/models/migrations/11848_alter_spatialview_geometrynode.py
```diff
@@ -0,0 +1,156 @@
+import django.db.models.deletion
+from django.db import migrations, models
+import pgtrigger
+
+
+class Migration(migrations.Migration):
+    drop_legacy_trigger = "drop trigger if exists __arches_trg_update_spatial_views on spatial_views; drop function if exists __arches_trg_fnc_update_spatial_views;"
+
+    create_legacy_trigger = """
```
bferguso, line 9: Just wondering if we want to start externalizing long-lasting SQL or if the plan is to keep it embedded in the .py migration file. My preference would be to start pulling it out for readability.

arches/app/models/trigger_functions.py
```diff
@@ -0,0 +1,106 @@
+ARCHES_UPDATE_SPATIAL_VIEWS_TRIGGER_FUNCTION = """
```
jacobtylerwalls: @bferguso this is where the long-lasting SQL is right now (the location in the migration .py file is only for the backwards migration, and will start fading away)

I like your idea -- we could this as a chance to change this .py file to a lintable .sql file (in some logical directory) and add the helper function you described in #11696 to read it into a python string. Would that be in line with your thinking?

bferguso: @jacobtylerwalls - That's mostly what I was thinking. Just wondering if there is a reason to not also externalize the reverse migration SQL as well. This would allow the changes to be easily seen starting at the current migration, not the next one.

Do we want to agree on an initial directory structure as a starting point? (Maybe something like `migrations/sql/v8.0`)?

jacobtylerwalls: > This would allow the changes to be easily seen starting at the current migration, not the next one.

I guess this is not quite how I see it -- I'm thinking there would only be one .sql file in the repo, no forward version, no reverse version, just a version. You'd use git checkouts if you needed to time travel, and this way you'd get nice git diffs in a PR, since you're diffing a single file.

bferguso: Ideally that would be nice, but unless I'm missing something in what you're suggesting, that can end up breaking the migrations over time as it can disrupt the order of operations if newer version of a particular SQL object depends on a previous migration. Eg:
```
0001__create_function_x.py
0002__add_column_y.py
0003__update_function_x_requiring_column_y.py
```

If migrations 1 & 3 point to the same SQL file, migration 0001 fails on a clean build because migation 0002 hasn't run yet.

arches/app/models/graph.py
```diff
@@ -118,6 +118,10 @@ def __init__(self, *args, **kwargs):
                 for card in args[0]["cards"]:
                     self.add_card(card)
 
+                for spatial_view in args[0]["spatial_views"]:
+                    spatial_view = models.SpatialView(**spatial_view)
+                    spatial_view.save()
```
jacobtylerwalls: `SpatialView` is one of the only arches models with a `clean()` method. This was to move validation logic out of the views and into the models, as well as to put it in a standard place that can be called by the django admin (this is also one of the few models that's enabled in the admin).

For that reason, I think `SpatialView` should follow the Django-recommended pattern[1], which is that when saving an object, unless you're going through a form or a (de)serializer that does something equivalent for you, you call `full_clean()` to get informative, localized failures before potentially failing at the db level.

(It will surely be a long time before we update other models to follow this pattern. We have follow-ups like #11965 for some of these validation gaps. But for a model that already has an intelligent `clean()` method I think it's time to use it 👍 )
```suggestion
                    spatial_view.full_clean()
                    spatial_view.save()
```

[1]: https://docs.djangoproject.com/en/5.2/ref/models/instances/#django.db.models.Model.full_clean

arches/app/models/graph.py
```diff
@@ -1960,6 +1964,10 @@ def serialize(
             else:
                 ret.pop("group_permissions", None)
 
+            ret["spatial_views"] = models.SpatialView.objects.select_related().filter(
+                Q(geometrynode__graph_id=self.source_identifier_id)
+                | Q(geometrynode__graph_id=self.graphid)
+            )
```
jacobtylerwalls:
```suggestion
            ret["spatial_views"] = models.SpatialView.objects.select_related().filter(
                geometrynode__graph__in=[self.source_identifier_id, self.graphid]
            )
```

releases/8.0.0.md
```diff
@@ -262,11 +263,12 @@ JavaScript:
             }
         }
         ```
-    - Change the "captcha" entry in `INSTALLED_APPS` to reflect its new name:
+    - Change the "captcha" entry in `INSTALLED_APPS` to reflect its new name and add pgtrigger:
         ```
         INSTALLED_APPS = [
             ...
             "django_recaptcha",  # was "captcha"
+            "pgtrigger"
```
jacobtylerwalls: trailing comma might help copy/pasters

arches/app/models/trigger_functions.py
jacobtylerwalls: Right, that's why I'm encouraging us to adopt `django-pgtrigger` (in this PR) and `django-pgviews` (eventually), and I suppose `django-sqlfun` for functions if need be, although it's not as mature, so we might end up rolling our own.

Those packages allow you to have natural git diffs on a single source because they still snapshot forward and backward SQL and pass that to the migration runner, but because it's been auto-generated from the source, you discourage the implementer from ever needing to consult it. This is basically how Django models work, you have a single models.py and all the forward/backward operations are encapsulated in the migration file, but you don't look there, you look at the models.

So in your example, when 0001 and 0003 run, what gets run is the chewed-up SQL string in the 0001 and 0003 migration files, but that's an artifact--instead the developer works with the single, unversioned SQL file, and delegates all the forward/backward stuff to the package.

arches/app/models/migrations/11848_alter_spatialview_geometrynode.py
```diff
@@ -0,0 +1,156 @@
...
+    ]
+
+    operations = [
+        migrations.RunSQL(drop_legacy_trigger, create_legacy_trigger),
+        pgtrigger.migrations.AddTrigger(
+            model_name="spatialview",
+            trigger=pgtrigger.compiler.Trigger(
+                name="__arches_trg_update_spatial_views",
+                sql=pgtrigger.compiler.UpsertTriggerSql(
+                    declare="DECLARE sv_perform text; valid_geom_nodeid boolean; has_att_nodes integer; valid_att_nodeids boolean; valid_language_count integer;",
```
jacobtylerwalls, line 146: For instance: This is the chewed up SQL that gets run, ensuring that 0001 and 0003 run distinct SQL, but the idea is that the developer would work with the single source of truth .sql file

arches/app/models/models.py
```diff
@@ -2383,6 +2387,21 @@ def __str__(self):
     class Meta:
         managed = True
         db_table = "spatial_views"
+        triggers = [
+            pgtrigger.Composer(
```
jacobtylerwalls, line 2388: For my own knowledge, can you walk me through why we needed `Composer`? Is it because it's a statement-level trigger?

arches/app/models/models.py
```diff
@@ -2383,6 +2387,21 @@ def __str__(self):
...
+                name="__arches_trg_update_spatial_views",
```
jacobtylerwalls: Looks like the final name in the db becomes `pgtrigger___arches_trg_update_spatial_views_7bca0`. I like `arches` so that apps are namespaced, but i wonder if this should just be `arches_update_spatial_views` ?

arches/app/models/trigger_functions.py
```diff
@@ -0,0 +1,106 @@
...
+        ----------------------------------------------------------------------------------------------
+        -- check language code is valid
+        select count(code)
+        into valid_language_count
+        from languages
+        where code = new.languageid;
+
+        if valid_language_count = 0 then
+            raise exception 'language is not valid for this graph';
+        end if;
```
jacobtylerwalls: Can you show me where this is checking language validity for a graph? From a glance it looks like it's just verifying the language exists, which we already have an FK for.

arches/app/models/migrations/11848_alter_spatialview_geometrynode.py
bferguso: Ok - so if I'm understanding correctly, the .py file would be a chewed up copy of what's in the .sql file?

jacobtylerwalls: Exactly

jacobtylerwalls: And sometimes there wouldn't be a SQL file if the developer never wrote SQL by hand in the first place

arches/app/models/models.py
aarongundel: I could not find a way with pgtrigger to remove the pgtrigger_ namespace

jacobtylerwalls: Oh that's fine, I just meant removing "trg" and the leading underscore but this is g2g

arches/app/models/models.py
aarongundel: Yes.  And readability wise, it makes it very easy to see what's going into this trigger's boilerplate.

arches/app/models/trigger_functions.py
aarongundel: I also want to note here that what's currently in this string is not executable sql by itself - it's not a full function because it's relying pgtrigger to generate the function boilerplate.  I think there's something to be gained from syntax highlighting in a sql file but this, by itself, is not imminently debuggable sql that you can run on your database and see a result from.

arches/app/models/migrations/11848_alter_spatialview_geometrynode.py
aarongundel: In this case, this sql is not particularly long lasting - this is just a rollback and a drop.  Past this point, the long lasting sql gets managed in the sql file.

arches/app/models/migrations/11848_alter_spatialview_geometrynode.py
aarongundel: re: conversation with @jacobtylerwalls to produce the full migration sql, you can run the `sqlmigrate` subcommand

`python manage.py sqlmigrate models [migration_name]`

arches/app/models/models.py
aarongundel: did that - more that we can save on chars, the better

arches/app/models/trigger_functions.py
aarongundel: fixed in clean method, yanked from trigger.

aarongundel:
investigating test failures

arches/app/models/graph.py
```diff
@@ -36,7 +36,7 @@
 from arches.app.utils.i18n import LanguageSynchronizer
 from django.utils.translation import gettext as _
 from pyld.jsonld import compact, JsonLdError
-from django.db.models.base import Deferred
+from django.db.models import Q
```
bferguso: This no longer needed?

bferguso (approved):
For the externalization of SQL part of this - I really like it! Thanks for getting that in there.
I haven't followed the discussion on the actual issue being being solved here so I'll leave it to others weigh in on the rest.

jacobtylerwalls (commented):
On the other PR I sketched out a little test -- I think it still makes sense to add one to cover the original issue.

arches/app/models/graph.py
```diff
@@ -118,6 +117,12 @@ def __init__(self, *args, **kwargs):
                 for card in args[0]["cards"]:
                     self.add_card(card)
 
+                if "spatial_views" in args[0]:
```
chrabyrd, line 122: this probably needs to be added to the block on line 87

chrabyrd (commented):
looks good, where are the spatial views being added to the serialized graph?

arches/app/models/graph.py
```diff
@@ -1960,6 +1965,9 @@ def serialize(
             else:
                 ret.pop("group_permissions", None)
 
+            ret["spatial_views"] = models.SpatialView.objects.select_related().filter(
```
aarongundel, line 1989: @chrabyrd this is where they're added to the serialized graph

arches/app/models/graph.py
aarongundel: @chrabyrd I didn't really understand that; can you explain why you're hanging stuff off the graph object there?

aarongundel:
@jacobtylerwalls I added a test there; then I had to modify a couple of others because of transaction block failures.

tests/models/spatialview_model_tests.py
```diff
@@ -192,20 +193,35 @@ def test_create_spatialview_mixed_geometry(self):
...
+        spatialview = self.generate_valid_spatialview()
+        spatialview.full_clean()
+        spatialview.save()
+
+        graph = Graph.objects.get(pk=spatialview.geometrynode.graph.pk)
+        draft_graph = graph.create_draft_graph()
+
+        # updating graph from draft graph removes all elements
+        # including the serialized graph - then recreates them
+        graph.update_from_draft_graph(draft_graph=draft_graph)
```
jacobtylerwalls, line 207: If we don't publish() here, then clean() crashes with `AttributeError: 'NoneType' object has no attribute 'publishedgraph_set'`. 

I think we should publish() to model correct usage.

arches/app/models/models.py
```diff
@@ -2383,6 +2384,37 @@ def __str__(self):
...
+                    format_file_into_sql(
+                        "arches_update_spatial_views.sql",
+                        "sql/triggers",
+                    )
+                ),
+            )
+        ]
+
+    def clean_fields(self, exclude=None):
+        if exclude is not None:
```
jacobtylerwalls, line 2410: Need to call `super().clean_fields(exclude=exclude)` so we don't evade all the other checks. This has the advantage of ensuring that self.language and self.geometrynode don't crash below if absent (well, unless they were excluded, so we should probably still guard for that and raise ValidationError instead of letting another exception escape)

jacobtylerwalls (changes requested):

tests/models/spatialview_model_tests.py
jacobtylerwalls: More context: what I was doing was fiddling with simplifying the next assertion to just `spatialview.refresh_from_db()`, which will raise if it's been deleted. I then added `full_clean()` right after and saw that it fails. This might be a nice update to the test.

jacobtylerwalls (approved):
Thanks!

arches/app/models/graph.py
aarongundel: discussed this; done now.

chrabyrd (approved):
lgtm! 🚀

arches/app/models/models.py
wesleykendall: @jacobtylerwalls @aarongundel, pgtrigger author here. My docs on composer were inaccurate. I just updated them [here](https://django-pgtrigger.readthedocs.io/en/4.15.3/statement/).

You'll want to do `level=pgtrigger.Statement` if you want this to execute once per statement instead of once per row. Just wanted to make sure you were aware of that.

`Composer` is a relatively new addition to the library to try to bridge the gap between row and statement level triggers, making it easy for libraries like django-pghistory to support conditional row and statement level triggers. One can write statement-level triggers with a basic `pgtrigger.Trigger` class. `Composer` just handles a lot of extra boilerplate and tries to make up for the fact that postgres doesn't support conditional statement level triggers.

Apologies for the confusion in the docs

wesleykendall:
@aarongundel @jacobtylerwalls I added a comment to a resolved thread [here](https://github.com/archesproject/arches/pull/12102/files#r2142579533) in case you didn't see it. Just wanted to be sure you all were aware that the trigger definition in this PR is a row-level trigger and not a statement-level one.

It looks like your former trigger was a row-level trigger, so probably not a mistake, but the convo made it sound like you were under the expectation that the trigger would be statement level.

I believe your old trigger was deferred. If you need to continue to support that, [pgtrigger can express deferred triggers for row level triggers too](https://django-pgtrigger.readthedocs.io/en/4.15.3/deferrable/). Thanks for using!

jacobtylerwalls:
Thanks for the package and for the helpful note @wesleykendall!

We were talking past each other a little bit--row-level is what we wanted (this isn't a hot path)--but thanks for those docs improvements. We adjusted to just use `Trigger` and took the opportunity to add deferral back in (thanks for the spot!)

PR #12106, opened 2025-05-22
drives term filter types from settings

### LINKED ISSUE #12105: Allow for settings to drive list of term search types
Currently, term search works with two fixed types of term results: "terms" and "concepts".  However custom ES indexes, mappings, and filters allow for the implementation of additional term types.  In order to facilitate taking advantage of these in the term filter dropdown, the set of term search types should be moved to settings.py so it can be extended by applications/projects.

robgaston:
### Types of changes

### Description of Change

adds a setting (`TERM_SEARCH_TYPES`) to allow for customizing the set of term search types to take advantage of custom indexes, filters, and mappings.

### Issues Solved

Closes #12105 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Found by: @robgaston 
*   Tested by: @robgaston 
*   Designed by: @robgaston 

### Further comments

I'm not wild about this as it perpetuates the old pattern of injecting these data into javascript.htm, but doing something async to get these data would require a much bigger rework of these existing KO modules.  I'm open to other options though of course.

### DIFF
--- a/arches/app/media/js/bindings/term-search.js
+++ b/arches/app/media/js/bindings/term-search.js
@@ -24,7 +24,7 @@ ko.bindingHandlers.termSearch = {
                 option.data = item;
                 searchbox.append(option);
                 searchbox.tags.push(option);
-            }); 
+            });
             searchbox.trigger('change');
             if (ko.unwrap(terms).length == 0 && searchbox.tags.length == 0){
                 searchbox.empty();
@@ -83,12 +83,11 @@ ko.bindingHandlers.termSearch = {
                         text: arches.translations.containsTerm(params.term),
                         value: params.term
                     });
-                    if(data.terms.length > 0){
-                        res.push({"text": arches.translations.termSearchTerm, "children": data.terms});
-                    }
-                    if(data.concepts.length > 0){
-                        res.push({"text": arches.translations.termSearchConcept, "children": data.concepts});
-                    }
+                    arches.termSearchTypes.forEach(function(searchType) {
+                        if(data[searchType.key].length > 0){
+                            res.push({"text": searchType.label, "children": data[searchType.key]});
+                        }
+                    });
                     return {
                         results: res
                     };
@@ -99,7 +98,7 @@ ko.bindingHandlers.termSearch = {
                     return result.text;
                 }
                 var markup = [];
-                var indent = result.type === 'concept' || result.type === 'term' ? 'term-search-item indent' : (result.type === 'string' ? 'term-search-item' : 'term-search-group');
+                var indent = arches.termSearchTypes.map((a) => a.type).indexOf(result.type) > -1 ? 'term-search-item indent' : (result.type === 'string' ? 'term-search-item' : 'term-search-group');
                 if (result.type === 'group') {
                     _.each(result.text, function(searchType, i) {
                         var label = searchType === 'concepts' ? arches.translations.termSearchConcept : arches.translations.termSearchTerm;
--- a/arches/app/media/js/bindings/term-search.js
+++ b/arches/app/media/js/bindings/term-search.js
@@ -98,7 +98,7 @@ ko.bindingHandlers.termSearch = {
                     return result.text;
                 }
                 var markup = [];
-                var indent = arches.termSearchTypes.map((a) => a.type).indexOf(result.type) > -1 ? 'term-search-item indent' : (result.type === 'string' ? 'term-search-item' : 'term-search-group');
+                var indent = arches.termSearchTypes.map((searchType) => searchType.type).indexOf(result.type) > -1 ? 'term-search-item indent' : (result.type === 'string' ? 'term-search-item' : 'term-search-group');
                 if (result.type === 'group') {
                     _.each(result.text, function(searchType, i) {
                         var label = searchType === 'concepts' ? arches.translations.termSearchConcept : arches.translations.termSearchTerm;
--- a/arches/app/media/js/views/components/search/term-filter.js
+++ b/arches/app/media/js/views/components/search/term-filter.js
@@ -57,7 +57,7 @@ const viewModel = BaseFilter.extend({
 
     updateQuery: function() {
         var terms = _.filter(this.filter.terms(), function(term){
-            return arches.termSearchTypes.map((a) => a.type).concat(['string']).indexOf(term.type) > -1;
+            return arches.termSearchTypes.map((searchType) => searchType.type).concat(['string']).indexOf(term.type) > -1;
         }, this);
 
         var queryObj = this.query();
--- a/arches/app/media/js/views/components/search/term-filter.js
+++ b/arches/app/media/js/views/components/search/term-filter.js
@@ -57,9 +57,9 @@ const viewModel = BaseFilter.extend({
 
     updateQuery: function() {
         var terms = _.filter(this.filter.terms(), function(term){
-            return term.type === 'string' || term.type === 'concept' || term.type === 'term';
+            return arches.termSearchTypes.map((a) => a.type).concat(['string']).indexOf(term.type) > -1;
         }, this);
-        
+
         var queryObj = this.query();
         if (terms.length > 0){
             queryObj[componentName] = ko.toJSON(terms);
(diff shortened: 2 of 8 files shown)


whatisgalen:
@robgaston [this](https://github.com/archesproject/arches/blob/dev/8.0.x/arches/app/views/search.py#L147-L155) may also be a good place to call out to these new settings

arches/app/media/js/bindings/term-search.js
```diff
@@ -99,7 +98,7 @@ ko.bindingHandlers.termSearch = {
                     return result.text;
                 }
                 var markup = [];
-                var indent = result.type === 'concept' || result.type === 'term' ? 'term-search-item indent' : (result.type === 'string' ? 'term-search-item' : 'term-search-group');
+                var indent = arches.termSearchTypes.map((a) => a.type).indexOf(result.type) > -1 ? 'term-search-item indent' : (result.type === 'string' ? 'term-search-item' : 'term-search-group');
```
chrabyrd: nit: single-letter variable naming

arches/app/media/js/views/components/search/term-filter.js
```diff
@@ -57,9 +57,9 @@ const viewModel = BaseFilter.extend({
 
     updateQuery: function() {
         var terms = _.filter(this.filter.terms(), function(term){
-            return term.type === 'string' || term.type === 'concept' || term.type === 'term';
+            return arches.termSearchTypes.map((a) => a.type).concat(['string']).indexOf(term.type) > -1;
```
chrabyrd: same

chrabyrd (changes requested):
booo further reliance on the `arches` object. I get it though, and I'm ok with it as long as it's not being used in Vue 👍 

small nit about single-letter var names, otherwise LGTM!

arches/app/views/search.py
apeters, line 1: I feel like if we're going to make term search derive from settings that we use those settings on line 164 (in search.py) , and if we do that just make the logic pluggable and associated with the term search type.  Then users could plug in any logic they want which might not even us ES.  Probably wouldn't be that much additional work.

robgaston: Yeah, this makes sense - I was overriding the term search route in controlled lists, but making this pluggable is nice.  I updated this accordingly, though perhaps the factoring is a bit off.  I built this into BaseIndex as that made sense given my use cases (and was convenient) but perhaps it's too prescriptive? I didn't really want to add another interface class here just for this one method, but maybe that's better.  What do you think @apeters?

apeters:
some potential refinements here: https://github.com/archesproject/arches/pull/12123

apeters (approved):
lgtm

arches/settings.py
```diff
@@ -105,6 +105,21 @@
 #     'should_update_asynchronously': False
 # }]
 
+TERM_SEARCH_TYPES = [
```
chrabyrd, line 108: arguably since this is meant to be modular and expandable ( I believe? ) , these modules should live in separate files. eg `arches.app.search.concept_search.ConceptSearch`. That provides a good structure/pattern for someone to add more modules without editing `search_term.py`

arches/app/search/search_term.py
```diff
@@ -1,7 +1,153 @@
...
+    return ret
+
+
+def get_resource_model_label(result):
+    if len(result["nodegroupid"]["buckets"]) > 0:
+        for nodegroup in result["nodegroupid"]["buckets"]:
+            nodegroup_id = nodegroup["key"]
+            node = Node.objects.get(nodeid=nodegroup_id)
+            graph = node.graph
+        return "{0} - {1}".format(graph.name, node.name)
```
chrabyrd: style nit for readability: `return f"{node.graph.name} - {node.name}"`

arches/app/search/search_term.py
```diff
@@ -1,7 +1,153 @@
...
+                        "nodegroupid": result["nodegroupid"]["buckets"][0]["key"],
+                    }
+                )
+                i = i + 1
+    return ret
+
+
+def get_resource_model_label(result):
+    if len(result["nodegroupid"]["buckets"]) > 0:
+        for nodegroup in result["nodegroupid"]["buckets"]:
```
chrabyrd: why loop here? It looks like it's unnecessary && the `.get()` call has the potential to hammer the db. Maybe some early nullguarding of the buckets combined with

```
node = Node.objects.select_related("graph").get(nodeid=buckets[0]["key"])
```

instead?

arches/app/search/search_term.py
```diff
@@ -1,7 +1,153 @@
...
+    results = query.search(index=index)
+    if results is not None:
+        for result in results["aggregations"]["value_agg"]["buckets"]:
+            if len(result["top_concept"]["buckets"]) > 0:
+                for top_concept in result["top_concept"]["buckets"]:
+                    top_concept_id = top_concept["key"]
+                    top_concept_label = get_preflabel_from_conceptid(
+                        top_concept["key"], lang=lang if lang != "*" else None
+                    )["value"]
+                    for concept in top_concept["conceptid"]["buckets"]:
```
chrabyrd: woof, triple-nested for-loop.

whatisgalen:
love these changes!

arches/app/search/search_term.py
```diff
@@ -1,7 +1,153 @@
...
+            query=searchString.lower(),
+            fuzziness="AUTO",
+            prefix_length=settings.SEARCH_TERM_SENSITIVITY,
+        )
+    )
+    boolquery.should(
+        Match(field="displayname.value", query=searchString, fuzziness=2, boost=2)
+    )
+
+    if index == "terms":
```
chrabyrd: probably just a lack of knowledge on my part: this will eval falsey for `ConceptSearch` yeah?

arches/app/search/search_term.py
```diff
@@ -1,7 +1,153 @@
...
+    base_agg.add_aggregation(max_score_agg)
+    base_agg.add_aggregation(top_concept_agg)
+    base_agg.add_aggregation(nodegroupid_agg)
+    query.add_aggregation(base_agg)
+
+    ret = []
+    results = query.search(index=index)
+    if results is not None:
+        for result in results["aggregations"]["value_agg"]["buckets"]:
+            if len(result["top_concept"]["buckets"]) > 0:
```
chrabyrd: is this return index-dependent? eg if searching the terms index this will eval as falsey right?

arches/app/search/search_term.py
```diff
@@ -1,7 +1,153 @@
...
+    def search_terms(search_string, lang, user, term_search_type):
+        return searchES(
+            index=term_search_type["key"],
+            searchString=search_string,
+            lang=lang,
+            user=user,
+        )
+
+
+def searchES(index, searchString, lang, user):
```
chrabyrd: This function is a bit of a hard read. It could def stand to be refactored / decomposed. 

I've made some assumptions in the comments below -- and if they hold true there's maybe a larger conversation to have, which is: why have this hard-coding split in the `searchES` function? A lot of the control flow here ( eg `if index == "terms"`, `if len(result["top_concept"]["buckets"]) > 0`, is smelling of breaking this thing out into two separate functions, each powering their respective class. 

And if those assumptions hold correct, splitting this into multiple functions also reinforces the expandability pattern of having this declared in settings.py

arches/app/search/search_term.py
```diff
@@ -1,7 +1,153 @@
...
+    i = 0
```
chrabyrd: boo single-letter variable.

chrabyrd (commented):
Looks good overall, just some conceptual questions on my end

arches/app/search/search_term.py
robgaston: the short answer to this (& most of your questions here, I think) is: this code was copy/pasted from search.py in order to centralize this logic.  I don't think there was any effort to refactor this code, but while were here I think we should consider what you're suggesting here (ie split out the logic for each term type)

arches/settings.py
robgaston: Yeah, this is a good point.  I don't think there was really a focus on the design here within Arches but rather on making this modular so that applications/projects could extend/configure term searches (see archesproject/arches-controlled-lists#110).  @apeters thoughts?

robgaston: @chrabyrd I refactored these modules per your suggestions!

chrabyrd (approved):
this lgtm! Thanks for getting those changes in 🚀

PR #12110, opened 2025-05-22
Avoid hand-escaping percent signs in I18n_JSON #12060

### LINKED ISSUE #12060: % signs in widget configs are ending up double-encoded
% signs are ending up multiply-encoded in graph json export/import

See examples in test fixtures, some of these predate 8.0.x, so doesn't sound like a recent regression:

https://github.com/archesproject/arches/blob/d854446d2f486d2184bc54cd1d5b68369b4f0f84/tests/fixtures/resource_graphs/branch_excel_test.json#L80

https://github.com/archesproject/arches/blob/d854446d2f486d2184bc54cd1d5b68369b4f0f84/tests/fixtures/resource_graphs/SpatialViews_Test_Model.json#L418

https://github.com/archesproject/arches/blob/d854446d2f486d2184bc54cd1d5b68369b4f0f84/tests/fixtures/resource_graphs/Search%20Sort%20Test%20Model.json#L55

This was first noticed here, so also check `placeholder`?
https://github.com/archesproject/arches/blob/d854446d2f486d2184bc54cd1d5b68369b4f0f84/arches/app/utils/data_management/resource_graphs/importer.py#L178-L184

aarongundel:
### Types of changes

### Description of Change

Change the i18n json code to unpack the %% as %.  This stops the percent sign explosion.  Though it seems like this should also affect 7.6, based on my testing it doesn't.  This probably has something to do with how publishing and such is done in 8 vs 7.  

### Issues Solved

Closes #12060

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/models/migrations/12002_percent_sign_fix.py
+++ b/arches/app/models/migrations/12002_percent_sign_fix.py
@@ -1,0 +1,28 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("models", "11848_alter_spatialview_geometrynode"),
+    ]
+
+    forward = """
+        UPDATE cards_x_nodes_x_widgets
+        SET config = jsonb_set(
+            config,
+            '{width}',
+            to_jsonb(
+                regexp_replace(config->>'width', '^(\d+)%{2,}$', '\1%', 'g')
+            )
+        )
+        WHERE config ? 'width'
+        AND config->>'width' ~ '^\d+%{2,}$';
+    """
+
+    operations = [
+        migrations.RunSQL(
+            forward,
+            migrations.RunSQL.noop,  # No reverse operation needed
+        ),
+    ]
--- a/arches/app/models/migrations/12002_percent_sign_fix.py
+++ b/arches/app/models/migrations/12002_percent_sign_fix.py
@@ -4,10 +4,10 @@ from django.db import migrations
 class Migration(migrations.Migration):
 
     dependencies = [
-        ("models", "11848_alter_spatialview_geometrynode"),
+        ("models", "11959_unique_node_source_widget"),
     ]
 
-    forward = """
+    forward = r"""
         UPDATE cards_x_nodes_x_widgets
         SET config = jsonb_set(
             config,
--- a/arches/app/models/fields/i18n.py
+++ b/arches/app/models/fields/i18n.py
@@ -245,9 +245,9 @@ class I18n_JSON(NothingNode):
 
         if isinstance(value, str):
             try:
-                ret = json.loads(value.replace("%%", "%"))
+                ret = json.loads(value)
             except:
-                ret = json.loads(json.dumps(value.replace("%%", "%")))
+                ret = json.loads(json.dumps(value))
         elif value is None:
             ret[lang] = None if use_nulls else ""
         elif isinstance(value, I18n_JSON):
@@ -321,7 +321,7 @@ class I18n_JSON(NothingNode):
                 ELSE %s
                 END
             """
-            params.append(json.dumps(self.to_localized_object()).replace("%", "%%"))
+            params.append(json.dumps(self.to_localized_object()))
 
         return sql, tuple(params)
 
--- a/arches/app/models/fields/i18n.py
+++ b/arches/app/models/fields/i18n.py
@@ -245,9 +245,9 @@ class I18n_JSON(NothingNode):
 
         if isinstance(value, str):
             try:
-                ret = json.loads(value)
+                ret = json.loads(value.replace("%%", "%"))
             except:
-                ret = json.loads(json.dumps(value))
+                ret = json.loads(json.dumps(value.replace("%%", "%")))
         elif value is None:
             ret[lang] = None if use_nulls else ""
         elif isinstance(value, I18n_JSON):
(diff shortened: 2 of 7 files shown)


aarongundel:
test failures appear unrelated to this PR

jacobtylerwalls:
> Though it seems like this should also affect 7.6, based on my testing it doesn't.

My hypothesis is that it's the same Django 5.2 bugfix mentioned in 28bb2e5

jacobtylerwalls:
I'm finding myself testing this manually. Would you mind adding a test? I can see this having the potential to regress in the future.

Maybe in tests/localization/field_tests.py?

jacobtylerwalls (commented):
It seems like this fix is located at loading from the db instead of on saving to the db. It would be nice to prevent the dupes from making it into the db at all.

I'm wondering if the fix is here:
https://github.com/archesproject/arches/blob/ad763385764efb6d19a1c6379ff0683e89c9333d/arches/app/models/fields/i18n.py#L324

since `params` is escaped by the db-adapter, we shouldn't need to hand-escape ourselves. I didn't get as far as to test it because I couldn't repro the issue with your note about editing the node description.

Are you able to confirm if it fixes the issue by removing this replace call?

aarongundel:
@jacobtylerwalls that is definitely where I found the issue to happen as well.  The change I made undoes that.  One thing to note - the "save save save" bit I mentioned in the description - you actually have to do that _then_ publish.  The saving has a compounding effect, then publishing it causes it to be committed to the serialized graph.  I will check on removing the escape there, though removing it in other places has a detrimental effect on saving to the db as "%"  in postgres denotes a parameter.  That's apparently why this was done in the first place.

jacobtylerwalls (changes requested):
Ah, I was editing the node description and checking the node description instead of checking the widget config.

> Are you able to confirm if it fixes the issue by removing this replace call?

Now that I retested, it looks like my suggestion fixes the bad data from ever getting into the db in the first place.

Agree this was harder to reproduce on 7.6, but from #12060 we have examples of this predating v8, so I think we should:
- clean up those fixtures now
- write a migration to fix the data in the database instead of trying to clean it when loading it from the db.
- add a regression test

jacobtylerwalls (approved):

jacobtylerwalls:
> write a migration to fix the data in the database instead of trying to clean it when loading it from the db.

happy to punt on that until we get reports of this in the wild. approved!

arches/app/models/migrations/12002_percent_sign_fix.py
```diff
@@ -0,0 +1,28 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("models", "11848_alter_spatialview_geometrynode"),
```
jacobtylerwalls: just merged another migration to dev/8, so will need adjusting

arches/app/models/migrations/12002_percent_sign_fix.py
```diff
@@ -0,0 +1,28 @@
...
+    ]
+
+    forward = """
+        UPDATE cards_x_nodes_x_widgets
```
jacobtylerwalls, line 11: I see the idea is this will eventually trickle down to published graphs.

arches/app/models/migrations/12002_percent_sign_fix.py
```diff
@@ -0,0 +1,28 @@
...
+        SET config = jsonb_set(
+            config,
+            '{width}',
+            to_jsonb(
+                regexp_replace(config->>'width', '^(\d+)%{2,}$', '\1%', 'g')
```
jacobtylerwalls, line 16: This changed my `"width": "100%"` to `"width": "\u0001%"`. Maybe it's overkill to add this if it's not safe.

arches/app/models/migrations/12002_percent_sign_fix.py
```diff
@@ -0,0 +1,28 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("models", "11848_alter_spatialview_geometrynode"),
+    ]
+
+    forward = """
```
jacobtylerwalls:
```suggestion
    forward = r"""
```
To avoid:
```py
/Users/jwalls/prj/arches/arches/app/models/migrations/12002_percent_sign_fix.py:10: SyntaxWarning: invalid escape sequence '\d'
  forward = """
```

jacobtylerwalls (changes requested):

arches/app/models/migrations/12002_percent_sign_fix.py
aarongundel: yep

aarongundel:
rebased on 8 to pull in test fixes.

arches/app/models/migrations/12002_percent_sign_fix.py
jacobtylerwalls: [this change](https://github.com/archesproject/arches/pull/12110#discussion_r2112377359) fixed it for me (on python 3.13)

jacobtylerwalls (approved):

PR #12123, opened 2025-05-28
some refinements to what Rob did with term search

apeters:
I basically extracted all the logic out of the view and put it into the utility.  This keeps the view pretty lean and removes all requirement for using ES if we wanted to in the future

### DIFF
--- a/arches/app/search/base_index.py
+++ b/arches/app/search/base_index.py
@@ -190,23 +190,6 @@ class BaseIndex(object):
         else:
             raise NotImplementedError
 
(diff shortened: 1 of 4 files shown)


apeters:
One potential further refinement could be to just use one unified class instead of the two that I created `arches.app.search.search_term.TermSearch` and `arches.app.search.search_term.ConceptSearch`.

apeters:
re #12106

robgaston (approved):

PR #12126, opened 2025-05-29
Bump minimum version of requests

jacobtylerwalls:
At least on 7.6 @bferguso reports oauth still works on these more recent versions of `requests` with security fixes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ dependencies = [
     "pytz==2023.3",
     "qrcode>=7.3.1",
     "rdflib==4.2.2",
-    "requests-oauthlib==1.3.1",
+    "requests-oauthlib==2.0.0",
     "requests[security]>=2.32.3",
     "semantic-version==2.10.0",
     "SPARQLWrapper==1.8.5",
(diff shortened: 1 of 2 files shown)


bferguso:
I'm wondering if we should be looking at bumping request-oauth -> 2.0.0 as well? `requests-oauthlib==2.0.0` + `requests[security]>=2.32.3` seems to work fine in Arches 7.6 with our OAuth config for both Django 4.2.21 and 5.2.1. If we don't want to do a major version update in 7.6, maybe worth doing it for 8.

`django-oauth-toolkit` is also quite stale - we're on 1.7.0 and 3.0.1 is now out. I'm unable to test that as it's used for oauth2 provider, not for external oauth user.

bferguso (approved):
LGTM! Maybe we should update `django-oauth-toolkit` and `requests-oauth` as well. These are major version updates so maybe best to target 8.0 for that.

jacobtylerwalls:
> django-oauth-toolkit is also quite stale - we're on 1.7.0 and 3.0.1 is now out. I'm unable to test that as it's used for oauth2 provider, not for external oauth user.

Right, I think we should hold that until someone has the ability to test.

Thanks for looking!

PR #12138, opened 2025-05-29
Fix importer treatment of missing key/value pairs in tile data #11836

### LINKED ISSUE #11836: import_business_data --overwrite in arches-json format corrupts tiles
#### To reproduce
- Import business data from arches-json where a tile with 10 data-collecting nodes has only 3 k/v pairs in the json
Example in arches-lingo:
```py
python manage.py packages -o import_business_data -s tests/fixtures/data/aat_entries_scheme_examples.json -ow overwrite
python manage.py packages -o import_business_data -s tests/fixtures/data/aat_entries_concept_examples.json -ow overwrite
```
- Result: 10 k/v pairs in the tile json, with the 7 missing keys initialized to None
- Repeat step 1
- Result: 3 k/v pairs in the tile json, the other 7 None's got deleted. Crashes in datatype methods.

jacobtylerwalls:
### Types of changes

### Description of Change
Before, the check in `TileModel.save()` that initialized empty node values to None only fired if the sortorder was `None` or it was fully provisional. Instead, `Tile()` instances get instantiated with a default sortorder of 0 in the business data importer, evading the check:

https://github.com/archesproject/arches/blob/6ec898d71a486f8920fdb13649c2452ae88a117e/arches/app/utils/data_management/resources/formats/archesfile.py#L229-L230

https://github.com/archesproject/arches/blob/6ec898d71a486f8920fdb13649c2452ae88a117e/arches/app/models/models.py#L1742

### Issues Solved

Closes #11836

### Checklist

### DIFF
--- a/arches/app/models/models.py
+++ b/arches/app/models/models.py
@@ -1375,7 +1375,7 @@ class TileModel(models.Model):  # Tile
         return any_key_set
 
     def save(self, *args, **kwargs):
-        if _any_key_set := self.set_missing_keys_to_none():
+        if self.set_missing_keys_to_none():
             add_to_update_fields(kwargs, "data")
         if self.sortorder is None or self.is_fully_provisional():
             sortorder_max = TileModel.objects.filter(
--- a/arches/app/models/models.py
+++ b/arches/app/models/models.py
@@ -1368,14 +1368,14 @@ class TileModel(models.Model):  # Tile
             if node.datatype == "semantic":
                 continue
             node_id_str = str(node.pk)
-            if not node_id_str in self.data:
+            if node_id_str not in self.data:
                 self.data[node_id_str] = None
                 any_key_set = True
 
         return any_key_set
 
     def save(self, *args, **kwargs):
-        if any_key_set := self.set_missing_keys_to_none():
+        if _any_key_set := self.set_missing_keys_to_none():
             add_to_update_fields(kwargs, "data")
         if self.sortorder is None or self.is_fully_provisional():
             sortorder_max = TileModel.objects.filter(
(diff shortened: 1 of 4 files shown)


arches/app/models/models.py
```diff
@@ -1360,14 +1360,24 @@ def __init__(self, *args, **kwargs):
...
 
+    def set_missing_keys_to_none(self):
+        any_key_set = False
+        if not self.nodegroup_id:
+            return any_key_set
+        for node in self.nodegroup.node_set.all():
+            if node.datatype == "semantic":
+                continue
+            node_id_str = str(node.pk)
+            if not node_id_str in self.data:
```
apeters: I think this reads much more naturally
```suggestion
            if node_id_str not in self.data:
```

arches/app/models/models.py
```diff
@@ -1360,14 +1360,24 @@ def __init__(self, *args, **kwargs):
...
+                self.data[node_id_str] = None
+                any_key_set = True
+
+        return any_key_set
+
     def save(self, *args, **kwargs):
+        if any_key_set := self.set_missing_keys_to_none():
```
apeters: doesn't look like `any_key_set` is used here:

```suggestion
        if self.set_missing_keys_to_none():
```

jacobtylerwalls: I guess I thought this was more clear, since from the method name you can't tell why it would have a return value.

apeters: I guess I assume that anything used in an "if" statement has a return value.  So that didn't bother me as much as the return value not being used.

jacobtylerwalls: [78f0f52](https://github.com/archesproject/arches/pull/12138/commits/78f0f524b628454ac595d93eba208c93bf600644)

apeters (approved):
lgtm

PR #12144, opened 2025-05-30
Filter resources by read perms

chiatt:
### Types of changes

-   [x] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
Verifies read permissions on resources requested from bulk disambiguated resource api endpoint

### Issues Solved

Closes #11787 

### Checklist

-   I targeted one of these branches:
    - [ ] dev/8.1.x (under development): features, bugfixes not covered below
    - [ ] dev/8.0.x (main support): regressions, crashing bugs, security issues, major bugs in new features
    - [x] dev/7.6.x (extended support): major security issues, data loss issues
-   [x] I added a changelog in arches/releases
-   [ ] I submitted a PR to arches-docs (if appropriate)
-   [x] Unit tests pass locally with my changes
-   [ ] I added tests that prove my fix is effective or that my feature works
-   [ ] My test fails on the target branch

#### Accessibility Checklist

[Developer Guide](https://arches.readthedocs.io/en/stable/developing/advanced/accessibility/)

| Topic            | Changed | Retested |
| ---------------- | ------- | -------- |
| Color contrast   |         |          |
| Form fields      |         |          |
| Headings         |         |          |
| Links            |         |          |
| Keyboard         |         |          |
| Responsive Design|         |          |
| HTML validation  |         |          |
| Screen reader    |         |          |

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

PR #12153, opened 2025-06-02
Extract constants to ease using legacy testing packages

### LINKED ISSUE #12137: is_fully_provisional() should check for None explicitly
By only checking truthiness, tiles with values such as 0, False, etc. will not be picked up as authoritative:

https://github.com/archesproject/arches/blob/6ec898d71a486f8920fdb13649c2452ae88a117e/arches/app/models/models.py#L1757-L1758

johnatawnclementawn:
### Types of changes

### Description of Change

Guard is_fully_provisional method from returning True if there are authoritative falsey edits and truthy provisional edits.

### Issues Solved

Closes #12137

### Checklist

#### Ticket Background

### DIFF
--- a/arches/app/const.py
+++ b/arches/app/const.py
@@ -1,4 +1,5 @@
 from enum import Enum, unique
+from uuid import UUID
 
 
 IntegrityCheckDescriptions = {
@@ -34,3 +35,11 @@ class ExtensionType(Enum):
     FUNCTIONS = "functions"
     SEARCH_COMPONENTS = "search_components"
     PERMISSIONS_FRAMEWORKS = "permissions"
+
+
+class ResourceLifecycleStates(Enum):
+    PERPETUAL = UUID("4e2a6b8e-2489-4377-9c9f-29cfbd3e76c8")
+    STANDARD = UUID("7e3cce56-fbfb-4a4b-8e83-59b9f9e7cb75")
+    DRAFT = UUID("9375c9a7-dad2-4f14-a5c1-d7e329fdde4f")
+    ACTIVE = UUID("f75bb034-36e3-4ab4-8167-f520cf0b4c58")
+    RETIRED = UUID("d95d9c0e-0e2c-4450-93a3-d788b91abcc8")
--- a/arches/app/const.py
+++ b/arches/app/const.py
@@ -37,7 +37,8 @@ class ExtensionType(Enum):
     PERMISSIONS_FRAMEWORKS = "permissions"
 
 
-class ResourceLifecycleStates(Enum):
+@unique
+class DefaultLifecycleStates(Enum):
     PERPETUAL = UUID("4e2a6b8e-2489-4377-9c9f-29cfbd3e76c8")
     STANDARD = UUID("7e3cce56-fbfb-4a4b-8e83-59b9f9e7cb75")
     DRAFT = UUID("9375c9a7-dad2-4f14-a5c1-d7e329fdde4f")
--- a/arches/app/models/models.py
+++ b/arches/app/models/models.py
@@ -1755,7 +1755,10 @@ class TileModel(SaveSupportsBlindOverwriteMixin, models.Model):  # Tile
         return None
 
     def is_fully_provisional(self):
-        return bool(self.provisionaledits and not any(self.data.values()))
+        return bool(
+            self.provisionaledits
+            and not any(val is not None for val in self.data.values())
+        )
 
     def save(self, **kwargs):
         if self.sortorder is None or self.is_fully_provisional():
--- a/arches/app/models/models.py
+++ b/arches/app/models/models.py
@@ -1755,9 +1755,8 @@ class TileModel(SaveSupportsBlindOverwriteMixin, models.Model):  # Tile
         return None
 
     def is_fully_provisional(self):
-        return bool(
-            self.provisionaledits
-            and not any(val is not None for val in self.data.values())
+        return bool(self.provisionaledits) and not any(
+            val is not None for val in self.data.values()
         )
 
     def save(self, **kwargs):
--- a/releases/8.0.0.md
+++ b/releases/8.0.0.md
@@ -68,6 +68,7 @@ Arches 8.0.0 Release Notes
 - Add UserPreference model for storing custom JSON configuration for a user [#11842](https://github.com/archesproject/arches/issues/11842)  
 - Give ETL module permissions to the dev user [#12016](https://github.com/archesproject/arches/pull/12016)
 - Update the ETL status page to display only the status from the current user [#12017](https://github.com/archesproject/arches/issues/12017)
+- Guard against falsey node values indicating that a tile is fully provisional[#12153](https://github.com/archesproject/arches/pull/12153)
 -   New option for resource-instance datatype advanced search [#11977](https://github.com/archesproject/arches/pull/11977)
 
 ### Dependency changes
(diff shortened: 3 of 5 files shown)


jacobtylerwalls:
I don't want to blow up the scope here, but we have the same problem in:
- BaseDataType.get_tile_data()
- Resource.get_documents_to_index()
- Resource.delete()
- Tile.is_blank()
- reverse_func in 9648 migration

arches/app/models/models.py
```diff
@@ -1755,7 +1755,10 @@ def find_nodegroup_alias(self):
         return None
 
     def is_fully_provisional(self):
-        return bool(self.provisionaledits and not any(self.data.values()))
+        return bool(
+            self.provisionaledits
+            and not any(val is not None for val in self.data.values())
+        )
```
jacobtylerwalls: I think I'd prefer to see `bool(self.provisionaledits)` rather than `bool(all of it)`, to clarify why the `bool` is even needed.

tests/models/tile_model_tests.py
```diff
@@ -45,18 +45,32 @@ class TileTests(ArchesTestCase):
...
-            resource_instance_lifecycle_state_id="4e2a6b8e-2489-4377-9c9f-29cfbd3e76c8",
+        resources = []
+        resources.append(
+            ResourceInstance(
+                pk="40000000-0000-0000-0000-000000000000",
+                legacyid="40000000-0000-0000-0000-000000000000",
+                graph_id="2f7f8e40-adbc-11e6-ac7f-14109fd34195",
+                createdtime="1/1/2000",
+                resource_instance_lifecycle_state_id="4e2a6b8e-2489-4377-9c9f-29cfbd3e76c8",
+            )
```
jacobtylerwalls: Since these are model instances and not filter() calls, can we use this opportunity to make all of these python UUID types?

jacobtylerwalls: While doing that, you can assign the graph id to some kind of constant so I can tell at a glance what graph I'm working with when I use this resource.

tests/models/tile_model_tests.py
```diff
@@ -45,18 +45,32 @@ class TileTests(ArchesTestCase):
         "rdf_export_document_model",
         "rdf_export_object_model",
         "Cardinality Test Model",
+        "All_Datatypes",
```
jacobtylerwalls, line 50: Any reason not to use Cardinality Test Model?

tests/models/tile_model_tests.py
```diff
@@ -383,6 +397,39 @@ def test_update_sortorder_provisional_tile(self):
...
+                "088e7d2c-c8c1-11ed-a172-0242ac130009": None,
+            },
+        }
+
+        authoritative_tile = Tile(json)
+        authoritative_tile.save(index=False)
+
+        user = User.objects.create_user(
+            username="testuser", password="TestingTesting123!"
+        )
```
jacobtylerwalls, line 422: Maybe you want to take a stab at reusing an existing util as suggested in https://github.com/archesproject/arches/pull/12144/files#r2123857272

tests/models/tile_model_tests.py
```diff
@@ -383,6 +397,39 @@ def test_update_sortorder_provisional_tile(self):
...
+        login = self.client.login(username="testuser", password="TestingTesting123!")
+        provisional_tile = Tile.objects.get(
+            resourceinstance_id=authoritative_tile.resourceinstance_id
+        )
```
jacobtylerwalls: Do you need to refetch here?

tests/models/tile_model_tests.py
```diff
@@ -383,6 +397,39 @@ def test_update_sortorder_provisional_tile(self):
...
+
+        provisional_tile.data["088e7d2c-c8c1-11ed-a172-0242ac130009"] = True
+
+        request = HttpRequest()
+        request.user = user
+        provisional_tile.save(index=False, request=request)
+        self.assertFalse(provisional_tile.is_fully_provisional())
```
jacobtylerwalls: When asserting against values produced by first-party code (i.e. us, and not django or python stdlib), I favor `assertIs(..., False)` for a better error message and more precision. (If we forget to `return` from `is_fully_provisional()`, None will make assertFalse pass but wont pass assertIs.)

jacobtylerwalls:
Oh, and do you know if ri-list and concept-list ever store empty arrays rather than None?

jacobtylerwalls:
Need a change log.

@chiatt should we target 7.6?

tests/models/tile_model_tests.py
johnatawnclementawn: That model only has string & semantic nodes 😞 - I considered editing that model, which could still be done...

jacobtylerwalls: No that's fine, just wondering

tests/models/tile_model_tests.py
jacobtylerwalls: Other tests call `call_command("add_test_users")`, maybe we should just have a helper method that does that.

chiatt:
> Need a change log.
> 
> @chiatt should we target 7.6?

It really would be nice to have it in 7.6 as well.

johnatawnclementawn:
> Oh, and do you know if ri-list and concept-list ever store empty arrays rather than None?

They're stored as `null` in the db, no empty arrays

johnatawnclementawn:
> I don't want to blow up the scope here, but we have the same problem in:
> 
> * BaseDataType.get_tile_data()
> * Resource.get_documents_to_index()
> * Resource.delete()
> * Tile.is_blank()
> * reverse_func in 9648 migration

Moved to #12159 because scope of this PR has already crept

arches/app/const.py
```diff
@@ -34,3 +35,11 @@ class ExtensionType(Enum):
     FUNCTIONS = "functions"
     SEARCH_COMPONENTS = "search_components"
     PERMISSIONS_FRAMEWORKS = "permissions"
+
+
+class ResourceLifecycleStates(Enum):
```
jacobtylerwalls: may as well decorate this with `@unique`,

also, suggest tweaking the name to not imply this is exhuastive -- these are merely the default states for the default lifecycle (something like that)

releases/8.0.0.md
```diff
@@ -68,6 +68,7 @@ Arches 8.0.0 Release Notes
 - Add UserPreference model for storing custom JSON configuration for a user [#11842](https://github.com/archesproject/arches/issues/11842)  
 - Give ETL module permissions to the dev user [#12016](https://github.com/archesproject/arches/pull/12016)
 - Update the ETL status page to display only the status from the current user [#12017](https://github.com/archesproject/arches/issues/12017)
+- Guard against falsey node values indicating that a tile is fully provisional[#12153](https://github.com/archesproject/arches/pull/12153)
```
jacobtylerwalls: this is now fixed in 7.6.11
```suggestion
```

tests/models/tile_model_tests.py
```diff
@@ -383,6 +399,36 @@ def test_update_sortorder_provisional_tile(self):
 
         self.assertEqual(obj.sortorder, 1)
 
+    def test_is_fully_provisional(self):
+        """
+        Tests that a tile is marked as fully provisional even if it has falsey values in its data.
+        """
+        json = {
+            "resourceinstance_id": AllDatatypesTestGraph.RESOURCE1.value,
+            "parenttile_id": "",
```
jacobtylerwalls, line 408: I don't think this is handled the same way as the pk.
```suggestion
```

tests/constants.py
```diff
@@ -0,0 +1,20 @@
+from enum import Enum
+from uuid import UUID
+
+"""
+Constants for use only in Arches test suite.
+Enumerations are grouped by test graph & should include things like:
```
jacobtylerwalls, line 6: cool! 😎

jacobtylerwalls (approved):
Awesome, nice QoL improvements! Left some minor suggestions

PR #12177, opened 2025-06-04
Match permission guard in POST to model method re #12176

### LINKED ISSUE #12176: Make active button returns HTML error response for provisional editor
log in as tester3/Test12345!

view resource editor

click "Make active"

- see "Sorry!" banner instead of informative permission info

jacobtylerwalls:
The model method called by this view checks `user_is_resource_reviewer`, but since this view only checks at the perimeter for `user_is_resource_editor` before returning a JSON error, provisional editors (e.g. `tester3/Test12345!` were sneaking through, raising `PermissionDenied` in the model method, which returns an HTML response.

An HTML response is displayed in knockout as "Request Failed ... please contact your administrator", when that's not desired.

If we need to check _both_ permissions, then let's make that change.

### DIFF
--- a/arches/app/views/api/resource.py
+++ b/arches/app/views/api/resource.py
@@ -31,7 +31,7 @@ from arches.app.utils.permission_backend import (
     get_nodegroups_by_perm,
 )
 from arches.app.utils.permission_backend import get_nodegroups_by_perm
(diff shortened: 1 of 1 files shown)


chrabyrd (approved):
good 🎣 !

PR #9776, opened 2023-07-14
updates publish modal #9775

chrabyrd:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Issues Solved

#9775

Issue #12062, opened 2025-05-13
import_graph() catches all exceptions inside its transaction


jacobtylerwalls:
Similar to #11806, `import_graph()` catches all exceptions, meaning transactions are never rolled back, and commands always exit with success even when there are failures.

https://github.com/archesproject/arches/blob/d854446d2f486d2184bc54cd1d5b68369b4f0f84/arches/app/utils/data_management/resource_graphs/importer.py#L244-L249

johnatawnclementawn:
+1 - ran into when testing #12063

Issue #11806, opened 2025-02-21
TileData.post catches all exceptions inside its transaction


jacobtylerwalls:
`TileData.post` looks like it uses a transaction to roll back failed writes, but it actually has no effect since all exceptions are caught inside its context.

Instead, we could let errors propagate and then transform them to a JSON response outside the transaction context manager.

https://github.com/archesproject/arches/blob/6d065afae35ab38d6c569a49d2e3dd06e1606ab8/arches/app/views/tile.py#L161-L190

PR #12178, opened 2025-06-05
Improves `draft_graph`s #12152

### LINKED ISSUE #12152: omnibus draft_graph improvements

chrabyrd:
### Types of changes

### Issues Solved

Closes #12152 

### Checklist
~~- better signposting in graph designer~~
~~- add help template entry~~
~~- disable permission tab w/ informative tooltip~~
~~- update language to make changes to current graph or similar~~
~~- have designer load on source graph instead of draft graph~~
~~- change graph designer warning modal text on publication update to be less severe~~
~~- move make changes to current graph into menu button~~
~~- remove published on date tag~~
~~- add UI flag for if editing draft/current graph~~
~~- move apply/revert changes button to share space with form-level buttons~~
~~- change button order to match rest of Arches ( eg swap save/cancel )~~
~~- fix dirty state calculation~~
~~- wrench down related graph model methods~~
~~- do not prevent users from accessing edit resource UX if graph dirty~~
~~- pull in #9776~~
~~- update editor to use published graph~~
~~- ticket => programatically enforce model changes to draft graph~~
~~- ticket => draft graphs completely ephemeral?~~
~~- ticket => documentation of automatic business migration rules~~


robgaston:
The "Make Active" flag currently does not update the dirty state when editing the current publication, so a user needs to refresh the page to update the publication and apply these changes

chrabyrd:
> The "Make Active" flag currently does not update the dirty state when editing the current publication, so a user needs to refresh the page to update the publication and apply these changes

ah yeah I remember that was originally by design to not have active state tracked

robgaston (changes requested):
This is a HUGE improvement in so many ways! a few things:
- The "Make Active" flag currently does not update the dirty state when editing the current publication, so a user needs to refresh the page to update the publication and apply these changes
- sometimes switching to the "Cards" tab will cause the publish/revert buttons to disappear
- automatically migrating data for a new graph publication that adds a node seems to always fail for me

chrabyrd:
@robgaston If it's alright I say we punt on `sometimes switching to the "Cards" tab will cause the publish/revert buttons to disappear` . I've filed a followup for it but I don't want to block getting this in asap for release testing sake

robgaston:
I haven't been able to come up with exact repro steps and the workaround is simply to refresh, so I totally agree. thanks!

robgaston (changes requested):
I'm still seeing an issue with the Active flag -- when I change it and then refresh the page the graph is reflecting changes to the current graph publication, ie the buttons for "Revert All Changes" and "Update Publication" are showing as if the current publication is dirty.  This is only reflected after refreshing the page, so I assume we are updating that bit on the backend when I flip the active flag

chrabyrd:
> I'm still seeing an issue with the Active flag -- when I change it and then refresh the page the graph is reflecting changes to the current graph publication, ie the buttons for "Revert All Changes" and "Update Publication" are showing as if the current publication is dirty. This is only reflected after refreshing the page, so I assume we are updating that bit on the backend when I flip the active flag

Huh, have you pulled latest and rebuilt? I'm not getting it locally.


https://github.com/user-attachments/assets/333029de-dae0-4715-85d2-d3066945264b

robgaston:
> > I'm still seeing an issue with the Active flag -- when I change it and then refresh the page the graph is reflecting changes to the current graph publication, ie the buttons for "Revert All Changes" and "Update Publication" are showing as if the current publication is dirty. This is only reflected after refreshing the page, so I assume we are updating that bit on the backend when I flip the active flag
> 
> Huh, have you pulled latest and rebuilt? I'm not getting it locally.
> 
>  Screen.Recording.2025-06-07.at.4.12.45.PM.mov

Sorry, yeah this is working after a rebuild.  my bad!

However, in my test w/ the latest when I force the migration of business data on publication of a new graph version the instances don't seem to reflect the new publication id and thus are flagged as not having been migrated.   Have you seen this?

edit: here's the error I'm seeing in the celery worker:
```
[2025-06-08 15:51:27,802: INFO/MainProcess] Task arches.app.tasks.update_resource_instance_data_based_on_graph_diff[e7f91822-d5fa-4e6f-a7cd-56c0d9d23ce1] received
2025-06-08 15:51:27,886 arches.app.tasks ERROR    'defaultValue'
[2025-06-08 15:51:27,886: ERROR/ForkPoolWorker-8] 'defaultValue'
[2025-06-08 15:51:27,901: INFO/ForkPoolWorker-8] Task arches.app.tasks.update_resource_instance_data_based_on_graph_diff[e7f91822-d5fa-4e6f-a7cd-56c0d9d23ce1] succeeded in 0.0724336669081822s: None
```
looks like this is related to the code that assigns default values from the widget config to tile data - probably just need to handle for cases where that key (`defaultValue`) is not assigned

chrabyrd:
> > > I'm still seeing an issue with the Active flag -- when I change it and then refresh the page the graph is reflecting changes to the current graph publication, ie the buttons for "Revert All Changes" and "Update Publication" are showing as if the current publication is dirty. This is only reflected after refreshing the page, so I assume we are updating that bit on the backend when I flip the active flag
> > 
> > 
> > Huh, have you pulled latest and rebuilt? I'm not getting it locally.
> > Screen.Recording.2025-06-07.at.4.12.45.PM.mov
> 
> Sorry, yeah this is working after a rebuild. my bad!
> 
> However, in my test w/ the latest when I force the migration of business data on publication of a new graph version the instances don't seem to reflect the new publication id and thus are flagged as not having been migrated. Have you seen this?
> 
> edit: here's the error I'm seeing in the celery worker:
> 
> ```
> [2025-06-08 15:51:27,802: INFO/MainProcess] Task arches.app.tasks.update_resource_instance_data_based_on_graph_diff[e7f91822-d5fa-4e6f-a7cd-56c0d9d23ce1] received
> 2025-06-08 15:51:27,886 arches.app.tasks ERROR    'defaultValue'
> [2025-06-08 15:51:27,886: ERROR/ForkPoolWorker-8] 'defaultValue'
> [2025-06-08 15:51:27,901: INFO/ForkPoolWorker-8] Task arches.app.tasks.update_resource_instance_data_based_on_graph_diff[e7f91822-d5fa-4e6f-a7cd-56c0d9d23ce1] succeeded in 0.0724336669081822s: None
> ```
> 
> looks like this is related to the code that assigns default values from the widget config to tile data - probably just need to handle for cases where that key (`defaultValue`) is not assigned

Do you have a specific resource you're hitting this with? I'm unable to repro with my ( albeit very basic ) models. Smells like maybe an issue with the assumption that there should always be a card_x_node_x_widget, which I think recently got added to 8.0?

Regardless, nullguarded those calls. Hopefully this'll work for you now 🤞

arches/app/tasks.py
```diff
@@ -629,6 +629,108 @@ def reverse_etl_load(loadid):
...
+                    updated_nodegroup["nodegroupid"]
+                )
+                == "n"
+            ):
+                nodegroups_with_extra_tiles.append(updated_nodegroup["nodegroupid"])
+
+        for tile in models.TileModel.objects.exclude(sortorder=0).filter(
+            nodegroup_id__in=nodegroups_with_extra_tiles
+        ):
+            tile.delete()
```
robgaston: I think that we may also need to delete tiles where nodegroupids may have changed; for example if a node is changed from being a "collector" to not and then one of it's parents is made a "collector" then we will need to delete all tiles for the original nodegroup.  I think that we can simply delete tiles for nodegroupids that don't match current nodegroups.  What do you think?

This will cascade to child tiles, correct?

robgaston (changes requested):
the migration works now for me - I've been testing w/ Lingo graphs.  Just some more questions about that - thanks!

arches/app/tasks.py
```diff
@@ -629,6 +629,108 @@ def reverse_etl_load(loadid):
...
+            if (
+                updated_nodegroup["cardinality"] == "1"
+                and initial_nodegroup_ids_to_cardinality.get(
+                    updated_nodegroup["nodegroupid"]
+                )
+                == "n"
+            ):
+                nodegroups_with_extra_tiles.append(updated_nodegroup["nodegroupid"])
+
+        for tile in models.TileModel.objects.exclude(sortorder=0).filter(
```
jacobtylerwalls: I don't think we can depend on the first tile having a sortorder of exactly zero. For a long time the bulk data manager created tiles with a null sortorder. A sortorder sequence could start anywhere besides 0, also.

arches/app/tasks.py
robgaston: another Q here: what about when a new nodegroup is introduced into an existing tree of nodes (ie flip an intervening node in a tree to "is collector")? this then orphans tiles in a sense as it can result in tiles whose specified parent is not of the right parent nodegroup.  The simple solution to just delete these tiles (tiles where tile.parenttile.nodegroupid != tile.nodegroup.parentnodegroupid) as well (and their children) I guess

chrabyrd:
Ok updated the logic what I believe you're suggesting @robgaston . Added a series of tests but

> The simple solution to just delete these tiles (tiles where tile.parenttile.nodegroupid != tile.nodegroup.parentnodegroupid) as well (and their children) I guess

is surprisingly hard to write a test for, so skipped for now for the sake of expediency

robgaston:
that's cool - I tested and anecdotally the task successfully deleted these kinds of tiles.  I did notice something in the process though, which is that migrated instances are rendered uneditable by migrations.  it does not seem to impact newly created resources for migrated models, but I see two errors, one when creating new tiles:
Screenshot 2025-06-09 at 5 49 40 PM
and another when editing existing tiles:
Screenshot 2025-06-09 at 5 49 57 PM
Are others seeing this? I'm okay with spinning this off into another ticket so we can move on from this PR

chrabyrd:
using my simple graphs I can't see it, but I believe you in that more complex graphs may cause trouble. IMO punt to a different ticket so this can get in and tested, and I'll roll onto addressing that ASAP

robgaston (approved):
This looks good, but there are still some issues with migrated resource data in some cases. will open a follow up for this

arches/app/views/graph.py
```diff
@@ -697,28 +664,58 @@ def post(self, request, graphid):
...
+                                )
+                            )
+
+                            update_resource_instance_data_based_on_graph_diff(
+                                initial_graph=published_source_graph.serialized_graph,
+                                updated_graph=updated_published_source_graph.serialized_graph,
+                                user=request.user,
+                            )
+                        except Exception as e:
+                            transaction.set_rollback(True)
```
jacobtylerwalls: Can we handle this without `set_rollback` by just handling the exception in the right place? We already handle the logging and response return value in L719 below, so it should just be a matter of factoring out a variable for the arguments to JSONErrorResponse. (The "low-level APIs" section for transaction management has a big "don't use this" disclaimer.)

chrabyrd: after some investigation looks like that's cruft 😄 . It was originally there because the draft_graph is consumed on publish and it wasn't being restored properly on error. All good now though 👍

chrabyrd: oh wait a sec I had the wrong repro steps, this is still borked with that line missing

jacobtylerwalls: Sweet. The suggestion also included removing the try/except, since if something does hit that except somehow, then it won't roll back. (We have a lot of these problems e.g. #11806 and #12062)

jacobtylerwalls: (or keeping the except but `raise`-ing again after whatever message variables we need to set are set)

jacobtylerwalls: > oh wait a sec I had the wrong repro steps, this is still borked with that line missing

Yeah that would be my expectation: we need to either remove the except or re-raise after parsing information

chrabyrd: mmk fixed 👍

jacobtylerwalls (approved):
thanks for looking at my little tangents 😃

Issue #9840, opened 2023-07-27
`I18N_string` and `I18N_json` are brittle against future Django ORM changes


jacobtylerwalls:
As part of #9799, I found it necessary to patch `I18N_string` and `I18N_json` to give them additional attributes expected by Django 4.0.

I asked on the Django bug tracker, and received a [response](https://code.djangoproject.com/ticket/34745) that we're using an unsupported pattern. So, although I have a [workaround](https://github.com/archesproject/arches/commit/a8be8d96a267b26b5e4ac14f5bb90208106fde6c) for Django 4.0, this is liable to keep happening. Suggesting we look into another solution.

Source for `I18N_string.as_sql()`:

https://github.com/archesproject/arches/blob/03b0b1d6211004fd353952a07473ed52652f00ad/arches/app/models/fields/i18n.py#L38-L54

PR #12199, opened 2025-06-10
Bump Django to 5.2.3

jacobtylerwalls:
- Fix I18n field issues uncovered by Django 5.2.3

### DIFF
--- a/arches/app/models/fields/i18n.py
+++ b/arches/app/models/fields/i18n.py
@@ -7,12 +7,15 @@ from django.db.migrations.serializer import BaseSerializer, Serializer
 from django.db.models import JSONField
 from django.db.models.sql import Query
 from django.db.models.sql.compiler import SQLInsertCompiler, SQLUpdateCompiler
-from django.db.models.sql.where import NothingNode
 from django.utils.translation import get_language
 
 
-class I18n_String(NothingNode):
-    """Subclassing NothingNode works around https://code.djangoproject.com/ticket/34745."""
+class I18n_String:
+    # Add attributes until https://github.com/archesproject/arches/issues/9840
+    contains_aggregate = False
+    contains_over_clause = False
+    contains_column_references = False
+    possibly_multivalued = False
 
     def __init__(self, value=None, lang=None, use_nulls=False, attname=None):
         self.attname = attname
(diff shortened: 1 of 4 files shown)


arches/app/models/fields/i18n.py
```diff
@@ -265,11 +285,20 @@ def _parse(self, value, lang, use_nulls):
...
             pass
 
+    def resolve_expression(
+        self, query=None, allow_joins=True, reuse=None, summarize=False, for_save=False
+    ):
+        """
+        Django needs this method to recognize this as both Resolvable & Compilable.
+        https://forum.djangoproject.com/t/revisiting-types-in-django-dep-14/37832/11
+        """
+        return self
```
jacobtylerwalls, line 295: Without adding this, I was getting:
```
django.db.utils.ProgrammingError: can't adapt type 'I18n_JSON'
```
... because Django is starting to remove hacks around SQL NULL != JSON NULL in JSONFields, and that's catching our fragile implementation here that is both a Resolvable _and_ a Compilable.

tests/localization/field_tests.py
```diff
@@ -234,6 +234,34 @@ def test_equality(self):
         translation.activate("de")
         self.assertEqual(value, "genau")
 
+    def test_bulk_update_heterogenous_values(self):
```
jacobtylerwalls, line 237: This is a port of the same unit test from the I18n_Json tests.

tests/localization/field_tests.py
```diff
@@ -467,8 +495,8 @@ def test_bulk_update_heterogenous_values(self):
         DDataType.objects.bulk_update(for_bulk_update, fields=["defaultconfig"])
 
         for i, obj in enumerate(for_bulk_update):
-            # None is transformed to {"en": ""}.
-            new_config_as_string = str(new_configs[i] or {"en": ""})
+            # None is transformed to {}.
+            new_config_as_string = str(new_configs[i] or {})
```
jacobtylerwalls, line 499: I don't think anyone was depending on this behavior: this test is new in Arches 8.

arches/app/models/fields/i18n.py
jacobtylerwalls: This is evident from the failing target branch job (hence why I'd like to get this in fast if at all possible)

aarongundel (approved):
This LGTM - I notice the target branch is failing here so I'm guessing that's what inspired you to do this!  IMO we should have #9840 on deck for 8.1

PR #12206, opened 2025-06-10
Avoid cascade deleting bulk data manager records on graph version restore #12032

### LINKED ISSUE #12032: Restoring graph version cascade deletes bulk data manager records
- Create a model with a number node called "number"
- Save edits
- Publish
- Make active
- Create a .csv with this invalid data:
```csv
resourceid,number
042f8268-51d6-4d6c-8dec-8ee6ab9b0af8,x
```
- bulk data manger > single csv import >
- match columns to nodes (use as an id, number)
- Import Data
- Notice validation errors in right hand side
- Graph Manager: Model history
- Restore prior version
- Refresh bulk data manger
- Notice left hand side shows failed load, but right hand side shows "No error found"

~Likely a cause of LoadStaging / LoadErrors being cascade deleted. Should perhaps be un-foreign keyed like we did tiles? Perhaps we could try with `db_constraint = False` and `DO_NOTHING` instead of `UUIDField`?~

jacobtylerwalls:
### Types of changes

### Description of Change
The restore graph version workflow deletes nodegroups, so we were getting cascade deletion of bulk data manager records due to the `on_delete=models.CASCADE`. While we want cascade deletes for graph deletion, I don't believe that's what we want for graph versioning.

### Issues Solved
Closes #12032

### Checklist

#### Ticket Background

### DIFF
--- a/arches/app/models/graph.py
+++ b/arches/app/models/graph.py
@@ -23,6 +23,7 @@ from contextlib import contextmanager
 from copy import deepcopy
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import transaction, connection
+from django.db.models import Q
 from django.db.utils import IntegrityError
 from arches.app.const import IntegrityCheck
 from arches.app.models import models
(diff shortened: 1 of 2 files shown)


arches/app/models/graph.py
```diff
@@ -1606,6 +1607,35 @@ def get_nodegroups(self, force_recalculation=False):
...
+            }
+            valid_stagings = {
+                obj for obj in staging_objs if obj.nodegroup_id in all_nodegroup_ids
+            }
+            invalid_errors = error_objs - valid_errors
+            invalid_stagings = staging_objs - valid_stagings
+            models.LoadErrors.objects.bulk_update(valid_errors, fields=["nodegroup"])
+            models.LoadStaging.objects.bulk_update(valid_stagings, fields=["nodegroup"])
+            models.LoadErrors.objects.filter(pk__in=invalid_errors).delete()
+            models.LoadStaging.objects.filter(pk__in=invalid_stagings).delete()
```
jacobtylerwalls: Can also just delete these two lines if we'd prefer leaving these records with null nodegroups. cc/ @chiatt

chrabyrd (approved):
lgtm!

arches/app/models/graph.py
jacobtylerwalls: Letting these stay null rather than delete after chatting w/ Cyrus.

PR #12237, opened 2025-06-12
Extend bulk data deletion module to remove tiles with empty tiledata

### LINKED ISSUE #12236: Tiles that have empty tiledata prevent the export functionality
When trying to export resources through the bulk exporters, if one of them has any tile with empty tiledata, the export fails and throws an error.

MikeGriniezakis:
### Types of changes

### Description of Change

Add an option to allow users to select whether they want to limit the deleted resources to the ones
that have empty tiledata.

### Issues Solved

Closes #12236

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/etl_modules/bulk_data_deletion.py
+++ b/arches/app/etl_modules/bulk_data_deletion.py
@@ -1,13 +1,15 @@
-from datetime import datetime
 import json
 import logging
-import pyprind
 import uuid
+
+import pyprind
 from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
 from django.db import connection
 from django.http import HttpRequest
 from django.utils.translation import gettext as _
+
+import arches.app.tasks as tasks
 from arches.app.etl_modules.base_data_editor import BaseBulkEditor
 from arches.app.etl_modules.decorators import load_data_async
 from arches.app.etl_modules.save import get_resourceids_from_search_url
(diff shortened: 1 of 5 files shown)


chiatt:
Because this PR introduces new functionality, needs to target the dev/8.1.x branch rather than dev/7.6.x.

chiatt (commented):
It seems potentially risky to allow users to delete empty tiles. If these tiles are serving as parent tiles, then deleting them could very negatively impact an Arches instance. Perhaps a better solution to #12236 would be to fix the exporter so that it doesn't break on export of empty tiles.

chiatt:
I just confirmed that a card with only a semantic node will save an empty object in tiledata `{}`. Such a tile can serve as the parent node of child tiles with data. If somebody were to unwittingly delete these empty parent tiles, the deletion would probably cascade to their child tiles or at least make their child tiles inaccessible in the system. While I appreciate the effort to solve #12236, I don't think we should follow this approach.

apeters:
@MikeGriniezakis I think we should focus on fixing https://github.com/archesproject/arches/issues/12236 so that it can handle for blank tiles.  Blank tiles themselves aren't an indication that something is wrong.  And deleting blank tiles that have child tiles will orphan those child tiles and make them inaccessible to the application (definitely something we don't want).

Because of this I will close this PR now until a proper fix for #12236 can be made.

MikeGriniezakis:
@apeters, thanks for the suggestions, I have updated the PR in Arches Provenance providing more context to the issue. Hope now it's more clear.

PR #12242, opened 2025-06-12
Keep spatial views trigger deferred re #11848

jacobtylerwalls:
Purely for the sake of future copy/paste, this is what I was getting at in this [comment](https://github.com/archesproject/arches/pull/12102#discussion_r2102745199) about why Composer instead of Trigger -- I think Trigger is the simple version that's more common.

Thanks to the maintainer for also observing we inadvertently removed the deferred timing. Now restored.


aarongundel:
@jacobtylerwalls looks like this is blowing up tests - possibly re: deferred?

jacobtylerwalls:
Taking a look

jacobtylerwalls:
@aarongundel appears to just be a test setup issue -- moving this test to TransactionTestCase helps (makes sense -- that's how it's doc'd -- for testing functionality that itself uses transactions, e.g. deferred triggers)

Part of what might be aggravating the issue here is that the triggers themselves run DDL to deploy views.

jacobtylerwalls:
(The test still fails if you comment out the spatial view save() added in #12097.)

aarongundel (approved):
LGTM - good feedback here that'll help establish patterns going forward.

PR #12247, opened 2025-06-12
Updates function manager buttons to be similar to graph designer

### LINKED ISSUE #12219: Update buttons should be present on the function manager
After saving a function, buttons should appear to to update the current publication like what happens now in the graph designer.

aarongundel:
### Types of changes

### Issues Solved

Closes #12219 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/models/models.py
```diff
@@ -668,7 +668,9 @@ def get_functions_x_graphs(self, force_recalculation=False):
                 for function_dict in function_slugs
             ]
         else:
-            return self.functionxgraph_set.all()
+            return [
```
chrabyrd, line 671: why this change?

arches/app/models/graph.py
```diff
@@ -156,7 +156,9 @@ def check_default_configs(default_configs, configs):
                 for node in self.nodes.values():
                     if node.istopnode:
                         self.root = node
-                self.functions_x_graphs = super().get_functions_x_graphs()
+                self.functions_x_graphs = super().get_functions_x_graphs(
+                    force_recalculation=True
```
chrabyrd: why this change?

chrabyrd (changes requested):
looks good overall just a couple of questions

arches/app/models/models.py
aarongundel: I sent you a message about this on slack; it's because of the call to append to the output of this function.  It'll blow up if that's a queryset.

arches/app/models/graph.py
aarongundel: This was necessary for a minute for me to break out of a bad published graph change where an id was causing a constraint violation on save (see L 392 of models/graph.py) but should no longer be necessary after that was removed.  There's another instance of this happening in one of these getter functions and the output was changed in the same way there.

chrabyrd (changes requested):
Code looks good, but if I refresh the page the update publication buttons do not persist.

chrabyrd (changes requested):
Great, now I can see the publish buttons persist.

One thing I did notice is that it's possible to show both dirty state buttons and publish buttons -- the condition to show dirty state buttons should override the condition to show the publication buttons.

chrabyrd (commented):
Added a small addition into this to handle for combination of dirty states of both draft and source ( that people can now hit via the function manager ). When all tests pass I'll approve/merge.

Thanks for getting this in 🚀

chrabyrd (approved):
lgtm!

PR #12251, opened 2025-06-13
Spruce up readme and release notes

jacobtylerwalls:
Copy over the note from the 7.6 release notes about the need for running `manage.py check` to generate a frontend configuration.

Any of the other manage.py commands will accomplish that as well, but the experience with 7.6 was that people prefer to see this spelled out so they can evaluate whether to skip steps.

### DIFF
--- a/releases/8.0.0.md
+++ b/releases/8.0.0.md
@@ -453,7 +453,7 @@ JavaScript:
 
     ```
 
-1. Update the JavaScript files in your project as follows. (Note that even projects without custom JavaScript may have been templated with a `media/js/reports/default.js` that needs updating.)
+1. Update the JavaScript files in your project as follows. (Note that even projects without custom JavaScript may have been templated with a `media/js/reports/default.js`. This file may be removed if it is not intentionally overriding default.js in core Arches. Otherwise, it must be updated.)
     1. Update the import and export logic. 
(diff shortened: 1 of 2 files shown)


releases/8.0.0.md
```diff
@@ -445,7 +453,7 @@ JavaScript:
 
     ```
 
-1. Update the JavaScript files in your project:
+1. Update the JavaScript files in your project as follows. (Note that even projects without custom JavaScript may have been templated with a `media/js/reports/default.js` that needs updating.)
```
chiatt: Maybe it's too much here, but it seems like it's worth mentioning that this default.js file can be removed. Maybe something like:

```Note that even projects without custom JavaScript may have been templated with a `media/js/reports/default.js`. This file may be removed if it is not intentionally being used to override default.js in core arches. If it is not removed, it must be updated as well.```

Comment (jacobtylerwalls) [inline releases/8.0.0.md]:
```suggestion
1. Update the JavaScript files in your project as follows. (Note that even projects without custom JavaScript may have been templated with a `media/js/reports/default.js` that needs updating or removing if not used to override default.js in core Arches.)
```

jacobtylerwalls: I'll go with more like this -- no room for misinterpretation.

chiatt (approved):
Looks good!

PR #12254, opened 2025-06-13
Remove module-level query in URL datatype #11821

### LINKED ISSUE #11821: Remove module-level query in datatypes/url.py
Module-level queries are [discouraged](https://docs.djangoproject.com/en/5.1/ref/applications/#troubleshooting) in the django docs since they might execute against the wrong database or cache stale data. They're supposed to emit `RuntimeWarning`, this one is likely not emitting that because in most cases the module is imported late, but there's no reason to expect that to always be the case, e.g. in a management command etc.

https://github.com/archesproject/arches/blob/d27ba10b53f8ca05d9c51ed77c9aab1516661b34/arches/app/datatypes/url.py#L42

jacobtylerwalls:
I don't think this is truly needed to register this datatype, since the 7224_url_datatype migration creates this datatype.

Leaving this module-level query here causes flaky `RuntimeWarning` depending on whether you happen to import this module:

```py
  File "/Users/jwalls/py313/lib/python3.13/site-packages/django/db/backends/utils.py", line 98, in _execute
    warnings.warn(self.APPS_NOT_READY_WARNING_MSG, category=RuntimeWarning)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeWarning: Accessing the database during app initialization is discouraged. To fix this warning, avoid executing queries in AppConfig.ready() or when your app modules are imported.
```

Closes #11821

Noticed this in archesproject/arches-querysets#22

~The only reason we weren't seeing this in core is because url.py has 0% test coverage~

### DIFF
--- a/arches/app/datatypes/url.py
+++ b/arches/app/datatypes/url.py
@@ -20,7 +20,6 @@ import re
 import json
 from arches.app.models.system_settings import settings
 from arches.app.datatypes.base import BaseDataType
-from arches.app.models.models import Widget
 from arches.app.search.elasticsearch_dsl_builder import Match, Exists, Term
 from arches.app.search.search_term import SearchTerm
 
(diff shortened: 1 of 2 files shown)


apeters (approved):
lgtm

apeters:
I'm not sure why the coverage check went down.  With removing untested code, I would've thought it would go up.

jacobtylerwalls:
> I'm not sure why the coverage check went down. With removing untested code, I would've thought it would go up.

I misspoke. I must have looked at a coverage report locally after I'd only run a subset of tests. url.py had about 58% coverage. So that module level details dict was probably counted toward coverage.

Why we weren't getting the RuntimeWarning in test, I don't know.

PR #12268, opened 2025-06-16
[7.6] Remove module-level query in URL datatype #11821

jacobtylerwalls:
I think backporting this fix to 7.6 is better than adding awkward workarounds in arches-querysets to enable testing on 7.6, see [logs](https://github.com/archesproject/arches-querysets/actions/runs/15681704641/job/44174483381).

### DIFF
--- a/arches/app/datatypes/url.py
+++ b/arches/app/datatypes/url.py
@@ -20,7 +20,6 @@ import re
 import json
 from arches.app.models.system_settings import settings
 from arches.app.datatypes.base import BaseDataType
-from arches.app.models.models import Widget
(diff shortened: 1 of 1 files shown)


johnatawnclementawn (approved):
I suppose we're seeing the same degradation in coverage as we did in the [original pr](https://github.com/archesproject/arches/pull/12254#issuecomment-2970848271)

johnatawnclementawn:
Need a release note?

jacobtylerwalls:
We've already published it in the 8.0 release note, so I guess I'd rather not double-document even though we're now backporting after the fact. idk.

PR #12291, opened 2025-06-24
Prevent key error during migration #12290

### LINKED ISSUE #12290: Migration to v8 fails if published graph does not have related functions
If a user attempts to migrate a graph that does not have related functions saved to it's publication, the migration will fail

chiatt:
### Types of changes

### Description of Change
Prevents KeyError when migrating graphs without a 'function_x_graphs' key in their publication.

### Issues Solved

Closes #12290

### DIFF
--- a/arches/app/models/models.py
+++ b/arches/app/models/models.py
@@ -648,18 +648,19 @@ class GraphModel(SaveSupportsBlindOverwriteMixin, models.Model):
             published_graph = self.get_published_graph()
 
             function_slugs = []
-            for function_dict in published_graph.serialized_graph["functions_x_graphs"]:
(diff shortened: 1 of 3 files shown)


aarongundel (approved):
LGTM

jacobtylerwalls:
Looks like we have to guard against this given graph publications in the wild now, but did you look into why the key was missing in the first place?

chiatt:
> Looks like we have to guard against this given graph publications in the wild now, but did you look into why the key was missing in the first place?

I couldn't track down why the key was missing because once I fixed the problem and could run migrations, the migration corrected it. I suspect the issue might have been with the system settings graph.

PR #12300, opened 2025-07-01
Fix dummy saves

jacobtylerwalls:
### Types of changes

### Description of Change
arches-querysets overrides `TileModel.save()` to orchestrate saving child tiles in bulk, but an [attempt is made](https://github.com/archesproject/arches-querysets/blob/214582e3bf6843c2d6d20e71db695d48aea0d6c6/arches_querysets/models.py#L501) to still run any other side effects from the `save()` method without re-saving the tile by passing `update_fields=set()`. This ran afoul of this helper's assumption any not-None value should have update fields appended. Falsy values like an empty set should be left alone: I feel in that case the caller knows what they're doing.

### Checklist

### DIFF
--- a/arches/app/models/utils.py
+++ b/arches/app/models/utils.py
@@ -3,7 +3,7 @@ def add_to_update_fields(kwargs, field_name):
     Update the `update_field` arg inside `kwargs` (if present) in-place
     with `field_name`.
     """
-    if (update_fields := kwargs.get("update_fields")) is not None:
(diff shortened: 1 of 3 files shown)


jacobtylerwalls:
Setting to draft until I can hear advice on the Django forum about the [catch-22 this exposed](https://forum.djangoproject.com/t/update-or-create-defaults-none-sends-update-fields-set-rather-than-none/41657).

PR #12302, opened 2025-07-03
Get RDFLib failures on Python 3.14

jacobtylerwalls:
Adding Python 3.14 support is going to force us to upgrade RDFLib, so we're going to need some expert help in that area.

[releases](https://github.com/RDFLib/rdflib/releases)


chiatt:
Resolved here https://github.com/archesproject/arches/pull/12790

PR #12312, opened 2025-07-16
Handle case where project_requirements is None

jacobtylerwalls:
`requires` can return None, was raising `NoneType is not iterable` in some cases.


chiatt (approved):
Looks good!

PR #12323, opened 2025-07-23
Lift up lifecycle handlers and edit log saves from Resource to ResourceInstance #12321

### LINKED ISSUE #12321: Lift up lifecycle methods from Resource to ResourceInstance
> The issue is that I'm imagining this proxy model as a successor to the old one, so I'd like to avoid calling into the old proxy model save() -- we'll never uncouple that in the future if we start now.

What I'd *love* to do is move this bit of logic from Resource to ResourceInstance. Then ResourceTileTree already subclasses _that_.

_Originally posted by @jacobtylerwalls in https://github.com/archesproject/arches-querysets/pull/75#discussion_r2223615834_

jacobtylerwalls:
### Types of changes

### Description of Change
Before, the lifecycle handlers and edit log saves were trapped on `Resource`, making it impossible to run these side effects from any other `ResourceInstance` subclasses.

Now, other subclasses can run these side effects.

This is part of the campaign to move any functionality that does not strictly depend on the data-fetching philosophy of the existing proxy models out of the proxy models.

### Issues Solved

Closes #12321

### Checklist

    - [x] dev/8.0.x (main support): regressions, crashing bugs, security issues, _major bugs in new features_

### DIFF
--- a/arches/app/models/models.py
+++ b/arches/app/models/models.py
@@ -1517,7 +1517,18 @@ class ResourceInstance(SaveSupportsBlindOverwriteMixin, models.Model):
 
         return creatorid
 
-    def save(self, **kwargs):
+    def save(
+        self,
+        *,
+        context=None,
+        index=True,
+        request=None,
+        transaction_id=None,
+        user=None,
+        should_update_resource_instance_lifecycle_state=False,
+        current_resource_instance_lifecycle_state=None,
+        **kwargs,
+    ):
         try:
             self.graph_publication = self.graph.publication
         except ResourceInstance.graph.RelatedObjectDoesNotExist:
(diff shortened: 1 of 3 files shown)


arches/app/models/models.py
```diff
@@ -1532,6 +1543,69 @@ def save(self, **kwargs):
...
+                newvalue=f"{self.resource_instance_lifecycle_state.name} ({self.resource_instance_lifecycle_state.id})",
+                transaction_id=transaction_id,
+            )
+            self.run_lifecycle_handlers(
+                current_resource_instance_lifecycle_state,
+                request=request,
+                context=context,
+            )
+            if index:
+                self.index(context)
```
jacobtylerwalls, line 1560: I added `if index:` -- alternatively we can raise if we want index=False to be incompatible with updating lifecycle state.

jacobtylerwalls: This is theoretical for now -- the only place that sends `should_update_resource_instance_lifecycle_state=True` today omits the index= arg and is thus `index=True`.

chrabyrd (approved):
lgtm

PR #12324, opened 2025-07-23
Remove no-op get_interchange_value()

jacobtylerwalls:
This abstract method was added in core to support rapid prototyping in arches-component-lab and arches-querysets, but the need for it is no more, see https://github.com/archesproject/arches-querysets/pull/64.

It would just be misleading to keep maintaining it here.


johnatawnclementawn (approved):
lgtm 👍🏼

PR #12359, opened 2025-08-05
Fix broken RunPython operation

jacobtylerwalls:
@aarongundel spotted that with unpublished graphs in your database, this RunPython operation crashed with:

```py
django.db.utils.ProgrammingError: column node_groups.groupingnodeid does not exist
LINE 1: ..."cardinality", "node_groups"."parentnodegroupid", "node_grou...
```


aarongundel (approved):
LGTM and thank you!

PR #12360, opened 2025-08-05
12336 - Add lifecycle state id to relation view update function

### LINKED ISSUE #12336: Instances relational view not populating resource_instance_lifecycle_state_id column
Need to update the `__arches_instance_view_update` function to set the `resource_instance_lifecycle_state_id` column in the `resource_instances` table when inserting or updating the `instances` relational view.

For discussion - should there be a default value on insert if that value is not provided?

jacobtylerwalls:
### Types of changes

### Description of Change
Update the relational view update function to set lifecyle state id.

_This adds a dependency `django-sql-migrate-deux` to generate forward & reverse Django migrations for functions from a single source of truth file like we're using for triggers._

### Issues Solved

Closes #12336

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/models/sql/functions/arches_instance_view_update.sql
```diff
@@ -0,0 +1,85 @@
+create or replace function __arches_instance_view_update() returns trigger as
```
bferguso: Small nit - can we make the the SQL filename with the DB object name?

bferguso (changes requested):
This is really cool. The `makemigrations --check` ensures that the latest version of the source SQL file is what is in the latest migration, which is great. Just wondering if we should make the SQL filename the same as the db object name.

arches/app/models/sql/functions/arches_instance_view_update.sql
jacobtylerwalls: Happy to change, but was just following the existing example in the sql/triggers folder. Quick head nod from @chiatt might be good about which way to go here (whether to add the `__` prefix).

chiatt: Given that so many of our functions are `__arches` prefixed, we should probably just stick with it. Some of our functions don't have any prefix at all. I think it would be nice at some point to rename all of our functions with just `arches_`, but maybe that should be done in a single, future PR.

releases/8.0.0.md
```diff
@@ -287,12 +288,13 @@ JavaScript:
             }
         }
         ```
-    - Change the `"captcha"` entry in `INSTALLED_APPS` to reflect its new name and add pgtrigger:
+    - Change the `"captcha"` entry in `INSTALLED_APPS` to reflect its new name, and add `"pgtrigger"` and `"django_sql_migrate"`:
```
chiatt, line 291: Really nice catch adding this to the 8.0.0 release notes.

chiatt (approved):
Looks good to me!

arches/app/models/sql/functions/README.md
```diff
@@ -0,0 +1,54 @@
+Adding an initial migration for existing database functions
```
jacobtylerwalls, line 1: Thanks @aarongundel for pointing out that manually copying over reverse sql was not the best path to take here. Updated this with a procedure that now results in two commits, one capturing the initial state, and one with a clean diff on the changes needed for the update.

aarongundel: Looks great - thanks for walking me through what the state migration bit is doing there.

aarongundel (approved):

arches/app/models/sql/functions/README.md
```diff
@@ -0,0 +1,54 @@
...
+                    sql="\nCREATE OR REPLACE FUNCTION ...
+                    reverse_sql="drop function __arches_instance_view_update;",
+                ),
+            ],
+        ),
+        django_migrate_sql.operations.AlterSQL(
+            name="__arches_instance_view_update",
+            sql="\ncreate or replace function __arches_instance_view_update() ...,
+            reverse_sql="\nCREATE OR REPLACE FUNCTION public.__arches_instance_view_update()...,
+            state_reverse_sql="drop function __arches_instance_view_update;",
```
jacobtylerwalls, line 52: a summary of convo with @aarongundel:

`reverse_sql` is what will get executed on the db in reverse.

`state_reverse_sql` is for Django internals only, and here it's just echoing back what we passed to the `reverse_sql` arg of `SQLItem()`, which is just a dummy since we didn't / aren't maintaining a whole separate .sql file for the first reverse sql in the chain.

PR #12363, opened 2025-08-07
Make `UserProfile.{*}able_nodegroups` a `@cached_property`

jacobtylerwalls:
### Types of changes

### Description of Change
arches-querysets has a use case for checking `.user.userprofile.editable_nodegroups` several times, and I've had to store off the result of this property to avoid repetitive queries, when really it would be a better DX to make sure that it was only calculated once at the source (per instance / request).


aarongundel (approved):
This is cool - LGTM.

Issue #12088, opened 2025-05-16
Updating a node in the Graph Designer causes psycopg2 deadlock


chrabyrd:
To reproduce:

1. Navigate to the Graph Designer and create a new graph
2. Add a node, change it to string.
3. Navigate to the cards tab
4. Notice the error 

Example stacktrace:

```
Traceback (most recent call last):
  File "/Users/christopherbyrd/Projects/arches-lingo/ENV/lib/python3.12/site-packages/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
psycopg2.errors.DeadlockDetected: deadlock detected
DETAIL:  Process 56162 waits for ShareLock on transaction 2471107; blocked by process 56163.
Process 56163 waits for ShareLock on transaction 2471108; blocked by process 56162.
HINT:  See server log for query details.
CONTEXT:  while updating tuple (1,18) in relation "graphs"


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/christopherbyrd/Projects/arches-lingo/ENV/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/christopherbyrd/Projects/arches-lingo/ENV/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/christopherbyrd/Projects/arches-lingo/ENV/lib/python3.12/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/christopherbyrd/Projects/arches-lingo/ENV/lib/python3.12/site-packages/django/utils/decorators.py", line 48, in _wrapper
    return bound_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #12142, opened 2025-05-30
Coerce uuid strings to UUID types when loading graph json


jacobtylerwalls:
One of the most common classes of bugs in Arches is type confusion between `str` and `UUID`. The entire system would be easier to reason about if we only had one type to contend with and every input boundary produced that one type.

The ORM already coerces to `UUID`. (It's lenient when you provide `str` to a queryset method like `filter`, but the behavior is undefined if you provide a `str` to a model field, see #10877.)

If we fixed every other input boundary besides the ORM to do the same, then we'd have far fewer bugs in this area.

Suggest we deserialize graph json with a decoder that always produces `UUID` types.

***

> Looks like the source of the `str` type is the serialized graph. I wonder if in a future version of arches we should just deserialize graph publications in a way that coerces types correctly. I'll bounce this to an enhancement ticket.

_Originally posted by @jacobtylerwalls in https://github.com/archesproject/arches/pull/12136#discussion_r2116136133_

PR #12383, opened 2025-08-15
Bulk save graph entities

jacobtylerwalls:
### Types of changes

### Description of Change
Bulk save graph entities.

I had this lying around from debugging #12088 earlier this year, but I think it's still useful.

### Checklist

### DIFF
--- a/arches/app/models/graph.py
+++ b/arches/app/models/graph.py
@@ -551,6 +551,8 @@ class Graph(models.GraphModel):
                     unique_fields={"pk"},
                 )
 
+            for edge in self.edges.values():
+                edge.clean()
             models.Edge.objects.bulk_create(
                 self.edges.values(),
                 update_conflicts=True,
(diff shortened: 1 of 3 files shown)


chrabyrd (changes requested):
mind resolving the merge conflicts?

chrabyrd (changes requested):
won't this skip any functionality that runs in the `save` and `delete` methods?

jacobtylerwalls:
Yep, one outcome of this conversation could be "don't do this", but then I was hoping to capture the "why not" in a test case that would fail if this is attempted in the future.

I just took a look: the only side effects in the `save()` methods in question here involve:
- minting new node aliases
- nulling the source_identifier field if the pk matches the source_identifier

I can update this PR to handle those cases. More interesting is _what about the future case?_

***
Make of this what you will, but I think there are several places where Arches is suffering from the lack of a form/serializer layer at the perimeter to validate inputs and coerce types. One angle is #12142, there are others. I totally understand why there is code to mint aliases and support graph versioning in the model save methods, but it feels like a compromise given the general lack of a form/serializer layer. To me, saying we shouldn't bulk save because we need to preserve the coupling of _cleaning data_ to _saving objects individually_ seems like memorializing the wrong design. I think instead we should encourage liberating side effects out of the model save and delete methods to make the models more bulk-friendly.
***
I could get you a little profile of the before & after performance impact here to help with a decision, just wanted to give you a frame to react to first!

PR #12420, opened 2025-08-27
Add Users Command

### LINKED ISSUE #12419: Allow for users to be loaded from an Arches Package
Currently there's no way to load a list of users from an arches package - it has to be done manually.  It would be nice if this were possible.

aarongundel:
### Types of changes

### Description of Change

Allows for a list of users to be loaded from an arches package

Example of new command usage.

`python manage.py add_users --power-user`  adds a power user (dev) and default number (3) of test users
`python manage.py add_users --power-user -c 0`  adds a power user (dev) and  no test users
`python manage.py add_users --operation csv_users -s ./users.csv`  loads users from a csv file
`python manage.py add_test_users` adds a power user and 3 test users - same behavior as current

### Issues Solved

Closes #12419 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/management/commands/add_users.py
+++ b/arches/management/commands/add_users.py
@@ -1,0 +1,207 @@
+"""
+ARCHES - a program developed to inventory and manage immovable cultural heritage.
+Copyright (C) 2013 J. Paul Getty Trust and World Monuments Fund
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as
+published by the Free Software Foundation, either version 3 of the
+License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+"""
+
+import argparse
+import csv
+import traceback
+from django.core.management.base import BaseCommand
+from django.contrib.auth.models import User
+from django.contrib.auth.models import Group
+from django.contrib.auth.models import Permission
+from django.utils.crypto import get_random_string
+from arches.app.utils.permission_backend import assign_perm
+from arches.app.models import models
+
+
+class Command(BaseCommand):
+    """
+    Commands for adding arches test users
+
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "-o",
+            "--operation",
+            action="store",
+            dest="operation",
+            choices=[
+                "test_users",
+                "csv_users",
+            ],
+            default="test_users",
+            help="Operation Type; "
+            + "'test_users'=Creates test users"
+            + "'csv_users'=Loads users from CSV file",
+        )
+
+        parser.add_argument(
+            "-s",
+            "--source",
+            action="store",
+            dest="csv_file",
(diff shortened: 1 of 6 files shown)


arches/management/commands/add_users.py
```diff
@@ -0,0 +1,189 @@
...
+                        assign_perm("change_plugin", user, plugin)
+                        assign_perm("add_plugin", user, plugin)
+                        assign_perm("delete_plugin", user, plugin)
+                        assign_perm("view_plugin", user, plugin)
+                    for etl_module in etl_modules:
+                        assign_perm("view_etlmodule", user, etl_module)
+                user.save()
+                print(
+                    f"Added test user: {user.username}, password: {profile['password']}"
+                )
```
jacobtylerwalls: Can we update the print statements throughout to `self.stdout.write()` while here? These print statements are noisy when run during the arches-querysets test suite.

aarongundel: good to see you still lurking!  😃  Done.

chiatt (commented):
Could you add a test?

arches/management/commands/add_users.py
```diff
@@ -0,0 +1,186 @@
...
+            self.load_users_from_csv(options)
+
+    def add_test_users(self, options):
+        user_count = options.get("user_count")
+        profiles = []
+        for current in range(1, user_count + 1):
+            profiles.append(
+                {
+                    "name": f"tester{current}",
+                    "email": f"tester{current}@test.com",
```
chiatt: If a user runs this command more than once because they will hit a unique constraint on the username and email fields because f"tester{current}" will always start at 1. To avoid that problem, I think you'd want to get the number of users already in they system and name the added users starting with that number. Something like this would work:
```suggestion
        user_count = options.get("user_count")
        profiles = []
        number_of_existing_users = User.objects.count()
        for current in range(1, user_count + 1):
            suffix = current + number_of_existing_users
            profiles.append(
                {
                    "name": f"tester{suffix}",
                    "email": f"tester{suffix}@test.com",
```

arches/management/commands/add_users.py
```diff
@@ -0,0 +1,186 @@
...
+            type=int,
+            help="The number of test users to add.  Not used for csv loading",
+        )
+
+        parser.add_argument(
+            "-p",
+            "--power-user",
+            action=argparse.BooleanOptionalAction,
+            help="Whether or not to create a power user",
+        )
```
chiatt, line 76: Could you add an example of the command and a description of what a Power User is in this PR? Then we could pass that along to @ekansa to make sure it gets added to the documentation.

chiatt (changes requested):

arches/management/commands/add_users.py
```diff
@@ -0,0 +1,207 @@
...
+            dest="operation",
+            choices=[
+                "test_users",
+                "csv_users",
+            ],
+            default="test_users",
+            help="Operation Type; "
+            + "'test_users'=Creates test users"
+            + "'csv_users'=Loads users from CSV file",
+        )
```
chiatt, line 51: Because 'test_users' as the default, it's possible to run the command like this:
```python manage.py add_users -s path/to/my.csv```

That will add three test users, but nothing from the csv, even though the csv was the intended source of the new users.

Removing this argument and just relying on the existence of the -s flag to determine the source of users might be a way to address this.

chiatt (approved):
Looks good to me

PR #12424, opened 2025-09-05
Fix test tile value

jacobtylerwalls:
:wave: Was running the arches test suite against some pending Django PRs and before remembering to uninstall arches-querysets, it choked on this invalid test data.


aarongundel (approved):
Thanks, Jacob!  LGTM.

Issue #8715, opened 2022-07-30
Move the default value from cards_x_nodes_x_widgets to the node


chiatt:
Currently the default value config is stored in the widget record. When a node is assigned a widget, a card_x_node_x_widget record is created that stores the default value. This pattern was practical when datatype classes did not exist and the only way to create new data (other than import) was to use the resource editor where the card_x_node_x_widget record is always available. Today, however, this pattern is problematic when trying to import data through the API or the database. In those cases a developer typically has easy access to the node, but has to do an additional work to get the default value from the card_x_node_x_widget record. Additionally, it makes the Arches data model clearer to store configs concerning data closer to the node/datatype and leave UI configs closer to the widgets and cards. 

For these reasons the config for the default value should be moved to the datatype and the default value should be saved to the node. 

To accomplish this we will need to:

1. Migrate the existing default value config to a new column on the datatype.
2. Modify the django view in which we save widget configs to save the default value to the node, rather than the card_x_nodes_x_widgets. Note - the user interface should not need to change. The best way to create default values is still the card manager because in the card manager users can create default values using a corresponding value for a datatype.
3. Ensure any places where we draw the default value from cards_x_nodes_x_widgets are updated to use the node's default value.

jacobtylerwalls:
Consider running datatype validation methods on incoming proposed default values to ensure they are valid (e.g. `""` is not saved as the default value for the number datatype, as we have currently).

PR #12437, opened 2025-09-11
Feat: Widgets configs through workflows

StuCM:
### Types of changes

### Description of Change

This PR adds the ability to provide configuration for individual widgets within a workflow by passing config options for the workflow directly in the workflow file.
This allows for different options to be applied to a node/widget in a workflow without editing the graph and affecting various workflows that may not need the applied config. For example disabling input or making adjustments to the label or placeholder string.
It modifies the card-component.js to take the input config and process this to overwrite the matching params that are passed to the widget on render. 
This is done through a series of handlers to create the correct data type depending on the param that is being overwritten.

An app has been setup with a workflow and models to test the functionality.

https://github.com/flaxandteal/arches_example.git

Follow the install instructions for the app and then in the `create-movie-workflow.js` some configuration options have been set for a node that can be adjusted.

The idea is that you add a `nodeOptions` param. Within that you set a key as the node to adjust and then you can overwrite the configs within that.

```
nodeOptions: {
    '6d5fc3d0-44a5-4f20-80b8-86bb42818eee': {
        disabled: true,
        allowInstanceCreation: false,
        widget: {
            visible: true
        },
        config: {
            label: "I have changed!",
            placeholder: "This is my new placeholder"
        },
        node: {
            isrequired: true
        }
    }
}
```

### Issues Solved

This is a new feature that we have been using to allow us to use the same node and apply different configurations to the widget dependent upon the workflow being used.

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches/app/media/js/viewmodels/card-component.js
```diff
@@ -316,5 +316,76 @@ export default function(params) {
         }
     };
 
+    this.getNodeOptions = (nodeId, widgetConfig) => {
+        const options = params.nodeOptions?.[nodeId] || {};
+        const nodeOptions = {};
+        
+        if (!options) return;
```
chiatt: I think you can remove this line because it doesn't look like `return` is ever reachable. In line 320, options is always a value or an object literal which isn't falsy.

chiatt (changes requested):
This is a very helpful feature, but I noticed an issue with required nodes. When I changed a required node to `isrequired: false`, it appeared in the card without the asterisk I'd expect, but when I saved the step, I got an error.

image

arches/app/media/js/viewmodels/card-component.js
StuCM: Just looking into this tile validation error and I can see this is happening as this is only editing outputs on the front end while a the model is required is being pulled from the data base. I can't see a way around this without modifying the save functions on the tile and trying to pass the override through the save request. Would this be valid to make adjustments to the tile model? Otherwise an option would be to remove node configuration and only handle aspects on the widget

chiatt: I don't think we should change the tile model to support overriding required nodes. If the node is required, the system should be able to rely on a value being there. In the future, [we should move the default value to the node config](https://github.com/archesproject/arches/issues/8715). Additionally we should probably require a default value for all required nodes (as Django does).

StuCM: I have removed the null check and removed the node handler to stop the conflict with the core node logic

chrabyrd (approved):
this LGTM!

chiatt (changes requested):
This looks good @StuCM. If you could add an entry in the changelog in `8.1.0.md` and we'll get this merged in.

chiatt (approved):

PR #12519, opened 2025-12-01
Adding a normalizer to the Elasticsearch "displayname" property

### LINKED ISSUE #11540: Sorting search results with accented characters doesn't return results in the "correct" order
See the issue described here: https://community.archesproject.org/t/sorting-search-results-with-accented-characters/2553/

For example, records that have values for the field used for Display Name that start with special characters (ex. “É”). end up being sorted sorted below everything else.

jkemper-pca:
This is a suggested change to the Elasticsearch mapping configuration to add an "ascii-folding" normalizer to the displayname property. This changes the alphabetical search results to display with accented characters treated as their "base" ascii character for sorting purposes.

### Types of changes

### Description of Change
The displayname field can contain many different values which can include "special" non-ascii characters. For sorting, these should be with the rest of the same "base" character. This PR was suggested based on feedback from my [original forum post](https://community.archesproject.org/t/sorting-search-results-with-accented-characters/2553) asking about this behaviour.

To compare, create any list of resources with at least one resource that use a display name that starts with a special characters (ex. "é") and then some that don't. Then on the search results page, sort the results by "Resource Name". On the target branch, this will list the record with the special character at the end after all the resources that start with an ascii character. With this change the resources with special characters will be sorted with their "base" characters.

Ex.
image

### Issues Solved

Closes [#11540](https://github.com/archesproject/arches/issues/11540)

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: Parks Canada Agency 
*   Found by: @jkemper-pca  
*   Tested by: @jkemper-pca  
*   Designed by: @jkemper-pca

### DIFF
--- a/arches/app/search/mappings.py
+++ b/arches/app/search/mappings.py
@@ -31,7 +31,14 @@ RESOURCES_INDEX = "resources"
 ANALYZER = {
     "analyzer": {
         "folding": {"tokenizer": "whitespace", "filter": ["lowercase", "asciifolding"]}
-    }
+    },
+    "normalizer": {
+        "folding_normalizer": {
+            "type": "custom",
+            "char_filter": [],
+            "filter": ["lowercase", "asciifolding"],
+        }
+    },
 }
 
 
@@ -154,7 +161,10 @@ def prepare_search_index(create=False):
                 "displayname": {
                     "type": "nested",
                     "properties": {
-                        "value": {"type": "keyword"},
+                        "value": {
+                            "type": "keyword",
+                            "normalizer": "folding_normalizer",
+                        },
                         "language": {"type": "keyword"},
                     },
                 },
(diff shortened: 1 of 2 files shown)


chiatt (changes requested):
Could you add a changelog entry in `arches/releases/8.0.8.md`? You can just follow the pattern used under the `### Bug Fixes and Enhancements` heading in `arches/releases/8.0.6.md`.

apeters (approved):
this looks good to me.  One question though, shouldn't this be targeting dev/8.1.x?

chiatt:
> this looks good to me. One question though, shouldn't this be targeting dev/8.1.x?

I have no objections to targeting 8.1. However, I saw this as a minor bug fix because if a user sorts alphabetically and does not see accented display names, they may think that their resource isn't in the system. At the same time, the issue doesn't seem serious enough to warrant a change to the LTS, so 8.0.x seemed like a good spot.

There is the question of whether a bug fix that requires indexing should be merged into a patch. Generally we wouldn't do that. However, this change doesn't necessarily require reindexing. It's only needed to see the fix. 

On the other hand, it might not be entirely prudent to introduce changes to the mapping file in a patch. If that's a concern at all, then I agree, it really should land in 8.1.

jkemper-pca:
I agree that dev/8.1.x is the better target for the PR because it does require a reindex of the database. I will get that changed.

chiatt:
Thanks for rebasing @jkemper-pca. Could you add the changelog entry here: https://github.com/archesproject/arches/blob/dev/8.1.x/releases/8.1.0.md#additional-highlights

It's always best to add it somewhere in the middle of the existing list to avoid merge conflicts.

chiatt (approved):

PR #12556, opened 2026-01-14
Fixes issue where term search may only return one result of many matching terms #12555

### LINKED ISSUE #12555: Term Search - Term Search returns incorrect or truncated results with spurious Nodegroup attribution.
**Arches Version/Branch:** stable/7.6.17

**Describe the Bug and how to reproduce it:**
When doing a term search with a multiple result set, only a single result is returned showing incorrect resource model label.

Image

Image

Image

khodgkinson-he:
### Types of changes

### Description of Change
This fixes an issue that occurs when a search term is matched multiple times, such as if there are multiple instances of a name across resources, it would only return the first match (whilst ascribing it the model label of the last results nodegroup), and may link to a different match. This fix ensures that all term matches are returned and are correctly linked.

### Issues Solved

Closes #12555 

### Checklist

    - [X] dev/7.6.x (extended support): major security issues, data loss issues
    
- [X] I added a changelog in arches/releases
- [ ] I submitted a PR to arches-docs (if appropriate)
- [ ] I added tests that prove my fix is effective or that my feature works
- [ ] My test fails on the target branch

### DIFF
--- a/arches/app/views/search.py
+++ b/arches/app/views/search.py
@@ -153,7 +153,9 @@ def search_terms(request):
     i = 0
     ret = {"terms": [], "concepts": []}
     if len(permitted_nodegroups) == 0:
-        return JSONResponse(ret)
+        return JSONResponse(ret)    
+    nodegroup_ids = set()
+    queries = {}
     for index in list(ret.keys()):
         query = Query(se, start=0, limit=0)
         boolquery = Bool()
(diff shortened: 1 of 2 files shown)


khodgkinson-he:
Despite the nature of the change, this seems to fail the python coverage test. Can this be safely ignored for this code correction? ^_^

arches/app/views/search.py
```diff
@@ -236,33 +236,27 @@ def search_terms(request):
...
-                            "value": result["key"],
-                            "nodegroupid": result["nodegroupid"]["buckets"][0]["key"],
-                        }
-                    )
-                    i = i + 1
+                    for nodegroup in result["nodegroupid"]["buckets"]:
+                        nodegroup_id = nodegroup["key"]
+                        node = Node.objects.get(nodeid=nodegroup_id)
+                        graph = node.graph
+                        context_label = "{0} - {1}".format(graph.name, node.name)
```
chiatt: It might be worth building a lookup dictionary of node and graph names with the nodegroup_id as the key so that you can move this query outside of the for loop.

khodgkinson-he: Single hit on the database sir? I'm convinced!  ^_^

arches/app/views/search.py
```diff
@@ -214,7 +216,24 @@ def search_terms(request):
...
+            for result in results["aggregations"]["value_agg"]["buckets"]:
+                if len(result["top_concept"]["buckets"]) == 0:
+                    for nodegroup in result["nodegroupid"]["buckets"]:
+                        nodegroup_ids.add(nodegroup["key"])
+
+    nodes = Node.objects.filter(nodeid__in=nodegroup_ids).select_related("graph")
+    node_lookup = {str(node.nodeid): (node.graph.name, node.name) for node in nodes}
+
+    i = 0
+    for index in list(ret.keys()):
```
chiatt: I think ret.keys() returns an iterable, so no need to cast to a list (I see, though, that somebody else used that syntax on 159).  

```suggestion
    for index in ret.keys():
```

chiatt (changes requested):
This looks great and works like a charm. Just one little suggested change and we can get this merged in.

chiatt (approved):
Looks great to me!

khodgkinson-he:
..thanks again for all your help, sir!  ^_^

Issue #12542, opened 2026-01-02
File List datatype `activateMax` field not configurable


johnatawnclementawn:
In the graph designer, interaction with the `activateMax` config option is unimplemented when configuring a new file list node. This means that even if a graph designer limits the number of files that can be saved for that node, the validation check [here](https://github.com/archesproject/arches/blob/28c74cb54fe9396f944c6cf19a2313d8724954a4/arches/app/datatypes/datatypes.py#L1179) will not be enforced. Since this param has not been implemented and I think most graph designers would just expect that supplying a max file count would enforce that limit, we should swap that part of the conditional for a check of truthy `maxFiles`.

PR #12558, opened 2026-01-16
Add language datatype #12557

### LINKED ISSUE #12557: Add a Language Datatype
Several projects and applications, notably Arches Lingo, have modeled or would like to model their data with a Language node as a sibling of other data collecting nodes in a nodegroup, rather than relying on the internationalization functionality currently in the String datatype. To date, Lingo has been using Arches Controlled Lists to define a controlled list of Languages that can be used for such a purpose, using URI's from the Getty's AAT. While the Library of Congress _does_ maintain ISO639 Language Codes that have URI's ([1. minimal two letter codes](https://id.loc.gov/vocabulary/iso639-1.html), [2. three letter codes](https://id.loc.gov/vocabulary/iso639-2.html), and [3. comprehensive 3 letter codes](https://id.loc.gov/vocabulary/iso639-3.html), these **do not** include subtags to indicate locale (i.e. British English vs American English), with these region codes being defined in [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2). The IANA maintains a [language subtag registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry), the issue being that these entities do not have URI's. Others have run into this issue [elsewhere](https://github.com/agroportal/project-management/issues/507), and there is an open request to [assign URI's for these subtags](https://github.com/w3c/i18n-discuss/issues/13). 

Suffice it to say, we have determined that:
1. We cannot rely on a controlled vocabulary of languages because of the requirement for URIs in a vocabulary
2. We ought to have one source of truth for languages - the Languages table
3. We will need to be able to make reference to the records in that table in the form of node values on a tile

Other references:
- [Arches Forum discussion](https://community.archesproject.org/t/modelling-and-displaying-multilingual-data-at-the-same-time-in-arches/2958)
- [W3C Internationalization Docs: Language tags in HTML & XML](https://www.w3.org/International/articles/language-tags/index.en.html)

johnatawnclementawn:
### Types of changes

### Description of Change

Add the Language Datatype

### Issues Solved

Closes #12557

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: GCI 
*   Found by: @johnatawnclementawn  
*   Tested by: @chrabyrd  


johnatawnclementawn:
Items for follow up / discussion:
- I think we need a better way of adding new languages to the Languages table and validating language tags & subtags
    - AFAIK, there are two paths for adding new languages:
        1.  Follow [localization instructions](https://arches.readthedocs.io/en/stable/developing/advanced/localizing-arches/) and [synchronize Languages table with languages in settings.py](https://github.com/archesproject/arches/blob/1386ccccb0cdc61abd98d7b49c5921ef61956f12/arches/app/utils/i18n.py#L327-L351).
        2. During an import of RDF XML into the RDM (or Arches Controlled Lists if you're using it), if a language code is encountered that is not in the database, [it is added provided the language info can be returned from the django translations util](https://github.com/archesproject/arches/blob/1386ccccb0cdc61abd98d7b49c5921ef61956f12/arches/app/utils/skos.py#L463-L486) (something I think we should get away from)
    - There are [python libraries](https://pypi.org/search/?q=language+subtag) that have already parsed the [IANA subtag registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry) and provide tools for validating subtags, so my initial idea is that we use the [library](https://pypi.org/project/iana-bcp47/) with the most recent updates and develop a method on the Language model that validates language tags (and region subtags if needed). This could then be used in the two methods referenced above as well as a management command for inserting new languages.
        - Something to consider is that a valid code != an usable code as mentioned in the [IANA-BCP47 readme](https://github.com/Masterain98/iana-bcp47?tab=readme-ov-file#usage) and [W3C's language tags article](https://www.w3.org/International/articles/language-tags/index.en.html#region).
- TODO: PR to docs is warranted, but I think getting Lingo configured with this new datatype & adding a new widget to Component Lab probably have priority over that. 
- [Follow up TODO](https://github.com/archesproject/arches/pull/12558/changes#diff-8f09e23698f4ac99abe8e670f189a7682e39895b5caaa4908b42b0737ff18b22R2636-R2637): add a method during RDF export that uses this node value as the language code for literals derived from sibling node values in a tile

arches/app/media/js/views/components/widgets/language.js
```diff
@@ -0,0 +1,17 @@
...
+import LanguageSelectViewModel from "viewmodels/language-select";
+import selectTemplate from "templates/views/components/widgets/select.htm";
+
+const viewModel = function(params) {
+    params.configKeys = ["defaultValue"];
+    
+    LanguageSelectViewModel.apply(this, [params]);
+    const self = this;
+
+    self.multiple = false; // TODO: allow multiselect?
```
chrabyrd: ehhhh YAGNI. I'd say pull this line entirely til we need it

arches/app/datatypes/datatypes.py
```diff
@@ -2592,3 +2593,81 @@ def default_es_mapping(self):
             }
         }
         return mapping
+
+
+class LanguageDataType(BaseDataType):
+    def __init__(self, model=None):
+        super().__init__()
```
chrabyrd: need to pass model into init I think

arches/app/datatypes/datatypes.py
```diff
@@ -2592,3 +2593,81 @@ def default_es_mapping(self):
...
+            if operation == "null" or operation == "not_null":
+                self.append_null_search_filters(value, node, query, request)
+            elif value["val"] != "":
+                field = f"tiles.data.{str(node.pk)}"
+                match_query = Term(field=field, term=value["val"])
+                (
+                    query.must(match_query)
+                    if "!" not in operation
+                    else query.must_not(match_query)
+                )
```
chrabyrd: nit: i'd say just break this out instead of a multi-line ternary

arches/app/media/js/viewmodels/language-select.js
```diff
@@ -0,0 +1,56 @@
+import ko from "knockout";
+import arches from "arches";
+import WidgetViewModel from "viewmodels/widget";
+
+let LANGUAGE_LOOKUP = {};
```
chrabyrd, line 5: ahhh global state. is this to handle multiple widgets per page? If so I'm not sure this'll work as-written -- without a locked/fetching state every widget that loads before the initial call returns will make the same call

arches/app/media/js/viewmodels/language-select.js
```diff
@@ -0,0 +1,56 @@
...
+    self.options = ko.observableArray();
+
+    self.displayValue = ko.computed(() => {
+        const selectedOption = self.options().find(
+            (option) => option.id === self.value()
+        );
+        return selectedOption ? selectedOption.text : "";
+    });
+
+    async function fetchLanguages() {
```
chrabyrd, line 21: needs error handling?

arches/app/models/migrations/12557_add_language_datatype.py
```diff
@@ -0,0 +1,50 @@
+# Generated by Django 5.2.3 on 2026-01-13 20:34
+
+from django.db import migrations
+from arches.app.models.models import DDataType, Widget
```
chrabyrd: Wow, I'm _truly_ surprised Django imported models into a migration this way 🤔 

Typically it's done via `apps.get_model`

chrabyrd (changes requested):
overall looks good, just some minor nits really. Good job 👍

arches/app/media/js/viewmodels/language-select.js
johnatawnclementawn: Actually I added this because it was fetching x2 when the display was instantiated and then again x2 when the 'editor' was instantiated, but for the same node. FWIW, we do the same [sort of thing for concepts](https://github.com/archesproject/arches/blob/360eca0f92bf684ec35ccd46f4734edcbdc14607/arches/app/media/js/viewmodels/concept-select.js#L6). I agree, I don't _love_ the pattern, but as Rob said this is just technical debt that we need to carry until full vue cutover.

arches/app/models/migrations/12557_add_language_datatype.py
johnatawnclementawn: whoops - that was me... and it worked for the file-list migration that I did for #12542 - I fixed both in the latest

arches/app/media/js/views/components/widgets/language.js
johnatawnclementawn: It's a required param for our [select template](https://github.com/archesproject/arches/blob/360eca0f92bf684ec35ccd46f4734edcbdc14607/arches/app/templates/views/components/widgets/select.htm#L20)

johnatawnclementawn:
A note about this PR and the [8.1 feature freeze](https://community.archesproject.org/t/version-8-1-feature-freeze/3071): this PR is a requirement for Lingo's v1 release in March and this feature has a very low impact on the Arches codebase as a whole and is not likely to affect existing projects. If anyone has objections to this being merged (after adequate review) after the feature freeze, please detail your reasoning.

chiatt:
> A note about this PR and the [8.1 feature freeze](https://community.archesproject.org/t/version-8-1-feature-freeze/3071): this PR is a requirement for Lingo's v1 release in March and this feature has a very low impact on the Arches codebase as a whole and is not likely to affect existing projects. If anyone has objections to this being merged (after adequate review) after the feature freeze, please detail your reasoning.

Thanks @johnatawnclementawn. I think the very low risk of this change combined with the clear benefit to Lingo (and possibly other projects as well) is a good justification for this to go into 8.1.

arches/app/media/js/viewmodels/language-select.js
chrabyrd: sure, fair enough 👍

chrabyrd (approved):
LGTM 👍

Issue #12513, opened 2025-11-18
Report node names not localizing


yshng:
Arches v8.0.5

When viewing a resource report `/report/`, node names are not localizing. 

<img width="720" height="365" alt="Image" src="https://github.com/user-attachments/assets/0629540d-50bb-410a-8730-13f9555c2d5a" />

Compare with `/resource/`

<img width="669" height="247" alt="Image" src="https://github.com/user-attachments/assets/69aaba49-44e6-425f-bb3c-aa496f0d3ab3" />

Similarly, everywhere else we would see the localized names of these nodes, the localization shows up correctly. Only reports have this issue.

yshng:
See #12502 for full reproduction recipe

PR #12561, opened 2026-01-21
Fix widget collection #12502, #12513

### LINKED ISSUE #12502: Rich text formatting not displaying in reports
In internationalized Arches 8.0.x, rich text is displaying correctly in the editor and in the edit view, but reports are showing HTML rendered as plain text. 

Image

Image

yshng:
### Types of changes

### Description of Change

Previously resource.py was looking for the non-existent "widgets" in the serialized graph, when it should be looking for "cards_x_nodes_x_widgets." As a result, no card/node/widget information was included in the API response and cards were not pulling the correct widget templates in the case of string fields (see issue #12502), nor including localized labels for widgets (see #12513).  This change gets the correct values so this information is included in the API response.  

### Issues Solved

Closes #12502 and #12513 

### Checklist

#### Ticket Background
*   Found by: @yshng 

### Further comments

No tests found when running `python manage.py test tests --settings="tests.test_settings"`

### DIFF
--- a/arches/test/views/resource_tests.py
+++ b/arches/test/views/resource_tests.py
@@ -1,0 +1,11 @@
+def test_api_resource_report_response(self):
+    """
+    Ensure api_resource_report returns proper response
+
+    """
+    self.client.login(username="admin", password="admin")
+    url = reverse(
+        "api_resource_report", kwargs={"resourceid": self.resource_instance_id}
+    )
+    response = self.client.get(url)
+    self.assertTrue(len(response.json()["cardwidgets"]) > 0)
--- a/arches/app/views/api/resource.py
+++ b/arches/app/views/api/resource.py
@@ -706,11 +706,11 @@ class ResourceReport(APIBase):
             permitted_card_ids = [card.pk for card in permitted_cards]
             cardwidgets = sorted(
                 [
-                    widget
-                    for widget in graph.widgets
-                    if widget["card_id"] in permitted_card_ids
+                    card_node_widget
+                    for card_node_widget in graph.cards_x_nodes_x_widgets
+                    if uuid.UUID(card_node_widget["card_id"]) in permitted_card_ids
                 ],
-                key=lambda widget: widget["sortorder"] or 0,
+                key=lambda card_node_widget: card_node_widget["sortorder"] or 0,
             )
 
             resp["cards"] = permitted_cards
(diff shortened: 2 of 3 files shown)


chiatt (changes requested):
This works great. I think a case could be made that it actually belongs in dev/8.0.x, but I can cherry pick it back into 8.0 later if others agree.

There is a merge conflict, probably in 8.1.0.md. You can sometimes prevent these conflicts if you add your changelog entry in the middle of the list of changes rather than at the end. 

I think a test will also be needed. Something simple like this would be ok: 

```
    def test_api_resource_report_response(self):
        """
        Ensure api_resource_report returns proper response

        """
        self.client.login(username="admin", password="admin")
        url = reverse(
            "api_resource_report", kwargs={"resourceid": self.resource_instance_id}
        )
        response = self.client.get(url)
        self.assertTrue(len(response.json()["cardwidgets"]) > 0)
``` 

You could add that to the `arches/tests/views/resource_tests.py` file.

yshng:
Thanks, @chiatt! Added the test and rebased the branch against dev/8.1.x. Hopefully this resolves the merge conflict as well... Makes sense that this should've been a bug fix on dev/8.0.x, though the intent was to make sure I could reproduce the bug and fix on the latest version before submitting.

chiatt:
> Thanks, @chiatt! Added the test and rebased the branch against dev/8.1.x. Hopefully this resolves the merge conflict as well... Makes sense that this should've been a bug fix on dev/8.0.x, though the intent was to make sure I could reproduce the bug and fix on the latest version before submitting.

Thanks @yshng, looks like there's a formatting issue with the build. Could you run `black .` on your branch? If you haven't already, you might consider installing the pre-commit hooks so this is done for you automatically in future commits.

```
pip install --group dev
pre-commit install
```

You can find more details about this in @jacobtylerwalls's post here: https://community.archesproject.org/t/pre-commit-hooks-for-code-formatting/2355

yshng:
Thanks @chiatt, sorry to not have that already installed. That's my bad. I will make sure my dev env is properly set up before any future PRs. It looks like I'm still not passing one check - I should be much more responsive today/this week, so do let me know if I need to do anything else.

chiatt:
> Thanks @chiatt, sorry to not have that already installed. That's my bad. I will make sure my dev env is properly set up before any future PRs. It looks like I'm still not passing one check - I should be much more responsive today/this week, so do let me know if I need to do anything else.

The test you added covers the modified code, so I think we can disregard that check.

chiatt (approved):
Looks good to me!

arches/test/views/resource_tests.py
```diff
@@ -0,0 +1,11 @@
+def test_api_resource_report_response(self):
```
jacobtylerwalls, line 1: Hey @yshng -- `arches/test` contains utilities for writing tests in arches & arches apps; it's not where tests themselves go. This would explain why your test didn't run. So that we don't lose your work, could you move it to `tests/views/api/test_resources.py`?

jacobtylerwalls: @chiatt also suggested arches/tests/views/resource_tests.py, but all I did was just grep for the `"api_resource_report"` arg to `reverse` to locate similar tests.

yshng:
@jacobtylerwalls Ah, thanks for catching that. What would be the preferred method of remedying this? Should I open a new PR which reverts the previous merge and implements everything at once, or should I submit a fix-up PR?

cc/ @chiatt

jacobtylerwalls:
If it were me, I'd just do the follow-up :)

PR #12567, opened 2026-01-23
Update version to 8.2.0 and add release notes

chiatt:
Update version to 8.2.0 and add release notes

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
         "webpack-bundle-tracker": "^3.1.0"
     },
     "devDependencies": {
-        "arches-dev-dependencies": "archesproject/arches-dev-dependencies#dev/8.1.x"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "arches",
-    "version": "8.1.0",
+    "version": "8.2.0",
```
apeters, line 3: should this be "8.2.0-alpha.0"

chiatt: We haven't worried about about iterating the pre-release here so far. If we were to release a bundle on npm, then I think we'd want to do that for sure, but because the Arches front-end is included in the python package which has the pre-release version we don't really have to worry about it here. I'd rather minimize the steps needed to do pre and patch releases, but once we move to automated releases for core, iterating the pre-release version wouldn't be any trouble, so we could consider doing that if it's helpful for people.

apeters (approved):
lgtm

PR #12569, opened 2026-01-23
adds resource identifiers api #12568

### LINKED ISSUE #12568: need basic resource identifiers API

chrabyrd:
### Types of changes

### Description of Change

Adds basic API for getting/setting resource identifiers. 

### Issues Solved

Closes #12568 

### Checklist

### Further comments

Ok so this is after feature freeze. I believe that although this did not meet the cutoff, it still deserves to be merged in. My reasoning is that:

1. This is additive, it does not affect functionality outside of the new feature.
2. We added the `ResourceIdentifier` class in 8.1, but not a way to interact with it. This allows users to interact with the new class.


arches/app/views/api/resource.py
```diff
@@ -476,6 +476,65 @@ def delete(self, request, resourceid, slug=None, graphid=None):
         return JSONResponse(status=200)
 
 
+@method_decorator(csrf_exempt, name="dispatch")
+class ResourceIdentifiers(APIBase):
+    def get(self, request, resourceid):
+        try:
+            resource_instance = models.ResourceInstance.objects.get(pk=resourceid)
+        except Exception as e:
+            return JSONErrorResponse(str(e), status=404)
```
chiatt: I think it's nice to be explicit about the error type when possible and for a little security, to avoid sending the Django error back in the response. Maybe something like this? https://github.com/archesproject/arches/blob/360eca0f92bf684ec35ccd46f4734edcbdc14607/arches/app/views/api/spatialview.py#L81-L86

arches/app/views/api/resource.py
```diff
@@ -476,6 +476,65 @@ def delete(self, request, resourceid, slug=None, graphid=None):
...
+
+        if not user_can_edit_resource(user=request.user, resource=resource_instance):
+            return JSONErrorResponse(
+                _("Request Failed"), _("Permission Denied"), status=403
+            )
+
+        try:
+            payload = JSONDeserializer().deserialize(request.body)
+        except Exception as e:
+            return JSONErrorResponse(str(e), status=400)
```
chiatt: Maybe a generic Exception is ok here because it's hard to say what could go wrong in the deserializer, but better to write a custom string than to pass the `e` back.

chiatt (changes requested):
This looks really good. Just a couple of minor changes.

chiatt (approved):
Looks good to me!

robgaston (approved):
looks good - thanks!

PR #12573, opened 2026-01-29
Fix directory rename notice during startproject

johnatawnclementawn:
### Types of changes

### Description of Change

Fixes notice of directory being renamed pointed out here: https://github.com/archesproject/arches-docs/issues/511

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @tris-ots & @yshng 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/install/arches_admin.py
+++ b/arches/install/arches_admin.py
@@ -131,6 +131,7 @@ class ArchesProjectCommand(TemplateCommand):
                     "If this is not desired, use the --directory option to create "
                     "a directory with the name you want. Consider using a name "
                     "distinct from your project name."
+                    "For more information, see https://github.com/archesproject/arches/issues/12028"
                 )
             )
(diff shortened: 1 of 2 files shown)


chiatt (changes requested):
I think in addition to this change, we should also explain why the folder name was changed so that people don't use the directory flag or rename the folder without understanding our reasoning. I think adding this link would be sufficient: https://github.com/archesproject/arches/issues/12028

I also have some concern about renaming the output directory without confirming whether a directory with that name already exists. I've tested this - and there is no error when it happens.

johnatawnclementawn:
@chiatt, this fix should be good for 8.0 - just the link to the original ticket and making sure the notice is printed

chiatt (approved):
Looks good to me!

PR #12577, opened 2026-02-04
Add better guardrails for directory creation during startproject #12028

### LINKED ISSUE #12622: Arches VERSION no longer accessible in 8.x
Apps like [Querysets](https://github.com/archesproject/arches-querysets/blob/6d178b6ab68184deb48e4b4067ec59ae7ea7efff/arches_querysets/models.py#L15) and Component Lab rely on the tuple version of the Arches version

Removed here: https://github.com/archesproject/arches/pull/12620/changes/ff8075c70693d27f6aed1f984b666d6808161625

johnatawnclementawn:
### Types of changes

### Description of Change

In #12029 and #12059, we took steps to help folks starting new Arches projects with package naming, with the idea that they should be more prepared to publish to PyPI as an Arches Application. As pointed out in this docs issue https://github.com/archesproject/arches-docs/issues/511, the UX here caused some confusion. 

### Issues Solved

Follow up to #12028 & #12059
Also Resolves #12622

### Checklist

#### Ticket Background
*   Sponsored by: GCI 
*   Found by: @tris-ots & @yshng 
*   Tested by: @chiatt  


johnatawnclementawn:
Hey @jacobtylerwalls, I know in #12059 you added the comment about not needing this when we get to Django 6. From my testing, Django doesn't do the same directory renaming that we're doing here, even at version 6. Obviously, it's nice to look out for folks starting a new Arches project that may want to eventually publish to PyPI, but if Django isn't doing this, I think @chiatt would argue that we should take the same path.

jacobtylerwalls:
Hey, thanks for the @ :-)

> Django doesn't do the same directory renaming that we're doing here

Right, it wouldn't rename anything, but my understanding (haven't retested) is that given some dir name, Django 6 will create that requested dir if it doesn't exist.

Before, it bombed out, so we had this code to handle the dir creation. Does that jibe with what you're seeing?

johnatawnclementawn:
> Hey, thanks for the @ :-)

Always nice to chat with ya! 


> Right, it wouldn't rename anything, but my understanding (haven't retested) is that given some dir name, Django 6 will create that requested dir if it doesn't exist.
> 
> Before, it bombed out, so we had this code to handle the dir creation. Does that jibe with what you're seeing?

Yes, that matches what I'm seeing. Cyrus and I were comparing behavior of starting a new Django project vs starting a new Arches project, not necessarily starting a new Arches project that used Django 5.2 vs 6, which is where I got confused. I did some local testing and I'm clear on what's supposed to be happening 👍🏼 

Thanks again!

arches/install/arches_admin.py
```diff
@@ -181,17 +178,33 @@ def command_startproject(args):
     options = vars(args)
     name = options["name"]
     make_directory = False
-    if not options["directory"] and "_" in name:
-        # The user is supposed to create the --directory themselves. But we
-        # should do it for them if the command is the one that invented the
-        # --directory argument.
-        # RemovedInArches81Warning (Django 6 creates target dir automatically)
+
+    options["project_name_kebab_case"] = name.replace("_", "-")
```
chiatt: I don't think it's necessary to add a key to `options`. Looks like you could just use a new variable.

```suggestion
    project_name_kebab_case = name.replace("_", "-")
```

arches/install/arches_admin.py
```diff
@@ -181,17 +178,33 @@ def command_startproject(args):
...
+    directory_name_will_be_changed = name != options["project_name_kebab_case"]
+
+    if not options["directory"] and directory_name_will_be_changed:
+        if not options.get("yes"):
+            response = input(
+                f"The project directory will be renamed from {name} to {options['project_name_kebab_case']}.\n"
```
chiatt:
```suggestion
                f"The project directory will be renamed from {name} to {project_name_kebab_case}.\n"
```

arches/install/arches_admin.py
```diff
@@ -181,17 +178,33 @@ def command_startproject(args):
...
+                "For more information, see https://github.com/archesproject/arches/issues/12028\n"
+                "Continue? (y/N):"
+            )
+            if response.lower() not in ["y", "yes"]:
+                print("Operation cancelled.")
+                sys.exit(0)
+
         make_directory = True
-        options["directory"] = name.replace("_", "-")
+        options["directory"] = options["project_name_kebab_case"]
```
chiatt:
```suggestion
        options["directory"] = project_name_kebab_case
```

arches/install/arches_admin.py
```diff
@@ -181,17 +178,33 @@ def command_startproject(args):
     options = vars(args)
     name = options["name"]
     make_directory = False
-    if not options["directory"] and "_" in name:
-        # The user is supposed to create the --directory themselves. But we
-        # should do it for them if the command is the one that invented the
-        # --directory argument.
-        # RemovedInArches81Warning (Django 6 creates target dir automatically)
+
```
chiatt: A pretty small nit, but you could probably remove this blank line

chiatt (changes requested):
This looks good - just a minor suggestion.

arches/install/arches_admin.py
johnatawnclementawn: `project_name_kebab_case` is first initialized in `command_startproject` but then needed later on in `handle`, where you've added the comment here, so I opted to keep it as a k/v in `options`

chiatt (approved):
lgtm!

tests/commands/test_arches_admin.py
```diff
@@ -0,0 +1,94 @@
...
+        if settings_file:
+            test_environ["DJANGO_SETTINGS_MODULE"] = settings_file
+        elif "DJANGO_SETTINGS_MODULE" in test_environ:
+            del test_environ["DJANGO_SETTINGS_MODULE"]
+
+        python_path = [base_dir, arches_dir, tests_dir]
+        test_environ["PYTHONPATH"] = os.pathsep.join(python_path)
+        test_environ["PYTHONWARNINGS"] = ""
+
+        process = subprocess.run(
```
jacobtylerwalls, line 28: Hi, one finding for you. This test takes about 45 seconds on my silicon mac, as the startproject command runs `npm install`, and there's no way to mock it out, since we're launching a subprocess. Can we call the functionality under test in a less roundabout way?

PR #12579, opened 2026-02-05
12578 - Support posting multiple tiles with file-list datatypes

### LINKED ISSUE #12578: File-list datatype does not allow for multiple tiles to be saved in one POST
With the addition of the new `arches-querysets` functionality it is now possible to POST an entire resource tree to be saved/updated. The current file datatype assumes that only one tile is being saved, and access to the file list is done like this:
```
            files = request.FILES.getlist(
                "file-list_" + nodeid + "_preloaded", []
            ) + request.FILES.getlist("file-list_" + nodeid, [])
```

However if there are 2 instances of the same card being saved at the same time (which should generate 2 tiles), there is no way to distinguish which files should be associated with which tile because the node id for both tiles is the same.

We need a way to add a unique identifier for each tile/file-list combination (and associated request.FILES values) to allow for this.

bferguso:
### Types of changes

### Description of Change
Initial cut of this is for discussion only. This brings knowledge of `arches-querysets` `aliased_data` into core, which shouldn't  be done. Will tidy up or close when a strategy has been developed.

### Issues Solved

Closes #12578 

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -1332,13 +1332,8 @@ class FileListDataType(BaseDataType):
                             except models.File.DoesNotExist:
                                 logger.exception(_("File does not exist"))
 
-            file_list_key = (
-                tile.aliased_data.site_images[0]["file_id"]
-                if hasattr(tile, "aliased_data")
-                and tile.aliased_data.site_images
-                and tile.aliased_data.site_images[0]["file_id"]
-                else "file-list_" + nodeid
-            )
+            # Try to get the files with the nodeid only. NB - this doesn't support saving multiple tiles in one POST
+            file_list_key = "file-list_" + nodeid
             files = request.FILES.getlist(
                 file_list_key + "_preloaded", []
             ) + request.FILES.getlist(
@@ -1346,9 +1341,18 @@ class FileListDataType(BaseDataType):
                 [],
             )
 
-            # files = request.FILES.getlist(
-            #     "file-list_" + nodeid + "_preloaded", []
-            # ) + request.FILES.getlist("file-list_" + nodeid, [])
+            # If they weren't available in the POST with the nodeid, try the tile-scoped file names.
+            # This adds support for saving multiple tiles in one POST
+            if len(files) == 0:
+                # First check to see if the files have been set using the tile ID
+                file_list_key = f"file-list_{tile.tileid}-{nodeid}"
+                files = request.FILES.getlist(
+                    file_list_key + "_preloaded", []
+                ) + request.FILES.getlist(
+                    file_list_key,
+                    [],
+                )
+
             tile_exists = models.TileModel.objects.filter(pk=tile.tileid).exists()
 
             for file_data in files:
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -1332,9 +1332,23 @@ class FileListDataType(BaseDataType):
                             except models.File.DoesNotExist:
                                 logger.exception(_("File does not exist"))
 
+            file_list_key = (
+                tile.aliased_data.site_images[0]["file_id"]
+                if hasattr(tile, "aliased_data")
+                and tile.aliased_data.site_images
+                and tile.aliased_data.site_images[0]["file_id"]
+                else "file-list_" + nodeid
+            )
             files = request.FILES.getlist(
-                "file-list_" + nodeid + "_preloaded", []
-            ) + request.FILES.getlist("file-list_" + nodeid, [])
+                file_list_key + "_preloaded", []
+            ) + request.FILES.getlist(
+                file_list_key,
+                [],
+            )
+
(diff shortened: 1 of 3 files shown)


apeters:
@bferguso arches-querysets already overrides some of the built-in arches datatype methods.  There's already one for files [here](https://github.com/archesproject/arches-querysets/blob/dev/1.1.x/arches_querysets/datatypes/file.py) 
I think we should just copy that updated method there and that way keep arches clean.

bferguso:
@apeters  - I guess my concern with this is if we start just overriding larger amounts of arches core functionality in the ancillary apps then then we're setting ourselves more more brittleness as both of them mature. With a new frontend being developed in v9 some of the existing assumptions may no longer be valid (in this case that only one tile will be updated / added at a time) and the earlier we get in front of those the better off we'll be.

apeters:
> @apeters - I guess my concern with this is if we start just overriding larger amounts of arches core functionality in the ancillary apps then then we're setting ourselves more more brittleness as both of them mature. With a new frontend being developed in v9 some of the existing assumptions may no longer be valid (in this case that only one tile will be updated / added at a time) and the earlier we get in front of those the better off we'll be.

I get that, but I feel like that ship has already sailed.  In fact I think it's probably the other way around.  I think that the querysets/component lab/modular reports apps will eventually become the new standard and that we'll be selectively porting code/developing new patterns in these new apps such that they ultimately represent "core" Arches.  However it happens these things will conceptually "merge" at some point.

I think that having references to TileTree attributes within core Arches when querysets is still very much in it's infancy and not required by core is potentially confusing.  Conversely, these overridden methods in querysets are already pretty well established and simply adding to that doesn't "break" from that.  

These, of course, are just my suggestions.  Happy to hear from others out there on this.

bferguso:
@apeters - This is a more generic approach for supporting multiple file-list tiles to be sent in a single POST. It defaults to the original nodeid-scoped naming convention, but falls back to the tileid / nodeid scoped if that fails.

arches/app/datatypes/datatypes.py
```diff
@@ -1320,9 +1320,27 @@ def post_tile_save(self, tile, nodeid, request):
...
+            ) + request.FILES.getlist(
+                file_list_key,
+                [],
+            )
+
+            # If they weren't available in the POST with the nodeid, try the tile-scoped file names.
+            # This adds support for saving multiple tiles in one POST
+            if len(files) == 0:
+                # First check to see if the files have been set using the tile ID
+                file_list_key = f"file-list_{tile.tileid}-{nodeid}"
```
apeters: what happens when new tiles are submitted and don't have a tileid?

apeters (commented):
I"m not able to produce a PUT request that can exercise this code.  Can you provide some instructions on how to repro this?

arches/app/datatypes/datatypes.py
bferguso: It's already caught above. I chose to use the existing implementation first so it would be a no-op when used by the existing the code base. Lines 1324-1330 implement the existing logic of nodeid-only namespacing.

apeters: So if submitting more than one new tile (w/o tileids) via querysets, then this will just revert back to the original logic which sill won't solve the issue of not being able to associate a tile with a file, correct?

bferguso: Although I haven't done any mods to querysets, the solution for this I think would be to pre-populate the tileid in querysets, and then the file-save logic would work in arches core. We have this working here, using querysets and pre-populating the the tileid even before it hits querysets. I think the approach in my mind is to have the apps (in this case querysets) prepare the data so it can be handled by the base datatype. This would eliminate having to override the whole save function and just focus on preparing the data so it can be consumed properly by the datatype.

Other option would be to pull the functionality to get the files from the request out into its own function so that could be overridden. I've checked that in as an option to consider:

https://github.com/archesproject/arches/pull/12579/changes#diff-8f09e23698f4ac99abe8e670f189a7682e39895b5caaa4908b42b0737ff18b22R1362-R1383

apeters: Yeah, I think breaking that out has no downside.  We should be mindful of populating tileids on the client as the backend logic might rely on the absence of tileids to inform its diffing logic.

apeters (approved):
looks good to me

chiatt (changes requested):
Could you add a changelog entry? The format of the 8.1.0.md file has significantly changed, so you'll probably want to pull and rebase off of dev/8.1.x before committing a change.

bferguso:
> Could you add a changelog entry? The format of the 8.1.0.md file has significantly changed, so you'll probably want to pull and rebase off of dev/8.1.x before committing a change.

Entry added. Sorry for the delay - this PR got lost in the mix.

chiatt (approved):

PR #12581, opened 2026-02-06
Revert unsafe change to 7787 migration #12580

### LINKED ISSUE #12580: Unsafe change made in 7787 migration, can't migrate forward on newer PG versions
In 8.0, I attempted to address a `SyntaxWarning` on Python 3.12 for "invalid escape sequences" by converting a triple-quoted string to a "raw" string that has different escaping rules. (We don't usually touch old migrations, but I was trying to head off future headaches when Python stops parsing strings with invalid escape sequences.)

https://github.com/archesproject/arches/blob/2f87c96be90c3b455c61f2039d5250728fc18289/arches/app/models/migrations/7787_relational_data_model.py#L11-L12

I thought this change was safe, following this blog post: https://adamj.eu/tech/2022/11/04/why-does-python-deprecationwarning-invalid-escape-sequence/

But today I discovered that now that the extra backslashes in this large string got treated as "raw", they made it to Postgres, and on PG 17(*) they crashed with:

```
django.db.utils.ProgrammingError: syntax error at or near "\"
LINE 458: ...                              )::json->\\'geometry\\' as geo...
```

So, we can either:
- audit of all backslashes here
- remove the `r` prefix, so it's no longer a raw string, and only fix the two locations that had invalid escape sequences.

I'll offer a PR for the second option.

(*) this is my theory for why CI isn't failing, not 100% sure though.

jacobtylerwalls:
### Types of changes

### Description of Change
Fix failure to run migrations or setup_db on more recent postgres versions.

### Issues Solved
Closes #12580

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/models/migrations/7787_relational_data_model.py
+++ b/arches/app/models/migrations/7787_relational_data_model.py
@@ -9,7 +9,7 @@ class Migration(migrations.Migration):
 
     operations = [
         migrations.RunSQL(
-            r"""
+            """
             create extension if not exists "unaccent";
 
             create or replace function __arches_slugify(
@@ -31,12 +31,12 @@ class Migration(migrations.Migration):
                 ),
                 -- replaces anything that's not a letter, number, hyphen('-'), or underscore('_') with an underscore('_')
                 "separated" as (
-                    select regexp_replace("value", '[^a-z0-9\\\\-_]+', '_', 'gi') as "value"
+                    select regexp_replace("value", '[^a-z0-9_-]+', '_', 'gi') as "value"
                     from "removed_quotes"
                 ),
                 -- trims hyphens('-') if they exist on the head or tail of the string
                 "trimmed" as (
-                    select regexp_replace(regexp_replace("value", '\-+$', ''), '^\-', '') as "value"
+                    select regexp_replace(regexp_replace("value", '-+$', ''), '^-', '') as "value"
                     from "separated"
                 )
             select "value"
--- a/arches/app/models/migrations/0003_40b4.py
+++ b/arches/app/models/migrations/0003_40b4.py
@@ -188,7 +188,7 @@ class Migration(migrations.Migration):
                     configname = NULL
                 WHERE datatype = 'domain-value-list';
             UPDATE d_data_types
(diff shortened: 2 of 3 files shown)


arches/app/models/migrations/7787_relational_data_model.py
```diff
@@ -31,12 +31,12 @@ class Migration(migrations.Migration):
...
                 -- replaces anything that's not a letter, number, hyphen('-'), or underscore('_') with an underscore('_')
                 "separated" as (
-                    select regexp_replace("value", '[^a-z0-9\\\\-_]+', '_', 'gi') as "value"
+                    select regexp_replace("value", '[^a-z0-9_-]+', '_', 'gi') as "value"
                     from "removed_quotes"
                 ),
                 -- trims hyphens('-') if they exist on the head or tail of the string
                 "trimmed" as (
-                    select regexp_replace(regexp_replace("value", '\-+$', ''), '^\-', '') as "value"
+                    select regexp_replace(regexp_replace("value", '-+$', ''), '^-', '') as "value"
```
jacobtylerwalls, line 39: Disclosure: ChatGPT offered me this fix. It thinks that `-` only needs to be escaped inside character sets, like the first instance here, but that can be worked around by placing it last. Then in the second instance here it doesn't need to be escaped at all.

You can verify the fix by running migrate with `python -Werror::SyntaxWarning manage.py setup_db` to elevate SyntaxWarnings to errors.

Without this tweak, and without the `r` prefix, you should have:
```py
  File "/Users/jwalls/arches/arches/app/models/migrations/7787_relational_data_model.py", line 12
    """
    ^^^
SyntaxError: invalid escape sequence '\-'
```

jacobtylerwalls: TIL about the hyphen at end trick
https://www.tutorialpedia.org/blog/regex-should-hyphens-be-escaped/?utm_source=chatgpt.com

robgaston (approved):
lgtm - thanks for this!

PR #12605, opened 2026-02-20
Allow new resource creation in tile api endpoint

chiatt:
### Types of changes

### Description of Change
Allows creation of new a new tile and resource when a user pushes a new tile for which no resource exists.

### Issues Solved

Closes # 12599

### DIFF
--- a/arches/app/utils/decorators.py
+++ b/arches/app/utils/decorators.py
@@ -163,7 +163,11 @@ def check_tile_permissions(func):
         if request.method == "POST":
             resourceid = request.POST.get("resourceinstanceid", None)
             tileid = request.POST.get("tileid", None) or kwargs.get("tileid")
-            if not resourceid and models.TileModel.objects.filter(pk=tileid).exists():
+            if (
+                not resourceid
+                and tileid
+                and models.TileModel.objects.filter(pk=tileid).exists()
+            ):
                 resourceid = models.TileModel.objects.get(pk=tileid).resourceinstance_id
             if not models.ResourceInstance.objects.filter(pk=resourceid).exists():
                 return func(request, *args, **kwargs)
--- a/arches/app/utils/decorators.py
+++ b/arches/app/utils/decorators.py
@@ -170,7 +170,9 @@ def check_tile_permissions(func):
             else:
                 permitted = user_can_edit_resource(user, resourceid)
         if request.method == "GET":
-            resourceid = request.GET.get("resourceinstanceid", None) or kwargs.get("tileid")
+            resourceid = request.GET.get("resourceinstanceid", None) or kwargs.get(
+                "tileid"
+            )
             permitted = user_can_read_resource(user, resourceid)
         if permitted:
             return func(request, *args, **kwargs)
--- a/arches/app/utils/decorators.py
+++ b/arches/app/utils/decorators.py
@@ -93,7 +93,6 @@ def can_edit_resource_instance(function):
             return function(request, *args, **kwargs)
         else:
             raise PermissionDenied
-        return function(request, *args, **kwargs)
 
     return wrapper
 
--- a/arches/app/utils/decorators.py
+++ b/arches/app/utils/decorators.py
@@ -158,15 +158,20 @@ def user_created_transaction_match(function):
 def check_tile_permissions(func):
     @functools.wraps(func)
     def wrapper(request, *args, **kwargs):
-        resourceid = request.POST.get("resourceinstanceid", None)
         permitted = False
-        if not resourceid:
-            tileid = request.POST.get("tileid", None) or kwargs.get("tileid")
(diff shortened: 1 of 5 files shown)


arches/app/views/tile.py
```diff
@@ -92,6 +93,9 @@ def handle_save_error(
             _(title), _(str(message)), {"message": message, "title": title}
         )
 
+    @method_decorator(
```
aarongundel: "name" isn't used when decorating a method, just a class

arches/app/views/tile.py
```diff
@@ -301,6 +305,9 @@ def post(self, request):
 
         return HttpResponseNotFound()
 
+    @method_decorator(
```
aarongundel: "name" isn't used when decorating a method, just a class

arches/app/utils/decorators.py
```diff
@@ -159,15 +158,22 @@ def wrapper(request, *args, **kwargs):
...
+            tileid = request.POST.get("tileid", None) or kwargs.get("tileid")
+            if not resourceid and models.TileModel.objects.filter(pk=tileid).exists():
+                resourceid = models.TileModel.objects.get(pk=tileid).resourceinstance_id
+            if not models.ResourceInstance.objects.filter(pk=resourceid).exists():
+                return func(request, *args, **kwargs)
+            else:
+                permitted = user_can_edit_resource(user, resourceid)
         if request.method == "GET":
-            permitted = user_can_read_resource(request.user, resourceid)
+            resourceid = request.GET.get("resourceinstanceid", None) or kwargs.get(
```
aarongundel: This seems like a bug with the tileid getting assigned as the fallback of the resourceid.  In the (unlikely but possible) chance of a collision you could unexpectedly return true if there was a collision between tile and resource id.

arches/app/utils/decorators.py
```diff
@@ -159,15 +158,22 @@ def wrapper(request, *args, **kwargs):
...
         permitted = False
-        if not resourceid:
-            tileid = request.POST.get("tileid", None) or kwargs.get("tileid")
-            resourceid = models.TileModel.objects.get(pk=tileid).resourceinstance_id
+        user = request.user
         if request.method == "POST":
-            permitted = user_can_edit_resource(request.user, resourceid)
+            resourceid = request.POST.get("resourceinstanceid", None)
+            tileid = request.POST.get("tileid", None) or kwargs.get("tileid")
+            if not resourceid and models.TileModel.objects.filter(pk=tileid).exists():
```
aarongundel: You could probably just do a single trip with a try/except to catch ObjectDoesNotExist with the .get call instead of two trips - one for exist followed by one for get.

arches/app/views/api.py
```diff
@@ -1701,6 +1701,9 @@ def get(self, request):
 
 @method_decorator(csrf_exempt, name="dispatch")
 @method_decorator(check_tile_permissions, name="dispatch")
```
aarongundel: You might want to swap the order here.  check_tile_permissions is more expensive than group_required.  If group_required runs first and bombs, we can skip the expensive check.

aarongundel (changes requested):
Looks good!  Couple of minor comments.

arches/app/views/api.py
```diff
@@ -1712,13 +1723,17 @@ def post(self, request):
...
         transaction_id = request.POST.get("transaction_id")
 
-        # get node model return error if not found
         try:
             node = models.Node.objects.get(nodeid=nodeid)
         except Exception as e:
             return JSONResponse(e, status=404)
 
-        # check if user has permissions to write to node
+        if resourceid and models.ResourceInstance.filter(pk=resourceid).exists():
```
aarongundel: Shouldn't this be ResourceInstance.objects.filter?

arches/app/views/api.py
```diff
@@ -1712,13 +1723,17 @@ def post(self, request):
...
+            if not user_can_edit_resource(request.user, resourceid):
```
aarongundel: I think this is where this landed, but correct me if I'm wrong, trying to summarize the logic - if the resourceid does not exist in the db yet, this will return None and not None is true so we'll 403 the call.  Which means posting a NodeValue to a nonexistent resource id will 403.

aarongundel (changes requested):
nits

arches/app/views/api.py
chiatt: You shouldn't reach this if the resource does not exist in the database because of the conditional above it - but if you could, this check surprisingly would not return `None`, it would return `True` (which should probably be changed - because it's obviously not intuitive). 
```
(Pdb) user_can_edit_resource(request.user, uuid.uuid1()) 
True
```

aarongundel (approved):
lgtm!

PR #12615, opened 2026-02-26
Iterate version to 8.0.10

chiatt:
Iterates arches version to next patch and moves version from `__init__.py` to `pypackage.toml` to standardize with Arches Applications facilitate automated releases.

### DIFF
--- a/arches/__init__.py
+++ b/arches/__init__.py
@@ -1,5 +1,5 @@
 from __future__ import absolute_import
-from arches.version import get_version
+import importlib.metadata
 
 try:
     # This will make sure the app is always imported when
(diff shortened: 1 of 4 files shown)


aarongundel (approved):
and me too!

arches/__init__.py
```diff
@@ -10,7 +10,4 @@
 except Exception as e:
     pass
 
-# VERSION[3] options = "alpha", "beta", "rc", or "final"
-VERSION = (8, 0, 9, "beta", 0)
```
jacobtylerwalls, line 14: This change makes it difficult to retrieve the correct version number when switching branches.

For example:
```
git checkout dev/8.0.x
pip install -e .
python -c "import arches; print(arches.__version__)"
```
8.0.10
```
git checkout dev/8.1.x
python -c "import arches; print(arches.__version__)"
```
8.0.10 💣

jacobtylerwalls: (Meaning that version checks in arches app code will not act correctly.)

jacobtylerwalls: Never mind, I guess it's just a new paradigm to get used to. I can start maintaining separate virtual envs for each version of Arches and just be careful never to git checkout across a version boundary. 👍

PR #12637, opened 2026-03-08
Avoid queries in LanguageDataType.lookup_language for None

jacobtylerwalls:
### Types of changes

### Description of Change
Noticed while updating arches-querysets to test against Arches 8.1 that there are avoidable queries in LanguageDataType for node values of None.


johnatawnclementawn (changes requested):
Thanks for noticing this @jacobtylerwalls - this looks good on all fronts except could you add a release note please

chiatt:
> Thanks for noticing this @jacobtylerwalls - this looks good on all fronts except could you add a release note please

Maybe a changelog entry is unnecessary in this case because this datatype is a new feature in 8.1 anyway.

johnatawnclementawn (approved):
looks good to me 👍🏼

PR #12644, opened 2026-03-11
Remove upgrade to v8 from updateproject

chiatt:
Remove function to upgrade project to version 8.0.0.


arches/management/commands/updateproject.py
```diff
@@ -18,7 +18,7 @@ class Command(BaseCommand):  # pragma: no cover
 
     def handle(self, *args, **options):
         answer = input(
-            "This operation will upgrade your project to version 8.0\n"
```
aarongundel, line 21: This is interesting.  Do you foresee modifying the updateproject command like this for other releases?  I'm not going to hold it up, but maybe we should start breaking out these methods (upgrade to 8.0, upgrade to 8.1) in their own files that updateproject calls in sequence to bring you up to date.

aarongundel (approved):

arches/management/commands/updateproject.py
chiatt: Yeah - that might be helpful for users that just installed 8.1, but need to upgrade from 7.6. But, we'd probably want to maintain our upgrade instructions in the release notes a bit differently to support that.

PR #12651, opened 2026-03-14
Update release notes with py and js dependency changes

chiatt:
Update release notes with py and js dependency changes


aarongundel (approved):
lgtm!

PR #12695, opened 2026-04-10
Improve resource and tile copying, including adding datatype-specific copy logic

### LINKED ISSUE #8352: Tile data is not properly copied in the Resource 'copy' method
When a user copies a resource instance the tile data is simply [copied verbatim from the source to the copy](https://github.com/archesproject/arches/blob/0273894ec4f4168514693e1bba14cc66e6b6bf25/arches/app/models/resource.py#L583-L584). This creates problems if the source data has values that should be unique to a single tile. This includes file datatype node values which has a fileid and the resource instance datatype which has a relationshipid. 

If a user copies an instance with these datatypes, issues described here can occur: https://github.com/archesproject/arches/issues/7533

To address this a copy method should be added to the tile model which can call a datatype function responsible for safely copying the data of complex datatypes like the file and resource instance datatypes. In the meantime, the copy function should be unavailable to users.

### LINKED ISSUE #10712: Resource.copy() does not work
Not urgent as the button is disabled since #8445, but should be solved with #8352

```py

>>> ri
>>> ri = ResourceInstance(pk='af792d07-e8a7-4f60-929e-591d2dfc4579')
>>> ri.refresh_from_db()
>>> ri.graph
<GraphModel: Instrument>
>>> r = Resource(ri.pk)
>>> r.graph = ri.graph
>>> r.copy()
  File "<console>", line 1, in <module>
  File "/web_root/arches/arches/app/models/resource.py", line 752, in copy
    new_resource.save(context="copy")
  File "/web_root/arches/arches/app/models/resource.py", line 249, in save
    self.index(context)
  File "/web_root/arches/arches/app/models/resource.py", line 337, in index
    document, terms = self.get_documents_to_index(datatype_factory=datatype_factory, node_datatypes=node_datatypes, context=context)
  File "/web_root/arches/arches/app/models/resource.py", line 383, in get_documents_to_index
    context["language"] = lang[0]
TypeError: 'str' object does not support item assignment
```

### LINKED ISSUE #12685: On copy of a resource instance node, a new resource_x_resource record(s) needs to be created
On copy of a resource instance node, a new resource_x_resource record needs to be created rather than simply duplicating the node value. Duplication results in the shared deletion of a resource_x_resource record if either resource's tile is deleted.

johnatawnclementawn:
### Types of changes

### Description of Change

Creates `copy` methods on ResourceInstance and TileModel base models which return unsaved ResourceInstance and TileModel instances, and adjusts the behavior of the Resource proxy model to use the aforementioned logic and run save, indexing, and edit_log entry creation side effects. This also provides a path for datatypes to perform their own logic during `copy`

### Issues Solved

Closes #10712, #8352, #12865

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Tested by: @chiatt / @apeters  

### DIFF
--- a/arches/app/functions/primary_descriptors.py
+++ b/arches/app/functions/primary_descriptors.py
@@ -19,7 +19,9 @@ class AbstractPrimaryDescriptorsFunction(BaseFunction):
         config -- the descriptor config which indicates how and what will define the descriptor
 
         Keyword Arguments:
-        context -- string such as "copy" to indicate conditions under which a resource participates in a function.
+        context -- Dictionary which may have:
+            language -- Language code in which the descriptor should be returned (e.g. 'en').
+            any key:value pairs needed to control the behavior of a custom descriptor function
         descriptor -- type of descriptor, e.g. "name", "map_popup", or "description"
         """
 
@@ -36,16 +38,22 @@ class PrimaryDescriptorsFunction(AbstractPrimaryDescriptorsFunction):
         config -- the descriptor config which indicates how and what will define the descriptor
 
         Keyword Arguments:
-        context -- string such as "copy" to indicate conditions under which a resource participates in a function.
+        context -- Dictionary which may have:
+            language -- Language code in which the descriptor should be returned (e.g. 'en').
+            any key:value pairs needed to control the behavior of a custom descriptor function
         descriptor -- type of descriptor, e.g. "name", "map_popup", or "description"
         """
 
+        if context is None:
+            context = {}
+        elif not isinstance(context, dict):
+            raise TypeError(
+                "Primary descriptor context must be a dict or None, "
+                f"got {type(context).__name__}."
+            )
+
         datatype_factory = None
-        language = (
-            context["language"]
-            if (context is not None and "language" in context)
-            else None
-        )
+        language = context.get("language", None)
         result = config["string_template"]
         updated = False
 
--- a/arches/app/functions/primary_descriptors.py
+++ b/arches/app/functions/primary_descriptors.py
@@ -46,11 +46,6 @@ class PrimaryDescriptorsFunction(AbstractPrimaryDescriptorsFunction):
 
         if context is None:
             context = {}
-        elif not isinstance(context, dict):
-            raise TypeError(
-                "Primary descriptor context must be a dict or None, "
-                f"got {type(context).__name__}."
-            )
 
         datatype_factory = None
         language = context.get("language", None)
--- a/arches/app/models/models.py
+++ b/arches/app/models/models.py
@@ -1,3 +1,4 @@
+import copy
 import datetime
 import json
 import logging
@@ -12,7 +13,7 @@ from django.contrib.gis.db import models
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.serializers.json import DjangoJSONEncoder
 from django.core.validators import RegexValidator, validate_slug
-from django.db import ProgrammingError, connection
+from django.db import ProgrammingError, connection, transaction
 from django.db.models import Case, F, JSONField, Max, Q, Value, When
 from django.db.models.constraints import UniqueConstraint
 from django.db.models.expressions import CombinedExpression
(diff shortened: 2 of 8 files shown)


arches/app/models/models.py
```diff
@@ -1619,6 +1619,40 @@ def save_edit(
         edit.edittype = edit_type
         edit.save()
 
+    @transaction.atomic
+    def copy(self):
```
jacobtylerwalls: Is the override on the proxy model needed now that it's implemented on the base model?

arches/app/models/models.py
```diff
@@ -1619,6 +1619,40 @@ def save_edit(
         edit.edittype = edit_type
         edit.save()
 
+    @transaction.atomic
```
jacobtylerwalls: This method is pretty heavy; I'd suggest holding a smaller transaction over just the writes instead of also over the reads. If you think it's rare that a resource with a huge quantity of tiles would be copied, then I suppose you can ignore this.

arches/app/models/models.py
```diff
@@ -1619,6 +1619,40 @@ def save_edit(
...
+        resourceid = self.resourceinstanceid
+        new_resource = self
+        new_resource.pk = None
+        new_resource._state.adding = True
+        new_resource.save()
+
+        new_tiles = []
+        for tile in TileModel.objects.filter(
+            resourceinstance_id=resourceid
+        ).prefetch_related("parenttile"):
```
jacobtylerwalls: Queries for related instances should start from `self`, e.g. `self.tilemodel_set`; starting from a fresh model manager frustrates natural caching or prefetching.

jacobtylerwalls: I think this is select_related() not prefetch_related() here but see next comment.

arches/app/models/models.py
```diff
@@ -1619,6 +1619,40 @@ def save_edit(
...
+            new_tile = tile
+            new_tile.pk = None
+            new_tile.resourceinstance = new_resource
+            new_tile._state.adding = True
+            new_tiles.append(new_tile)
+            id_map[tile.pk] = new_tile
+
+        for tile in new_tiles:
+            if tile.parenttile:
```
jacobtylerwalls: If you check parenttile_id here then you don't need to fetch the related parent tile row.

tests/models/resource_test.py
```diff
@@ -778,3 +779,34 @@ def test_user_can_see_edit_history_if_resource_editor(self, mock_has_perm):
         mock_has_perm.assert_any_call(
             "read_nodegroup", self.test_resource.tiles[0].nodegroup
         )
+
+
+class ContextDictContractTests(SimpleTestCase):
```
jacobtylerwalls: This looks a little AI-fluffy to me; if you remove the helper then you don't need to test the helper.

arches/app/models/tile.py
```diff
@@ -433,14 +433,24 @@ def datatype_post_save_actions(self, request=None):
             datatype = self.datatype_factory.get_instance(node.datatype)
             datatype.post_tile_save(self, nodeid, request)
 
+    @staticmethod
+    def _ensure_context_dict(context: None | dict) -> dict:
```
jacobtylerwalls: I think these two helpers are an unnecessary abstraction.

arches/app/models/tile.py
```diff
@@ -433,14 +433,24 @@ def datatype_post_save_actions(self, request=None):
...
+
     def save(self, **kwargs):
         request = kwargs.pop("request", None)
         index = kwargs.pop("index", True)
         user = kwargs.pop("user", None)
         new_resource_created = kwargs.pop("new_resource_created", False)
         resource_creation = kwargs.pop("resource_creation", False)
         note = "resource creation" if resource_creation else None
-        context = kwargs.pop("context", None)
+        context = self._ensure_context_dict(kwargs.pop("context", None))
```
jacobtylerwalls: Just provide a fresh dict here (no helper):
```suggestion
        context = kwargs.pop("context", {})
```

tests/models/resource_test.py
jacobtylerwalls: Instead of testing an implementation detail, I'd expect to see a test for the functionality e.g. the snippet from the bug report.

jacobtylerwalls (commented):
Hey @johnatawnclementawn, awesome to see this fixed! I hope you don't mind me adding some advice, let me know if any of this is out of scope for your current purposes :-)

arches/app/models/models.py
```diff
@@ -1619,6 +1619,40 @@ def save_edit(
...
+                tile.parenttile = id_map[tile.parenttile_id]
+
+        TileModel.objects.bulk_create(new_tiles)
```
jacobtylerwalls: Looks like sort order isn't set.

arches/app/models/resource.py
```diff
@@ -143,10 +143,10 @@ def get_descriptor_language(self, context):
         if self.name is None:
             self.name = {}
 
-        requested_language = None
+        if context is None:
+            context = {}
 
-        if context and "language" in context:
-            requested_language = context["language"]
+        requested_language = context.get("language")
```
apeters: could be more simply 
`langauge = context.get("language", get_language())`

arches/app/functions/primary_descriptors.py
```diff
@@ -36,16 +38,22 @@ def get_primary_descriptor_from_nodes(
...
         descriptor -- type of descriptor, e.g. "name", "map_popup", or "description"
         """
 
+        if context is None:
+            context = {}
+        elif not isinstance(context, dict):
+            raise TypeError(
+                "Primary descriptor context must be a dict or None, "
+                f"got {type(context).__name__}."
+            )
```
apeters: No where else, do we actually check if context is a "dict".  The only advantage you get here is a specific error message.  I'd either get rid of the check here and just let it throw a generic "AttributeError" or add checks of the context as you do here in "tile.py" and "resource.py", just for consistencies sake.

apeters (changes requested):

arches/app/models/resource.py
```diff
@@ -1020,36 +1020,38 @@ def get_relations(
...
-            id_map[tile.pk] = new_tile
-
-        for tile in new_resource.tiles:
-            if tile.parenttile:
-                tile.parenttile = id_map[tile.parenttile_id]
-
-        with transaction.atomic():
-            new_resource.save(context="copy")
+        new_resource.tiles = top_level_tiles
+        new_resource.save()
```
apeters: I don't think copy should save the new resource.

apeters (changes requested):

apeters (approved):
lgtm

PR #12700, opened 2026-04-14
Fix SQL call in ResourceInstanceDataType.post_tile_save

### LINKED ISSUE #12610: Resource relation is not removed until tile is deleted
Found in Arches for HERs 1.1 (7.6.x)

When relating a Consultation to an Application Area via the "Related Application Area" widget it shows up correctly in the "Resource Reference By" section (this is fine and expected behaviour). However, when you remove the relationship data from the tile but don't delete the tile this relationship remains in the "Resource Reference By" section. Only when removing the entire tile does the referenced relationship disappear.

The reason I think this is a core issue is because the removed relationship appears in both `response.related_resources.resource_relationships` and `response.related_resources.related_resources`.

See the video below for a visual demonstration of the issue 

https://github.com/user-attachments/assets/c0230402-a185-4ee1-b21b-a6ae37356bba

_Originally posted by @samuelscandrett in https://github.com/archesproject/arches-her/pull/1436#pullrequestreview-3848510038_

bferguso:
### Types of changes

### Description of Change

Fixes SQL call that refreshe the resource_x_resource values for a tile

### Issues Solved

Closes #12610

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @ 

### Further comments
This should probably be applied to dev/7.6.x, dev/8.0.x, dev/8.2.x

On further analysis, the reason the change to `__arches_refresh_tile_resource_relationships` is failing is because that function doesn't apply the default relationships types from the node to the `resource_x_resource` relationship columns. That appears to be a bug in itself, but I also think there is a larger issue as to why we have both the `__arches_refresh_tile_resource_relationships` and the `__arches_create_resource_x_resource_relationships` functions as they are essentially doing the same thing. There was concern noted about the bulk uploader, but the I'm not sure there is any substantive difference between the two functions other than the above difference and the create version short-circuits the delete/insert if the resource-instance value is null.

It in fact appears that the only place the `__arches_create_resource_x_resource_relationships` is used in the entire system is in the one post_tile_save python function, and that everywhere else uses the `__arches_refresh_tile_resource_relationships` flavour. As such I think the right thing to do is:
1. Remove the `__arches_create_resource_x_resource_relationships` function
2. Modify the `__arches_refresh_tile_resource_relationships` function to set the default relationships as the create version was doing.


bferguso:
Trying to understand the difference between the `__arches_create_resource_x_resource_relationships` and the `__arches_refresh_tile_resource_relationships` functions. Making that change seems to do the right thing looking manually at the resource_x_resource table, but it's now failing the CI tests so maybe that's not the right move.

This is the part in the create function that stops it from deleting existing tiles when deleting a relationship:

```
            select count(*) into relational_count from tiles where tiledata::text like '%resourceX%' and tileid = tile_id;
            IF relational_count = '0' 
                THEN 
                RETURN false;
            END IF; 
```

bferguso:
Didn't delete the `__arches_create_resource_x_resource_relationships` yet as we probably don't want to do that in a patch migration?

bferguso:
I'll need to move the changelog entry to 8.1.3 if this looks good otherwise.

releases/8.1.2.md
```diff
@@ -3,6 +3,7 @@
 ## Bug Fixes
 
 - Generate static and media urls for all applications/extensions[#12670](https://github.com/archesproject/arches/pull/12670)
+- Clear relationships when ResourceInstance node value is deleted[#12610](https://github.com/archesproject/arches/issues/12610)
```
chiatt, line 6: Could you move to 8.1.3.md and add an additional entry regarding default relationship types? Both entries can point to the same PR.

chiatt:
> Didn't delete the `__arches_create_resource_x_resource_relationships` yet as we probably don't want to do that in a patch migration?

Yeah, I think we should remove that in 8.2 in case it's being used in a project customization (unlikely as that might be).

arches/app/datatypes/datatypes.py
```diff
@@ -2222,7 +2222,7 @@ def pre_tile_save(self, tile, nodeid):
     def post_tile_save(self, tile, nodeid, request):
         ret = False
         sql = """
-            SELECT * FROM __arches_create_resource_x_resource_relationships('%s') as t;
+            SELECT * FROM __arches_refresh_tile_resource_relationships('%s') as t;
```
chiatt, line 2225: This change could have been cherry-picked back to 7.6, but it's in the same commit as the changelog entry. No change needed here, but just noting that it's nice to keep changelog commits separate because it offers more flexibility.

arches/app/models/sql/functions/__arches_refresh_tile_resource_relationships.sql
```diff
@@ -2,51 +2,85 @@ CREATE OR REPLACE FUNCTION __arches_refresh_tile_resource_relationships(tile_id
...
+                   SELECT graphs->>'ontologyProperty'
+                   FROM jsonb_array_elements(fr.config->'graphs') AS graphs
+                   WHERE graphs->>'graphid' = fr.to_graphid::text
+               ) AS defaultOntologyProperty,
+               (
+                   SELECT graphs->>'inverseOntologyProperty'
+                   FROM jsonb_array_elements(fr.config->'graphs') AS graphs
+                   WHERE graphs->>'graphid' = fr.to_graphid::text
+               ) AS defaultInverseOntologyProperty
+        FROM relationships2 fr
```
chiatt, line 44: It seems like the root of this problem is that the default relationship must not be getting saved to the tile as it should be. It wouldn't be necessary to pull the relationship from the node config if the relationship was saved to the tile as expected. 

Another issue with fixing this here is that the node config stores the relationship type on either the ontologyProperty attribute or the relationshipConcept attribute depending on the source of the relationship. This fix won't address relationshipConcepts which are now the more common flavor of relationship type. Also, I believe that tiles still only store relationship types on the ontology and inverseOntology attributes. I don't recall why that wasn't changed, but in any case it means that if the relationship type always comes from the tile, then we wouldn't need any conditional logic here to handle for both kinds of default relationships types.

chiatt (changes requested):

releases/8.1.2.md
chiatt: Correction - 8.1.3 is released, can you move to 8.1.4.md?

bferguso:
This PR is just being left open as a reminder for when work is done to support `arches-controlled-lists` based relationship types - likely in 8.2.

PR #12706, opened 2026-04-16
Allow datatypes to be handled differently during resource copy

### LINKED ISSUE #8352: Tile data is not properly copied in the Resource 'copy' method

johnatawnclementawn:
### Types of changes

### Description of Change

Adds a `copy` method to the base datatype class to be able to control what happens to individual node values when a resource or tile are cloned.

### Issues Solved

Closes #8352

### Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/datatypes/base.py
+++ b/arches/app/datatypes/base.py
@@ -89,6 +89,15 @@ class BaseDataType(object):
         """
         return value
 
+    def copy(self, value, **kwargs):
+        """
+        Allows for transformation of a value when copying a resource.
+        By default, the value is returned unchanged, but this can be overridden to allow
+        for transformations during copy operations
+        (eg: updating resource to resource records or duplicating files during copy)
+        """
+        return value
+
     def get_bounds(self, tile, node):
         """
         Gets the bounds of a geometry if the datatype is spatial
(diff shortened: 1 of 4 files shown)


apeters (commented):
@johnatawnclementawn I think the major issue I have with this is that the only way to get a complete "copy" of a resource is to save it first.  I'd expect that we'd have a "copy" method on the "tile.py" model that calls "datatype.copy" that is currently found in the "tile.save" method.

What I'd imagined is that a call to "Resource.copy()" would return a copy of that resource and all associated copied tiles that you could subsequently modify and then finally save.

johnatawnclementawn:
Merging this into #12695 for sake of simple review

PR #12710, opened 2026-04-16
Exclude nodegroups from search export that user doesn't have access to

### LINKED ISSUE #12707: Search Export doesn't respect no_access_to_nodegrooup permission
When exporting search results, it appears that all exportable nodes are exported, even those that the current user doesn't have access to.

In our example we have set the `Internal Remark` and `Site Record Admin` to No Access for the Guest group:

Image

But when exporting resources as anonymous (with the Guest group) the nodes under those nodegroups are exported.

bferguso:
### Types of changes

### Description of Change

Restricts the exported nodegroups to those that are visible by user

### Issues Solved

Closes #12707

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: Self Funded
*   Found by: @bferguso
*   Tested by: @bferguso
*   Designed by: @bferguso

### DIFF
--- a/arches/app/search/search_export.py
+++ b/arches/app/search/search_export.py
@@ -32,6 +32,7 @@ import arches.app.utils.zip as zip_utils
 from arches.app.datatypes.datatypes import DataTypeFactory
 from arches.app.models import models
 from arches.app.models.system_settings import settings
+from arches.app.utils.permission_backend import get_nodegroups_by_perm
 from arches.app.utils.betterJSONSerializer import JSONDeserializer
 from arches.app.utils.data_management.resources.exporter import ResourceExporter
 from arches.app.utils.flatten_dict import flatten_dict
@@ -77,10 +78,31 @@ class SearchResultsExporter(object):
             main_card_list[index_number:index_number] = sub_cards_to_add
         return subcards_added
 
-    def return_ordered_header(self, graphid, export_type):
+    def get_headers(self, graph, export_type, fields):
+
+        restricted_ids = set(
+            get_nodegroups_by_perm(self.search_request.user, "no_access_to_nodegroup")
+        )
+        if not settings.EXPORT_DATA_FIELDS_IN_CARD_ORDER or export_type == "tilexl":
+            fields_to_export = fields if type(fields) is list else [fields]
+            if len(fields_to_export) == 1:
+                return list(
+                    graph.node_set.filter(exportable=True)
+                    .exclude(nodegroup_id__in=restricted_ids)
+                    .values_list(*fields, flat=True)
+                )
+            else:
+                return list(
+                    graph.node_set.filter(exportable=True)
+                    .exclude(nodegroup_id__in=restricted_ids)
+                    .values(*fields)
+                )
+
         subcard_list_with_sort = []
-        all_cards = models.CardModel.objects.filter(graph=graphid).select_related(
-            "nodegroup"
+        all_cards = (
+            models.CardModel.objects.filter(graph=graph)
+            .exclude(nodegroup_id__in=restricted_ids)
+            .select_related("nodegroup")
         )
         all_card_list_with_sort = all_cards.exclude(sortorder=None).order_by(
             "sortorder"
@@ -211,14 +233,7 @@ class SearchResultsExporter(object):
                     resource["Link"] = f"{export_namespace}{report_url}"
 
             if format == "geojson":
-                if settings.EXPORT_DATA_FIELDS_IN_CARD_ORDER is True:
-                    headers = self.return_ordered_header(graph_id, "csv")
-                else:
-                    headers = list(
-                        graph.node_set.filter(exportable=True).values_list(
-                            "name", flat=True
-                        )
-                    )
+                headers = self.get_headers(graph, "csv", ["name"])
 
                 if (report_link == "true") and ("Link" not in headers):
                     headers.append("Link")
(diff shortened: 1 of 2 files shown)


arches/app/search/search_export.py
```diff
@@ -77,10 +80,48 @@ def insert_subcard_below_parent_card(
...
+                group__in=self.search_request.user.groups.all(),
+            ).values_list("object_pk", flat=True)
+        ) | set(
+            UserObjectPermission.objects.filter(
+                content_type=ct,
+                permission__codename="no_access_to_nodegroup",
+                user=self.search_request.user,
+            ).values_list("object_pk", flat=True)
+        )
+
```
chiatt, line 97: I think it would be good to take advantage of the `get_nodegroups_by_perm_for_user_or_group` method which will use the nodegroup permission cache. It will also check permissions on the groups a user belongs to.
```
from arches.app.permissions.arches_permission_base import get_nodegroups_by_perm_for_user_or_group

def get_restricted_nodegroups(self):
    result = get_nodegroups_by_perm_for_user_or_group(
        self.search_request.user, "no_access_to_nodegroup"
    )
    return {str(nodegroup.pk) for nodegroup in result if result[nodegroup]}
```

My guess is that you just weren't aware of that method, but you may have had a reason for not using it. If so, it might be nice to union the queries together and then cast to a set. I doubt there is any real performance improvement, but I think it's just a little bit more readable.
```
groups = GroupObjectPermission.objects.filter(
    content_type=ct,
    permission__codename="no_access_to_nodegroup",
    group__in=self.search_request.user.groups.all(),
).values_list("object_pk", flat=True)

users = UserObjectPermission.objects.filter(
    content_type=ct,
    permission__codename="no_access_to_nodegroup",
    user=self.search_request.user,
).values_list("object_pk", flat=True)

return set(groups.union(users))
```

chiatt (changes requested):
This works great and is a nice refactor in general. I have just a small suggestion on getting restricted nodegroup ids.

arches/app/search/search_export.py
bferguso: I originally tried to use that function, but that it isn't implemented as I was expecting. It doesn't seem to iterate through the groups a user has access to, so unless that permission is applied directly to the user (which we're trying not to do) that call permits the user to see the node.

I think long term it would be really nice to have a core function that asks the question "does user x have access to y" which looks at the user and all the user's groups to determine if the "y" thing is accessible to the user, but I didn't want to open the can of worms here.

I've clarified the return value as suggested.

chiatt: I think it really should be checking the group perms for the user. It's using the Guardian permission checker: https://github.com/django-guardian/django-guardian/blob/af363dcc62c5061280499156749c9676a408c56d/guardian/core.py#L176-L183.

chiatt: I'll do a little testing and see what might be going wrong.

chiatt: I did some more testing with the anonymous user and the `get_nodegroups_by_perm_for_user_or_group` is working as expected. By default every nodegroup has explicit 'read' permission guardian record for the anonymous user. If 'no access' is applied to the 'Guest' group on a nodegroup, the anonymous user can still access the nodegroup because user perms trump group perms. 

If we query the Guardian tables for 'no_access_to_nodegroup' for groups and users and union them as done here, that would allow group permissions to trump user permissions and that would be inconsistent with how permissions are applied elsewhere. In other words a user would be able to see tiles in a report, but be unable to edit them.

bferguso:
After the conversation w/ @chiatt yesterday, I did some more digging to determine the scope of the issue.

This issue appears to be localized to the `anonymous` user (which is good!). I did a test by adding the `no_access_to_node_group` to the `Guest` group and added the group to an authenticated user with no explicit permissions set directly on the user. This restricted the nodegroups as expected.

This still opens the potential to expose data to anonymous users that it shouldn't (which is bad), but the workaround for right now is to apply the permissions to the `anonymous` user directly, not to the group.

The "right" thing would be to fix this behaviour in the permissions functions, however there is concern that this may cause undesirable effects in the rest of the system as there is some magic that applies special logic to some Groups rather than using the guardian functionality directly. (For example, the `Resource Editor` group doesn't have any entries in the guardian tables, but somewhere in the codebase it's given RWD permissions.

This discussion also brought up some larger issues that the group came across when implementing the Default Deny model and trying to apply that model for objects other than resources (eg nodegroups).

The current implementation also shed light on the issue that it implements both additive permissions (eg `read access`, `create/update access`, `delete access` and subtractive permissions (eg `no access to nodegroup`) which allows for conflicting permissions that have no logical answer:

```
Group 1 (read on nodegroup x)-------
                                    |-> User 1
Group 2 (no access on nodegroup x) -
```
Should User 1 have read `access` or `no access` on that nodegroup?

In future implementation/cleanup it might make sense to:
1. have _only additive_ or _only subtractive_ permissions for a scheme (eg Default Allow would have _only subtractive_ and Default Deny would have _only additive_. This would remove any potential for contradicting permissions and might help keep the permissions logic clean.
2. Make all permissions explicit (ie get rid of any special logic such as "if `Resource Editor` then" and use the guardian functionality directly. This would apply the guardian permissions to all the Groups/Objects explicitly rather than spreading that logic throughout the code making it easier to understand and maintain.
3. Move to a default deny model as part of the removal of Default Allow to be consistent with the resource-based permission model.

Fixing the above is likely to be a pretty heavy lift and could potentially change the existing behaviour so should probably be done as part of a major release.

For this PR I'll see if I can make the `anonymous` user behaviour align with other users, just for the search export functionality so it doesn't create unexpected changes to the rest of the codebase. If that isn't easily done, perhaps this should be turned into a larger PR to simplify the permissions logic, targeting v9.

chiatt:
> After the conversation w/ @chiatt yesterday, I did some more digging to determine the scope of the issue.
> 
> This issue appears to be localized to the `anonymous` user (which is good!). I did a test by adding the `no_access_to_node_group` to the `Guest` group and added the group to an authenticated user with no explicit permissions set directly on the user. This restricted the nodegroups as expected.
> ...

Thanks for documenting all of this Brett!

chiatt:
Resolved in: https://github.com/archesproject/arches/pull/12843

PR #12713, opened 2026-04-20
Elevate level of django request logger to ERROR in tests

jacobtylerwalls:
### Types of changes

### Description of Change
Running the tests, we're accumulating more tests that aren't capturing warnings for 4xx responses, so if we just elevate the log level a little bit (as we're already doing for warnings from arches itself), then, we can fix this and also stop asserting over the logs so much.


chiatt (approved):
lgtm!

PR #12725, opened 2026-04-25
Move default value to node config

### LINKED ISSUE #8715: Move the default value from cards_x_nodes_x_widgets to the node

chiatt:
### Types of changes

-   [] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change
Moves the default value stored in cards_x_nodes_x_widgets to the node config. 

### Issues Solved

Closes #8715

### Checklist

### DIFF
--- a/arches/app/models/migrations/12587_move_default_value_to_node_config.py
+++ b/arches/app/models/migrations/12587_move_default_value_to_node_config.py
@@ -54,14 +54,19 @@ def _move_default_value_from_nodes_to_widgets(serialized_graph):
 
 def _update_published_graphs(apps, graph_transformer):
     PublishedGraph = apps.get_model("models", "PublishedGraph")
+    batch_size = 25
     graphs_to_update = []
 
-    for published_graph in PublishedGraph.objects.exclude(
-        serialized_graph__isnull=True
-    ):
+    queryset = PublishedGraph.objects.exclude(serialized_graph__isnull=True)
+    for published_graph in queryset.iterator(chunk_size=batch_size):
         serialized_graph = published_graph.serialized_graph
         if graph_transformer(serialized_graph):
             graphs_to_update.append(published_graph)
+            if len(graphs_to_update) == batch_size:
+                PublishedGraph.objects.bulk_update(
+                    graphs_to_update, ["serialized_graph"]
+                )
+                graphs_to_update = []
 
     if graphs_to_update:
         PublishedGraph.objects.bulk_update(graphs_to_update, ["serialized_graph"])
--- a/arches/app/models/migrations/12587_move_default_value_to_node_config.py
+++ b/arches/app/models/migrations/12587_move_default_value_to_node_config.py
@@ -1,0 +1,125 @@
+from django.db import migrations
+
+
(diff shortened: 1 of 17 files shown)


chiatt:
Still need to implement @jacobtylerwalls suggestion here: https://github.com/archesproject/arches/issues/8715#issuecomment-3033192406

aarongundel (changes requested):
mentioned two nits over message - datatype.validate in the forward migration to provide a punchlist of correctable items.  second - potential memory issues with the publication bulk_update.

arches/app/models/migrations/12587_move_default_value_to_node_config.py
```diff
@@ -54,14 +54,19 @@ def _move_default_value_from_nodes_to_widgets(serialized_graph):
...
 
-    for published_graph in PublishedGraph.objects.exclude(
-        serialized_graph__isnull=True
-    ):
+    queryset = PublishedGraph.objects.exclude(serialized_graph__isnull=True)
+    for published_graph in queryset.iterator(chunk_size=batch_size):
         serialized_graph = published_graph.serialized_graph
         if graph_transformer(serialized_graph):
             graphs_to_update.append(published_graph)
+            if len(graphs_to_update) == batch_size:
```
jacobtylerwalls, line 65: What about the last batch that doesn't fill the batch size?

chiatt: Those should be handled by the final bulk_update called further down on line 71

chiatt:
> mentioned two nits over message - datatype.validate in the forward migration to provide a punchlist of correctable items. second - potential memory issues with the publication bulk_update.

Addressed the memory issues with a batch size of 25. For the validation punchlist, I considered doing a check during the migration, but some of the validation commands use model instances that won't use historical versions in the migration. As an alternative I've added a default value validation operation to the validate command that admins can run after the migration to address any default value issues.

aarongundel (approved):
lgtm!

PR #12730, opened 2026-04-28
External domain datatype

johnatawnclementawn:
### Types of changes

### Description of Change

The underlying use case of the Language datatype is to use values from a database table as tile data without having to ETL those data into controlled lists or concepts (and keep them in sync). This PR proposes the External Domain datatype to provide baseline functionality for that use case.

Rather than a concrete datatype, External Domain is an abstract base class. Subclasses implement at minimum, four methods: `get_options_queryset()`, `get_option_id()`, `get_option_display()`, and `get_lookup_filter()` and can override the traditional Arches datatype methods like `validate`, `transform_value_for_tile`, `append_to_document`, and `append_search_filters`.

### Issues Solved

Closes #

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


johnatawnclementawn:
Still open to naming changes... other option I considered `Table-Backed Domain`

arches/app/datatypes/core/__init__.py
chrabyrd, line 1: nit: I'd prefer `abstract` over core but def won't block on this 👍

arches/app/datatypes/core/external_domain.py
```diff
@@ -0,0 +1,118 @@
...
+
+class BaseExternalDomainDataType(BaseDataType):
+    """Base datatype for domain-like values sourced from an external DB table.
+
+    Subclasses define which table/queryset to use, how to extract the stored
+    value (option id) and display value from each record, and how records are
+    looked up by either id or display text.
+    """
+
+    def __init__(self, model=None):
```
chrabyrd, line 18: one thing:

I think a little cleaner pattern is to bulk-load the options upfront in one DB query and be done with it, eg:

```
@cached_property
def _options_index(self):
    index = {}
    for record in self.get_options_queryset():
        index[self.get_option_id(record)] = record
        index[self.get_option_display(record)] = record
    return index

def lookup_record(self, value):
    if isinstance(value, list) and len(value) > 0:
        value = value[0]
    if not value:
        return None
    return self._options_index.get(value)
```

Keeps the DB noise down and makes things a little simpler.

chrabyrd (changes requested):
overall looks great! 2 minor tweaks but this is a nice pattern 👍

PR #12749, opened 2026-05-14
Re-enable resource copy from UI

### LINKED ISSUE #12746: Reinstate the ability to copy a resource from the resource editor menu
there is already comment code that removed that menu item.  We should uncomment that and confirm copying works from there.

johnatawnclementawn:
### Types of changes

### Description of Change

Re-enables the `Copy Resource` button in the resource editor UI

### Issues Solved

Closes #12746

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/views/resource/editor.js
+++ b/arches/app/media/js/views/resource/editor.js
@@ -247,7 +247,6 @@ var vm = {
                         null,
                         function () { }
                     ));
-                    window.open(arches.urls.resource_editor + data.resourceid, '_blank');
                 },
                 error: function () {
                     vm.alert(new AlertViewModel('ep-alert-red', arches.translations.resourceCopyFailed.title, arches.translations.resourceCopyFailed.text, null, function () { }));
--- a/arches/app/media/js/views/resource/editor.js
+++ b/arches/app/media/js/views/resource/editor.js
@@ -243,10 +243,11 @@ var vm = {
                     vm.alert(new AlertViewModel(
                         'ep-alert-blue',
                         arches.translations.resourceCopySuccess.title,
-                        "<a style='color: #fff; font-weight: 700;' target='_blank' href=" + arches.urls.resource_editor + data.resourceid + ">" + arches.translations.resourceCopySuccess.text + "</a>",
+                        "<a style='color: #fff; font-weight: 700;' target='_blank' href='" + arches.urls.resource_editor + data.resourceid + "'>" + arches.translations.resourceCopySuccess.text + "</a>",
                         null,
                         function () { }
                     ));
+                    window.open(arches.urls.resource_editor + data.resourceid, '_blank');
                 },
                 error: function () {
                     vm.alert(new AlertViewModel('ep-alert-red', arches.translations.resourceCopyFailed.title, arches.translations.resourceCopyFailed.text, null, function () { }));
--- a/arches/app/models/resource.py
+++ b/arches/app/models/resource.py
@@ -234,6 +234,7 @@ class Resource(models.ResourceInstance):
         user=None,
         should_update_resource_instance_lifecycle_state=False,
         current_resource_instance_lifecycle_state=None,
+        edit_type="create",
         **kwargs,
     ):
         """
@@ -259,6 +260,8 @@ class Resource(models.ResourceInstance):
             self.principaluser_id = user.id
             kwargs = add_to_update_fields(kwargs, "principaluser_id")
 
+        note = kwargs.pop("note", "")
+
         super(Resource, self).save(
             context=context,
             index=index,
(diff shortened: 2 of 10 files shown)


arches/app/media/js/views/resource/editor.js
```diff
@@ -243,10 +243,11 @@ var vm = {
                     vm.alert(new AlertViewModel(
                         'ep-alert-blue',
                         arches.translations.resourceCopySuccess.title,
-                        "<a style='color: #fff; font-weight: 700;' target='_blank' href=" + arches.urls.resource_editor + data.resourceid + ">" + arches.translations.resourceCopySuccess.text + "</a>",
+                        "<a style='color: #fff; font-weight: 700;' target='_blank' href='" + arches.urls.resource_editor + data.resourceid + "'>" + arches.translations.resourceCopySuccess.text + "</a>",
                         null,
                         function () { }
                     ));
+                    window.open(arches.urls.resource_editor + data.resourceid, '_blank');
```
apeters: I'm not sure we should open a new window with the copied resource, since we pop up a banner on the current resource that says the copy happened and giving a link to it.

apeters (changes requested):

apeters (changes requested):
could we make the edit log entires  for the copy command, links to the copied resource?

arches/app/models/resource.py
```diff
@@ -234,6 +234,7 @@ def save(
         user=None,
         should_update_resource_instance_lifecycle_state=False,
         current_resource_instance_lifecycle_state=None,
+        edit_type=None,
```
apeters: I wonder if this is best expressed as "edit_log_kwargs" which could be an object with "edit_type, note, newvalue, oldvalue" keys?  Not sure if that's the best pattern, though.  What I'm not fond of is hidden kwargs (like "note" for example), but putting all those in the method signature seems like a lot of extra cruft, hence the question about grouping all edit_log pass-through kwargs in a dict.

apeters: One issue with the "edit_log_kwargs" idea is that there still would be kwargs (like "user") that are used in the edit_log.save method that wouldn't be in that dict

arches/app/views/resource.py
```diff
@@ -653,6 +676,27 @@ def get(
...
                     permitted_edits.append(edit)
+
+            # Process copy edits to extract descriptor in requested language
+            language = translation.get_language()
+            for edit in permitted_edits:
+                if edit.edittype == "copy" and edit.newvalue is not None:
+                    if (
+                        "descriptors" in edit.newvalue
+                        and "resourceinstanceid" in edit.newvalue
+                    ):
```
apeters: I think this could be collapsed into 1 "if" statement:

eg:
```
if (
        edit.edittype == "copy"
        and edit.newvalue is not None
        and "descriptors" in edit.newvalue
        and "resourceinstanceid" in edit.newvalue
    ):
```

arches/app/models/resource.py
```diff
@@ -259,6 +260,13 @@ def save(
             self.principaluser_id = user.id
             kwargs = add_to_update_fields(kwargs, "principaluser_id")
 
+        note = kwargs.pop("note", "")
+        newvalue = kwargs.pop("newvalue", None)
+        oldvalue = kwargs.pop("oldvalue", None)
+        edit_type = (
+            "create" if self.createdtime is None and not edit_type else edit_type
+        )
+
```
apeters: can we move this block to just above line 285?

arches/app/models/resource.py
apeters: Can we add the edit log params to the doc string as well?

apeters (changes requested):

arches/app/models/resource.py
chrabyrd: Personally I'd go the other way, make this super explicit. eg:

```
edit_log_type=None
edit_log_note=None
edit_log_oldvalue=None
etc etc
```

Because yeah I agree that leaning on `**kwargs` for this makes the code a bit harder to grok, and I personally don't like grouping these into a dict without it being some sort of `@dataclass` authority that we could use as a source of truth for shape across the app.

apeters: Thanks @chrabyrd for your input.  I'm down with your suggestion.  I'd actually considered the same myself.  Prefixing the params like.  @johnatawnclementawn can you make that change, and also add to the doc string?

arches/app/models/resource.py
johnatawnclementawn: dropped the kwarg pops, but this is needed above the `super.save()` call because the django internals indicate it's not a new instance after that call

apeters (approved):
lgtm

PR #12772, opened 2026-05-27
Fix project template license

chiatt:
Make project template's license AGPL instead of GPL to be consistent with core arches


aarongundel (approved):
lgtm!

PR #12789, opened 2026-06-02
Fix: Front end cardinality n tile error

### LINKED ISSUE #12795: Issue in 8.2 with cardinality n tiles duplicating values
#12725 ("Move default value to node config") changed defaultValue from per-widget config to node.config.defaultValue. In viewmodels/widget.js, that value is wrapped in a ko.observable stored on the shared graph node object, so every widget instance for a node references the same observable.

The pre-existing two-way sync between value and defaultValue is gated by if (!self.form). That guard was safe when defaultValue was per-instance, but it's now a cross-tile conduit: resource-editor display widgets receive no form, so they bind their per-tile value to the one shared defaultValue. Editing any tile writes into it; every sibling's subscription writes it back into their value.

StuCM:
### Types of changes

### Description of Change
When updating tile data on a node with cardinality n, the last tile save would update all the values for the tiles so that they were identical. This was an issue with the default value being shared across the tiles. This was happening on the front end and not directly affecting the database but saving in certain cards would then update data.

The default value is only needed for the graph designer page, a check was added against to check for an attribute only present within the designer so that the default value would not then be shared across tile data.

### Issues Solved

Closes #12795

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/media/js/viewmodels/widget.js
+++ b/arches/app/media/js/viewmodels/widget.js
@@ -126,7 +126,7 @@ var WidgetViewModel = function(params) {
             this.value(defaultValue);
         }
 
-        if (!self.form) {
+        if (params.hasOwnProperty('graphDesignerHasDirtyWidget')) {
             if (ko.isObservable(self.value)) {
                 self.valueSubscription = self.value.subscribe(function(val){
                     if (self.defaultValue() != val) {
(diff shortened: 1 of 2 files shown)


arches/app/media/js/viewmodels/widget.js
```diff
@@ -126,7 +126,7 @@ var WidgetViewModel = function(params) {
             this.value(defaultValue);
         }
 
-        if (!self.form) {
+        if (params.hasOwnProperty('graphDesignerHasDirtyWidget')) {
```
chiatt, line 129: This is okay because we are doing it in other places in this file, but it would be nice (in a future PR) to add a more explicit attribute that we could use throughout the file like `self.inGraphDesigern = params.hasOwnProperty('graphDesignerHasDirtyWidget')`.

chiatt (approved):
lgtm!

PR #12790, opened 2026-06-02
Upgrade rdflib && support Python 3.14

### LINKED ISSUE #12540: Support Python 3.14
Python 3.14 isn't supported by default because rdflib needs an upgrade.  This will be a problem for any devs who upgrade to MacOS 26.2 because system python will upgrade to 3.14 by default - meaning any new projects will (default) to a venv at 3.14.  

Temporary workaround is to install through homebrew and run 

`python312` or `python313` to setup the virtual environment.

johnatawnclementawn:
### Types of changes

### Description of Change

Upgrades rdflib & supports Python 3.14

### Issues Solved

Closes #12540

### Checklist

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -881,12 +881,18 @@ class DateDataType(BaseDataType):
         # type and the number as a numeric literal (as this is how it is in the JSON)
         g = Graph()
         if edge_info["range_tile_data"] is not None:
+            value = edge_info["range_tile_data"]
+            literal_datatype = XSD.dateTime
+            if isinstance(value, str):
+                valid_date_format, valid = self.get_valid_date_format(value)
+                if valid and valid_date_format == "%Y-%m-%d":
+                    literal_datatype = XSD.date
             g.add((edge_info["d_uri"], RDF.type, URIRef(edge.domainnode.ontologyclass)))
             g.add(
                 (
                     edge_info["d_uri"],
                     URIRef(edge.ontologyproperty),
-                    Literal(edge_info["range_tile_data"], datatype=XSD.dateTime),
+                    Literal(value, datatype=literal_datatype),
                 )
             )
         return g
--- a/arches/management/commands/load_ontology.py
+++ b/arches/management/commands/load_ontology.py
@@ -188,7 +188,22 @@ class Command(BaseCommand):
     def add_ontology(
         self, id=None, data_source=None, version=None, name=None, parentontology=None
     ):
-        self.graph.parse(data_source)
+        extension_format_map = {
+            ".rdfs": "xml",
+            ".rdf": "xml",
+            ".owl": "xml",
+            ".ttl": "turtle",
+            ".n3": "n3",
+            ".nt": "nt",
+            ".jsonld": "json-ld",
+        }
+        _, extension = os.path.splitext(data_source)
+        rdf_format = extension_format_map.get(extension.lower())
+        (
+            self.graph.parse(data_source, format=rdf_format)
+            if rdf_format
+            else self.graph.parse(data_source)
+        )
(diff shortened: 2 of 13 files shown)


johnatawnclementawn:
Hey @njkim & @apeters - I'm looking for your opinion on [580b7ae](https://github.com/archesproject/arches/pull/12790/commits/580b7aee87399dde4df964e1841d6065a0c2d2cd) - rdflib 7.6.0 normalizes dates to datetime and since we're [always formatting incoming dates as datetime](https://github.com/archesproject/arches/blob/6ba10be7852259550bd6bca3fad49389fa0c2bda/arches/app/datatypes/datatypes.py#L889) in `to_rdf`, the returned date vals in those tests are datetime, even if the input vals don't originally have the time digits. I'm inclined to say that this is erroneous data mutation and implement [c796b00](https://github.com/archesproject/arches/pull/12790/commits/c796b00cceb47d40278a2ddbfb5fbe494de4f5a1)

arches/management/commands/load_ontology.py
```diff
@@ -188,7 +188,22 @@ def run_loader(
     def add_ontology(
         self, id=None, data_source=None, version=None, name=None, parentontology=None
     ):
-        self.graph.parse(data_source)
+        extension_format_map = {
```
chrabyrd: is this doing anything that [guess_format](https://rdflib.readthedocs.io/en/stable/apidocs/rdflib.util/#rdflib.util.guess_format) can't handle? I'm typically not a fan of hardcoded mappings if they can be avoided

arches/app/datatypes/datatypes.py
```diff
@@ -881,12 +881,18 @@ def to_rdf(self, edge_info, edge):
         # type and the number as a numeric literal (as this is how it is in the JSON)
         g = Graph()
         if edge_info["range_tile_data"] is not None:
+            value = edge_info["range_tile_data"]
+            literal_datatype = XSD.dateTime
+            if isinstance(value, str):
+                valid_date_format, valid = self.get_valid_date_format(value)
+                if valid and valid_date_format == "%Y-%m-%d":
+                    literal_datatype = XSD.date
```
chrabyrd, line 889: is it kosher to have function now emit different datatypes? This affects any consumer that would expect this to return a dateTime regardless of format

arches/management/commands/load_ontology.py
```diff
@@ -188,7 +188,22 @@ def run_loader(
...
+            ".nt": "nt",
+            ".jsonld": "json-ld",
+        }
+        _, extension = os.path.splitext(data_source)
+        rdf_format = extension_format_map.get(extension.lower())
+        (
+            self.graph.parse(data_source, format=rdf_format)
+            if rdf_format
+            else self.graph.parse(data_source)
+        )
```
chrabyrd: nit: in this case I think a simple if/else is more readable than a ternary

chrabyrd (changes requested):
overall looks good! Just a few minor questions

arches/app/datatypes/datatypes.py
johnatawnclementawn: Sniffing for date/dateTime like this keeps the data in the format that they were originally created in. If you're expecting a dateTime, be intentional about including time values, but if we don't determine the format and act accordingly, I think consumers will be more upset that dates `2019-01-11` erroneously have time values appended rather than having more descriptive typed literals. Arches doesn't currently make the distinction between date & dateTime, we just store what's provided by the user, so I think this is the best way to avoid data mutation when exporting to rdf formats.

chrabyrd: yeah, makes sense 👍

chrabyrd (approved):
looks good! 👍

PR #12811, opened 2026-06-22
Fix/12810 map editor geometry error

### LINKED ISSUE #12810: Map Editor widget error on single points
## Summary
The map editor widget does not have a guard against a missing properties key in the feature object. It tries to assign a nodeId that lives within the properties object, if this does not exist, the widget throws an error. 

It seems as if several other functions guard against this by creating an empty properties object. The solution seems to be to add this to the `getDrawFeatures` function as well

StuCM:
### Types of changes

### Description of Change

For single point geometries, we were experience a widget error for the map editor where the widget was not displaying. This was due to an issue within the data of the point, the `properties` key was missing within the `feature`. Although this is an issue with the data, several functions within the map editor guard against this by creating a blank `properties` object. It seemed the `getDrawFeatures` should also implement a similar guard.

### Issues Solved

Closes #12810 

### Checklist

#### Ticket Background
*   Found by: @StuCM 
*   Tested by: @StuCM 

### Further comments

I targeted 8.1 but unsure if this needs to go on into several versions

### DIFF
--- a/arches/app/datatypes/core/geojson_feature_collection.py
+++ b/arches/app/datatypes/core/geojson_feature_collection.py
@@ -78,6 +78,16 @@ class GeojsonFeatureCollectionDataType(BaseDataType):
         if value is not None:
             for feature in value["features"]:
                 try:
+                    feature["properties"]
+                    feature["type"]
+                except KeyError:
+                    message = _("Feature missing properties or type attribute")
+                    title = _("Invalid geojson feature")
+                    error_message = self.create_error_message(
+                        value, source, row_number, message, title
+                    )
+                    errors.append(error_message)
+                try:
                     geom = GEOSGeometry(JSONSerializer().serialize(feature["geometry"]))
                     if geom.valid:
                         validate_geom_bbox(geom)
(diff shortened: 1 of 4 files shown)


chiatt (changes requested):
It seems like this bug is really an issue with our datatype validation because a geojson feature without a properties value is invalid. 

I think we can merge this change, but it should accompany a validation fix so we don't end up masking invalid data. Could you add something like this to the [geojson datatype's validation method](https://github.com/archesproject/arches/blob/30e3fc8ce2ffec14b07371572a76c0e805be9459/arches/app/datatypes/core/geojson_feature_collection.py#L80):

```
try:
    feature["properties"]
    feature["type"]
except KeyError:
    message = _("Feature missing properties or type attribute")
    title = _("Invalid geojson feature")
    error_message = self.create_error_message(
        value, source, row_number, message, title
    )
    errors.append(error_message)
```

StuCM:
Added in this additional validation check and added a test for the validation

chiatt (approved):
lgtm!

PR #12633, opened 2026-03-05
yanks semver in favor of packaging

aarongundel:
### Types of changes

-   [x] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

Fixes bug with 8.1.0a10 release where pre release version parsing caused issues with comparison when greater than 9

### Issues Solved

Closes #

### Checklist

-   I targeted one of these branches:
    - [ ] dev/8.1.x (under development): features, bugfixes not covered below
    - [ ] dev/8.0.x (main support): regressions, crashing bugs, security issues, major bugs in new features
    - [ ] dev/7.6.x (extended support): major security issues, data loss issues
-   [ ] I added a changelog in arches/releases
-   [ ] I submitted a PR to arches-docs (if appropriate)
-   [ ] Unit tests pass locally with my changes
-   [ ] I added tests that prove my fix is effective or that my feature works
-   [ ] My test fails on the target branch

#### Accessibility Checklist

[Developer Guide](https://arches.readthedocs.io/en/stable/developing/advanced/accessibility/)

| Topic            | Changed | Retested |
| ---------------- | ------- | -------- |
| Color contrast   |         |          |
| Form fields      |         |          |
| Headings         |         |          |
| Links            |         |          |
| Keyboard         |         |          |
| Responsive Design|         |          |
| HTML validation  |         |          |
| Screen reader    |         |          |

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

PR #12814, opened 2026-06-25
Restore the VERSION tuple

jacobtylerwalls:
### Types of changes

### Description of Change
Shortly after 1f5f9bb021878f92663ab5d3f0a96473799b6906 we discussed in a standup how restoring the VERSION tuple would simplify comparisons against pre-releases:

```py
>>> # has 8.0 source_identifier field?
>>> Version("8.0.0a0") > Version("8.0")
False
>>> (8, 0, 0, "alpha", 0)  >= (8, 0)
True
```

To make the first flavor work, where the right-hand side is a `Version`, as far as I understand you would need to definitively know how the pre-releases are named ("a", "a0", "alpha", "a1", etc.)

Seems like in harmony with #12633?

This reverts commit 1f5f9bb021878f92663ab5d3f0a96473799b6906.

### Further comments
Unblocks archesproject/arches-querysets#175

### DIFF
--- a/arches/__init__.py
+++ b/arches/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 import importlib.metadata
+from packaging.version import Version
 
 try:
     # This will make sure the app is always imported when
@@ -11,3 +12,16 @@ except Exception as e:
     pass
 
 __version__ = importlib.metadata.version(__package__)
+
+version = Version(__version__)
+major, minor, micro = version.major, version.minor, version.micro
+
+if version.pre:
+    pre_type_mapping = {"a": "alpha", "b": "beta", "rc": "rc"}
+    pre_type = pre_type_mapping.get(version.pre[0], version.pre[0])
+    pre_num = version.pre[1]
+else:
+    pre_type = "final"
+    pre_num = 0
+
+VERSION = (major, minor, micro, pre_type, pre_num)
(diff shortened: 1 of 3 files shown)


chiatt:
The issue that I see is that the following should return `False` because the alpha would be released before 8.0 unless we make the assumption that 8.0 refers to any version of 8.0 forward or backward:
```
>>> (8, 0, 0, "alpha", 0) >= (8, 0)
True
```
This seems more explicit - no assumption needed:
```
>>> Version("8.0.0a0") >= Version("8.0.0a0")
True
```
`Version` is actually pretty forgiving, so the right hand can be expressed in several ways. The following will all return `True`:
```
Version("8.0.0a0") >= Version("8.0.0a")
Version("8.0.0a0") >= Version("8.0.0a0")
Version("8.0.0a0") >= Version("8.0.0alpha")
Version("8.0.0a0") >= Version("8.0.0alpha0")
``` 
I'm not strongly opposed to bringing back the tuple, but it seems nice to have a single approach to doing version comparisons, and relying on `Version` seems more standard.

jacobtylerwalls:
> Version is actually pretty forgiving, so the right hand can be expressed in several ways. The following will all return True:

Ah, that works for me! I was worried about variance on the left-hand side, but I see that using "a" alone on the right-hand side would be enough.

Issue #12681, opened 2026-04-06
File upload creates duplicate "orphan" files and File objects


mradamcox:
I've found an issue with the file upload widget in two Arches 6.2 installations I've been working on. I don't have a later installation to test now right now, but wanted to log the details here for future reference.

I have a file-list node that accepts many files. In the data entry interface, I add these 4 images **one by one**--drop in one file, then click the "add more files" button, add another single file, and so on. The entry looks like this before saving:

<img width="1453" height="346" alt="Image" src="https://github.com/user-attachments/assets/c9009ccb-51cf-4286-bbe2-1f9c9154eb89" />

At that point, my local `uploadedfiles` directory is empty (good!). I click Save and the UI looks good, no errors:

<img width="1435" height="311" alt="Image" src="https://github.com/user-attachments/assets/754bb7cb-488a-477e-86b0-674a79cc0a0a" />

However, in `uploadedfiles` I now have a ton of duplicate files saved.

<img width="707" height="277" alt="Image" src="https://github.com/user-attachments/assets/4658ef48-3456-435b-a0a4-dbc504672836" />

A more detailed look at the timestamps shows that the duplication behavior cascades; after the first file is uploaded, it is uploaded again when the second one is uploaded, then the first and second are both uploaded again when the third is uploaded, etc.

```bash
/uploadedfiles$ ls -trla --full-time
2026-04-06 09:35:22.737229390 -0500 photo1.png
2026-04-06 09:35:22.742322493 -0500 photo1_6QNVqq5.png
2026-04-06 09:35:22.744428724 -0500 photo2.png
2026-04-06 09:35:22.747188478 -0500 photo1_MDasuZP.png
2026-04-06 09:35:22.749521236 -0500 photo2_6tM8rP7.png
2026-04-06 09:35:22.751300246 -0500 photo3.png
2026-04-06 09:35:22.755066149 -0500 photo1_aW4bDNz.png
2026-04-06 09:35:22.757565714 -0500 photo2_boy7be0.png
2026-04-06 09:35:22.759408501 -0500 photo3_mwNo1ez.png
2026-04-06 09:35:22.761332989 -0500 photo4.png
```

This means that if a user were to upload 10 files, the result would be 10 + 9 + 8 +7 + 6 + 5 + 4 + 3 + 2 + 1 = 55 new images stored.

[...]

PR #12826, opened 2026-07-05
Fix for duplicating files when adding or updating an existing list of…

bferguso:
… files

### Types of changes

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -1509,6 +1509,14 @@ class FileListDataType(BaseDataType):
                 file_info if isinstance(file_info, str) else file_info.get("name")
             )
             original_file_path = file_path
+
+            # If file_info is a dict with an existing file_id, the file is already
+            # stored on the server. Pass it through without creating a new File record.
+            if isinstance(file_info, dict) and file_info.get("file_id"):
+                tile_file = {**file_info}
+                tile_data.append(tile_file)
+                continue
+
             tile_file = {}
             try:
                 file_stats = os.stat(file_path)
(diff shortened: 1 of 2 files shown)


chiatt (changes requested):
I'm still seeing [#12681](https://github.com/archesproject/arches/issues/12681) when I add files to the FileList widget the way @mradamcox describes. His issue seems like it may have something to do with the front-end because it only occurs when adding one file at a time and then saving. If the user adds multiple files to the browser at the same time - there's no problem. 

I suspect the fix in this PR solves a different problem because the change here is to `transform_value_for_tile` method which is only called during ETL, the tile API, or using Arches Querysets. Tile `save` doesn't call it because data coming from the UI is always in the proper format and does not need transformation.

bferguso:
> I'm still seeing [#12681](https://github.com/archesproject/arches/issues/12681) when I add files to the FileList widget the way @mradamcox describes. His issue seems like it may have something to do with the front-end because it only occurs when adding one file at a time and then saving. If the user adds multiple files to the browser at the same time - there's no problem.
> 
> I suspect the fix in this PR solves a different problem because the change here is to `transform_value_for_tile` method which is only called during ETL, the tile API, or using Arches Querysets. Tile `save` doesn't call it because data coming from the UI is always in the proper format and does not need transformation.

Quite right. We're not actually seeing this behaviour in our implementation, but we're using S3 so that could explain why  - I'd thought the transform function ran in the OOB front end as well - my mistake.

bferguso:
Removed the link to origin issue - this PR solves file duplication when persisting through `arches-querysets`.

chiatt (approved):
lgtm!

PR #12841, opened 2026-07-21
Revert js dependency upgrades

chiatt:
Reverts jquery-validation and underscore upgrades


package.json
```diff
@@ -84,7 +84,7 @@
         "requirejs-text": "~2.0.16",
         "select-woo": "jadu/selectWoo",
         "shpjs": "~6.1.0",
-        "underscore": "^1.13.8",
+        "underscore": "^1.13.6",
```
chrabyrd, line 87: if this is being downgraded may want to remove the caret then

chrabyrd (changes requested):

package.json
chrabyrd: nevermind this is just a reversion

chrabyrd (approved):
LGTM

PR #7432, opened 2021-05-14
7428 adv styling geom clone

whatisgalen:
### Types of changes

-   [x] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Issues Solved

fixes #7428 

### Checklist

-   [ ] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @

Issue #7428, opened 2021-05-13
geom datatype advanced style cloning bug


whatisgalen:
**Describe the bug**

Somewhat convoluted, but bear with me:
If you have a graph called Historic District and you end up configuring advanced styling in its Resource Layer (in the Map Layer Manager), e.g. fill-color and what not, this advanced styling gets saved in a config object where the map source makes reference to `resource-[nodeid]`. 

If you clone Historic District and call it Historic District v2, it retains this same advanced styling however the references to the map source names remain the same. The problem here is now you're pointing to a source generated for a `nodeid` that is on a different graph (and may no longer exist). This can break the map filter in search if you end up marking the first graph "Inactive" because there still remain references to this map source which cannot be found.

**To Reproduce**

**Screenshots**

**Expected behavior**

**Your Arches Information**

*   Version used:
*   Operating System and version (desktop or mobile):
*   Browser Name and version:
*   Link to your Arches Install (optional):

**Additional context**

**Ticket Background**
*   Found by: @

chiatt:
This is likely related to #5902

whatisgalen:
A more fundamental solution here would be to implement graph versioning so that graphs wouldn't have to be cloned to be iterated over.

That notwithstanding, other than manually doing a find-and-replace of one nodeid with another in the json, there's literally no way to preserve the styling without breaking the map. To cause the least harm, I think when a graph gets cloned the map styling should just be removed altogether with some kind of alert to the user like "You will need to set map styling for this new graph." 

Otherwise we should look at alternative ways of generating these layer names that do not include a nodeid / some value unique to a graph.

chiatt:
Resolved in #7432

PR #12848, opened 2026-07-23
12847 - Don't blank out advanced styling config on graph copy

### LINKED ISSUE #12847: Create resource model draft loses map styling
When creating a new draft of a resource model that has one or more `geojson-feature-collection` nodes, the advanced styling is set to "False" and style set to empty string.

This was historically used so that the styling wasn't stale when a new branch was created, however since versioned graphs, it also has the undesirable effect of wiping out any advanced styling every time a new draft is created/published.

For now, the decision was to remove the explicit removal of those styles until a new pattern is set for how we manage resource styling in the future.

bferguso:
### Types of changes

### Description of Change
Remove explicit removal of geojson node config when copying graph

### Issues Solved

Closes #12847

### Checklist

#### Accessibility Checklist

#### Ticket Background
*   Sponsored by: Self Funded
*   Found by: @bferguso
*   Tested by: @bferguso
*   Designed by: @bferguso

### DIFF
--- a/arches/app/models/graph.py
+++ b/arches/app/models/graph.py
@@ -1005,10 +1005,6 @@ class Graph(models.GraphModel):
         for node in copy_of_self.nodes.values():
             node.is_immutable = bool(node.is_immutable or self.is_copy_immutable)
 
-            if node.datatype == "geojson-feature-collection":
-                node.config["advancedStyle"] = ""
-                node.config["advancedStyling"] = False
-
         copy_of_self.pk = uuid.uuid4()
         node_map = {}
         card_map = {}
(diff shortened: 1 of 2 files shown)


chiatt (approved):
lgtm!

chiatt:
The removal of the advanced styling (#7432) during graph copy was done to prevent an error caused by the fact that the nodeids in the advanced styling would match the original graph's nodeids - not the clone's. Graph copy is now used in the process of graph versioning in which we need the advanced styling to persist because the clone (draft graph) is just a temporary state, and a newly publish graph will have nodeids that must match the advanced styling. 

The change in this PR addresses the problem with graph versioning/publication because graph versioning is a much more common operation than standard cloning, but consequently it reintroduces the original problem. As @whatisgalen points out (#7428), a long-term solution is still needed regardless. This issue is not exclusive to the geojson datatype. The node-value nodes also share this problem because they store nodeids in their config. A safer solution might be to store node aliases rather than ids in node configs as they won't mutate during copy.

PR #12852, opened 2026-07-24
Prevent error when new project is created in a dev branch

chiatt:
### Types of changes

### Description of Change
Version must be a pre-release, otherwise the arches path in package.json will point to the stable version which has not yet been released. This will cause an error when a user attempts to create a new project on a dev branch.


njkim (approved):
lgtm

PR #12860, opened 2026-07-27
Remove release notes for unmaintained versions.

chiatt:
These release files remain available in the dev branches of their respective major/minor versions.


apeters:
@chiatt, is there a reason for this?  Especially if these files are available in other branches?  Taking a quick peek at the django project, they don't seem to do this.

PR #12869, opened 2026-07-29
Remove obsolete POSTGIS_VERSION setting

jacobtylerwalls:
The `POSTGIS_VERSION` [setting](https://docs.djangoproject.com/en/6.0/ref/contrib/gis/testing/#postgis-version) is used to lie to Django about the PostGIS version to potentially save a query, however, in current versions of Django with PostGIS, it's not used for anything.

I was debugging recently and thought that this was my PostGIS version, but I was actually on 3.5.5. Since it's potentially confusing and has no current benefit, my thought was to remove it.


chiatt (approved):
Looks good!

Issue #12694, opened 2026-04-10
Make Search backends swappable


chiatt:
On tile save Arches indexes resource data into ElasticSearch. If an application does not use the ElasticSearch index (e.g. Lingo) an admin should be able to disable indexing altogether. However, if an admin wants to index into a different backend, for example a triplestore or [arches-search](https://github.com/archesproject/arches-search/), the indexing methods in Arches should run the indexing logic of the respective search backend rather than those hard-coded for Elastic.

As part of this effort, we should make the current Search a plugin rather than a baked in page. The indexing methods can be moved into an indexing class that a search backend can inherit from and override methods as needed.

PR #12883, opened 2026-08-06
Make index optional in etl cmds

chiatt:
### Types of changes

### Description of Change
Optionally index when running etl commands, e.g.
```
etl tile-excel-importer -s /Users/cyrus/Documents/projects/getty/afrc/rascolls-data-pkg/Place.xlsx -mp -mxp 3 --no-index
```

### Checklist

### DIFF
--- a/arches/app/etl_modules/base_import_module.py
+++ b/arches/app/etl_modules/base_import_module.py
@@ -246,6 +246,7 @@ class BaseImportModule:
         loadid,
         multiprocessing=False,
         max_subprocesses=0,
+        index=True,
     ):
         try:
             with connection.cursor() as cursor:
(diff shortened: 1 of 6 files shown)


johnatawnclementawn (approved):
This looks good! 
 
One improvement that I think falls outside of scope, but should we support choosing which index(es) values are indexed to? i.e. a project has implemented Arches Search Extension, they may want to skip the ES index, but ensure resources are index to the search tables. My impression is that we haven't implemented that in Arches Search, is that correct?

chiatt:
> 
> One improvement that I think falls outside of scope, but should we support choosing which index(es) values are indexed to? i.e. a project has implemented Arches Search Extension, they may want to skip the ES index, but ensure resources are index to the search tables. My impression is that we haven't implemented that in Arches Search, is that correct?

Yeah - currently the arches-search extension index is a different command that isn't called following a bulk load. It has to be run optionally anyway. The plan is to make them swappable: https://github.com/archesproject/arches/issues/12694

When that happens, at least as I currently imagine it, a user will only index one using index methods of the search framework. If it's done that way, there will only every be one index to perform.

PR #12888, opened 2026-08-07
Filter resource queryset by perm

chiatt:
### Types of changes

### Description of Change
Introduces a method that applies a permission filter to a resource queryset and returns the filtered queryset. 

### Checklist

### DIFF
--- a/arches/app/permissions/arches_default_allow.py
+++ b/arches/app/permissions/arches_default_allow.py
@@ -203,7 +203,7 @@ class ArchesDefaultAllowPermissionFramework(ArchesPermissionBase):
             == "arches_default_allow.ArchesDefaultAllowPermissionFramework"
         ):
             raise Exception(
-                f"The default allow permission framework does not support this method. Use the default deny"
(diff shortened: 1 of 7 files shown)


arches/app/permissions/arches_default_allow.py
```diff
@@ -191,6 +191,18 @@ def get_filtered_instances(
...
+        user: User,
+        queryset,
+        resourceinstance_field: str = "resourceinstanceid",
+        permission: str = "models.view_resourceinstance",
+    ):
+        """
+        Default allow is deprecated; return the queryset unchanged.
+        """
+        return queryset
+
```
apeters, line 210: Even though this is "deprecated", it's still a viable option to use.  I think we should still return a queryset that respects any permissions that have been set on specific instances.

apeters: right now if I choose to use default allow, I'm going to leak instances that should be hidden.

chiatt: I've added an exception to deter people from staying on default allow. I pointed people to the default deny docs if they hit this exception.

apeters (approved):
lgtm

PR #12889, opened 2026-08-08
Bump Django to 6.1

jacobtylerwalls:
As part of the march toward supporting the 6.2 LTS, expected April 2027 ([roadmap](https://www.djangoproject.com/download/#supported-versions)).

The interesting changes are:
- updating to `MAILERS` (migration guide linked in changelog)
- removing deprecated no-argument form of `select_related()` (Alternatives: name fields explicitly, or use the `FETCH_PEERS` fetch mode, which is not exactly equivalent but provides a similar convenience--no explicit field naming.)


arches/install/arches-templates/project_name/settings.py-tpl
```diff
@@ -324,14 +324,18 @@ ENABLE_CAPTCHA = False
...
 
-# EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'  #<-- Only need to uncomment this for testing without an actual email server
-# EMAIL_USE_TLS = True
-# EMAIL_HOST = 'smtp.gmail.com'
-EMAIL_HOST_USER = "xxxx@xxx.com"
-# EMAIL_HOST_PASSWORD = 'xxxxxxx'
-# EMAIL_PORT = 587
-
-DEFAULT_FROM_EMAIL = EMAIL_HOST_USER
+MAILERS = {
```
jacobtylerwalls, line 327: Reusable arches apps should probably just delete any email settings.

arches/install/arches-templates/project_name/settings.py-tpl
```diff
@@ -324,14 +324,18 @@ ENABLE_CAPTCHA = False
...
+        "OPTIONS": {
+            "use_tls": True,
+            # "host": "smtp.gmail.com",
+            "username": "xxxx@xxx.com",
+            # "password" = "xxxxxxx",
+        },
+    },
+}
+
+DEFAULT_FROM_EMAIL = MAILERS["default"]["OPTIONS"]["username"]
```
jacobtylerwalls, line 338: Some manual testing of email functionality is probably warranted.

arches/app/models/graph.py
```diff
@@ -2039,9 +2039,10 @@ def serialize(
             else:
                 ret.pop("group_permissions", None)
 
-            ret["spatial_views"] = models.SpatialView.objects.select_related().filter(
-                geometrynode__graph__in=[self.source_identifier_id, self.graphid]
-            )
+            # Bark if any related fields are unintentionally fetched.
```
jacobtylerwalls, line 2042: This looked like over-fetching, so I added a performance guard to bark if I'm wrong.

tests/base_test.py
```diff
@@ -103,32 +95,19 @@ def loadOntology(cls):
...
-            VALUES (
-                44, '{oauth_client_id}', 'http://localhost:8000/test', 'public', 'client-credentials',
-                '{oauth_client_secret}',
-                'TEST APP', {user_id}, false, '1-1-2000', '1-1-2000', '{jwt_algorithm}', '', true, '*')
-            ON CONFLICT DO NOTHING;
-        """
-
-        sql = sql.format(
+    def setUpTestData(cls):
+        Application.objects.filter(id=44).update_or_create(
```
jacobtylerwalls, line 99: Relying on the models means we can pick up new defaults transparently instead of mantaining hard-coded SQL.

chiatt (changes requested):
Thanks @jacobtylerwalls, this looks great. There's just one test failing in `
/tests/models/tile_model_tests.py`

```
FAIL: test_update_sortorder_provisional_tile (tests.models.tile_model_tests.TileTests.test_update_sortorder_provisional_tile)
  File "/tile_model_tests.py", line 389, in test_update_sortorder_provisional_tile
    with self.assertWarns(FutureWarning):
         ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
AssertionError: FutureWarning not triggered
```
I think this line: https://github.com/archesproject/arches/blob/86a7cf4299da4a22344383f55bfb54309f7fe58e/tests/models/tile_model_tests.py#L389-L395

Might need to be changed to something like this to ensure that `TileModel.set_next_sortorder` is called on save.
```
        obj = TileModel.objects.get(pk=provisional_tile.pk)
        obj.save()
        obj.refresh_from_db()

        self.assertEqual(obj.sortorder, 1)
```

jacobtylerwalls:
I adjusted the test to do what the release notes suggest if you were really intending to use `update_or_create()`. Basically you have to hint whether or not you want a field in `update_fields` to opt in to slower writes. I can also leave it out of this PR if you like.

jacobtylerwalls:
@chiatt Thanks for looking. Is the plan still to raise the upper bound on Django during 8.2 so that it can pick up the 6.2 LTS?

Also, 📰 : https://www.djangoproject.com/weblog/2026/aug/10/annual-release-cycle/

chiatt:
> I adjusted the test to do what the release notes suggest if you were really intending to use `update_or_create()`. Basically you have to hint whether or not you want a field in `update_fields` to opt in to slower writes. I can also leave it out of this PR if you like.

Thanks - let's leave it in. It provides a nice example.

chiatt:
> @chiatt Thanks for looking. Is the plan still to raise the upper bound on Django during 8.2 so that it can pick up the 6.2 LTS?
> 
> Also, 📰 : https://www.djangoproject.com/weblog/2026/aug/10/annual-release-cycle/

I'm a little unsure about loosening the upper bound until we can test with an 8.2 release candidate to make sure we don't need to adjust anything and ship the Django upgrade in an Arches 8.2 patch (8.2.0 will be out in December). If you think raising the upper bound is pretty safe though, I'm okay with increasing the Django range to 6.2. 

Thanks for sharing that post. I need to write a post on the Arches forum explaining why we need to track the Django cadence and this gives me something nice to point to.

jacobtylerwalls:
Oh for sure no need to raise the upper bound yet, just flagging it could be good to do in April.

chiatt (approved):
This looks good to me. 

There is a failing test (`tests.bulkdata.jsonld_import_tests.JSONLDImportTests.test_write`), however, all tests pass locally, so the issue is likely with our CI workflow rather than any changes in this PR.

chiatt:
This should address the failing jsonld importer test: https://github.com/archesproject/arches/pull/12892 per @jacobtylerwalls's [suggestion](https://github.com/archesproject/arches/pull/12891#issuecomment-5260383887).

PR #12891, opened 2026-08-11
Prevent load of duplicate data

chiatt:
Just testing if preventing load of duplicate data allows jsonld importer tests to pass

### DIFF
--- a/tests/bulkdata/jsonld_import_tests.py
+++ b/tests/bulkdata/jsonld_import_tests.py
@@ -8,6 +8,7 @@ from pathlib import Path
 
 from arches.app.etl_modules.jsonld_importer import JSONLDImporter
 from arches.app.models.models import (
+    Concept,
     EditLog,
     ETLModule,
     GraphModel,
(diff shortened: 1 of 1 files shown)


tests/bulkdata/jsonld_import_tests.py
```diff
@@ -53,15 +54,23 @@ def setUp(self):
...
-        )
-        skos.save_concepts_from_skos(rdf)
+        if not Concept.objects.filter(
+            pk="7c90899a-dbe9-4574-9175-e69481a80b3c"
+        ).exists():
+            skos = SKOSReader()
+            rdf = skos.read_file(
+                "tests/fixtures/jsonld_base/rdm/jsonld_test_thesaurus.xml"
+            )
+            skos.save_concepts_from_skos(rdf)
```
chiatt, line 64: Maybe we don't want to hard code conceptids here, but this at least hints at the cause of the test failure.

jacobtylerwalls:
Which test failure was this for, the one on the Django 6.1 PR? That's reproducible if you merge dev/8.2.x, there was a oversight in 851cc0494b304668626c284e0444f22c32d4d117 where the signature of `save_to_tiles()` changed without fixing the override in the jsonld writer.

Blame the Github downtime, otherwise you'd have had the failures at push time.

PR #12892, opened 2026-08-12
Fix missing arg following #12883

chiatt:
### Types of changes

### Description of Change
Adds index argument in the `save_to_tiles` method of the jsonld importer that became required as a result of making indexing optional during bulk load #12883.


jacobtylerwalls (approved):
Nice. 🚢 

re: failures, I guess I should have used ~=6.0.0 instead of ~=6.0 when doing the version specifier last time. I can just fix on the Django 6.1 branch.

jacobtylerwalls:
@chiatt you could try a rebase?

PR #12897, opened 2026-08-18
Fix search export error when a linked tile is missing for a node-value node

### LINKED ISSUE #12574: Search export error when a node-value datatype references a deleted tile
If a node with the `node-value` datatype contains a reference to a tile that no longer exists, then any search export with that resource will return a 500 error when Arches attempts to flatten it into an exportable format.  

To use an Arches for HERs example: a Monument has a construction phase with a component, and the construction phase is deleted. The component still exists, and the reference to that construction phase's tileid still exists in the component's tiledata. 

When `search_export` attempts to get the correct db value, using `get_display_value`, it raises an Exception:
`Exception(f'Node with name "{node.name}" is not configured correctly.')`

We probably do still want to throw an exception here, but it might be better to replace this with a more specific `tile.DoesNotExist` exception. We could then catch this more specific exception one level up in `flatten_tiles` (within `search_export.py`), and set `node_value` to None.  The search then works. 

Another option would be to automatically clear any node-value references to a tile when it's deleted, but I'm not sure whether that's practical given the complexity and overhead of first finding all the relevant nodes and tiles.

CWDamm-Kint:
### Types of changes

### Description of Change
This PR handles the scenario where a node with the datatype `node-value` is included in search export, but a linked tile recorded using this node has been deleted. As described in #12574, this causes an error when the datatype `get_display_value` is run, as part of `flatten_tiles`. 

The solution catches the specific error for a missing Tile, and returns "Linked Tile Not Found" for inclusion in the csv. 

Running the search export tests was also creating zip files in `tests/fixtures/data/export_deliverables/` every time `test_write_export_to_zip` was run. I've added a snippet to delete these during `tearDownClass`, but could move this to another PR if shouldn't be included here. 

### Issues Solved
Closes #12574

### Checklist

#### Ticket Background
*   Sponsored by: K-int
*   Found by: @CWDamm-Kint 
*   Tested by: @CWDamm-Kint 
*   Designed by: @CWDamm-Kint

### DIFF
--- a/arches/app/datatypes/datatypes.py
+++ b/arches/app/datatypes/datatypes.py
@@ -2439,6 +2439,10 @@ class NodeValueDataType(BaseDataType):
                 datatype = datatype_factory.get_instance(value_node.datatype)
                 return datatype.get_display_value(value_tile, value_node)
             return ""
+
+        except models.TileModel.DoesNotExist:
+            return "Linked Tile Not Found"
+
         except:
             raise Exception(
                 f'Node with name "{node.name}" is not configured correctly.'
--- a/releases/7.6.26.md
+++ b/releases/7.6.26.md
@@ -1,8 +1,7 @@
 # Arches 7.6.26 Release Notes
 
 ## Bug Fixes
-
--
+- Handle missing linked tiles for node-value nodes during csv search export [#12897](https://github.com/archesproject/arches/pull/12897)
 
 ## Dependency changes
 
--- a/tests/fixtures/resource_graphs/Search Test Model.json
+++ b/tests/fixtures/resource_graphs/Search Test Model.json
@@ -5,6 +5,62 @@
             "cards": [
                 {
                     "active": true,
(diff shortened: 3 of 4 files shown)


chiatt (changes requested):
Could you add a changelog entry?

samuelscandrett:
Looks good to me, and appropriate for a 7.6.x patch as it's a critical, data-related bug. I could foresee it occurring on an AfHER deployment and requiring backend intervention to fix.

Happy to approve once a changelog entry is added, and the questions I left in a [comment](https://github.com/archesproject/arches/issues/12574#issuecomment-5397707426) on the issue are looked into (I saw your reply to say you were looking into them)

CWDamm-Kint:
Just to note, this fix also affects what shows up in the report:
image

samuelscandrett (approved):
LGTM!

chiatt (approved):

PR #12903, opened 2026-08-28
fix: resolve project src before installed module in tsconfig paths

StuCM:
### Types of changes

### Description of Change
tsconfig.paths mapped @/arches/* and @/<application>/* only at the installed module, so a file a project overrode under its own src was never resolved — type checking ran against the original instead of the override.

Fix is to add the project's src path first for each namespace, so it is the first location checked. paths entries are tried in order and a path that doesn't resolve falls through, so where there is no override (or no project at all) resolution falls back to the installed module exactly as before. This matches how the @ alias in webpack.common.js already resolves.

Also puts ../node_modules/@types/* ahead of ../node_modules/* so a plain-JS package resolves to its bundled declarations rather than succeeding at the package itself.

### Issues Solved

Closes #

### Checklist


chrabyrd (changes requested):
overall looks good! mind adding to the release notes?

chrabyrd:
superseded by: https://github.com/archesproject/arches/pull/12918

PR #12944, opened 2026-09-17
Package migrations rebuild

### LINKED ISSUE #9054: Design and implement a pattern for database migrations in a project to migrate tiles

chrabyrd:
### Types of changes

### Issues Solved

Closes #9054 

### Further comments

Versions an Arches application's graphs the way Django migrations version schema. Rebuilds the approach from #12840.

An application ships graphs. When version 2 changes one, every install that already has version 1 needs its graph rows updated, its tile data brought in line, the graph republished, and its resources moved onto the new publication. This does that, with a generator, a runner, and a ledger.

The committed graph JSON under pkg/graphs/** is the desired state, the same files load_package reads. makepkgmigrations replays the app's existing package migrations into an in-memory state, projects the committed JSON into the same shape, diffs them, and writes migrations. 

commands:

```
makepkgmigrations <app> [--name] [--dry-run] [--check]
migratepkg [<app>] [<migration>|zero] [--fake] [--plan] [--force] [--database]
showpkgmigrations [<app>] [--list] [--plan]
```

guards:

- **divergence.** These migrations are generated against migration history, not against your database. If a curator deleted a node in the Designer, an `AlterNode` would update zero rows and report success. `migratepkg` checks the rows a plan is about to write and refuses, naming them.
- **authoring machine.** If everything a plan creates is already present, it says so and tells you to record it with `--fake` instead.
- **faking data.** Faking structural work is safe when the rows exist. Faking tile work skips the only thing that would do it, so it is refused, with the two commands that do it properly.
- **reference data.** A graph points at an ontology, a template, widgets and card components. Those come from `load_package`. Missing ones are named before anything runs, instead of surfacing as `Ontology.DoesNotExist` inside `Graph.publish()`.
- **half reversals.** `migratepkg <app> zero` refuses up front if any migration in the plan cannot be reversed.
- **duplicate or malformed exports.** Two files carrying the same graphid is an error, because the winner would otherwise depend on filename order.
- **changes that cannot be converted automatically.** A datatype change, or a node moving between nodegroups, warns that stored values need a hand written `RunPackagePython` migration.

testing in rascolls:

Install normally, then adopt:

```bash
python manage.py setup_db --force
python manage.py packages -o load_package -s arches_rascolls/pkg -y
python manage.py makepkgmigrations arches_rascolls
python manage.py migratepkg arches_rascolls --fake
python manage.py showpkgmigrations arches_rascolls
```

Make a change and ship it. Add a node to an existing card in the Designer, publish, then:

```bash
GRAPH=$(psql arches_rascolls -tAc "SELECT graphid FROM graphs WHERE slug='collection_or_set';")
python manage.py packages -o export_graphs -d arches_rascolls/pkg/graphs/resource_models -g "$GRAPH"
python manage.py makepkgmigrations arches_rascolls
python manage.py migratepkg arches_rascolls --plan
```

Your own database already has the rows, so record the graph migration and run the data one:

```bash
python manage.py migratepkg arches_rascolls <last graph migration> --fake
python manage.py migratepkg arches_rascolls
```

Optional, and the path nothing exercised before. Blow the database away and let the migrations create every graph:

```bash
python manage.py setup_db --force
python manage.py load_ontology -s arches_rascolls/pkg/ontologies/linkedart
python manage.py migratepkg arches_rascolls --plan
time python manage.py migratepkg arches_rascolls
```

The ontology has to be loaded first. Graphs point at one by foreign key, and the pre-flight check refuses without it rather than failing inside `Graph.publish()`.

known limitaitons:

- card constraints are not migrated yet. `ConstraintModel` is not a concrete field of `CardModel`, so the projection does not see it.
- `migratepkg` does not reindex. Reindex affected resources afterwards or search will reflect the old graph.
- datatype changes and nodes moving between nodegroups warn rather than convert. Both need a decision a diff cannot make.
- migrations own graphs, and nothing else. Ontologies, controlled lists, concepts, functions, plugins and widgets still come from `load_package`. A release that adds a node using a new controlled list has to ship the list separately, and nothing orders the two.


chiatt (changes requested):
For string formatting - could you use f-strings or `.format()` for localized strings?

